PR #1, opened 2020-01-17
Add Batchwriter.

nevinheintze:
Initial commit to harbourbridge repository with license/contrib files
an initial README (mostly empty), and the first chunk of code: batchwriter.

Batchwriter is an abstraction on top of the standard Spanner Client
library that batches mutations together and manages the writing of
those mutations to Spanner. Harbourbridge uses this to effeciently
write data to Spanner.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+harbourbridge
+vendor/
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,0 @@
-harbourbridge
-vendor/
-
--- a/spanner/batchwriter.go
+++ b/spanner/batchwriter.go
@@ -1,0 +1,320 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package spanner provides high-level abstractions for working with
+// Cloud Spanner that are not available from the core Cloud Spanner
+// libraries.
+package spanner
+
+import (
+	"fmt"
+	"sync"
+	"sync/atomic"
+	"time"
+	"unsafe"
+
+	sp "cloud.google.com/go/spanner"
+)
+
+// Parameters used to control building batches to write to Spanner.
+// Batches are built by adding rows until we hit one of the thresholds.
+// Bigger batches are usually more efficient, but we need to be careful
+// not to exceed Spanner's limits. Also, sending huge RPCs is potentially
+// unreliable.
+const (
+	kb             = 1 << 10
+	mb             = 1 << 20
+	countThreshold = 10 * kb // Spanner per-operation limit is 20K.
+	byteThreshold  = 20 * mb // Spanner per-operation limit is 100MB.
+)
+
+// BatchWriter accumulates Spanner mutations (via AddRow) and assembles
+// them into batches that it asynchronously writes to Spanner. If Spanner
+// returns an error for a batch, BatchWriter splits the batch into
+// smaller chunks to retry, as it attempts to isolate which row(s) in a
+// batch is bad. BatchWriter respects Spanner's limits on byte size and
+// mutation count and has configurable limits on the number of
+// in-progress writes, amount of data buffered and retry behavior.
+// BatchWriter is not threadsafe: only one call to AddRow or Flush should
+// be active at any time.  See ExampleBatchWriter (batchwriter_test.go)
+// for sample usage code.
+type BatchWriter struct {
+	rows       []*row                     // Buffered rows.
+	rBytes     int64                      // Estimate of bytes for buffered rows.
+	rCount     int64                      // Mutation count for buffered rows.
+	write      func([]*sp.Mutation) error // Typically a closure that calls client.Apply, but structured this way for testing.
+	wg         sync.WaitGroup             // Tracks in-progress writes.
+	writeLimit int64                      // Limit on number of in-progress writes.
+	bytesLimit int64                      // Limit on bytes buffered. AddRow blocks if rBytes exceeded this value.
+	retryLimit int64                      // Limit on retries.
+	verbose    bool                       // If true, print out messages about each write batch.
+	a          asyncState
+}
+
+type row struct {
+	table string
+	cols  []string
+	vals  []interface{}
+}
+
+// Fields in this struct are modified asynchronously e.g. by go routines writing
+// data to Spanner. Either hold a lock or use atomics, as detailed below.
+type asyncState struct {
+	writes       int64            // Number of in-progress writes; access using atomic.
+	retries      int64            // Number of retries; access using atomic.
+	lock         sync.Mutex       // Protects errors and badRows
+	errors       map[string]int64 // Errors encountered; protected by lock.
+	badRows      []*row           // Rows that generated errors; protected by lock.
+	badRowsBytes int64            // Estimate of bytes for badRows; protected by lock.
+	droppedRows  map[string]int64 // Count of dropped rows, broken down by table.
+}
+
+// BatchWriterConfig specifies parameters for configuring BatchWriter.
+type BatchWriterConfig struct {
+	WriteLimit int64                      // Limit on number of in-progress writes.
+	BytesLimit int64                      // Limit on bytes buffered.
+	RetryLimit int64                      // Limit on retries.
+	Write      func([]*sp.Mutation) error // Function to call to write to Spanner (typically a closure that calls client.Apply).
+	Verbose    bool                       // If true, print out messages about each write batch.
+}
+
+// NewBatchWriter returns a new BatchWriter with parameters defined by config.
+func NewBatchWriter(config BatchWriterConfig) *BatchWriter {
+	return &BatchWriter{
+		write:      config.Write,
+		writeLimit: config.WriteLimit,
+		bytesLimit: config.BytesLimit,
+		retryLimit: config.RetryLimit,
+		verbose:    config.Verbose,
+		a: asyncState{
+			errors:      make(map[string]int64),
+			droppedRows: make(map[string]int64),
+		},
+	}
+}
+
+// AddRow appends a new row of data to bw's buffer of rows. Depending on the
+// state of BatchWriter, AddRow may immediately return, or it may initiate writes,
+// or it may block (waiting for some of the writes already in progress to
+// complete) and then initiate writes.
+func (bw *BatchWriter) AddRow(table string, cols []string, vals []interface{}) {
+	r := &row{table, cols, vals}
+	bw.rows = append(bw.rows, r)
+	bw.rBytes += byteSize(r)
+	bw.rCount += int64(len(r.cols))
+	bw.writeData()
+}
+
+// Flush initiates writes to Spanner of all buffered rows of data, and waits
+// for them to complete.
+func (bw *BatchWriter) Flush() {
+	for len(bw.rows) > 0 {
+		if atomic.LoadInt64(&bw.a.writes) < bw.writeLimit {
+			m, count, bytes := bw.getBatch()
+			if bw.verbose {
+				fmt.Printf("Starting write of %d rows to Spanner (%d bytes, %d mutations) [%d in progress]\n",
+					len(m), bytes, count, atomic.LoadInt64(&bw.a.writes))
+			}
+			bw.startWrite(m)
+		} else {
+			time.Sleep(10 * time.Millisecond)
+		}
+	}
+	bw.wg.Wait()
+}
+
+// DroppedRows returns a map of tables to counts of dropped rows.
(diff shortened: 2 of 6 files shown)


spanner/batchwriter.go
```diff
@@ -0,0 +1,320 @@
...
+			bw.startWrite(m)
+		} else {
+			time.Sleep(10 * time.Millisecond)
+		}
+	}
+	bw.wg.Wait()
+}
+
+// DroppedRows returns a map of tables to counts of dropped rows.
+func (bw *BatchWriter) DroppedRows() map[string]int64 {
```
lgruen: Should this indicate in the function name that _counts_ are returned? (E.g. in contrast, `BadRows` below returns the actual row contents.)

spanner/batchwriter.go
```diff
@@ -0,0 +1,320 @@
...
+				fmt.Printf("Have hit %d retries: will no do any more\n", atomic.LoadInt64(&bw.a.retries))
+			}
+			return
+		}
+		// Split into 10 pieces and retry. This is useful
+		// if a batch contains a bad data row (Spanner
+		// will fail the entire batch). In effect we attempt
+		// to narrow down which row (or rows) are bad, and
+		// write the 'good' rows to Spanner.
+		k := 1 + len(rows)/10
```
lgruen: How about splitting into pieces of 10 rows each, instead of 10 pieces?

spanner/batchwriter.go
```diff
@@ -0,0 +1,320 @@
...
+	for _, x := range rows {
+		m = append(m, sp.InsertOrUpdate(x.table, x.cols, x.vals))
+	}
+	if err := bw.write(m); err != nil {
+		hitRetryLimit := atomic.LoadInt64(&bw.a.retries) >= bw.retryLimit
+		retry := len(rows) > 1 && !hitRetryLimit
+		bw.errorStats(rows, err, retry)
+		if !retry {
+			if hitRetryLimit && bw.verbose {
+				fmt.Printf("Have hit %d retries: will no do any more\n", atomic.LoadInt64(&bw.a.retries))
```
lgruen: typo: will **not** do any more

spanner/batchwriter.go
```diff
@@ -0,0 +1,320 @@
...
+			return j
+		}
+		for i := 0; i < len(rows); i += k {
+			atomic.AddInt64(&bw.a.retries, 1)
+			bw.doWriteAndHandleErrors(rows[i:min(i+k, len(rows))])
+		}
+	}
+}
+
+// Note: backgroundWrite must be thread-safe.
```
lgruen: "must be thread-safe" -- why?

spanner/batchwriter.go
```diff
@@ -0,0 +1,320 @@
...
+func (bw *BatchWriter) backgroundWrite(rows []*row) {
+	defer bw.wg.Done()
+	defer atomic.AddInt64(&bw.a.writes, -1)
+	bw.doWriteAndHandleErrors(rows)
+}
+
+// startWrite initiates an asynchronous write of m to Spanner.
```
lgruen: write of `m` => write of `rows`?

spanner/batchwriter.go
```diff
@@ -0,0 +1,320 @@
...
+		bw.errorStats(rows, err, retry)
+		if !retry {
+			if hitRetryLimit && bw.verbose {
+				fmt.Printf("Have hit %d retries: will no do any more\n", atomic.LoadInt64(&bw.a.retries))
+			}
+			return
+		}
+		// Split into 10 pieces and retry. This is useful
+		// if a batch contains a bad data row (Spanner
+		// will fail the entire batch). In effect we attempt
```
lgruen: Based on our conversation in person, we should check whether Spanner actually fails the entire batch. Maybe the batch needs to be wrapped in a transaction?

.gitignore
```diff
@@ -0,0 +1,3 @@
+harbourbridge
+vendor/
```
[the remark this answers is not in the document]
nevinheintze: vendor: go used to use a vendor directory for managing dependencies (vendor directory had a copy of dependency source trees). That's gone away with go's new dependency system (use of vendor directory is off by default). So its not needed anymore. Early versions of this codebase used the older go system, and then switched to the new dependency system.

harbourbridge: I often build a binary within the git repository directory e.g. go build harbourbridge will do this. But for regular uses its probably not needed.

In short: I'll get rid of both and drop .gitignore.

spanner/batchwriter.go
```diff
@@ -0,0 +1,320 @@
...
+
+// Parameters used to control building batches to write to Spanner.
+// Batches are built by adding rows until we hit one of the thresholds.
+// Bigger batches are usually more efficient, but we need to be careful
+// not to exceed Spanner's limits. Also, sending huge RPCs is potentially
+// unreliable.
+const (
+	kb             = 1 << 10
+	mb             = 1 << 20
+	countThreshold = 10 * kb // Spanner per-operation limit is 20K.
```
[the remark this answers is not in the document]
nevinheintze: Thanks -- I like your suggestion of folding each definition into single line.

Also, cleanup up the definitions a little e.g. count limit is actually 20,000, so using that explicitly.

spanner/batchwriter.go
```diff
@@ -0,0 +1,320 @@
...
+)
+
+// BatchWriter accumulates Spanner mutations (via AddRow) and assembles
+// them into batches that it asynchronously writes to Spanner. If Spanner
+// returns an error for a batch, BatchWriter splits the batch into
+// smaller chunks to retry, as it attempts to isolate which row(s) in a
+// batch is bad. BatchWriter respects Spanner's limits on byte size and
+// mutation count and has configurable limits on the number of
+// in-progress writes, amount of data buffered and retry behavior.
+// BatchWriter is not threadsafe: only one call to AddRow or Flush should
```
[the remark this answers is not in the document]
nevinheintze: It would be a moderate amount of work. All access to rows, rBytes, rCount would need to be either protected by locks or use atomics. Somewhat more code and complexity. Also, the current locking is not on the critical path (its only for handling errors). But to make the library thread-safe, we'd need to protect rows, which is in the critical path. It might make sense to use go channels instead i.e. there's probably some thinking to do to make this change.

Slight preference for not doing it now since it isn't needed for harbourbridge. I think the trigger to do it would be if there is interest in using this code elsewhere.

spanner/batchwriter.go
```diff
@@ -0,0 +1,320 @@
...
+	bytesLimit int64                      // Limit on bytes buffered. AddRow blocks if rBytes exceeded this value.
+	retryLimit int64                      // Limit on retries.
+	verbose    bool                       // If true, print out messages about each write batch.
+	a          asyncState
+}
+
+type row struct {
+	table string
+	cols  []string
+	vals  []interface{}
```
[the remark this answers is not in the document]
nevinheintze: Yes. We don't check that: I think it is checked in the Spanner client library. Note that this struct mimics InsertOrUpdate (https://godoc.org/cloud.google.com/go/spanner#InsertOrUpdate).

spanner/batchwriter.go
```diff
@@ -0,0 +1,320 @@
...
+	for len(bw.rows) > 0 {
+		if atomic.LoadInt64(&bw.a.writes) < bw.writeLimit {
+			m, count, bytes := bw.getBatch()
+			if bw.verbose {
+				fmt.Printf("Starting write of %d rows to Spanner (%d bytes, %d mutations) [%d in progress]\n",
+					len(m), bytes, count, atomic.LoadInt64(&bw.a.writes))
+			}
+			bw.startWrite(m)
+		} else {
+			time.Sleep(10 * time.Millisecond)
```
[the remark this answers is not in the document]
nevinheintze: Flush needs to wait till completion in any case.

The sleep here (and the one in writeData) are because we don't want to have too many pending writes out to Spanner. The main reason for this is that if we just churned through the pg_dump file and generated lots of pending mutations all waiting to be written to Spanner, we'd consume considerable memory in harbourbridge and we'd OOM on larger pg_dumps.

To make harbourbridge more robust and limit its memory footprint, we limit the number of pending Spanner writes and the amount of data we keep in rows, and if we exceed these limits, we throttle our processing of the pg_dump i.e. we leave the data on-disk until we have capacity to process it.

spanner/batchwriter.go
```diff
@@ -0,0 +1,320 @@
...
+	rows       []*row                     // Buffered rows.
+	rBytes     int64                      // Estimate of bytes for buffered rows.
+	rCount     int64                      // Mutation count for buffered rows.
+	write      func([]*sp.Mutation) error // Typically a closure that calls client.Apply, but structured this way for testing.
+	wg         sync.WaitGroup             // Tracks in-progress writes.
+	writeLimit int64                      // Limit on number of in-progress writes.
+	bytesLimit int64                      // Limit on bytes buffered. AddRow blocks if rBytes exceeded this value.
+	retryLimit int64                      // Limit on retries.
+	verbose    bool                       // If true, print out messages about each write batch.
+	a          asyncState
```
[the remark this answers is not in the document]
nevinheintze: I'm assuming your comment is about the single letter struct field name?

Yes, happy to change this, but the single letter name here was motivated by go's naming conventions/guidelines. In particular, the type here is all we need to understand what this thing does: it represents async state for this struct.
We could have:

      async          asyncState

But that's kind of redundant.

In general, my take is:
- prefer shorter names where possible, especially for local variables with limited scope
- longer names for things that are used further from their declaration, especially global variables
- for a struct, if the type is specific/descriptive (and only one of them in the struct) use a shorter name for the field, but if the type is very generic and there are many things of that type in the struct, be more descriptive e.g. use more descriptive names for int fields, but shorter for things like asyncState

But I'm not a go expert, so will defer to others on making this call.

spanner/batchwriter.go
nevinheintze: Hmm. Yes you are right -- the terminology is pretty inconsistent. Even having badRows and droppedRows in the struct is pretty confusing -- bad and dropped refer to the essentially the same thing (rows that are bad and got dropped), but there is a subtle difference that isn't at all apparent.
I'll figure out something clearer and more consistent.

spanner/batchwriter.go
nevinheintze: Done - thanks!

spanner/batchwriter.go
nevinheintze: We could have up to 10,000 rows in a mutation.
Splitting immediately into 10 rows would mean we'd send 1000 re-try operations.

Splitting into 10 pieces is a bit more divide-and-conquer.

Was your concern here about making code as simple as possible, or something else?

lgruen: This was more about making the "bad rows" diagnostics more useful -- if there are 10,000 rows in a mutation and just one of them is bad, it might be really helpful to narrow this down. If we know that it's at least one bad row in a set of 10, that's likely easier to diagnose than 1 in 1000. I assume that we don't expect these retries to be necessary often, so sending 1000 retry operations might be ok?

spanner/batchwriter.go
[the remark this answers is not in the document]
lgruen: Completely agree that it's a good idea to limit the memory usage.

Instead of "polling" `bw.a.writes`, could this be implemented through channels?

spanner/batchwriter.go
nevinheintze: backgroundWrite and doWriteAndHandleErrors are run via go-routines.
So there will be multiple instances of these functions running.

The boundary between single-threaded vs go routines is the call to go bw.backgroundWrite in startWrite.

spanner/batchwriter.go
nevinheintze: Thanks for catching!

spanner/batchwriter.go
lgruen: As discussed in person: this is actually already recursive (divide-and-conquer), so please disregard my previous comment.

spanner/batchwriter.go
lgruen: Might be worth mentioning that in the comment?

spanner/batchwriter.go
lgruen: As discussed in person: the batch gets written in a transaction. Therefore marking this as resolved.

spanner/batchwriter_test.go
```diff
@@ -0,0 +1,234 @@
...
+	config := BatchWriterConfig{
+		BytesLimit: 100 << 20,
+		Verbose:    false,
+		RetryLimit: 1000,
+	}
+	for _, tc := range tests {
+		data, limit := generateRows(tc.count, tc.rowSize)
+		goodRows, badRows := partitionRows(tc.badRowIndex, data)
+		badMutations := toMutations(badRows)
+		var mutex = &sync.Mutex{}
```
larkee: Can this not be ``mutex := &sync.Mutex{}``?

nevinheintze: Done x 2.
Thanks!

spanner/batchwriter.go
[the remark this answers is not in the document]
nevinheintze: Done.

spanner/batchwriter.go
nevinheintze: Done. Comment for this function now reads:
// Note: backgroundWrite must be thread-safe because it is run as
// a go routine.

Added similar comment to previous function.

spanner/batchwriter.go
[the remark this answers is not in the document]
nevinheintze: As discussed, let's defer for now, and I'll add a feature request to make this whole library thread-safe, and (probably) use a channel for submitting rows to batchwriter.

PR #5, opened 2020-01-23
Add internal utilities.

nevinheintze:
Add package internal, which implements database-agnostic functionality.

### DIFF
--- a/internal/progress_test.go
+++ b/internal/progress_test.go
@@ -22,13 +22,13 @@ import (
 )
 
 func TestProgress(t *testing.T) {
-	var total int64 = 4321
+	total := int64(4321)
 	p := NewProgress(total, "Progress", false)
 	assert.Equal(t, 0, p.pct)
 	time.Sleep(500 * time.Millisecond)
-	for _, i := range []int64{1000, 2000, 3000} {
-		p.MaybeReport(i)
-		assert.Equal(t, int((100*i)/total), p.pct)
+	for _, v := range []int64{1000, 2000, 3000} {
+		p.MaybeReport(v)
+		assert.Equal(t, int((100*v)/total), p.pct)
 		time.Sleep(500 * time.Millisecond)
 	}
 	pct := p.pct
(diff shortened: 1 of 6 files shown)


internal/progress_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestProgress(t *testing.T) {
+	var total int64 = 4321
```
larkee: ``total := int64(4321)``

internal/progress_test.go
```diff
@@ -0,0 +1,43 @@
...
+	p := NewProgress(total, "Progress", false)
+	assert.Equal(t, 0, p.pct)
+	time.Sleep(500 * time.Millisecond)
+	for _, i := range []int64{1000, 2000, 3000} {
```
larkee: Prefer ``v`` over ``i`` to avoid confusion over index vs value.

internal/progress_test.go
```diff
@@ -0,0 +1,43 @@
...
+		p.MaybeReport(i)
+		assert.Equal(t, int((100*i)/total), p.pct)
+		time.Sleep(500 * time.Millisecond)
+	}
+	pct := p.pct
+	p.MaybeReport(2000)
+	assert.Equal(t, pct, p.pct) // pct is monotonic.
+	p.MaybeReport(5000)
+	assert.Equal(t, 100, p.pct) // Never exceed 100%.
+	p.Done()
```
larkee: Should this line be here? Are you missing a subsequent check?

internal/progress_test.go
nevinheintze: Done.
Sorry, I missed this - thanks!

internal/progress_test.go
nevinheintze: Done.

internal/progress_test.go
nevinheintze: Got rid of this, and added a separate test for Done.

PR #10, opened 2020-01-24
Add tablemap.

nevinheintze:
Tablemap maintains the mappings between PostgreSQL tables and columns
and Spanner tables and columns. Note that PostgreSQL table and column
names can be essentially any string, but Spanner table and column
names must use a limited character set. So when we convert PostgreSQL
to Spanner, we potentially have to remap PostgreSQL table and column
names to make them legal Spanner.

### DIFF
--- a/internal/postgres/tablemap.go
+++ b/internal/postgres/tablemap.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,8 +16,9 @@ package postgres
 
 import (
 	"fmt"
-	"harbourbridge/internal"
 	"strconv"
+
+	"harbourbridge/internal"
 )
 
 // GetSpannerTable maps a PostgreSQL table name into a legal Spanner table
@@ -30,7 +31,7 @@ import (
 // c) we consistently return the same name for this table.
 func GetSpannerTable(conv *Conv, pgTable string) (string, error) {
 	if pgTable == "" {
-		return "", fmt.Errorf("Bad parameter: table string is empty")
+		return "", fmt.Errorf("bad parameter: table string is empty")
 	}
 	if sp, found := conv.toSpanner[pgTable]; found {
 		return sp.name, nil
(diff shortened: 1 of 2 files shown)


internal/postgres/tablemap_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetSpannerTable(t *testing.T) {
+	conv := MakeConv()
+	basicTests := []struct {
+		name    string // Name of test.
+		pgTable string // PostgreSQL table name to test.
+		error   bool   // Whether an error is expected.
+		changed bool   // Should table name to be changed.
```
rghetia: instead of 'changed' if it was 'expectedTableName' then uniqueness/algorithm test can be covered here.

internal/postgres/tablemap_test.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func TestGetSpannerCol(t *testing.T) {
+	conv := MakeConv()
+	basicTests := []struct {
+		name    string // Name of test.
+		pgTable string // PostgreSQL table name to test
+		pgCol   string // PostgreSQL col name to test
+		error   bool   // Whether an error is expected.
+		changed bool   // Should col name to be changed.
```
rghetia: same comment as above

internal/postgres/tablemap.go
```diff
@@ -0,0 +1,116 @@
+// Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
nevinheintze: Done.

internal/postgres/tablemap.go
```diff
@@ -0,0 +1,116 @@
...
+// name. Note that PostgreSQL column names can be essentially any string, but
+// Spanner column names must use a limited character set. This means that
+// getSpannerTable may have to change a name to make it legal, we must ensure
+// that:
+// a) the new table name is legal
+// b) the new table name doesn't clash with other Spanner table names
+// c) we consistently return the same name for this table.
+func GetSpannerTable(conv *Conv, pgTable string) (string, error) {
+	if pgTable == "" {
+		return "", fmt.Errorf("Bad parameter: table string is empty")
```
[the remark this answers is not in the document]
nevinheintze: Done.
Oops, sorry. I fixed a bunch of them elsewhere, but missed these.

Curiously I can't get golint to catch these?
I'm running golang/lint - is there something better? I tried playing with min_confidence but still no luck.

internal/postgres/tablemap.go
```diff
@@ -0,0 +1,116 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package postgres
+
+import (
+	"fmt"
+	"harbourbridge/internal"
```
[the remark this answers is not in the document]
nevinheintze: Done.
I got a bit lazy relying on gofmt to deal with imports.

internal/postgres/tablemap_test.go
nevinheintze: Done.
I was trying to specify the tests in a way that wouldn't be dependent on the specific collision avoidance algorithm. But I'm not sure it is worth doing that. Moreover, the specifics of collision avoidance are API visible in any case, so not unreasonable to test the specifics.

internal/postgres/tablemap_test.go
nevinheintze: Done.

PR #11, opened 2020-01-24
Add statement.

nevinheintze:
Statement extracts schema information and data from PostgreSQL
statements

### DIFF
--- a/internal/postgres/statement.go
+++ b/internal/postgres/statement.go
@@ -1,0 +1,696 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package postgres
+
+import (
+	"fmt"
+	"harbourbridge/internal"
+	"harbourbridge/spanner/ddl"
+	"reflect"
+	"strconv"
+	"strings"
+	"time"
+	"unicode"
+
+	nodes "github.com/lfittl/pg_query_go/nodes"
+)
+
+type copyOrInsert struct {
+	stmt    stmtType
+	spTable string   // Spanner table name.
+	pgTable string   // Postgres table name.
+	cols    []string // Spanner cols names.
+	vals    []string // Empty for COPY-FROM.
+}
+
+type stmtType int
+
+const (
+	copyFrom stmtType = iota
+	insert
+)
+
+// processStatements extracts schema information and data from PostgreSQL
+// statements, updating Conv with new schema information, and returning
+// copyOrInsert if a COPY-FROM or INSERT statement is encountered.
+// Note that the actual parsing/processing of COPY-FROM data blocks is
+// handled elsewhere (see process.go).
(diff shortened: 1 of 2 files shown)


internal/postgres/statement.go
```diff
@@ -0,0 +1,697 @@
...
+				statementError(conv, n, fmt.Errorf("found %s node in Arg", reflect.TypeOf(c)))
+				return
+			}
+		}
+	}
+}
+
+// getSpannerType determines the Spanner type for the scalar postgres type
+// defined by id and mods. This is the core postgres-to-Spanner type mapping.
+// getSpannerType returns the Spanner type and a bool to indicate whether
```
larkee: `// getSpannerType returns the Spanner type and a bool`
No it doesn't?

internal/postgres/statement.go
```diff
@@ -0,0 +1,697 @@
...
+		}
+		ids = append(ids, s)
+	}
+	if len(ids) > 1 && ids[0] == "pg_catalog" {
+		ids = ids[1:]
+	}
+	return strings.Join(ids, "."), nil
+}
+
+func prType(pgCol pgColDef) string {
```
larkee: Change to printType. I thought this was a typo for pgType.

internal/postgres/statement.go
```diff
@@ -0,0 +1,697 @@
...
+
+type constraint struct {
+	ct   nodes.ConstrType
+	keys []string
+}
+
+// analyzeConstraints traverses a list of nodes (expecting them to be
+// Constraint nodes), and collects the contraints they represent as
+// a list of constraint-type/key-list pairs.
+func analyzeConstraints(conv *Conv, n nodes.Node, pgTable string, l []nodes.Node) (cs []constraint) {
```
larkee: Would extractConstraints make sense? It would make it clear what the return type is. To me, analyzeConstraints doesn't sound like it would return something.

internal/postgres/statement.go
rghetia: return type does not match description.

internal/postgres/statement.go
```diff
@@ -0,0 +1,697 @@
...
+		c.keys = []string{pgCol}
+		cs = append(cs, c)
+	}
+	return cs
+}
+
+// getConstraint finds the constraint-type ct in the constraint list cs.
+// It returns true (along with the associated keys) if it finds ct,
+// and false otherwise.
+func getConstraint(ct nodes.ConstrType, cs []constraint) (bool, []string) {
```
larkee: Perhaps switch the order of the results:
`([]string, bool)`
It's more consistent with the golang map lookup syntax of `val, ok := m['key']`

internal/postgres/statement.go
```diff
@@ -0,0 +1,697 @@
...
+				}
+			default:
+				conv.unexpected(fmt.Sprintf("Processing %v statement: found %s node in ValuesList", reflect.TypeOf(n), reflect.TypeOf(v)))
+			}
+		}
+	}
+	return values
+}
+
+func statementError(conv *Conv, n nodes.Node, err error) {
```
larkee: Perhaps logStatementError or logStmtError.

internal/postgres/statement.go
nevinheintze: Done.
Thanks for catching.

nevinheintze: Fixed.
Thanks for catching!

internal/postgres/statement.go
nevinheintze: Done.

internal/postgres/statement.go
nevinheintze: Done.
Yes, I think that is a more intuitive name.

internal/postgres/statement.go
nevinheintze: Done.

internal/postgres/statement.go
nevinheintze: Done.

PR #13, opened 2020-01-24
Add data.

nevinheintze:
Data implements data conversion i.e. it maps PostgreSQL data to
Spanner data, based on the Spanner schema generated by schema
conversion.

### DIFF
--- a/internal/postgres/data.go
+++ b/internal/postgres/data.go
@@ -1,0 +1,266 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package postgres
+
+import (
+	"encoding/hex"
+	"fmt"
+	"harbourbridge/spanner/ddl"
+	"math/bits"
+	"reflect"
+	"strconv"
+	"strings"
+	"time"
+
+	"cloud.google.com/go/civil"
+)
+
+// ConvertData maps the PostgreSQL data in vals into Spanner data,
+// based on the specified Spanner and PostgreSQL schemas. Note that since entries
+// in vals may be empty, we also return the list of columns (empty cols are dropped).
+func ConvertData(conv *Conv, spTable, pgTable string, cols []string, vals []string) ([]string, []interface{}, error) {
+	spannerSchema, ok1 := conv.spSchema[spTable]
+	pgSchema, ok2 := conv.pgSchema[pgTable]
+	if !ok1 || !ok2 {
+		return []string{}, []interface{}{}, fmt.Errorf("Can't find table %s in schema", spTable)
+	}
+	var c []string
+	var v []interface{}
+	if len(cols) != len(vals) {
+		return []string{}, []interface{}{}, fmt.Errorf("Bad parameters: cols and vals have different lengths: len(cols)=%d, len(vals)=%d", len(cols), len(vals))
+	}
+	for i, col := range cols {
+		if vals[i] == "\\N" { // PostgreSQL representation of empty column in COPY-FROM blocks.
+			continue
+		}
+		cd, ok := spannerSchema.Cds[col]
+		if !ok {
+			return []string{}, []interface{}{}, fmt.Errorf("Can't find col %s in schema", col)
+		}
+		pgCol, ok := pgSchema.cols[col]
+		if !ok {
+			return []string{}, []interface{}{}, fmt.Errorf("Can't find col %s in pgSchema", col)
+		}
+		var x interface{}
+		var err error
+		if cd.IsArray {
+			x, err = convArray(cd.T, pgCol, conv.location, vals[i])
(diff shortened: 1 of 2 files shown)


internal/postgres/data.go
```diff
@@ -0,0 +1,266 @@
...
+	b, err := strconv.ParseBool(val)
+	if err != nil {
+		return b, fmt.Errorf("Can't convert to bool: %w", err)
+	}
+	return b, err
+}
+
+func convBytes(val string) ([]byte, error) {
+	if val[0:3] != `\\x` {
+		return []byte{}, fmt.Errorf("Can't convert bytea data to bytes: doesn't start with \\x prefix")
```
rghetia: typo? 
s/bytea data to/data to/

internal/postgres/data.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	b, err := hex.DecodeString(val[3:])
+	if err != nil {
+		return b, fmt.Errorf("Can't convert bytea data to bytes: %w", err)
```
rghetia: same.

internal/postgres/data_test.go
```diff
@@ -0,0 +1,237 @@
...
+
+	timestampTests := []struct {
+		name string
+		pgt  string
+		in   string
+		e    interface{}
+	}{
+		{"timestamptz hour timezone", "timestamptz", "2019-10-29 05:30:00+10:00", getTime(t, "2019-10-29T05:30:00+10:00")},
+		{"timestamptz hour/min timezone", "timestamptz", "2019-10-29 05:30:00+10:30", getTime(t, "2019-10-29T05:30:00+10:30")},
+		{"timestamptz no timezone", "timestamptz", "2019-10-29 05:30:00", getTime(t, "2019-10-29T05:30:00+11:00")},
```
rghetia: will this fail if I run the test in PST because it is taking local time?

internal/postgres/data_test.go
```diff
@@ -0,0 +1,237 @@
...
+		},
+		{
+			name: "Error in float64",
+			cols: []string{"a", "b", "c"},
+			vals: []string{"6", "6.6e", "true"},
+		},
+		{
+			name: "Error in bool",
+			cols: []string{"a", "b", "c"},
+			vals: []string{" 6", "6.6", "truee"},
```
rghetia: 'Error in bool' is returned because column 'c' is evaluated first? The reason I am asking is because column 'a' also has error in int64.

internal/postgres/data.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+		v = append(v, x)
+		c = append(c, col)
+	}
+	if aux, ok := conv.syntheticPKeys[spTable]; ok {
+		c = append(c, aux.col)
+		v = append(v, int64(bits.Reverse64(uint64(aux.sequence))))
+		aux.sequence++
+		conv.syntheticPKeys[spTable] = aux
+	}
```
rghetia: what is syntheticPKeys for?

internal/postgres/data.go
```diff
@@ -0,0 +1,266 @@
+// Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
nevinheintze: Done.

internal/postgres/data.go
```diff
@@ -0,0 +1,266 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package postgres
+
+import (
+	"encoding/hex"
+	"fmt"
+	"harbourbridge/spanner/ddl"
```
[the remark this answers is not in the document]
nevinheintze: Done.

internal/postgres/data.go
nevinheintze: Done.

Actually this was intended: 'bytea' is the PostgreSQL type we are converting from here (https://www.postgresql.org/docs/9.0/datatype-binary.html).
But after looking at the other comments in the convXXX functions, I think it makes sense to have all of these errors use a consistent format.

internal/postgres/data.go
nevinheintze: Done.

Actually this was intended: 'bytea' is the PostgreSQL type we are converting from here (https://www.postgresql.org/docs/9.0/datatype-binary.html).
But after looking at the other comments in the convXXX functions, I think it makes sense to have all of these errors use a consistent format.

internal/postgres/data_test.go
nevinheintze: No. We explicitly set the location in the test body to ensure that the test is robust: see conv.SetLocation(loc) a few lines below (line 83).

Added a comment to clarify this.

internal/postgres/data_test.go
nevinheintze: Fixed.
Good catch -- thanks! Should be just a single error in the third arg (bool).

internal/postgres/data.go
nevinheintze: Spanner requires primary keys for all tables. PostgreSQL recommends the use of
primary keys for all tables, but does not enforce this. When converting a PostgreSQL table
without a primary key, we need to create a new primary key -- these are called synthetic keys.
They are introduced in statement.go (see https://github.com/cloudspannerecosystem/harbourbridge/pull/11).
We need to deal with them in data because we need to generate values for these new columns (since
they don't appear in the PostgreSQL data).

internal/postgres/data_test.go
rghetia: Ahha!, missed the line 83.

internal/postgres/data.go
rghetia: Thx for the explanation.

PR #26, opened 2020-01-28
Add process.

nevinheintze:
Process reads pg_dump data and does schema and data conversion. It
parses the pg_dump statements (using the pg_query_go library as well
as hand-coded parsing of COPY-FROM data blocks), calls into statements
and data to do schema and data conversion, and then writes data to
Spanner (by calling conv.dataSink). In other words, it glues together
lower-layer components into overall 'process pg_dump' functionality.

Note that that actual code here is fairly small (about 160 lines). The
bulk of this pull-request is tests. This is because we want to test much of the
lower-level functionality via the public API provided by process.

### DIFF
--- a/internal/postgres/process.go
+++ b/internal/postgres/process.go
@@ -1,0 +1,163 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package postgres implements PostgreSQL-specific schema conversion
+// and data conversion for HarbourBridge.
+package postgres
+
+import (
+	"fmt"
+	"strings"
+
+	pg_query "github.com/lfittl/pg_query_go"
+	nodes "github.com/lfittl/pg_query_go/nodes"
+
+	"harbourbridge/internal"
+)
+
+// ProcessPgDump reads pg_dump data from r and does schema or data conversion,
+// depending on whether conv is configured for schema mode or data mode.
+// In schema mode, ProcessPgDump incrementally builds a schema (updating conv).
+// In data mode, ProcessPgDump uses this schema to convert PostgreSQL data
+// and writes it to Spanner, using the data sink specified in conv.
+func ProcessPgDump(conv *Conv, r *internal.Reader) {
+	for {
+		startLine := r.LineNumber
+		startOffset := r.Offset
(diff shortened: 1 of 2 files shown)


internal/postgres/process.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+		if r.EOF {
+			conv.unexpected("Reached eof while parsing copy-block")
+			return
+		}
+		conv.statsAddRow(c.spTable, conv.schemaMode())
+		// We have to read the copy-block data so that we can process the remaining
+		// pg_dump content. However, if we don't want the data, stop here.
+		// In particular, avoid the strings.Split and ProcessRow calls below, which
+		// weill be expensive for huge datasets.
```
rghetia: typo s/weill/will/
I am assuming that there is not schema after copy-block hence you can stop processing if it is only schema mode.

internal/postgres/process.go
```diff
@@ -0,0 +1,163 @@
...
+		if !conv.dataMode() {
+			continue
+		}
+		// COPY-FROM blocks use tabs to separate data items. Note that space within data
+		// items is significant e.g. if a table row contains data items "a ", " b "
+		// it will be show in the COPY-FROM block as "a \t b ".
```
rghetia: typo s/show/shown/

internal/postgres/process_test.go
```diff
@@ -0,0 +1,589 @@
...
+						"b_b": ddl.ColumnDef{Name: "b_b", T: ddl.String{Len: ddl.MaxLength{}}},
+						"n_n": ddl.ColumnDef{Name: "n_n", T: ddl.Int64{}},
+					},
+					Pks: []ddl.IndexKey{ddl.IndexKey{Col: "a__"}}}},
+		},
+		// The "Data conversion: ..." cases check data conversion for each type.
+		{
+			name: "Data conversion: bool, bigserial, char, bytea",
+			input: `
+CREATE TABLE test (id integer PRIMARY KEY, a bool, b bigserial, c char, d bytea);
```
rghetia: accidental newline? other places as well.

internal/postgres/process.go
nevinheintze: Done.

Yes, that's right. The copy-block just contains data, so we skip it in schema mode.

But we do have to start processing again after the end of the copy-block to process any statements that follow the copy-block (which can contain more scheme info).

internal/postgres/process.go
nevinheintze: Done.
Thanks for catching!

internal/postgres/process_test.go
nevinheintze: Actually I formatted the COPY-FROM tests like this on purpose.

The reason is that the copy-from block (the bit with the data that is ended with \.) has to be very specifically formatted. In particular, it has to use \t to separate data elements, and any whitespace is significant. So this part was easier to format using raw string literals. Also, its easier to cut-and-paste examples from pg_dump files.

I added an extra new line at the start so everything would line up -- extra whitespace is fine in CREATE TABLE statements.

But I'm open to changing this?

rghetia: that makes sense. You can leave it as is.

PR #42, opened 2020-02-12
fix: update to exit early when the file is not parsable.

### LINKED ISSUE #15: Early exit if pg_dump not parsable
If pg_dump is not parsable, we still proceed and create a DB (with empty schema) and then go on to do data conversion (which will fail since there is no schema). This all takes time and the output from the tool is very clear about what happened i.e. it says 'no schema and no data', but doesn't pinpoint the likely cause.

A better user experience would be to recognize that the pg_dump wasn't parsable, print an appropriate error and exit early.

hengfengli:
This change fixes an issue that when the file is not parsable, we should not create the database and exit early. 

Fixes #15 

- [X] Tests pass

### DIFF
--- a/internal/postgres/process.go
+++ b/internal/postgres/process.go
@@ -31,11 +31,14 @@ import (
 // In schema mode, ProcessPgDump incrementally builds a schema (updating conv).
 // In data mode, ProcessPgDump uses this schema to convert PostgreSQL data
 // and writes it to Spanner, using the data sink specified in conv.
-func ProcessPgDump(conv *Conv, r *internal.Reader) {
+func ProcessPgDump(conv *Conv, r *internal.Reader) error {
 	for {
 		startLine := r.LineNumber
 		startOffset := r.Offset
-		b, stmts := readAndParseChunk(conv, r)
+		b, stmts, err := readAndParseChunk(conv, r)
+		if err != nil {
+			return err
+		}
 		ci := processStatements(conv, stmts)
 		internal.VerbosePrintf("Parsed SQL command at line=%d/fpos=%d: %d stmts (%d lines, %d bytes) ci=%v\n", startLine, startOffset, len(stmts), r.LineNumber-startLine, len(b), ci != nil)
 		if ci != nil {
@@ -53,6 +56,8 @@ func ProcessPgDump(conv *Conv, r *internal.Reader) {
 	if conv.schemaMode() {
 		conv.AddPrimaryKeys()
 	}
+
+	return nil
 }
 
 // ProcessRow converts a row of data and writes it out to Spanner.
@@ -98,7 +103,7 @@ func byteSize(r *row) int64 {
 
 // readAndParseChunk parses a chunk of pg_dump data, returning the bytes read,
 // the parsed AST (nil if nothing read), and whether we've hit end-of-file.
-func readAndParseChunk(conv *Conv, r *internal.Reader) ([]byte, []nodes.Node) {
+func readAndParseChunk(conv *Conv, r *internal.Reader) ([]byte, []nodes.Node, error) {
 	var l [][]byte
 	for {
 		b := r.ReadLine()
@@ -117,7 +122,7 @@ func readAndParseChunk(conv *Conv, r *internal.Reader) ([]byte, []nodes.Node) {
 			}
 			tree, err := pg_query.Parse(string(s))
 			if err == nil {
-				return s, tree.Statements
+				return s, tree.Statements, nil
 			}
 			// Likely causes of failing to parse:
 			// a) complex statements with embedded semicolons e.g. 'CREATE FUNCTION'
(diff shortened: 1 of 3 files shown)


internal/postgres/process.go
```diff
@@ -127,10 +132,7 @@ func readAndParseChunk(conv *Conv, r *internal.Reader) ([]byte, []nodes.Node) {
 			conv.stats.reparsed++
 		}
 		if r.EOF {
-			if len(l) != 0 {
-				fmt.Printf("Error parsing last %d line(s) of input\n", len(l))
-			}
-			return nil, nil
+			return nil, nil, fmt.Errorf("Error parsing input: %q", l)
```
nevinheintze: 'l' could be huge: maybe MB of data.
Probably don't want to put it all in the error message.
Either don't put it in, or limit it.

internal/postgres/process.go
```diff
@@ -31,11 +31,14 @@ import (
...
 // In data mode, ProcessPgDump uses this schema to convert PostgreSQL data
 // and writes it to Spanner, using the data sink specified in conv.
-func ProcessPgDump(conv *Conv, r *internal.Reader) {
+func ProcessPgDump(conv *Conv, r *internal.Reader) error {
 	for {
 		startLine := r.LineNumber
 		startOffset := r.Offset
-		b, stmts := readAndParseChunk(conv, r)
+		b, stmts, err := readAndParseChunk(conv, r)
+		if err != nil {
```
nevinheintze: This change means that if there is any error during parsing, we'd return and immediately exit (with the changes in main.go). In other words, we'd drop any chunks of pg_dump data that we'd successfully parsed.

It might be better to proceed with processing partial data (since that might be useful), and only exit if we didn't parse anything (i.e. we have the empty schema, which is basically useless).

internal/postgres/process.go
hengfengli: Thanks for pointing out. I changed it back to use `len(l)` for the error info. 

One question I have with this method is that e.g., use the following example: 

```sql
this is not a parsable file;

CREATE TABLE public.testusers (
    id integer NOT NULL,
    active boolean,
    name text
);

ALTER TABLE public.testusers OWNER TO postgres;

COPY public.testusers (id, active, name) FROM stdin;
1	t	foo
2	f	bar
\.

ALTER TABLE ONLY public.testusers
    ADD CONSTRAINT testusers_pkey PRIMARY KEY (id);
```

The first line is not a valid statement, but in this case, `l` will still try to include more statements and keep passing it to `pg_query.Parse(string(s))` until containing the content of the entire file. If the entire file is in GB size, we got an error at first place and can we fail it quickly? 

I understand that there might be cases the statement itself contains `;`. I just wonder if there is a better solution for this.

internal/postgres/process.go
hengfengli: According to `readAndParseChunk`, if an error happens, that means it has already tried to parse the content from where the error happens up to the end of the file, but failed to do so. 

From my understanding, the workflow looks like this: 

```bash

=== first pass begins ===

sql file: 
valid statement #1 (successfully parsed)
invalid statement #2 (error happens)
valid statement #3 (unable to parse this statement)

=== first pass ends ===

=== generate schema file ===

=== create database ===

=== second pass begins === (read data statements and insert/update rows)

=== second pass ends ===

```

In this case, the schema file is not empty, which has the valid statement \# 1 but not \# 2 and \# 3. Do we still need to create the database? and how about the second pass (read data and insert/update rows)?

Please forgive me if I misunderstood some info.

internal/postgres/process.go
nevinheintze: Yes, that's correct: if we hit an error, we'll keep re-trying.
And, as you say, if the error is on the first line of the file, we'll eventually try to parse the entire file. If its big, that won't end well -- we can probably only parse files up to around a few hundred MB because of the overheads of the pg AST.

There are a couple of options:
a) we could put a limit on the amount of the file we try to parse: say bail out if we're trying to parse more than a few MB.
b) try to do something more intelligent for recognizing end-of-statement: this is hard, and potentially very fragile.
c) don't use pg_dump output and instead use more direct introspection into the PostgreSQL DB.

Longer-term, I think option (c) makes the most sense. But happy to do a quick version of (a) in the short-term. I don't think (b) is a great option.

internal/postgres/process.go
nevinheintze: Yes, that's correct. We'll process up to the first error, and ignore the rest of the input, both schema and data.

My original thinking was that we should try to make as much progress as possible i.e. that even partial conversion was still useful. But maybe that's not worth doing, since there is a potential cost of increased complexity of behaviour: if you hit a parse error, and some tables were missing, wouldn't that be confusing? Perhaps its simpler to just unambiguously fail. What do you think?

hengfengli: I agree that it's simpler to fail quickly. An incomplete schema does not make sense to continue the migration. 

Still use my previous example: 

```sql
CREATE TABLE public.testusers (
    id integer NOT NULL,
    active boolean,
    name text
);

ALTER TABLE public.testusers OWNER TO postgres;

# Error happens here!!!
COPY public.testusers (id, active, name) FROM stdin;
1	t	foo
2	f	bar
\.

ALTER TABLE ONLY public.testusers
    ADD CONSTRAINT testusers_pkey PRIMARY KEY (id);
```

In the above example, if we only have `CREATE TABLE` (the first statement) in our schema file without `ALTER TABLE` (the last one), this schema file is invalid. The table we create would be wrong without a primary key. In this case, we should not try to create the database and insert data.

PR #45, opened 2020-02-14
fix: update to exit early when the file is not parsable.

hengfengli:
This is a revert for #44. The new added test is broken due to the last change, which has been fixed.

### DIFF
--- a/internal/postgres/process.go
+++ b/internal/postgres/process.go
@@ -31,11 +31,14 @@ import (
 // In schema mode, ProcessPgDump incrementally builds a schema (updating conv).
 // In data mode, ProcessPgDump uses this schema to convert PostgreSQL data
 // and writes it to Spanner, using the data sink specified in conv.
-func ProcessPgDump(conv *Conv, r *internal.Reader) {
+func ProcessPgDump(conv *Conv, r *internal.Reader) error {
 	for {
 		startLine := r.LineNumber
 		startOffset := r.Offset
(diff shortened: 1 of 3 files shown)


hengfengli:
@nevinheintze We should configure running unit tests in PR checks. PR must pass unit tests before merging. 

Sorry for the inconvenience. I saw a `green` "squash and merge" button soI thought the unit tests are passed.

lgruen:
Good idea, but we'd need to find a way to trigger the "ci/circleci:
build_and_test" check before the merge, but after approval. We don't want
to run it for any PR as that would risk leaking credentials.

hengfengli:
@lgruen I think we can have two CircleCi jobs, i.e., unit tests and integration tests (or builds). We only run unit tests for forked PRs (a PR from a forked repo). When you enable CI for forked PRs, it should provide an option “Pass secrets to builds from forked pull requests”, which is disabled by default. So the environment variables are not visible in the unit tests in forked PRs. 

When the forked PR is approved and merged with the master, this becomes a push to the master repo (not a forked PR) so it will trigger both unit tests and integration tests and the secrets will be visible to them. 

Please see the example in the reference. 

### Reference
* https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/

hengfengli:
I set up a small demo: https://github.com/hengfengli/test-circleci. It works well.

PR #46, opened 2020-02-18
test: add the first integration test

hengfengli:
I made the following changes: 

* Add a simple integration test
    * If `HARBOURBRIDGE_TESTS_GCLOUD_PROJECT_ID` and `HARBOURBRIDGE_TESTS_GCLOUD_INSTANCE_ID` are not set, then the integration tests will be skipped. 
    * Or, `-short` flag will also trigger to skip integration tests. 
* Refactor the `main.go`
    * Transform to `main` to a seperate method for testing purpose (I don't want to make too much refactoring in this PR). 
    * Use a struct to pass IO streams to take input and output. 
    * Move `init` and `usage` to the top. I think this is a convention in go. 

- [X] Tests pass


hengfengli:
@lgruen Instead of using the `+build` build comment, I follow the same approach in `google-cloud-go`. Basically, if `HARBOURBRIDGE_TESTS_GCLOUD_PROJECT_ID` and `HARBOURBRIDGE_TESTS_GCLOUD_INSTANCE_ID` are not set, integration tests will be skipped. 

So we don't need to make any circleci config changes: 

1) `Forked PRs`: no secrets will be passed to them, so integration tests will be skipped. 
2) `A push/commit to master`: secrets will be available, so integration tests will run.

main.go
```diff
@@ -52,80 +52,123 @@ var (
...
 	instanceOverride string
+	projectOverride  string
 	filePrefix       = ""
 	verbose          bool
 )
 
+func init() {
+	flag.StringVar(&dbNameOverride, "dbname", "", "dbname: name to use for Spanner db")
+	flag.StringVar(&instanceOverride, "instance", "", "instance: Spanner instance to use")
+	flag.StringVar(&projectOverride, "project", "", "project: Google Cloud project to use")
```
nevinheintze: Why do we need projectOverride?
Can we use existing GCLOUD_PROJECT environment variable?

Background: there was a conscious push to limit the number of flags for HarbourBridge, especially when there were standard environment variables for the same thing.

main.go
```diff
@@ -52,80 +52,123 @@ var (
...
 		panic(fmt.Errorf("can't set up log file"))
 	}
 	defer close(lf)
-	project, err := getProject()
+
+	err = process(projectOverride, instanceOverride, dbNameOverride, nil, filePrefix)
 	if err != nil {
-		fmt.Printf("\nCan't get project: %v\n", err)
-		panic(fmt.Errorf("can't get project"))
+		panic(fmt.Errorf("failed to run the job: %v", err))
```
nevinheintze: Not sure "failed to run the job" is the right message here e.g. not clear what "job" means in this context.
Maybe we just use the err returned and don't wrap it in anything else?

main.go
```diff
@@ -52,80 +52,123 @@ var (
...
 	if err != nil {
 		fmt.Printf("\nCan't set up log file: %v\n", err)
 		panic(fmt.Errorf("can't set up log file"))
 	}
 	defer close(lf)
-	project, err := getProject()
+
+	err = process(projectOverride, instanceOverride, dbNameOverride, nil, filePrefix)
 	if err != nil {
-		fmt.Printf("\nCan't get project: %v\n", err)
```
nevinheintze: I agree that doing the printf and then the panic looks kind of redundant, but it helps generate a readable error message.

Currently, this prints something like:

    Generating schema: 100%
    Wrote schema to file 'pg_dump_2020-02-19_817a-c14b.schema.txt'.
    Creating new database pg_dump_2020-02-19_817a-c14b in instance e2e-spanner-testing with default permissions ... 
    Can't create database: can't build CreateDatabaseRequest: rpc error: code = Unauthenticated desc = transport: oauth2: cannot fetch token: 400 Bad Request
    Response: {
      "error": "invalid_grant",
      "error_description": "reauth related error (invalid_rapt)",
      "error_subtype": "invalid_rapt"
    }.
    
    Possible cause: credentials are mis-configured. Do you need to run
    
      gcloud auth application-default login
    
    or configure environment variable GOOGLE_APPLICATION_CREDENTIALS.
    See https://cloud.google.com/docs/authentication/getting-started.
    
    panic: can't create database

    goroutine 1 [running]:
    main.main()
    	/usr/local/google/home/nch/cloud-spanner/postgres-to-spanner/harbourbridge/main.go:102 +0xd73
    exit status 2

But switching to just the panic produces:
    
    Generating schema: 100%
    Wrote schema to file 'pg_dump_2020-02-19_6248-fecd.schema.txt'.
    Creating new database pg_dump_2020-02-19_6248-fecd in instance e2e-spanner-testing with default permissions ... panic: can't create database: can't build CreateDatabaseRequest: rpc error: code = Unauthenticated desc = transport: oauth2: cannot fetch token: 400 Bad Request
    Response: {
      "error": "invalid_grant",
      "error_description": "reauth related error (invalid_rapt)",
      "error_subtype": "invalid_rapt"
    }.
    
    Possible cause: credentials are mis-configured. Do you need to run
    
      gcloud auth application-default login
    
    or configure environment variable GOOGLE_APPLICATION_CREDENTIALS.
    See https://cloud.google.com/docs/authentication/getting-started.
    
    
    goroutine 1 [running]:
    main.main()
    	/usr/local/google/home/nch/cloud-spanner/postgres-to-spanner/harbourbridge/main.go:103 +0xd26
    exit status 2

which is harder to read and figure out what went wrong.

What we should really do is use log.Fatal, but because of the noise of the client library (it prints out spammy logs), we probably don't want to switch to that right now.

main.go
```diff
@@ -56,76 +56,128 @@ var (
...
 	internal.VerboseInit(verbose)
-	now := time.Now()
 	lf, err := setupLogFile()
 	if err != nil {
 		fmt.Printf("\nCan't set up log file: %v\n", err)
 		panic(fmt.Errorf("can't set up log file"))
 	}
 	defer close(lf)
+
+	// Get project ID
```
nevinheintze: Do you think this is needed?
In general I thought go discouraged comments that are fairly clear from context.
If you think emphasizing the ID is important, how about renaming getProject to getProjectID?

main.go
```diff
@@ -56,76 +56,128 @@ var (
...
 	bw := secondPass(f, client, conv, rows)
-	report(bw, n, now, db, conv)
+	report(bw, n, now, db, conv, ouputFilePrefix+reportFile, ouputFilePrefix+badDataFile)
+	return nil
 }
 
-func report(bw *spanner.BatchWriter, bytesRead int64, now time.Time, db string, conv *postgres.Conv) {
-	fileName := filePrefix + reportFile
-	f, err := os.Create(fileName)
+func report(bw *spanner.BatchWriter, bytesRead int64, now time.Time, db string, conv *postgres.Conv, reportFileName string, badDataFileName string) {
```
nevinheintze: s/reportFileName/fileName/
since report is redundant here.

main.go
```diff
@@ -56,76 +56,128 @@ var (
...
 	_, err = f.Seek(0, 0)
 	if err != nil {
 		fmt.Printf("\nCan't seek to start of file (preparation for second pass): %v\n", err)
-		panic(fmt.Errorf("can't seek to start of file"))
+		return fmt.Errorf("can't seek to start of file")
 	}
 	rows := conv.Rows()
 	bw := secondPass(f, client, conv, rows)
-	report(bw, n, now, db, conv)
+	report(bw, n, now, db, conv, ouputFilePrefix+reportFile, ouputFilePrefix+badDataFile)
```
nevinheintze: Just checking: your motivation for passing these parameters is so that we can later refactor process into a separate module, and the parameters will capture everything we need from main.

integration_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+func TestMain(m *testing.M) {
+	cleanup := initIntegrationTests()
+	res := m.Run()
+	cleanup()
+	os.Exit(res)
+}
+
+func initIntegrationTests() (cleanup func()) {
+	testProjectID = os.Getenv("HARBOURBRIDGE_TESTS_GCLOUD_PROJECT_ID")
```
nevinheintze: s/testProjectID/projectID/
I think test is redundant here.
Same for testInstanceID.

main.go
hengfengli: I agree that the comment is not necessary, but I feel that some comments can help developers to scan code quickly when the code is long. In other words, dividing the long code into different sections by using comments can improve readability. This is just my personal thought. 

For `project` or `projectID`, my original concern actually is with `instance` because we have an import alias `instance` on the top. So at the beginning, I changed all `project` and `instance` to `projectID` and `instanceID`. After that, I feel `project` and `instance` are less verbose so I changed them back. 

In addition, `getProject()` gives me a feeling that it should return a project instance. It is not very clear about what it returns from the function name.

integration_test.go
```diff
@@ -0,0 +1,284 @@
...
+	dbName, _ := getDatabaseName(now)
+	dbPath := fmt.Sprintf("projects/%s/instances/%s/databases/%s", testProjectID, testInstanceID, dbName)
+
+	dataFilepath := "test_data/pg_dump.test.out"
+	filePrefix = dbName + "."
+	// Be aware that when testing with the command, the time `now` might be
+	// different between file prefixes and the contents in the files. This
+	// is because file prefixes use `now` from here (the test function) and
+	// the generated time in the files uses a `now` inside the command, which
+	// can be different.
```
nevinheintze: So the test will still pass (since we don't check the contents of the files), but the cleanup may fail?

hengfengli: The cleanup will not fail because we explicitly pass the file prefix as a command parameter here. We can cleanup files according to the given file prefix. 

In my current version, I created a temporary directory so I can delete the directory easily. However, the main purpose of temporary directory is to solve a possible conflict in parallel tests. 

If we want to write some tests to check contents based on the time `now`, we should be aware of this issue that we can't pass a time via command line.

main.go
hengfengli: Yes. Later, when we move `process` into a seperate module, adding these parameters makes this method to be more testable (relying on function parameters instead of global variables).

PR #52, opened 2020-03-16
refactor: refactor the main.go

hengfengli:
Basically, I refactor to have a main entrance `toSpanner()`, which includes the following abstract steps: 

1. Run schema conversion (`schemaConv()`)
2. Create spanner database
3. Run data conversion (`dataConv()`)
4. Generate report

Schema & data conversion will be routed to different methods based on the driver name. Currently, we support two drivers: 

* pgdump
* posgres

For example, `schemaConv` will choose different methods to run according to the driver name. 

* posgres: schemaConvFromSourceDB()
* pgdump: schemaConvFromFileStream()

schemaConvFromSourceDB() will intialize different driver config according to the driver name. E.g, if the driver name is "posgres", the program will call `pgDriverConfig()` for make a db connection. In the future, we only need to add a `mysqlDriverConfig()` for mysql extension. 

We can do the same thing for `dataConv`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -48,6 +48,13 @@ import (
 	"github.com/cloudspannerecosystem/harbourbridge/spanner/ddl"
 )
 
+const (
+	// PGDUMP is the driver name for pg_dump.
+	PGDUMP string = "pgdump"
+	// POSTGRES is the driver name for postgresSQL.
+	POSTGRES string = "posgres"
+)
+
 var (
 	badDataFile      = "dropped.txt"
 	schemaFile       = "schema.txt"
@@ -89,6 +96,7 @@ func main() {
 	}
 	defer close(lf)
 
+	ioHelper := &ioStreams{in: os.Stdin, out: os.Stdout}
 	project, err := getProject()
 	if err != nil {
 		fmt.Printf("\nCan't get project: %v\n", err)
@@ -98,14 +106,14 @@ func main() {
 
 	instance := instanceOverride
 	if instance == "" {
-		instance, err = getInstance(project)
+		instance, err = getInstance(project, ioHelper.out)
 		if err != nil {
 			fmt.Printf("\nCan't get instance: %v\n", err)
 			panic(fmt.Errorf("can't get instance"))
 		}
 	}
 	fmt.Printf("Using Spanner instance: %s\n", instance)
-	printPermissionsWarning()
+	printPermissionsWarning(ioHelper.out)
 
 	now := time.Now()
 	dbName := dbNameOverride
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -48,6 +48,13 @@ import (
 	"github.com/cloudspannerecosystem/harbourbridge/spanner/ddl"
 )
 
+const (
+	// PGDUMP is the driver name for pg_dump.
+	PGDUMP string = "pgdump"
+	// POSTGRES is the driver name for postgresSQL.
```
nevinheintze: PostgreSQL

main.go
```diff
@@ -124,118 +132,174 @@ func main() {
...
+	banner := getBanner(now, db)
+	report(bw.DroppedRowsByTable(), 0, banner, conv, outputFilePrefix+reportFile, ioHelper.out)
+	writeBadData(bw, conv, banner, outputFilePrefix+badDataFile, ioHelper.out)
+	return nil
+}
+
+func schemaConv(driver string, ioHelper *ioStreams) (*internal.Conv, error) {
+	switch driver {
+	case POSTGRES:
+		return schemaConvFromSourceDB(driver)
```
nevinheintze: This is kind of a long name for go.
How about just schemaFromSQL?
And schemaFromPgDump instead of schemaConvFromFileStream.

main.go
```diff
@@ -124,118 +132,174 @@ func main() {
...
+		return schemaConvFromFileStream(ioHelper)
+	default:
+		return nil, fmt.Errorf("schema conversion for driver %s not supported", driver)
+	}
+}
+
+func dataConv(driver string, ioHelper *ioStreams, client *sp.Client, conv *internal.Conv) (*spanner.BatchWriter, error) {
+	switch driver {
+	case POSTGRES:
+		return dataConvFromSourceDB(ioHelper, client, conv)
```
nevinheintze: Same here: how about dataFromSQL and dataFromPgDump?

main.go
```diff
@@ -124,118 +132,174 @@ func main() {
...
 	}
-	writeSchemaFile(conv, now, outputFilePrefix+schemaFile)
-	spannerDB, err := createDatabase(projectID, instanceID, dbName, conv)
-	if err != nil {
-		fmt.Printf("\nCan't create database: %v\n", err)
-		return fmt.Errorf("can't create database")
+	return conv, nil
+}
+
+func dataConvFromSourceDB(ioHelper *ioStreams, client *sp.Client, conv *internal.Conv) (*spanner.BatchWriter, error) {
```
nevinheintze: I don't think the ioHelper would ever be needed here?
For data conversion, we only use ioHelper.seekableIn, and that should only happen when we're reading pgdump data i.e. dataConvFromFileStream.

main.go
```diff
@@ -124,118 +132,174 @@ func main() {
...
+	config := spanner.BatchWriterConfig{
```
nevinheintze: The core parts of config are common between here and secondPass.

How about we factoring out the common part (first four components).
Maybe adding the following to the start of dataConv:

config := spanner.BatchWriterConfig{
		BytesLimit: 100 * 1000 * 1000,
		WriteLimit: 40,
		RetryLimit: 1000,
		Verbose:    internal.Verbose()}

and then passing config to dataConvFromSourceDB and dataConvFromFileStream.
They can then do:

config.Write = ...
spanner.NewBatchWriter(config)

main.go
```diff
@@ -124,118 +132,174 @@ func main() {
...
-		return fmt.Errorf("can't get seekable input file")
+		printSeekError(err, ioHelper.out)
+		return nil, fmt.Errorf("can't get seekable input file")
 	}
-	defer f.Close()
+	ioHelper.seekableIn = f
+	ioHelper.bytesRead = n
 	conv := internal.MakeConv()
 
 	err = firstPass(f, n, conv)
```
nevinheintze: How about just inlining firstPass?
The concept of 'first pass' has outlived its usefulness.
And the split between these two functions is pretty arbitrary and both functions are small.

main.go
```diff
@@ -124,118 +132,174 @@ func main() {
...
-	client, err := getClient(db)
-	if err != nil {
-		fmt.Printf("\nCan't create client for db %s: %v\n", db, err)
-		return fmt.Errorf("can't create Spanner client")
-	}
-	_, err = f.Seek(0, 0)
+	return conv, nil
+}
+
+func dataConvFromFileStream(ioHelper *ioStreams, client *sp.Client, conv *internal.Conv) (*spanner.BatchWriter, error) {
```
nevinheintze: Same here: just inline secondPass I think.

main.go
```diff
@@ -124,118 +132,174 @@ func main() {
...
 	}
 	rows := conv.Rows()
-	bw := secondPass(f, client, conv, rows)
-	banner := getBanner(now, db)
-	// TODO(hengfeng): When we refactor `process` into a separate module, and
-	// the parameters will capture everything we need from main.
-	report(bw.DroppedRowsByTable(), n, banner, conv, outputFilePrefix+reportFile)
-	writeBadData(bw, conv, banner, outputFilePrefix+badDataFile)
-	return nil
+	return secondPass(ioHelper.seekableIn, client, conv, rows), nil
```
nevinheintze: Same here: just inline secondPass I think.

PR #56, opened 2020-04-06
Fix pg_dump backslashes in IMPORT-FROM and fix several array issues.

nevinheintze:
Includes:
- fix for escape sequences in IMPORT-FROM statements (double-backslash issue)
- adds handling of quoted array values
- adds handling of empty arrays
- adds handling of NULL in arrays

### DIFF
--- a/internal/data_test.go
+++ b/internal/data_test.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/civil"
+	"cloud.google.com/go/spanner"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/cloudspannerecosystem/harbourbridge/schema"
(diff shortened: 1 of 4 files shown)


internal/data_test.go
```diff
@@ -69,21 +70,44 @@ func TestConvertData(t *testing.T) {
 		e       interface{} // Expected result.
 	}{
 		{"bool", ddl.Bool{}, false, "", "true", true},
-		{"bytes", ddl.Bytes{Len: ddl.MaxLength{}}, false, "", `\\x0001beef`, []byte{0x0, 0x1, 0xbe, 0xef}},
+		{"bytes", ddl.Bytes{Len: ddl.MaxLength{}}, false, "", `\x0001beef`, []byte{0x0, 0x1, 0xbe, 0xef}},
```
hengfengli, line 73: Should we also change the return value of scalar types to spanner.Null*? Because now they are inconsistent, e.g., convert a single bool value (bool type in Go) and an array of bool values (an array of spanner.NullBool).

internal/pgdump_test.go
```diff
@@ -431,14 +432,21 @@ COPY test (id, a, b, c, d, e) FROM stdin;
 				spannerData{
 					table: "test", cols: []string{"int8", "float8", "bool", "timestamp", "date", "bytea", "arr", "synth_id"},
 					vals: []interface{}{int64(7), float64(42.1), true, getTime(t, "2019-10-29T05:30:00Z"),
-						getDate("2019-10-29"), []byte{0x0, 0x1, 0xbe, 0xef}, []int64{42, 6}, bitReverse(0)}},
+						getDate("2019-10-29"), []byte{0x0, 0x1, 0xbe, 0xef},
+						[]spanner.NullInt64{spanner.NullInt64{Int64: 42, Valid: true}, spanner.NullInt64{Int64: 6, Valid: true}},
```
hengfengli, line 436: We can change this to: 

```
[]spanner.NullInt64{ {42, true}, {6, true} }
```

internal/pgdump_test.go
```diff
@@ -431,14 +432,21 @@ COPY test (id, a, b, c, d, e) FROM stdin;
...
 				spannerData{table: "test", cols: []string{"int8", "synth_id"}, vals: []interface{}{int64(7), bitReverse(1)}},
 				spannerData{table: "test", cols: []string{"float8", "synth_id"}, vals: []interface{}{float64(42.1), bitReverse(2)}},
 				spannerData{table: "test", cols: []string{"bool", "synth_id"}, vals: []interface{}{true, bitReverse(3)}},
 				spannerData{table: "test", cols: []string{"timestamp", "synth_id"}, vals: []interface{}{getTime(t, "2019-10-29T05:30:00Z"), bitReverse(4)}},
 				spannerData{table: "test", cols: []string{"date", "synth_id"}, vals: []interface{}{getDate("2019-10-29"), bitReverse(5)}},
 				spannerData{table: "test", cols: []string{"bytea", "synth_id"}, vals: []interface{}{[]byte{0x0, 0x1, 0xbe, 0xef}, bitReverse(6)}},
-				spannerData{table: "test", cols: []string{"arr", "synth_id"}, vals: []interface{}{[]int64{42, 6}, bitReverse(7)}},
+				spannerData{table: "test", cols: []string{"arr", "synth_id"},
+					vals: []interface{}{
+						[]spanner.NullInt64{
```
hengfengli, line 446: same here.

internal/data_test.go
nevinheintze: So for scalars we already handle the case of NULL in ConvertData: NULL columns are represented by pgdump using "\\n" which we handle in the main for loop (line 89). Basically the for loop builds up both a list of columns and values, and simply skips columns if the data is NULL.

Array, on the other hand, are different: if you have an array value that is NULL, you can't just skip it!

Having said that, I might need to revisit this when dealing with data returned from select statements. But let's leave leave as is for now.

internal/pgdump_test.go
nevinheintze: Done.

PR #57, opened 2020-04-08
Add COPY-FROM test with renamed table/cols.

nevinheintze:
Add pgdump test for when we ingest COPY-FROM data and the table and col
need to be renamed.

### DIFF
--- a/internal/pgdump_test.go
+++ b/internal/pgdump_test.go
@@ -216,6 +216,18 @@ func TestProcessPgDump(t *testing.T) {
 				spannerData{table: "test", cols: []string{"a", "b", "n"}, vals: []interface{}{"a22", "b99", int64(6)}}},
 		},
 		{
+			name: "COPY FROM with renamed table/cols",
+			input: "CREATE TABLE _test (_a text, b text, n bigint);\n" +
+				"ALTER TABLE ONLY _test ADD CONSTRAINT test_pkey PRIMARY KEY (_a, b);" +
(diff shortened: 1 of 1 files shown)


internal/pgdump_test.go
```diff
@@ -215,6 +215,18 @@ func TestProcessPgDump(t *testing.T) {
...
+		{
+			name: "COPY FROM with renamed table/cols",
+			input: "CREATE TABLE _test (_a text, b text, n bigint);\n" +
+				"ALTER TABLE ONLY _test ADD CONSTRAINT test_pkey PRIMARY KEY (_a, b);" +
+				"COPY public._test (_a, b, n) FROM stdin;\n" +
+				"a1	b1	42\n" +
+				"a22	b99	6\n" +
+				"\\.\n",
+			expectedData: []spannerData{
+				spannerData{table: "Atest", cols: []string{"Aa", "b", "n"}, vals: []interface{}{"a1", "b1", int64(42)}},
```
hengfengli, line 227: Just curious: how do we decide to rename `_` to `A`?

nevinheintze: This is done by internals/fixname.go (https://github.com/cloudspannerecosystem/harbourbridge/blob/master/internal/fixname.go).

Basically it maps table/column/index names into something Spanner will accept (which ust start with {a-z|A-Z}). Collisions are handled by adding a postfix.

PR #58, opened 2020-04-20
Complete information-schema mode: add data processing.

nevinheintze:
In information-schema mode we use SQL queries to get schema
information and data from a source database. A previous commit had
added schema processing (reading information_schema tables). This
commit adds the data processing part (reading data from tables using
"SELECT *" queries).

### DIFF
--- a/internal/convert.go
+++ b/internal/convert.go
@@ -179,6 +179,19 @@ func (conv *Conv) GetDDL(c ddl.Config) []string {
 	return ddl
 }
 
+// WriteRow calls dataSink and updates row stats.
+func (conv *Conv) WriteRow(srcTable, spTable string, spCols []string, spVals []interface{}) {
+	if conv.dataSink == nil {
+		msg := "Internal error: ProcessDataRow called but dataSink not configured"
+		VerbosePrintf("%s\n", msg)
+		conv.unexpected(msg)
+		conv.statsAddBadRow(srcTable, conv.dataMode())
+	} else {
+		conv.dataSink(spTable, spCols, spVals)
+		conv.statsAddGoodRow(srcTable, conv.dataMode())
+	}
+}
+
 // Rows returns the total count of data rows processed.
 func (conv *Conv) Rows() int64 {
 	n := int64(0)
@@ -220,6 +233,18 @@ func (conv *Conv) StatementErrors() int64 {
 // encountered during processing.
 func (conv *Conv) Unexpecteds() int64 {
 	return int64(len(conv.stats.unexpected))
+}
+
+// CollectBadRows updates the list of bad rows, while respecting
+// the byte limit for bad rows.
+func (conv *Conv) CollectBadRow(srcTable string, srcCols, vals []string) {
+	r := &row{table: srcTable, cols: srcCols, vals: vals}
+	bytes := byteSize(r)
+	// Cap storage used by badRows. Keep at least one bad row.
+	if len(conv.sampleBadRows.rows) == 0 || bytes+conv.sampleBadRows.bytes < conv.sampleBadRows.bytesLimit {
+		conv.sampleBadRows.rows = append(conv.sampleBadRows.rows, r)
+		conv.sampleBadRows.bytes += bytes
+	}
 }
 
 // SampleBadRows returns a string-formatted list of rows that generated errors.
@@ -300,6 +325,10 @@ func (conv *Conv) statsAddRow(srcTable string, b bool) {
 	}
 }
 
+func (conv *Conv) statsAddRows(srcTable string, count int64) {
+	conv.stats.rows[srcTable] += count
+}
+
 // statsAddGoodRow increments the good-row stats for 'srcTable' if b
 // is true.  See statsAddRow comments for context.
 func (conv *Conv) statsAddGoodRow(srcTable string, b bool) {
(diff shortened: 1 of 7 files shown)


internal/convert.go
```diff
@@ -179,6 +179,19 @@ func (conv *Conv) GetDDL(c ddl.Config) []string {
 	return ddl
 }
 
+// WriteRow calls dataSink and updates row stats.
```
nevinheintze, line 182: WriteRow and CollectBadRows are chunks of code moved from data.go so they can be shared with infoschema.go.

internal/data_test.go
```diff
@@ -283,9 +283,9 @@ func buildConv(spTable ddl.CreateTable, srcTable schema.Table) *Conv {
 	conv.srcSchema[srcTable.Name] = srcTable
 	conv.toSource[spTable.Name] = nameAndCols{name: srcTable.Name, cols: make(map[string]string)}
 	conv.toSpanner[srcTable.Name] = nameAndCols{name: spTable.Name, cols: make(map[string]string)}
-	for _, c := range spTable.ColNames {
-		conv.toSource[spTable.Name].cols[c] = c
-		conv.toSpanner[srcTable.Name].cols[c] = c
+	for i := range spTable.ColNames {
```
nevinheintze, line 286: Extends buildConv so that it can handle renamed columns.

internal/infoschema_test.go
```diff
@@ -153,4 +139,237 @@ func TestProcessInfoSchema(t *testing.T) {
...
+	db := mkMockDB(t, ms)
+	conv := MakeConv()
+	conv.SetDataMode()
+	SetRowStats(conv, db)
+	assert.Equal(t, int64(5), conv.stats.rows["test1"])
+	assert.Equal(t, int64(142), conv.stats.rows["test2"])
+	assert.Equal(t, int64(0), conv.Unexpecteds())
+}
+
+type mockSpec struct {
```
hengfengli, line 352: Maybe consider to put this on the top for a better reading flow.

internal/infoschema_test.go
```diff
@@ -153,4 +139,237 @@ func TestProcessInfoSchema(t *testing.T) {
...
+			Name:     tableName,
+			ColNames: []string{col},
+			ColDefs:  map[string]ddl.ColumnDef{col: ddl.ColumnDef{Name: col, T: tc.spType, IsArray: tc.isArray}}}
+		ac, av, err := ConvertSqlRow(conv, tableName, cols, srcSchema, tableName, cols, spSchema, []interface{}{tc.in})
+		assert.Equal(t, cols, ac)
+		assert.Equal(t, []interface{}{tc.e}, av)
+		assert.Nil(t, err)
+	}
+
+	// End-to-end test: builds schema from information_schema
```
hengfengli, line 270: Nits: Should we split this into a separate test?

internal/infoschema.go
```diff
@@ -48,6 +53,150 @@ func ProcessInfoSchema(conv *Conv, db *sql.DB) error {
...
+		v, iv := buildVals(len(srcCols))
+		for rows.Next() {
+			err := rows.Scan(iv...)
+			if err != nil {
+				conv.unexpected(fmt.Sprintf("Couldn't process sql data row: %s", err))
+				// Scan failed, so we don't have any data to add to bad rows.
+				conv.statsAddBadRow(srcTable, conv.dataMode())
+				continue
+			}
+			cvtCols, cvtVals, err := ConvertSqlRow(conv, srcTable, srcCols, srcSchema, spTable, spCols, spSchema, v)
```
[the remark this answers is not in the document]
nevinheintze: This is discussed in the documentation for ConvertSqlRow.
In short, cvtCols is a subset of spCols: cols can change when we add a column (synthetic primary key) or because we drop columns (handling of NULL values).

internal/infoschema.go
```diff
@@ -48,6 +53,150 @@ func ProcessInfoSchema(conv *Conv, db *sql.DB) error {
...
+		if err != nil {
+			conv.unexpected(fmt.Sprintf("Couldn't get data for table: %s", err))
+			continue
+		}
+		defer rows.Close()
+		srcTable := buildTableName(t.schema, t.name)
+		srcCols, err1 := rows.Columns()
+		spTable, err2 := GetSpannerTable(conv, srcTable)
+		spCols, err3 := GetSpannerCols(conv, srcTable, srcCols)
+		spSchema, ok1 := conv.spSchema[spTable]
```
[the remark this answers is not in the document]
nevinheintze: These should be safe because:
a) spTable is a string, so if GetSpannerTable fails, it returns empty string. We will look up an empty string in conv.spSchema which will return not-found i.e. ok1 will be false.
b) if rows.Columns fails, srcCols will be the empty slice; GetSpannerCols will iterate over this and return an empty spCols.

internal/infoschema_test.go
nevinheintze: Thanks - done!

internal/infoschema.go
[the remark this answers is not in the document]
nevinheintze: Based on our vc, we think we're good here. Resolving the thread.

internal/infoschema.go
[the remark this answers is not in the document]
nevinheintze: The other option would be to pull the synthetic-primary key stuff out of CollectBadRows, and handle NULL values using nil.
But ProcessSqlData is already a large function.
Let's leave this till we address the bigger refactoring issue i.e. using a transaction for all of the schema and data queries.

internal/infoschema_test.go
nevinheintze: Done.

PR #60, opened 2020-05-30
Postgres - common changes

Dharmil-searce:
Refactored convert.go to make functions public which would be useful to make a separate postgres package

### DIFF
--- a/internal/convert.go
+++ b/internal/convert.go
@@ -16,12 +16,8 @@ package internal
 
 import (
 	"fmt"
-	"reflect"
 	"sort"
-	"strings"
 	"time"
-
-	nodes "github.com/lfittl/pg_query_go/nodes"
 
 	"github.com/cloudspannerecosystem/harbourbridge/schema"
 	"github.com/cloudspannerecosystem/harbourbridge/spanner/ddl"
@@ -30,16 +26,17 @@ import (
 // Conv contains all schema and data conversion state.
 type Conv struct {
 	mode           mode                                // Schema mode or data mode.
-	spSchema       map[string]ddl.CreateTable          // Maps Spanner table name to Spanner schema.
-	syntheticPKeys map[string]syntheticPKey            // Maps Spanner table name to synthetic primary key (if needed).
-	srcSchema      map[string]schema.Table             // Maps source-DB table name to schema information.
-	issues         map[string]map[string][]schemaIssue // Maps source-DB table/col to list of schema conversion issues.
-	toSpanner      map[string]nameAndCols              // Maps from source-DB table name to Spanner name and column mapping.
-	toSource       map[string]nameAndCols              // Maps from Spanner table name to source-DB table name and column mapping.
+	SpSchema       map[string]ddl.CreateTable          // Maps Spanner table name to Spanner schema.
+	SyntheticPKeys map[string]SyntheticPKey            // Maps Spanner table name to synthetic primary key (if needed).
+	SrcSchema      map[string]schema.Table             // Maps source-DB table name to schema information.
+	Issues         map[string]map[string][]SchemaIssue // Maps source-DB table/col to list of schema conversion issues.
+	ToSpanner      map[string]NameAndCols              // Maps from source-DB table name to Spanner name and column mapping.
+	ToSource       map[string]NameAndCols              // Maps from Spanner table name to source-DB table name and column mapping.
 	dataSink       func(table string, cols []string, values []interface{})
-	location       *time.Location // Timezone (for timestamp conversion).
+	Location       *time.Location // Timezone (for timestamp conversion).
 	sampleBadRows  rowSamples     // Rows that generated errors during conversion.
-	stats          stats
+	Stats          stats
+	TimezoneOffset string // Timezone offset for timestamp conversion.
 }
 
 type mode int
@@ -49,37 +46,37 @@ const (
 	dataOnly
 )
 
-// syntheticPKey specifies a synthetic primary key and current sequence
+// SyntheticPKey specifies a synthetic primary key and current sequence
 // count for a table, if needed. We use a synthetic primary key when
 // the source DB table has no primary key.
-type syntheticPKey struct {
-	col      string
-	sequence int64
+type SyntheticPKey struct {
+	Col      string
+	Sequence int64
 }
 
-type schemaIssue int
+type SchemaIssue int
 
 // Defines all of the schema issues we track. Includes issues
 // with type mappings, as well as features (such as source
 // DB constraints) that aren't supported in Spanner.
 const (
-	defaultValue schemaIssue = iota
-	foreignKey
-	missingPrimaryKey
-	multiDimensionalArray
-	noGoodType
-	numeric
-	numericThatFits
-	serial
-	timestamp
-	widened
+	DefaultValue SchemaIssue = iota
+	ForeignKey
+	MissingPrimaryKey
+	MultiDimensionalArray
+	NoGoodType
+	Numeric
+	NumericThatFits
+	Serial
+	Timestamp
+	Widened
 )
 
-// nameAndCols contains the name of a table and its columns.
+// NameAndCols contains the name of a table and its columns.
 // Used to map between source DB and Spanner table and column names.
-type nameAndCols struct {
-	name string
-	cols map[string]string
+type NameAndCols struct {
+	Name string
+	Cols map[string]string
 }
 
 type rowSamples struct {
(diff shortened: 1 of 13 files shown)


Dharmil-searce:
@googlebot I signed it!

hengfengli:
@Dharmil-searce Is this the first step in our planning doc? It seems that it includes the first two steps in this PR. Or maybe I am wrong.

Dharmil-searce:
@hengfengli Yes, it includes 2 steps from the doc. 1st commit changes are useful for moving Postgres in different package. Only 1st commit standalone is not useful and justified. That is the reason I have added 2 steps in this PR for 'postgres package'. If you suggest then we can also do different PR for each commit.

hengfengli:
I would prefer to split them into separate PRs so that each PR can become smaller and better for code reviews. Thanks for doing that.

Dharmil-searce:
@hengfengli Please check. I have removed 2nd commit and I will make separate PR for postgres package.

internal/convert.go
```diff
@@ -373,38 +419,66 @@ func (conv *Conv) getStatementStat(s string) *statementStat {
...
+func (conv *Conv) SchemaMode() bool {
 	return conv.mode == schemaOnly
 }
 
-func (conv *Conv) dataMode() bool {
+func (conv *Conv) DataMode() bool {
 	return conv.mode == dataOnly
 }
+
+func byteSize(r *row) int64 {
```
nevinheintze, line 475: Just noting that byteSize used to be defined in data.go and used in convert.go.
In short, this is a cleanup, and allows us to keep byteSize private.

internal/convert.go
```diff
@@ -201,6 +203,11 @@ func (conv *Conv) Rows() int64 {
 	return n
 }
 
+// GetRows return count of data rows for a srcTable
+func (conv *Conv) GetRows(srcTable string) int64 {
```
nevinheintze, line 207: In general, we've avoided adding accessors to convert.go, and instead have directly accessed the subfields of conv as needed. Note that there are functions like BadRows and Statements, but they all do some computation e.g. sum up bad rows etc.

For consistency I think we should continue down that path. That means we'd have to make stats public. Given that much of Conv is being made public anyway, that seems reasonable.

In other words, if anything in convert.go would need an accessor, let's just make it public and avoid the accessor. Same comment for the other GetNNN functions added below.

Hengfeng -- do you have any thoughts on this?

internal/convert.go
```diff
@@ -311,57 +342,72 @@ func (conv *Conv) unexpected(u string) {
...
 	if b {
 		conv.stats.rows[srcTable]++
 	}
 }
 
-func (conv *Conv) statsAddRows(srcTable string, count int64) {
+// StatsAddRows increments the count of rows for 'srcTable' by the count specified
+func (conv *Conv) StatsAddRows(srcTable string, count int64) {
 	conv.stats.rows[srcTable] += count
 }
```
nevinheintze, line 362: Suggest dropping this function and just inlining the access to conv.stats.rows (assuming we make stats public) where the function is used.

internal/convert.go
```diff
@@ -311,57 +342,72 @@ func (conv *Conv) unexpected(u string) {
...
 // statsAddGoodRow increments the good-row stats for 'srcTable' if b
-// is true.  See statsAddRow comments for context.
+// is true.  See StatsAddRow comments for context.
 func (conv *Conv) statsAddGoodRow(srcTable string, b bool) {
 	if b {
 		conv.stats.goodRows[srcTable]++
 	}
 }
 
-func (conv *Conv) statsAddGoodRows(srcTable string, count int64) {
```
nevinheintze, line 340: Same here.

internal/convert.go
```diff
@@ -373,38 +419,66 @@ func (conv *Conv) getStatementStat(s string) *statementStat {
 	return conv.stats.statement[s]
 }
 
-func (conv *Conv) skipStatement(l []nodes.Node) {
-	if conv.schemaMode() { // Record statement stats on first pass only.
+func (conv *Conv) SkipStatement(l []nodes.Node) {
```
nevinheintze, line 422: Yeah, the existing code here is kind of broken: we shouldn't reference nodes.Node in convert.go.
All of that should live in pgdump.go.

I think what should happen longer-term is that prNodes and prNodeType should be moved into pgdump.go.

Let's go with what you have for the moment, but could you add a TODO something like:

// TODO: Move prNodes to pgdump.go, remove all references to nodes.Node in convert.go,
// and just have one version of SkipStatement, ErrorInStatement, SchemaStatement and DataStatement.

internal/data.go
```diff
@@ -380,14 +380,3 @@ func processQuote(s string) (string, error) {
 	}
 	return s, nil
 }
-
-func byteSize(r *row) int64 {
-	n := int64(len(r.table))
```
nevinheintze, line 385: Just noting: this is now in convert.go.

internal/convert.go
```diff
@@ -373,38 +419,66 @@ func (conv *Conv) getStatementStat(s string) *statementStat {
...
+	if conv.SchemaMode() { // Record statement stats on first pass only.
 		s := prNodes(l)
 		VerbosePrintf("Skipping statement: %s\n", s)
 		conv.getStatementStat(s).skip++
 	}
 }
+func (conv *Conv) SkipStatementMySQL(stmtType string) {
```
hengfengli, line 429: Is it possible to merge `SkipStatement` and `SkipStatementMySQL`? What I mean is to move `s := prNodes(l)` out from this function and then `SkipStatement` should work for both `Postgres` and `Mysql`. Same for `ErrorInStatement ` and `ErrorInStatementMYSQL`. 

If this has been done in next PRs, then I am fine with this right now.

internal/convert.go
```diff
@@ -134,6 +135,7 @@ func MakeConv() *Conv {
 			statement:  make(map[string]*statementStat),
 			unexpected: make(map[string]int64),
 		},
+		Offset: "",
```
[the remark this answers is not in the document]
Dharmil-searce: MySQL timezone are just offsets, example "+05:30". The statement to set timezone in mysql is like `SET time_zone = "+9:50";` and we would not be getting a timezone ID as in postgres. Thus, we will capture the value of offset while parsing SET statement and use conv.Offset for timestamp conversion.

internal/convert.go
```diff
@@ -201,6 +203,11 @@ func (conv *Conv) Rows() int64 {
 	return n
 }
 
+// GetRows return count of data rows for a srcTable
```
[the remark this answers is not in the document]
Dharmil-searce: Removed GetX() function in a new commit

internal/convert.go
Dharmil-searce: I had created new functions with MySQL suffix because using one version needed changes in pgdump.go.
Sure, now you can check in new commit that I have moved prNodes and prNodeType to pgdump.go and created one version for SkipStatement, ErrorInStatement, SchemaStatement and DataStatement functions.

internal/convert.go
Dharmil-searce: Agreed. Now that we are making Stats as public, functions except the one's using bool parameter are not required. I have made changes accordingly.

internal/convert.go
```diff
@@ -224,17 +231,36 @@ func (conv *Conv) Statements() int64 {
...
 	return int64(len(conv.stats.unexpected))
 }
 
+// GetUnexpected returns all count of unexpected conditions,
+// broken down by condition description
+func (conv *Conv) GetUnexpected() map[string]int64 {
+	return conv.stats.unexpected
+}
+
+func (conv *Conv) StatsAddUnexpecteds(u string, count int64) {
```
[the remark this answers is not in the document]
Dharmil-searce: I removed this function as we have made Unexpected public now.

internal/convert.go
```diff
@@ -263,34 +289,39 @@ func (conv *Conv) SampleBadRows(n int) []string {
...
 	}
 }
 
 // SetLocation configures the timezone for data conversion.
 func (conv *Conv) SetLocation(loc *time.Location) {
-	conv.location = loc
+	conv.Location = loc
+}
+
+// SetOffset configures the offset for data conversion.
```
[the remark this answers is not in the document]
Dharmil-searce: Ok, I removed this function and instead will use conv.Offset.

internal/convert.go
Dharmil-searce: I didn't merge it at first because it needed changes in pgdump. But now after your suggestion, I have merged these functions in latest commit.

internal/convert.go
Dharmil-searce: Ok, I have made stats and it's fields public and removed accessors i.e GetNNN functions.

internal/convert.go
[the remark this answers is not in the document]
hengfengli: Then, we should be more specific to change its name to `TimezoneOffset`.

hengfengli:
@Dharmil-searce Thanks for resolving the comments. It can be merged once @nevinheintze approves it.

Dharmil-searce:
@nevinheintze Sure, we can discuss on our weekly sync up call. Handling of timezone is different in mysql and postgres but sure I will check on how we can have only one timezone spec in convert.go.
@hengfengli I have renamed renaming Offset to TimezoneOffset.
PR can me merged now

PR #61, opened 2020-06-10
Added postgres package

Dharmil-searce:
Moved postgres specific files to postgres package. Changes in files of `postgres` package to access functions() and variables from `internal` package.

### DIFF
--- a/internal/report.go
+++ b/internal/report.go
@@ -168,7 +168,7 @@ func buildTableReportBody(conv *Conv, srcTable string, issues map[string][]Schem
 				if err != nil {
 					conv.Unexpected(err.Error())
 				}
-				srcType := printSourceType(srcSchema.ColDefs[srcCol].Type)
+				srcType := srcSchema.ColDefs[srcCol].Type.Print()
 				spType := spSchema.ColDefs[spCol].PrintColumnDefType()
 				// A note on case: Spanner types are case insensitive, but
 				// default to upper case. In particular, the Spanner AST uses
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -25,6 +25,12 @@
 // schema, and potentially get rid of the ddl package.
 package schema
 
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
 // Table represents a database table.
 type Table struct {
 	Name        string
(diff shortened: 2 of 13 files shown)


Dharmil-searce:
@googlebot I fixed it.

internal/convert.go
```diff
@@ -339,6 +342,19 @@ func (conv *Conv) StatsAddBadRow(srcTable string, b bool) {
 	}
 }
 
+func PrNodeType(n nodes.Node) string {
```
nevinheintze, line 345: Apologies this is causing churn.

As discussed on PR#60, the original code was kind of broken: we should not have been referencing nodes.Node in convert.go, and it shouldn't have been part of the convert.go APIs or used in convert_test.go.
In fact the only reason that we referenced nodes.Node in convert_test was because we needed to construct a nodes.Node to pass to conv.errorInStatement, conv.schemaStatement etc in TestStatments and TestStatementErrors. And even worse, we needed to call pg_query_go in convert_test.go.

The convert.go and convert_test.go files should be database-agnostic, and shouldn't depend on anything pg related.

I think the way forward is to:
a) Move prNodeType and prNodes back to pgdump.go (sorry for the work!)
b) Change convert_test.go to just use strings (that's effectively what prNodes(parse(t, "CREATE TABLE cart (pid text NOT NULL);")) is doing for us. In fact the test doesn't depend on the content of the string. It just counts the number of entries. Suggest changing TestStatements to:

func TestStatements(t *testing.T) {
	conv := MakeConv()
	conv.ErrorInStatement("ErrorStatement")
	conv.SchemaStatement("SchemaStatement")
	conv.DataStatement("DataStatement")
	conv.SkipStatement("SkipStatement")
	assert.Equal(t, int64(4), conv.Statements())
}

and TestStatementErrors to:

func TestStatementErrors(t *testing.T) {
	conv := MakeConv()
	conv.ErrorInStatement("ErrorStatement")
	assert.Equal(t, int64(1), conv.StatementErrors())
}

Dharmil-searce: Agreed. I will go ahead and make the changes

internal/report.go
```diff
@@ -501,3 +502,26 @@ func writeHeading(w *bufio.Writer, s string) {
 		s, "\n",
 		"----------------------------\n"}, ""))
 }
+
+func PrintSourceType(ty schema.Type) string {
+	s := ty.Name
```
Dharmil-searce, line 507: @nevinheintze Should I move it to schema/schema.go ? or This is good?

nevinheintze: Yes, 100% agreed, it should be moved to schema/schema.go.

I actually made the same comment when reviewing a few hours ago, but I don't see the comment! Must have messed up somehow.

In short, please go ahead and make that change -- the code logically belongs in schema.go.

Dharmil-searce: oh ok!
I have moved this function to schema/schema.go. Please review and merge this PR if everything else looks good.

schema/schema.go
```diff
@@ -73,3 +79,26 @@ type Ignored struct {
 	Exclusion  bool
 	ForeignKey bool
 }
+
+func PrintSourceType(ty Type) string {
```
nevinheintze, line 83: I think this should would be clearer using a receiver type.
Also, given this function is in the schema package, we can shorten the function name.

func (ty *Type) Print() {
  ...
}

Dharmil-searce: I have converted to use receiver type and changed function name to Print(). Please merge this PR, so then I will send `mysqldump` PR

PR #63, opened 2020-06-25
test: fix the failed integration test

hengfengli:


integration_test.go
```diff
@@ -157,7 +157,7 @@ func TestIntegration_Command(t *testing.T) {
 	// is because file prefixes use `now` from here (the test function) and
 	// the generated time in the files uses a `now` inside the command, which
 	// can be different.
-	cmd := exec.Command("bash", "-c", fmt.Sprintf("go run github.com/cloudspannerecosystem/harbourbridge -instance %s -dbname %s -prefix %s < %s", instanceID, dbName, filePrefix, dataFilepath))
+	cmd := exec.Command("bash", "-c", fmt.Sprintf("go run github.com/cloudspannerecosystem/harbourbridge -driver %s -instance %s -dbname %s -prefix %s < %s", "pgdump", instanceID, dbName, filePrefix, dataFilepath))
```
nevinheintze, line 160: I think instead of this change, we should make "pgdump" the default for this flag.
That should be the default -- it preserves the "it just runs" behaviour for pgdump.

main.go
```diff
@@ -79,7 +79,7 @@ func init() {
 	flag.StringVar(&dbNameOverride, "dbname", "", "dbname: name to use for Spanner DB")
 	flag.StringVar(&instanceOverride, "instance", "", "instance: Spanner instance to use")
 	flag.StringVar(&filePrefix, "prefix", "", "prefix: file prefix for generated files")
-	flag.StringVar(&driverName, "driver", "", "driver name: experimental flag for accessing source DB via database/sql driver (accepted values are \"postgres\" and \"mysql\")")
+	flag.StringVar(&driverName, "driver", "", "driver name: experimental flag for accessing source DB or dump files (accepted values are \"pgdump\", \"postgres\", \"mysqldump\", and \"mysql\")")
```
nevinheintze, line 82: Drop experimental.

Also, suggest line 74 be:

	driverName       = "pgdump"

git is being difficult and won't allow me to put a comment there (weird -- maybe too far from the changes in file?)

PR #66, opened 2020-06-26
Added support for MySQL Information Schema

Dharmil-searce:
Imported [GO-MySQL-Driver](https://github.com/go-sql-driver/mysql) to use [database/sql](https://golang.org/pkg/database/sql/) package.

### DIFF
--- a/mysql/infoschema.go
+++ b/mysql/infoschema.go
@@ -62,8 +62,8 @@ func ProcessSQLData(conv *internal.Conv, db *sql.DB) {
 	}
 	for _, t := range tables {
 		srcTable := buildTableName(t.schema, t.name)
-		srcSchema, ok1 := conv.SrcSchema[srcTable]
-		if !ok1 {
+		srcSchema, ok := conv.SrcSchema[srcTable]
+		if !ok {
 			conv.Stats.BadRows[srcTable] += conv.Stats.Rows[srcTable]
 			conv.Unexpected(fmt.Sprintf("Can't get schemas for table %s", srcTable))
 			continue
@@ -94,8 +94,8 @@ func ProcessSQLData(conv *internal.Conv, db *sql.DB) {
 			conv.Unexpected(fmt.Sprintf("Couldn't get spanner columns for table %s : err = %s", t.name, err))
 			continue
 		}
-		spSchema, ok2 := conv.SpSchema[spTable]
-		if !ok2 {
+		spSchema, ok := conv.SpSchema[spTable]
+		if !ok {
 			conv.Stats.BadRows[srcTable] += conv.Stats.Rows[srcTable]
 			conv.Unexpected(fmt.Sprintf("Can't get schemas for table %s", srcTable))
 			continue
@@ -159,7 +159,7 @@ func SetRowStats(conv *internal.Conv, db *sql.DB) {
 		if rows.Next() {
 			err := rows.Scan(&count)
 			if err != nil {
-				fmt.Printf("Can't get row count: %s\n", err)
+				conv.Unexpected(fmt.Sprintf("Can't get row count: %s", err))
 				continue
 			}
 			conv.Stats.Rows[tableName] += count
@@ -222,7 +222,7 @@ func processTable(conv *internal.Conv, db *sql.DB, table schemaAndName) error {
 func getColumns(table schemaAndName, db *sql.DB) (*sql.Rows, error) {
 	q := `SELECT c.column_name, c.data_type, c.column_type, c.is_nullable, c.column_default, c.character_maximum_length, c.numeric_precision, c.numeric_scale
               FROM information_schema.COLUMNS c
-              where table_schema = ? and table_name = ? ORDER BY c.ordinal_position;`
+              where table_schema = $1 and table_name = $2 ORDER BY c.ordinal_position;`
 	return db.Query(q, table.schema, table.name)
 }
 
@@ -235,7 +235,7 @@ func processColumns(conv *internal.Conv, cols *sql.Rows, constraints map[string]
 	for cols.Next() {
 		err := cols.Scan(&colName, &dataType, &columnType, &isNullable, &colDefault, &charMaxLen, &numericPrecision, &numericScale)
 		if err != nil {
-			fmt.Printf("Can't scan: %v\n", err)
+			conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
 			continue
 		}
 		unique := false
@@ -286,11 +286,11 @@ func getConstraints(conv *internal.Conv, db *sql.DB, table schemaAndName) ([]str
 	for rows.Next() {
 		err := rows.Scan(&col, &constraint)
 		if err != nil {
-			fmt.Printf("Can't scan: %v\n", err)
+			conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
 			continue
 		}
 		if col == "" || constraint == "" {
-			fmt.Printf("Got empty col or constraint\n")
+			conv.Unexpected(fmt.Sprintf("Got empty col or constraint"))
 			continue
 		}
 		switch constraint {
@@ -329,12 +329,11 @@ func toNotNull(conv *internal.Conv, isNullable string) bool {
 	return false
 }
 
-// buildVals contructs interface{} value containers to scan row
+// buildVals constructs []sql.RawBytes value containers to scan row
 // results into.  Returns both the underlying containers (as a slice)
 // as well as an interface{} of pointers to containers to pass to
 // rows.Scan.
 func buildVals(n int) (v []sql.RawBytes, iv []interface{}) {
-	// Make a slice for the values
 	v = make([]sql.RawBytes, n)
 	// rows.Scan wants '[]interface{}' as an argument, so we must copy the
 	// references into such a slice.
--- a/mysql/infoschema.go
+++ b/mysql/infoschema.go
@@ -84,6 +84,7 @@ func ProcessSQLData(conv *internal.Conv, db *sql.DB) {
 			continue
 		}
 		defer rows.Close()
+		srcCols, _ = rows.Columns()
 		spTable, err := internal.GetSpannerTable(conv, srcTable)
 		if err != nil {
 			conv.Unexpected(fmt.Sprintf("Couldn't get spanner table : %s", err))
@@ -222,7 +223,7 @@ func processTable(conv *internal.Conv, db *sql.DB, table schemaAndName) error {
 func getColumns(table schemaAndName, db *sql.DB) (*sql.Rows, error) {
 	q := `SELECT c.column_name, c.data_type, c.column_type, c.is_nullable, c.column_default, c.character_maximum_length, c.numeric_precision, c.numeric_scale
               FROM information_schema.COLUMNS c
-              where table_schema = $1 and table_name = $2 ORDER BY c.ordinal_position;`
+              where table_schema = ? and table_name = ? ORDER BY c.ordinal_position;`
 	return db.Query(q, table.schema, table.name)
 }
 
--- a/mysql/infoschema.go
+++ b/mysql/infoschema.go
@@ -1,0 +1,364 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mysql
+
+import (
+	"database/sql"
+	"fmt"
+	"strings"
+
+	"github.com/cloudspannerecosystem/harbourbridge/internal"
+	"github.com/cloudspannerecosystem/harbourbridge/schema"
+	_ "github.com/go-sql-driver/mysql" // The driver should be used via the database/sql package.
+	_ "github.com/lib/pq"
+)
+
+// ProcessInfoSchema performs schema conversion for source database
+// 'db'. We assume that the source database supports information
+// schema tables. These tables are a broadly supported ANSI standard,
+// and we use them to obtain source database's schema information.
+func ProcessInfoSchema(conv *internal.Conv, db *sql.DB) error {
+	tables, err := getTables(db)
+	if err != nil {
+		return err
+	}
+	for _, t := range tables {
+		if err := processTable(conv, db, t); err != nil {
+			return err
+		}
+	}
+	schemaToDDL(conv)
+	conv.AddPrimaryKeys()
+	return nil
+}
+
+// ProcessSQLData performs data conversion for source database
+// 'db'. For each table, we extract data using a "SELECT (colNamesList)" query,
+// convert the data to Spanner data (based on the source and Spanner
+// schemas), and write it to Spanner.  If we can't get/process data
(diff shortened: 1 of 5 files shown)


mysql/infoschema.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+// ProcessSQLData performs data conversion for source database
+// 'db'. For each table, we extract data using a "SELECT (colNamesList)" query,
+// convert the data to Spanner data (based on the source and Spanner
+// schemas), and write it to Spanner.  If we can't get/process data
+// for a table, we skip that table and process the remaining tables.
+//
+// Note that the database/sql library has a somewhat complex model for
+// returning data from rows.Scan. We pass *sql.RawBytes to rows.scan.
```
nevinheintze, line 56: Suggest dropping "Note that the database/sql library has a somewhat complex model for
returning data from rows.Scan." 

That comment was from the postgres code where we had to deal with different encodings for scalars and arrays.

mysql/infoschema.go
```diff
@@ -0,0 +1,367 @@
...
+	_ "github.com/go-sql-driver/mysql" // The driver should be used via the database/sql package.
+	_ "github.com/lib/pq"
+)
+
+// ProcessInfoSchema performs schema conversion for source database
+// 'db'. We assume that the source database supports information
+// schema tables. These tables are a broadly supported ANSI standard,
+// and we use them to obtain source database's schema information.
+func ProcessInfoSchema(conv *internal.Conv, db *sql.DB) error {
+	// TODO: refactor to use the set of tables computed by
```
nevinheintze, line 33: Cut-and-paste error?
This comment doesn't make sense here (there is already a copy of this comment in ProcessSQLData, where it belongs).

mysql/infoschema.go
```diff
@@ -0,0 +1,367 @@
...
+	"database/sql"
+	"fmt"
+	"strings"
+
+	"github.com/cloudspannerecosystem/harbourbridge/internal"
+	"github.com/cloudspannerecosystem/harbourbridge/schema"
+	_ "github.com/go-sql-driver/mysql" // The driver should be used via the database/sql package.
+	_ "github.com/lib/pq"
+)
+
```
nevinheintze, line 27: There is significant overlap between mysql/infoschema.go and postgres/inforschema.go.

Can you summarize the differences?

Probably worth a TODO to capture this.

Dharmil-searce: Main differences between postgres/infoschema.go and mysql/infoschema.go:

- SQL statements.
   - Difference in sql statements `ProcessSQLData` and `getColumns`.
   - In `SetRowStats` -  dropped the `" "` from table_schema and table_name as per mysql syntax.

- Different way to read data i.e. instead of interface{} we are using sql.RawBytes{}.

- Different ways of handling data conversion. Instead of `ConvertSQLRow`,`cvtSQLScalar` and `cvtSQLArray` function used for data conversion in postgres, mysql is converting row values to string and then using `ProcessDataRow` function of `mysql/data.go` file for data conversion. Hence, removed the redundancy of data conversion functions by directly using functions in `mysql/data.go`.

- Added a function `buildColNameList` to handle spatial datatypes. 

- Different list of system tables to be ignored in `getTables` function. 
   - Mysql - `{"information_schema", "mysql", "performance_schema", "sys"}` 
   - Postgres - `{"information_schema", "postgres", "pg_catalog", "pg_temp_1", "pg_toast", "pg_toast_temp_1"}`

- Do not need to drop `public` prefix in table name because mysql do not support `public`.

- `valsToStrings` is different because postgres uses `[]interface{}` argument, while mysql uses `[]sql.RawBytes` argument.

mysql/infoschema.go
Dharmil-searce: My Bad. Removed it.

mysql/infoschema.go
Dharmil-searce: Done. Formatted the comment.

mysql/toddl.go
```diff
@@ -136,10 +136,10 @@ func toSpannerType(conv *internal.Conv, id string, mods []int64) (ddl.ScalarType
 		}
 		return ddl.String{Len: ddl.MaxLength{}}, nil
 	case "text", "tinytext", "mediumtext", "longtext":
-		maxExpectedMods(0)
+		maxExpectedMods(1)
```
nevinheintze, line 139: Can you give some background on this change: can text types have a modifier?
Can you give an example?

Dharmil-searce: Though we would not be specifying mods for `text` type, but MySQL by default will put max value for `text` type in `CHARACTER_MAXIMUM_LENGTH` column of `information_schema.COLUMNS` table. Eg: 
```
mysql> select c.column_name,c.data_type,c.is_nullable,c.column_default, c.character_maximum_length, c.numeric_precision, c.numeric_scale from information_schema.COLUMNS c where table_schema='testdb' and COLUMN_NAME LIKE '%text';
+-------------+------------+-------------+----------------+--------------------------+-------------------+---------------+
| COLUMN_NAME | DATA_TYPE  | IS_NULLABLE | COLUMN_DEFAULT | CHARACTER_MAXIMUM_LENGTH | NUMERIC_PRECISION | NUMERIC_SCALE |
+-------------+------------+-------------+----------------+--------------------------+-------------------+---------------+
| text        | text       | YES         | NULL           |                    65535 |              NULL |          NULL |
| tinytext    | tinytext   | YES         | NULL           |                      255 |              NULL |          NULL |
| mediumtext  | mediumtext | YES         | NULL           |                 16777215 |              NULL |          NULL |
| longtext    | longtext   | YES         | NULL           |               4294967295 |              NULL |          NULL |
+-------------+------------+-------------+----------------+--------------------------+-------------------+---------------+
4 rows in set (0.00 sec)
```

In function `toType()`, when case is `charLen.Valid` we return `schema.Type{Name: dataType, Mods: []int64{charLen.Int64}}`.  Thus, for `text` type, mods will be allocated, and hence I have changed to`maxExpectedMods(1)` in `mysql/toddl.go` file.

mysql/infoschema.go
nevinheintze: Many thanks for the detailed response.

What's your take on RawBytes vs interface{}?

My understanding was that interface{} would return values from the underlying driver without conversion.
But RawBytes effectively forces conversion to a string format. Which means that if the value is an int64, then we do int64 -> string -> int64.

But I 100% agree with you about the redundant data conversion code.

Should we just do both postgres and mysql via string? Any thoughts on the additional processing/conversion cost?

mysql/infoschema.go
```diff
@@ -0,0 +1,364 @@
...
+// getConstraints returns a list of primary keys and by-column map of
+// other constraints.  Note: we need to preserve ordinal order of
+// columns in primary key constraints.
+func getConstraints(conv *internal.Conv, db *sql.DB, table schemaAndName) ([]string, map[string][]string, error) {
+	q := `SELECT k.COLUMN_NAME, t.CONSTRAINT_TYPE
+              FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS AS t
+                INNER JOIN INFORMATION_SCHEMA.KEY_COLUMN_USAGE AS k
+                  ON t.CONSTRAINT_NAME = k.CONSTRAINT_NAME AND t.CONSTRAINT_SCHEMA = k.CONSTRAINT_SCHEMA AND t.TABLE_NAME=k.TABLE_NAME
+              WHERE k.TABLE_SCHEMA = ? AND k.TABLE_NAME = ? ORDER BY k.ordinal_position;`
```
nevinheintze, line 277: Why use '?' instead of $1, $2 for the parameters?

mysql/infoschema.go
```diff
@@ -0,0 +1,364 @@
...
+		return nil, nil, err
+	}
+	defer rows.Close()
+	var primaryKeys []string
+	var col, constraint string
+	m := make(map[string][]string)
+	for rows.Next() {
+		err := rows.Scan(&col, &constraint)
+		if err != nil {
+			fmt.Printf("Can't scan: %v\n", err)
```
nevinheintze, line 289: This is going to potentially spam stdout with lots of errors.
Maybe better as:

conv.Unexpected(fmt.Sprintf("Can't scan: %v\n", err))

This is also an issue in postgres/infoschema.go.

Same comment in other uses of fmt.Printf in this file.

mysql/infoschema.go
Dharmil-searce: MySQL does not support bind parameters `$1,$2..` like postgres. It throws an error when such bind parameters are used:
```
panic: couldn't get schema for table testdb.Persons: Error 1054: Unknown column '$1' in 'where clause'
```
Thus, used `?` instead of `$1,$2..`.

mysql/infoschema.go
Dharmil-searce: Agreed. Initially, I had put it as unexpected only but then changed it to fmt.Printf() after comparing the code with postgres/infoschema.go.
Sure, I will update `fmt.Printf()` usage with `conv.Unexpected(fmt.Sprintf())` in both the packages.

mysql/infoschema.go
```diff
@@ -0,0 +1,364 @@
...
+	case "YES":
+		return false
+	case "NO":
+		return true
+	}
+	conv.Unexpected(fmt.Sprintf("isNullable column has unknown value: %s", isNullable))
+	return false
+}
+
+// buildVals contructs interface{} value containers to scan row
```
nevinheintze, line 332: You are constructing []RawBytes containers.

How about rephrasing:

// buildVals constructs []sql.RawBytes value containers to scan row

mysql/infoschema.go
```diff
@@ -0,0 +1,364 @@
...
+// results into.  Returns both the underlying containers (as a slice)
+// as well as an interface{} of pointers to containers to pass to
+// rows.Scan.
+func buildVals(n int) (v []sql.RawBytes, iv []interface{}) {
+	// Make a slice for the values
```
nevinheintze, line 337: Suggest deleting this comment -- its not adding much, given the function-level comment.

mysql/infoschema_test.go
```diff
@@ -0,0 +1,304 @@
...
+		"bs": []internal.SchemaIssue{internal.DefaultValue},
+		"f4": []internal.SchemaIssue{internal.Widened},
+		"i4": []internal.SchemaIssue{internal.Widened},
+		"i2": []internal.SchemaIssue{internal.Widened},
+		"si": []internal.SchemaIssue{internal.Widened, internal.DefaultValue},
+		"ts": []internal.SchemaIssue{internal.Datetime},
+	}
+	assert.Equal(t, expectedIssues, conv.Issues["test.test"])
+	assert.Equal(t, int64(0), conv.Unexpecteds())
+}
```
nevinheintze, line 144: Should there be a TestProcessSqlData similar to the one in postgres/infoschema_test.go?

mysql/infoschema_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+// TODO : Remove this test (Not needed)
```
nevinheintze, line 146: Yes, agreed. I think we can remove this since it is redundant.

mysql/infoschema.go
Dharmil-searce: Done.

mysql/infoschema.go
```diff
@@ -0,0 +1,364 @@
...
+	// TODO: refactor to use the set of tables computed by
+	// ProcessInfoSchema instead of computing them again.
+	tables, err := getTables(db)
+	if err != nil {
+		conv.Unexpected(fmt.Sprintf("Couldn't get list of table: %s", err))
+		return
+	}
+	for _, t := range tables {
+		srcTable := buildTableName(t.schema, t.name)
+		srcSchema, ok1 := conv.SrcSchema[srcTable]
```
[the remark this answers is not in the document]
Dharmil-searce: Done.

mysql/infoschema.go
Dharmil-searce: Done.

mysql/infoschema.go
Dharmil-searce: Updated in both `mysql/infoschema.go` and `postgres/infoschema.go`.

mysql/toddl.go
nevinheintze: Thanks for all of the details and the example.

So this doesn't show up in mysql dump parsing, just in the info schema tables?

In effect we're relaxing the error checking for mysql dump generated schemas, but I don't see a huge issue in doing that.

mysql/infoschema.go
nevinheintze: Ah right. Got it. That's super annoying.
I thought mysql did support $1, $2 etc.
Maybe its an issue with the golang database/sql driver?

mysql/toddl.go
Dharmil-searce: Correct, In` mysqldump` create statements will have `text` and `blob` datatype without any mods. It's also mentioned in pingcap documentation under data types section ([LINK](https://docs.pingcap.com/tidb/dev/sql-statement-create-table)).
Even if mysqldump has a modifier for `text` & `blob` datatype then also pingcap parser will ignore it and return the parsed statement without any modifier ( I have tested the same for both datatypes). Hence, we are not relaxing error checking for mysqldump. 
One work around could be to explicitly put a condition when datatype is `text` or `blob` and return `schema.Type{Name: dataType}` in mysql/infoschema.go  and then we do not need to change `maxExpectedMods(0)` but I do not see much benefit of doing that and also it would not be a good code design.

Thus, I think we can move forward with this change of error checks.

mysql/infoschema.go
Dharmil-searce: Yeah, maybe it's not supported by golang driver.

mysql/infoschema_test.go
```diff
@@ -0,0 +1,304 @@
...
+			ColDefs:  map[string]ddl.ColumnDef{col: ddl.ColumnDef{Name: col, T: tc.spType, IsArray: tc.isArray}}}
+		conv.TimezoneOffset = "+05:30"
+		_, ac, av, err := ConvertData(conv, tableName, cols, srcSchema, tableName, cols, spSchema, []string{tc.in})
+		assert.Equal(t, cols, ac)
+		assert.Equal(t, []interface{}{tc.e}, av)
+		assert.Nil(t, err)
+	}
+}
+
+func TestConvertSqlRow_MultiCol(t *testing.T) {
```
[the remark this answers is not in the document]
Dharmil-searce: Done. Renamed.

mysql/infoschema_test.go
Dharmil-searce: In `ProcessSQLData` function of `mysql/infoschema.go` we are calling `ProcessDataRow` of `mysql/data.go` unlike in postgres where `ConvertSQLRow` is being called. Hence, adding `TestProcessSqlData` is not adding any benefit because test cases for `ProcessDataRow` are covered in `mysql/data_test.go` file.

Still, as per your suggestion I have added `TestProcessSQLData`.

Dharmil-searce:
> I'm happy with the PR.
> Please merge when all comments are resolved.

@nevinheintze Only those with write access to the repo can merge PRs. I do not have that access so request you to merge the PR.

> Should we just do both postgres and mysql via string? Any thoughts on the additional processing/conversion cost?

We can discuss this on the group.

mysql/infoschema.go
Dharmil-searce: Main benefit to use RawBytes is to avoid redundancy in code and reuse data conversion functions. Other benefit is that we can avoid complex switch conditions in `convScalar` and `convArray`.
It would definitely be easier and in sync to use RawBytes in postgres as well. [This](github.com/lib/pq) library is currently used in postgres. Need to check if it supports RawBytes. Also, in the documentation of current library another library([LINK](https://github.com/jackc/pgx)) is recommended to be used. In addition to that,array structure is a bit complex in postgres compared to mysql need to check that also.

We can discuss more about additional processing/conversion cost on google chat group.

PR #67, opened 2020-06-30
Fix "dirver" typo.

nevinheintze:
dirver -> driver

### DIFF
--- a/main.go
+++ b/main.go
@@ -189,7 +189,7 @@ func toSpanner(driver, projectID, instanceID, dbName string, ioHelper *ioStreams
 }
 
 func schemaConv(driver string, ioHelper *ioStreams) (*internal.Conv, error) {
-	fmt.Println("dirver:", driver)
+	fmt.Println("driver:", driver)
 	switch driver {
 	case POSTGRES, MYSQL:
 		return schemaFromSQL(driver)
--- a/main.go
+++ b/main.go
@@ -105,13 +105,13 @@ func main() {
 	defer close(lf)
 
 	ioHelper := &ioStreams{in: os.Stdin, out: os.Stdout}
-
+	fmt.Println("Using driver (source DB):", driverName)
 	project, err := getProject()
 	if err != nil {
 		fmt.Printf("\nCan't get project: %v\n", err)
 		panic(fmt.Errorf("can't get project"))
 	}
-	fmt.Printf("Using project: %s\n", project)
+	fmt.Println("Using Google Cloud project:", project)
 
 	instance := instanceOverride
 	if instance == "" {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -189,7 +189,7 @@ func toSpanner(driver, projectID, instanceID, dbName string, ioHelper *ioStreams
 }
 
 func schemaConv(driver string, ioHelper *ioStreams) (*internal.Conv, error) {
-	fmt.Println("dirver:", driver)
+	fmt.Println("driver:", driver)
```
hengfengli, line 192: Is this a debugging code? should we remove it?

nevinheintze: That's a good question. I think we probably need something to indicate how harbourbridge is behaving.
@Dharmil-searce -- what's your take?

Dharmil-searce: @hengfengli Actually this line was added by you in #63.

I agree with @nevinheintze that it would be good to indicate the `driver` same as we are indicating `project` and `instance`.

One suggestion - To be in sync with the previous print statements, I think `Using driver : drivername` would be more appropriate instead of just `driver : drivername`. What do you think ? Final output would be like :

```
Using project: test-project
Using Spanner instance: test-instance

WARNING: Please check that permissions for this Spanner instance are
appropriate. Spanner manages access control at the database level, and the
database created by HarbourBridge will inherit default permissions from this
instance. All data written to Spanner will be visible to anyone who can
access the created database. Note that mysqldump table-level and row-level
ACLs are dropped during conversion since they are not supported by Spanner.

Using driver: mysqldump
Generating schema: 100%
Wrote schema to file 'mysqldump_2020-07-01_2c3d-d1b2.schema.txt'.
Creating new database mysqldump_2020-07-01_2c3d-d1b2 in instance test-instance with default permissions ... done.
Writing data to Spanner: 100%
Processed 283 bytes of mysqldump data (1 statements, 3 rows of data, 0 errors, 0 unexpected conditions).
Schema conversion: EXCELLENT (all columns mapped cleanly).
Data conversion: EXCELLENT (all 3 rows written to Spanner).
See file 'mysqldump_2020-07-01_2c3d-d1b2.report.txt' for details of the schema and data conversions.
```

hengfengli: Ah, I forgot to delete this when I submitted the fix for the failed IT test. If we need this, I would move it to `toSpanner()` before the call of `schemaConv()`.

Dharmil-searce: Yes, it would be good to have it.

nevinheintze: Done.
I made it first, so we now have a block of "Using":

  Using driver (source DB): mysqldump
  Using Google Cloud project: e2e-debugging
  Using Cloud Spanner instance: e2e-spanner-testing

followed by the warning about permissions for the instance.
I clarified the other message a bit too:
  "project" --> "Google Cloud project"
  "Spanner instance" --> "Cloud Spanner instance"

PR #68, opened 2020-07-02
Fix DEFAULT NULL reporting issue in mysqldump

Dharmil-searce:
`DEFAULT NULL` is explicitly defined by MySQL ([LINK](https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html#:~:text=If%20a%20data%20type%20specification,with%20no%20explicit%20DEFAULT%20clause.)). Ignored this case from issue reporting of `DEFAULT` condition.

### DIFF
--- a/mysql/mysqldump.go
+++ b/mysql/mysqldump.go
@@ -341,7 +341,15 @@ func updateColsByOption(conv *internal.Conv, tableName string, col *ast.ColumnDe
 		case ast.ColumnOptionNotNull:
 			column.NotNull = true
 		case ast.ColumnOptionDefaultValue:
-			column.Ignored.Default = true
+			// If a data type specification includes no explicit DEFAULT
+			// value, MySQL determines if the column can take NULL as a value
+			// and the column is defined with DEFAULT NULL clause in mysqldump.
+			// This case is ignored from issue reporting of 'Default' value.
+			v, ok := elem.Expr.(*driver.ValueExpr)
+			nullDefault := ok && v.GetValue() == nil
+			if !nullDefault {
+				column.Ignored.Default = true
+			}
 		case ast.ColumnOptionUniqKey:
 			column.Unique = true
 		case ast.ColumnOptionCheck:


mysql/mysqldump.go
```diff
@@ -342,6 +342,14 @@ func updateColsByOption(conv *internal.Conv, tableName string, col *ast.ColumnDe
 			column.NotNull = true
 		case ast.ColumnOptionDefaultValue:
 			column.Ignored.Default = true
+			// If a data type specification includes no explicit DEFAULT
+			// value, MySQL determines if the column can take NULL as a value
+			// and the column is defined with DEFAULT NULL clause in mysqldump.
+			// This case is ignored from issue reporting of 'Deafult' value.
```
hengfengli, line 348: A typo.

mysql/mysqldump.go
```diff
@@ -342,6 +342,14 @@ func updateColsByOption(conv *internal.Conv, tableName string, col *ast.ColumnDe
...
+			defaultValue := fmt.Sprintf("%v", elem.Expr.(*driver.ValueExpr).GetValue())
```
nevinheintze, line 349: Probably best to avoid the type assertion here (which could trigger panic), and instead use the "ok" version so we can test the assertion.
Also let's avoid the Sprintf and do direct value comparison.
i.e. elem.Expr.(*driver.ValueExpr).GetValue() == nil.
And simplest to make the assignment to true conditional, instead of having two assignments.

How about:

			v, ok := elem.Expr.(*driver.ValueExpr)
			if !ok || v.GetValue() != nil {
				column.Ignored.Default = true
			}

or maybe a bit more readable as:

			// If a data type specification includes no explicit DEFAULT
			// value, MySQL determines if the column can take NULL as a value
			// and the column is defined with DEFAULT NULL clause in mysqldump.
			// This case is ignored from issue reporting of 'Default' value.
  			v, ok := elem.Expr.(*driver.ValueExpr)
                        nullDefault := ok && v.GetValue() == nil
                        if !nullDefault {
                                column.Ignored.Default = true
                        }

hengfengli: I like the second one.

Dharmil-searce: Sure, updated according to second one.

PR #62, opened 2020-06-11
Added support for Mysqldump

Dharmil-searce:
Added `mysql` package to support mysqldump. [pingcap parser](https://github.com/pingcap/parser) is used to parse mysqldump file

PR #69, opened 2020-07-02
Added MySQL README support.

Dharmil-searce:


README.md
```diff
@@ -21,35 +22,41 @@ produces to complete your evaluation. For example, while HarbourBridge preserves
...
 indexes](https://cloud.google.com/spanner/docs/secondary-indexes) if your SQL
-queries rely on PostgreSQL indexes that have been dropped. HarbourBridge is not
+queries rely on PostgreSQL/MySQL indexes that have been dropped. HarbourBridge is not
 intended for production database migration.
 
 To use the tool on a PostgreSQL database called mydb, run
 
 ```sh
-pg_dump mydb | harbourbridge
+pg_dump mydb | harbourbridge -driver=pgdump
```
hengfengli, line 31: This is not needed because by default the driver is `pgdump`.

hengfengli: We can say: 

```sh
# By default, the driver is "pgdump".
pg_dump mydb | harbourbridge
# Or, 
pg_dump mydb | harbourbridge -driver=pgdump
```

README.md
```diff
@@ -21,35 +22,41 @@ produces to complete your evaluation. For example, while HarbourBridge preserves
...
+
+HarbourBridge accepts pg_dump/mysqldump's standard plain-text format, but not archive or
+custom formats. More details on usage can be found in [Example usage](#example-usage) section.
 
 HarbourBridge automatically determines the cloud project and Spanner instance to
-use, and generates a new Spanner database name (prefixed with `pg_dump_` and
+use, and generates a new Spanner database name (prefixed with `{driver}_` and
 today's date). Command-line flags can be used to explicitly set the Spanner
-instance or database name.
+instance or database name. See [Options](#options)
```
hengfengli, line 46: Add `.` at the end.

README.md
```diff
@@ -219,6 +237,10 @@ dbname.
 created in this instance. If not specified, the tool automatically determines an
 appropriate instance using gcloud.
 
+`-driver` Specifies the driver to use for schema and data conversion. Supported drivers
```
hengfengli, line 240: By default, the driver is `pgdump`. Please update this statement accordingly.

README.md
```diff
@@ -228,236 +250,52 @@ messages about the conversion.
...
-```
-
-By default, HarbourBridge will generate a new Spanner database name to populate.
-You can override this and specify the database name to use by:
+Driver types (source DB) that are supported by Harbourbridge:
 
-```sh
-pg_dump mydb | harbourbridge -dbname my-spanner-database-name
-```
+1. mysqldump
```
hengfengli, line 255: nits: can we follow the same order for drivers: `postgres`, `pgdump`, `mysql`, `mysqldump`?

README.md
```diff
@@ -228,236 +250,52 @@ messages about the conversion.
...
+4. postgres
 
-HarbourBridge generates a report file, a schema file, and a bad-data file (if
-there are bad-data rows). You can control where these files are written by
-specifying a file prefix. For example,
-
-```sh
-pg_dump mydb | harbourbridge -prefix mydb.
-```
+Please refer below mentioned links for example usage of MySQL and PostgreSQL database:
```
hengfengli, line 260: Nits: Same here. We always talk about PostgresSQL/MySQL. Suddenly, the order has changed. I would prefer the same order, `PostgresSQL` and then `MySQL`.

mysql/README.md
```diff
@@ -0,0 +1,233 @@
...
+some decimal types (e.g. `DECIMAL(7, 3))` this mapping will preserve precision.
+But for others, the decimal type does not fit in `FLOAT64`; HarbourBridge
+generates a warning in such cases. In general, mapping `DECIMAL` to `FLOAT64`
+can be useful for evaluation purposes, but it is not recommended for production
+use. Note that in MySQL, NUMERIC is implemented as DECIMAL, so the remarks about DECIMAL apply equally to NUMERIC.
+
+### `TIMESTAMP` and `DATETIME`
+
+MySQL has two timestamp types: `TIMESTAMP` and `DATETIME`. Both provide microsecond resolultion, but neither
+actually stores a timezone with the data. The key difference between the two types is that MySQL converts `TIMESTAMP` values from the current time zone to UTC for storage, and back from UTC to the current time zone for retrieval. This does not occur `DATETIME` and data is returned
```
hengfengli, line 108: Can we reformat this line and other lines in this file to max line length of 80 characters?

README.md
```diff
@@ -1,14 +1,15 @@
-# HarbourBridge: Turnkey Postgres-to-Spanner Evaluation
+# HarbourBridge: Turnkey Postgres-to-Spanner/MySQL-to-Spanner Evaluation
```
nevinheintze, line 1: How about just:

# HarbourBridge: Turnkey Spanner Evaluation

README.md
```diff
@@ -1,14 +1,15 @@
...
+
 [![cloudspannerecosystem](https://circleci.com/gh/cloudspannerecosystem/harbourbridge.svg?style=svg)](https://circleci.com/gh/cloudspannerecosystem/harbourbridge)
 
 HarbourBridge is a stand-alone open source tool for Cloud Spanner evaluation,
-using data from an existing PostgreSQL database. The tool ingests pg_dump
-output, automatically builds a Spanner schema, and creates a new Spanner
-database populated with data from pg_dump.
+using data from an existing PostgreSQL/MySQL database. The tool ingests from
```
nevinheintze, line 6: Suggest:

"PostgreSQL/MySQL" --> "PostgreSQL or MySQL"

README.md
```diff
@@ -1,14 +1,15 @@
...
+database/dump file output, automatically builds a Spanner schema, and creates
```
nevinheintze, line 7: How about:

The tool ingests schema and data from either a pgdump/mysqldump file or directly from the database,
automatically builds ...

README.md
```diff
@@ -1,14 +1,15 @@
...
+a new Spanner database populated with data from the selected approach.
```
nevinheintze, line 8: Suggest:

"the selected approach" --> "the source database"

README.md
```diff
@@ -1,14 +1,15 @@
...
 
 HarbourBridge is designed to simplify Spanner evaluation, and in particular to
-bootstrap the process by getting moderate-size PostgreSQL datasets into Spanner
-(up to a few GB). Many features of PostgreSQL, especially those that don't map
+bootstrap the process by getting moderate-size PostgreSQL/MySQL datasets into Spanner
+(up to a few GB). Many features of PostgreSQL/MySQL, especially those that don't map
```
nevinheintze, line 12: Given the experiments you have run and also ones we've run, should we consider relaxing this a bit.
Maybe say: "up to a few tens of GB"?

README.md
```diff
@@ -21,35 +22,41 @@ produces to complete your evaluation. For example, while HarbourBridge preserves
...
 anyone who can access the created database.**
 
-As it processes the pg_dump data, HarbourBridge reports on progress, provides
+As it processes the PostgreSQL/MySQL data, HarbourBridge reports on progress, provides
 stats on the schema and data conversion steps, and an overall assessment of the
 quality of the conversion. It also generates a schema file and report file (and
 a bad-data file if data was dropped). See
 [Files Generated by HarbourBridge](#files-generated-by-harbourbridge). Details
-of how PostgreSQL is mapped to Spanner can be found in the
+of how current schema is mapped to Spanner can be found in the
```
nevinheintze, line 59: Suggest:

"current schema" --> "source database's schema"

README.md
Dharmil-searce: What I recall is that we once had a discussion and default `pgdump` was removed. Currently there is no default condition and hence specifying a driver is mandatory. We can have couple of options:

-  We can add a error condition if driver is NULL and suggest user to pick up a driver:

```
func main() {
...

if driverName == "" {
fmt.Printf("\nNo driver (source DB) selected. Please specify a driver using '--driver' flag\n")
panic(fmt.Errorf("No driver (source DB) selected"))
}
fmt.Println("Using driver (source DB):", driverName)

...
}
```

- We can again add default condition for `pgdump` :
```
func main() {
...

if driverName == "" {
  driverName=PGDUMP
}
fmt.Println("Using driver (source DB):", driverName)

...
}
```

What I feel is that instead of having default condition (`pgdump` driver) and then error thrown by the code if dump file is not `pgdump` or the selected approach is `postgres` information schema, it's better to mandate the driver(which is currently done) and put a error condition for not specifying a driver.
What are your views on this? @hengfengli @nevinheintze

README.md
Dharmil-searce: As discussed in previous comment, currently there is no default driver and specifying a driver is must. Let me know what do you think and then I will update accordingly.

nevinheintze: And remove the "It is mandatory to specify a driver."

mysql/README.md
```diff
@@ -0,0 +1,233 @@
+## MySQL Example Usage
```
nevinheintze, line 1: I think we should have a top-level header and a bit of explanatory text. How about something like:

# HarbourBridge: Turnkey MySQL-to-Spanner Evaluation

HarbourBridge is a stand-alone open source tool for Cloud Spanner evaluation, using data from an existing PostgreSQL or MySQL database. This README provides details of the tools MySQL capabilities. For
general HarbourBridge information see <pointer-to-main-README>.

## Example MySQL Usage

postgres/README.md
```diff
@@ -0,0 +1,226 @@
+## PostgreSQL Example Usage
```
nevinheintze, line 1: Same comment as mysql/README.md: I think we should have a top-level header and a bit of explanatory text. How about something like:

# HarbourBridge: Turnkey PostgreSQL-to-Spanner Evaluation

HarbourBridge is a stand-alone open source tool for Cloud Spanner evaluation, using data from an existing PostgreSQL or MySQL database. This README provides details of the tool's PostgreSQL capabilities. For
general HarbourBridge information see <pointer-to-main-README>.

## Example PostgreSQL Usage

postgres/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+The following examples assume `harbourbridge` has been added to your PATH
+environment variable.
+
+To use HarbourBridge on a PostgreSQL database called mydb using pgdump output, run:
```
nevinheintze, line 6: pgdump --> pg_dump

postgres/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+```sh
+pg_dump mydb | harbourbridge -driver=pgdump
```
nevinheintze, line 9: Quick question: for consistency should we use "pg_dump" for the driver name (i.e. change the PGDUMP constant in main.go).
I don't feel too strongly, but the inconsistency kind of stands out in the README.
And if we want to make this change we should do it now.

mysql/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+The following examples assume `harbourbridge` has been added to your PATH
+environment variable.
+
+To use HarbourBridge on a MySQL database called mydb using mysqldump, run:
```
nevinheintze, line 6: I think it might be worth highlighting the mysqldump vs direct method a bit more. In the current text it is kind of lost at the bottom of the page. How about:

1) Inserting a sentence at the start of the paragraph something like: "HarbourBridge can either be used with mysqldump or it can be run directly on a MySQL database (via go's database/sql package)."
2) Add a sub-heading: "Using HarbourBridge with mysqldump"
3) Add a sub-heading: "Directly connecting to a MySQL database"

mysql/README.md
```diff
@@ -0,0 +1,233 @@
...
+would write the files into the directory `~/spanner-eval-mydb/`. Note
+that HarbourBridge will not create directories as it writes these files.
+
+To use the tool directly on a MySQL database called mydb, run
+
+```sh
+harbourbridge -driver=mysql
+```
+
+It is assumed that MySQL database supports information schema tables and
```
nevinheintze, line 55: "It is assumed that MySQL database supports information schema tables" -- we can drop this since we're specifically discussion MySQL and I think all supported MySQL version have information schema.

Side note: we use to have infoschema.go as a generic file (and so the comment in that file about assuming information schema is supported made sense). But I'm wondering if we should drop the comment in those files now since the inforschema.go files are under mysql/ and postgres/ and both of those DBs support infoschema.

mysql/README.md
```diff
@@ -0,0 +1,233 @@
...
+_MYSQLHOST_, _MYSQLPORT_, _MYSQLUSER_, _MYSQLDATABASE_ environments variables are set.
+Password can be specified either in the _MYSQL_PWD_ environment variable or need to
+enter password when prompted by harbourbridge.
```
nevinheintze, line 58: Suggest replacing:
"or need to enter password when prompted by harbourbridge."
by
"or provided at the password prompt."

mysql/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+Rest of the above mentioned usage of options also apply to Information schema approach.
```
nevinheintze, line 60: Suggest replacing with:

Note that all of the options described in the previous section on using mysqldump can also be used with "-driver=mysql".

mysql/README.md
```diff
@@ -0,0 +1,233 @@
...
+Spanner does not support decimal types, so these are mapped to `FLOAT64`. For
+some decimal types (e.g. `DECIMAL(7, 3))` this mapping will preserve precision.
+But for others, the decimal type does not fit in `FLOAT64`; HarbourBridge
+generates a warning in such cases. In general, mapping `DECIMAL` to `FLOAT64`
+can be useful for evaluation purposes, but it is not recommended for production
+use. Note that in MySQL, NUMERIC is implemented as DECIMAL, so the remarks about DECIMAL apply equally to NUMERIC.
+
+### `TIMESTAMP` and `DATETIME`
+
+MySQL has two timestamp types: `TIMESTAMP` and `DATETIME`. Both provide microsecond resolultion, but neither
```
nevinheintze, line 107: typo: resolultion

mysql/README.md
nevinheintze, line 108: Missing "for" in "occur `DATETIME`".

mysql/README.md
```diff
@@ -0,0 +1,233 @@
...
+the semantics of MySQL's `VARCHAR(n)`. However there are some minor
+differences. For example, even `VARCHAR(n)` has some special treatment of
+spaces: string with trailing spaces in excess of the column length are truncated
+prior to insertion and a warning is generated.
+
+### `SET`
+
+MySQL `SET` is a string object that can hold muliple values, each of which must be
+chosen from a list of permitted values specified when the table is created. `SET`
+is being mapped to `ARRAY<STRING>` spanner type. Validation of `SET` element values
```
nevinheintze, line 140: Suggest replacing:

`SET` is being mapped to `ARRAY<STRING>` spanner type.

by 

"`SET` is mapped to the Spanner type `ARRAY<STRING>`."

mysql/README.md
```diff
@@ -0,0 +1,233 @@
...
+will be dropped in spanner. Thus for production use, validation needs to be done
```
nevinheintze, line 141: spanner --> Spanner

Can you check this elsewhere in the file? 
Spanner should be consistently capitalized.

mysql/README.md
```diff
@@ -0,0 +1,233 @@
...
+in the application.
+
+### `Spatial datatype`
+
+MySQL spatial datatypes are used to represent geographic feature/geospatial feature.
+It includes `GEOMETRY`, `POINT`, `LINESTRING`, `POLYGON`, `MULTIPOINT`, `MULTIPOLYGON`
+and `GEOMETRYCOLLECTION` datatypes. Spanner does not support Geospatial data types.
```
nevinheintze, line 148: Geospatial -> geospatial?
I don't think we need to capitalize this?

mysql/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+MySQL `SET` is a string object that can hold muliple values, each of which must be
+chosen from a list of permitted values specified when the table is created. `SET`
+is being mapped to `ARRAY<STRING>` spanner type. Validation of `SET` element values
+will be dropped in spanner. Thus for production use, validation needs to be done
+in the application.
+
+### `Spatial datatype`
+
+MySQL spatial datatypes are used to represent geographic feature/geospatial feature.
```
nevinheintze, line 146: Suggest replacing "geographic feature/geospatial feature" with just "geographic features", based on https://dev.mysql.com/doc/refman/8.0/en/spatial-types.html.

README.md
Dharmil-searce: Fair enough.

README.md
```diff
@@ -228,236 +250,52 @@ messages about the conversion.
...
-
-To specify a particular Spanner instance to use, run:
-
-```sh
-pg_dump mydb | harbourbridge -instance my-spanner-instance
-```
-
-By default, HarbourBridge will generate a new Spanner database name to populate.
-You can override this and specify the database name to use by:
+Driver types (source DB) that are supported by Harbourbridge:
```
nevinheintze, line 253: Suggest deleting the paragraph -- I don't see what it adds here, and it kind of repeats the list from a few lines above.

Driver types (source DB) that are supported by Harbourbridge:
1. postgres
2. pgdump
3. mysql
4. mysqldump

mysql/README.md
Dharmil-searce: Done.

postgres/README.md
Dharmil-searce: Done.

README.md
```diff
@@ -228,236 +250,52 @@ messages about the conversion.
...
-sequences, procecdures, triggers, (non-primary) indexes and views. The tool does
-not support these and the relevant statements are dropped during schema
-conversion.
-
-See
-[Migrating from PostgreSQL to Cloud Spanner](https://cloud.google.com/spanner/docs/migrating-postgres-spanner)
-for a general discussion of PostgreSQL to Spanner migration issues.
-HarbourBridge follows most of the recommendations in that guide. The main
-difference is that we map a few more types to `STRING(MAX)`.
+Please refer below mentioned links for details on schema conversion of PostgreSQL
```
nevinheintze, line 267: Suggest replacing with:

Details on HarbourBridge schema conversion for PostgreSQL and MySQL can be found in the [PostgreSQL schema conversion](postgres/README.md#schema-conversion) and [MySQL schema conversion](mysql/README.md#schema-conversion) sections respectively.

mysql/README.md
Dharmil-searce: Done. Updated `postgres/README.go` as well.

README.md
```diff
@@ -228,236 +250,52 @@ messages about the conversion.
...
+- [MySQL schema conversion](mysql/README.md#schema-conversion)
 
 ## Data Conversion
 
-HarbourBridge converts PostgreSQL data to Spanner data based on the Spanner
+HarbourBridge converts PostgreSQL/MySQL data to Spanner data based on the Spanner
 schema it constructs. Conversion for most data types is fairly straightforward,
 but several types deserve discussion. Note that HarbourBridge is not intended
-for databases larger than a couple of GB.
+for databases larger than a couple of GB. Please refer below mentioned links for details on
```
nevinheintze, line 278: Replace "Please refer below mentioned links" with "Please refer to the links below".

mysql/README.md
Dharmil-searce: Sure, I had kept this statement to be in sync with the comment in `infoschema.go` files. 
Both Postgres and Mysql DB supports Information schema, so it would be reasonable to drop "assuming information schema.." comment from both package's `infoschema.go` file and from README as well.

README.md
nevinheintze: Yes, sorry, we should have discussed this in a bit more detail.

https://github.com/cloudspannerecosystem/harbourbridge/pull/63 added the default.

What happened is that #62 broke the integration test. Hengfeng and I discussed various options for fixing. In the end we decided to have pgdump as the default to preserve the original harbourbridge behaviour i.e. harbourbridge just works out-of-the-box with pg_dump, no flags, no args, no configs.

This is done by initializing driverName to "pgdump" (line 74 of main.go).

mysql/README.md
Dharmil-searce: Done. Updated the same for postgres/README.md.

mysql/README.md
Dharmil-searce: Checked and capitalized Spanner at all the places.

README.md
```diff
@@ -228,236 +250,52 @@ messages about the conversion.
...
 
-First, check that pg_dump is correctly configured to connect to your PostgreSQL
-database. Note that pg_dump uses the same options as psql to connect to your
-database. See the [psql](https://www.postgresql.org/docs/9.3/app-psql.html) and
-[pg_dump](https://www.postgresql.org/docs/9.3/app-pgdump.html) documentation.
+- **pgdump**: pg_dump uses the same options as psql to connect to your
+  database. See the [psql](https://www.postgresql.org/docs/9.3/app-psql.html) and
+  [pg_dump](https://www.postgresql.org/docs/9.3/app-pgdump.html) documentation.
 
 Access to a PostgreSQL database is typically configured using the following
```
nevinheintze, line 300: Formatting seems broken here: the "Access to a PostgreSQL database ..." is part of the pgdump bullet, but it is not indented?

Same for the other bullets for postgres, mysqldump and mysql.

README.md
```diff
@@ -228,236 +250,52 @@ messages about the conversion.
...
 
 ## Troubleshooting Guide
 
 The following steps can help diagnose common issues encountered while running
 HarbourBridge.
 
-### 1. Verify pg_dump configuration
+### 1. Verify driver configuration
+
+If you are using the dump approach then check that pg_dump/mysqldump is correctly
```
nevinheintze, line 291: Maybe start this paragraph with:

First, check that the driver is correctly configured to connect to your database. Driver configuration varies depending on the driver is used.

#### pgdump

If you are using pg_dump (-driver=pgdump), check that pg_dump is
correctly configured to connect to your PostgreSQL
database. Note that pg_dump uses the same options as psql to connect to your
database. See the [psql](https://www.postgresql.org/docs/9.3/app-psql.html) and
[pg_dump](https://www.postgresql.org/docs/9.3/app-pgdump.html) documentation.

Access to a PostgreSQL database is typically configured using the following
environment variables, which are standard across PostgreSQL utilities including
pg_dump:

```sh
PGHOST
PGPORT
PGUSER
```

It is also possible to configure access via pg_dump's command-line options
`--host`, `--port`, and `--username`.

#### Direct access to PostgreSQL

In this case, HarbourBridge connects directly to the PostgreSQL database
to retrieve table schema and data using the settings of the standard _PGHOST_,
_PGPORT_, _PGUSER_, _PGDATABASE_ environments variables. Password can be specified either in the _PGPASSWORD_ environment variable or provided at the password prompt.

and similarly for the other drivers.

README.md
Dharmil-searce: Yeah, adding subsections would be better.

README.md
Dharmil-searce: Done. You can check here ([LINK](https://github.com/cloudspannerecosystem/harbourbridge/blob/7ec5817781e9ecb8194e62918100a3c73a3b42f5/README.md#1-verify-driver-configuration)).

README.md
Dharmil-searce: Ah! ok. I didn't notice that change. Thanks for the update. 
Maybe it's worth adding the point (default driver is PGDUMP) in README. I will update it.

README.md
Dharmil-searce: Just added it for better clarity, if directly referring to `example usage` section. But yeah sure agreed, it's kind of redundant. I will delete this paragraph.

README.md
Dharmil-searce: Done. Updated as per the suggestion.

postgres/README.md
Dharmil-searce: Yes agreed. The command that is being used by PostgreSQL is `pg_dump` and driver that we have used is `pgdump` and hence the inconsistency. If we change in the code right now, then all functionality of postgres needs to be tested once after changing it.

I also don't feel too strongly but sure it would be better from consistency point of view. Let me know if you think that we should change it. 

One suggestion - If you think that we should change `pgdump-->pg_dump`, then it would be a better idea to do a separate PR for that change and keep this PR focused on documentation.

nevinheintze: Yes, I think that's a good idea.
Let's do it in a separate PR.

README.md
```diff
@@ -228,233 +253,40 @@ messages about the conversion.
...
-
-```sh
-pg_dump mydb | harbourbridge -prefix ~/spanner-eval-mydb/
-```
-
-would write the files into the directory `~/spanner-eval-mydb/`. Note
-that HarbourBridge will not create directories as it writes these files.
+Details on HarbourBridge example usage for PostgreSQL and MySQL can be
+found in the [PostgreSQL example usage](postgres/README.md#example-postgresql-usage)
+and [MySQL example usage](mysql/README.md#example-mysql-usage)respectively.
```
nevinheintze, line 258: Missing space before "respectively."

README.md
```diff
@@ -467,57 +299,86 @@ pg_dump:
 PGHOST
```
nevinheintze, line 299: Maybe format the environment variables in the same way as they are formatted in the following section (1.2) i.e. list them within the paragraph, using the same phrasing and formatting.

README.md
```diff
@@ -467,57 +299,86 @@ pg_dump:
...
+
+Access to a MySQL database is configured via mysqldump's command-line options `--host`,
+`--port`, and `--user`.
+
+#### 1.4 Direct access to MySQL
+
+In this case, HarbourBridge connects directly to the MySQL
+database to retrieve table schema and data using the setting of
+_MYSQLHOST_, _MYSQLPORT_, _MYSQLUSER_, _MYSQLDATABASE_ environment
+variables. Password can be specified either in the _MYSQL_PWD_
```
nevinheintze, line 332: Why is there an underscore in MYSQL_PWD?
None of the others have it -- shouldn't this just be MYSQLPWD or MYSQLPASSWORD for consistency?

README.md
```diff
@@ -467,57 +299,86 @@ pg_dump:
...
+
+#### 1.3 mysqldump
+
+If you are using mysqldump (-driver=mysqldump), check that mysqldump is
+correctly configured to connect to your MySQL
+database. Note that mysqldump uses the same options as mysql to connect to your
+database. See the [mysql](https://dev.mysql.com/doc/refman/8.0/en/mysql-commands.html) and
+[mysqldump](https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html) documentation.
+
+Access to a MySQL database is configured via mysqldump's command-line options `--host`,
```
nevinheintze, line 324: Maybe merge this sentence in with the previous paragraph? It seems kind of an afterthought here?

How about:

If you are using mysqldump (-driver=mysqldump), check that mysqldump is
correctly configured to connect to your MySQL database via the
command-line options `--host`, `--port`, and `--user`.
Note that mysqldump uses the same options as mysql to connect to your
database. See the [mysql](https://dev.mysql.com/doc/refman/8.0/en/mysql-commands.html) and
[mysqldump](https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html) documentation.

README.md
```diff
@@ -228,233 +253,40 @@ messages about the conversion.
...
-Spanner requires that `STRING` values be UTF-8 encoded. All Spanner functions
-and operators that act on `STRING` values operate on Unicode characters rather
-than bytes. Since we map many PostgreSQL types (including `TEXT` and `CHAR`
-types) to Spanner's `STRING` type, HarbourBridge is effectively a UTF-8 based
-tool.
-
-Note that the tool itself does not do any encoding/decoding or UTF-8 checks: it
-passes through data from pg_dump to Spanner. Internally, we use Go's string
-type, which supports UTF-8.
+for databases larger than a couple of GB. Details on HarbourBridge data conversion
```
nevinheintze, line 271: for consistency, lets use "few tens of GB".

README.md
```diff
@@ -113,16 +124,27 @@ The tool should now be installed as `$GOPATH/bin/harbourbridge`
 To use the tool on a PostgreSQL database called mydb, run
 
 ```sh
-pg_dump mydb | $GOPATH/bin/harbourbridge
```
nevinheintze, line 116: Just wondering about whether we should remove the $GOPATH/bin prefix on the command here.
If your PATH environment variable is set up "properly" to include $GOPATH/bin, then this will work nicely.

But I'm not sure whether we can assume this?
I'm wondering if golang new users would have things set up this way (I found many golang quickstart guides saying to add /usr/local/go/bin to PATH, but none mentioned adding GOPATH/bin).
Maybe leaving $GOPATH/bin/harbourbridge is more foolproof: regular golang users who have their path set up properly will know they can drop this, so the extra verbosity isn't a huge issue.
Thoughts?

README.md
Dharmil-searce: `MYSQL_PWD` is among the few standard variable environment available for connecting to mysql. Rest of the variables as we had previously discussed are not standard because in MySQL connection is usually done by command line arguments.
Thus, I thought it would be a good idea to follow the standard and keep it as `MYSQL_PWD` instead of following the consistency and keeping it as `MYSQLPASSWORD`.

README.md
Dharmil-searce: Most of the users would have set up their PATH environment variable 'properly' and it is mentioned in installation guide as well ([Install GO](https://golang.org/doc/install)). I think it would be fair enough to assume this. Also, I have not come across go packages in which for running the tool they are using `$GOPATH/bin/tool_name` command. Mostly it is being left to the user to configure PATH properly.

Maybe we can make some update our README and explicitly specify to configure PATH. Below changes should be done in that case :

1. Update the following instruction in `Quickstart guide`/`Before you begin` section :
'configure the GOPATH environment variable if it is not already configured' --> 'configure the PATH environment variable to '/usr/local/go/bin' if it is not already configured'

2. Update the following instruction in `Quickstart guide`/`Installing HarbourBridge` section :
'The tool should now be installed as `$GOPATH/bin/harbourbridge`' --> 'The tool should now be installed as `harbourbridge` in your PATH environment variable'

what do you think?

README.md
Dharmil-searce: Yup, looks good. Done.

README.md
hengfengli: I think it is fine to keep this way. If I have configured `$GOPATH/bin/` in my `$PATH`, I would know that I can directly run `harbourbridge` command without the prefix `$GOPATH/bin/`. In the current way, it is more clear where the tool has been installed.

README.md
hengfengli: This mixed style of environment variables looks confusing and inconsistent. We probably should agree on what kind of convention we should follow: 

1. Use the same style either `MYSQLPWD` or `MYSQL_PWD` for all environment variables. 
2. Environment variables should be consistent with the convention of the corresponding database, e.g., 
    * PostgreSQL: use _PGHOST_, _PGPORT_, _PGUSER_, _PGDATABASE_, _PGPASSWORD_. 
    * MySQL: use _MYSQL_HOST_, _MYSQL_PORT_, _MYSQL_USER_, _MYSQL_DATABASE_, _MYSQL_PWD_. 

(1) or (2)? 

I personally prefer (1).

Dharmil-searce: Ok sure. I also think (1) would be preferable. I will change it to `MYSQLPWD`.

README.md
Dharmil-searce: Ok sure. I will undo current change.

PR #71, opened 2020-07-06
Fix : mysql-infoschema getTables() and mysqldump SET TIMEZONE issue

Dharmil-searce:
Changes done in this PR:

- Updated query to fetch table in `getTables()`. Added a conditon of table_schema = `dbName`. `dbName` is passed as function argument and it's value is set to MYSQLDATABASE(env variable)

- Remove ignored schemas.

- Removed `buildTableName()` function.

- Removed `logStmtError()` in mysqldump for the case when value in `SET TIMEZONE` is a variable expression.

### DIFF
--- a/main.go
+++ b/main.go
@@ -269,7 +269,8 @@ func schemaFromSQL(driver string) (*internal.Conv, error) {
 		return nil, err
 	}
 	conv := internal.MakeConv()
-	err = ProcessInfoSchema(driver, conv, sourceDB)
+	tableSchema := os.Getenv("MYSQLDATABASE")
+	err = ProcessInfoSchema(driver, conv, sourceDB, tableSchema)
 	if err != nil {
 		return nil, err
 	}
@@ -290,7 +291,8 @@ func dataFromSQL(driver string, config spanner.BatchWriterConfig, client *sp.Cli
 	if err != nil {
 		return nil, err
 	}
-	err = SetRowStats(driver, conv, sourceDB)
+	tableSchema := os.Getenv("MYSQLDATABASE")
+	err = SetRowStats(driver, conv, sourceDB, tableSchema)
 	if err != nil {
 		return nil, err
 	}
@@ -312,7 +314,7 @@ func dataFromSQL(driver string, config spanner.BatchWriterConfig, client *sp.Cli
 		func(table string, cols []string, vals []interface{}) {
 			writer.AddRow(table, cols, vals)
 		})
-	err = ProcessSQLData(driver, conv, sourceDB)
+	err = ProcessSQLData(driver, conv, sourceDB, tableSchema)
 	if err != nil {
 		return nil, err
 	}
@@ -772,10 +774,10 @@ func ProcessDump(driver string, conv *internal.Conv, r *internal.Reader) error {
 }
 
 // ProcessInfoSchema invokes process infoschema function from a sql package based on driver selected.
-func ProcessInfoSchema(driver string, conv *internal.Conv, db *sql.DB) error {
+func ProcessInfoSchema(driver string, conv *internal.Conv, db *sql.DB, tableSchema string) error {
 	switch driver {
 	case MYSQL:
-		return mysql.ProcessInfoSchema(conv, db)
+		return mysql.ProcessInfoSchema(conv, db, tableSchema)
 	case POSTGRES:
 		return postgres.ProcessInfoSchema(conv, db)
 	default:
@@ -784,10 +786,10 @@ func ProcessInfoSchema(driver string, conv *internal.Conv, db *sql.DB) error {
 }
 
 // SetRowStats invokes SetRowStats function from a sql package based on driver selected.
-func SetRowStats(driver string, conv *internal.Conv, db *sql.DB) error {
+func SetRowStats(driver string, conv *internal.Conv, db *sql.DB, tableSchema string) error {
 	switch driver {
 	case MYSQL:
-		mysql.SetRowStats(conv, db)
+		mysql.SetRowStats(conv, db, tableSchema)
 	case POSTGRES:
 		postgres.SetRowStats(conv, db)
 	default:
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -269,7 +269,8 @@ func schemaFromSQL(driver string) (*internal.Conv, error) {
 		return nil, err
 	}
 	conv := internal.MakeConv()
-	err = ProcessInfoSchema(driver, conv, sourceDB)
+	tableSchema := os.Getenv("MYSQLDATABASE")
```
hengfengli, line 272: Two issues:

- This function is a generic method for all drivers. We shouldn't read a specific env var for all drivers. 
- `tableSchema` is not clear and might be under a specific context, e.g., MySQL background.

hengfengli: If I am using `pgdump` and `postgres`, I should not read `MYSQLDATABASE` env var here. What does `tableSchema` mean for `pgdump` and `postgres`?

main.go
```diff
@@ -772,10 +774,10 @@ func ProcessDump(driver string, conv *internal.Conv, r *internal.Reader) error {
 }
 
 // ProcessInfoSchema invokes process infoschema function from a sql package based on driver selected.
-func ProcessInfoSchema(driver string, conv *internal.Conv, db *sql.DB) error {
+func ProcessInfoSchema(driver string, conv *internal.Conv, db *sql.DB, tableSchema string) error {
 	switch driver {
 	case MYSQL:
-		return mysql.ProcessInfoSchema(conv, db)
+		return mysql.ProcessInfoSchema(conv, db, tableSchema)
```
hengfengli, line 780: Remove the `tableSchema` parameter. 

Change to: 

```
return mysql.ProcessInfoSchema(conv, db, os.Getenv("MYSQLDATABASE"))
```

main.go
```diff
@@ -784,10 +786,10 @@ func ProcessInfoSchema(driver string, conv *internal.Conv, db *sql.DB) error {
 }
 
 // SetRowStats invokes SetRowStats function from a sql package based on driver selected.
-func SetRowStats(driver string, conv *internal.Conv, db *sql.DB) error {
+func SetRowStats(driver string, conv *internal.Conv, db *sql.DB, tableSchema string) error {
 	switch driver {
 	case MYSQL:
-		mysql.SetRowStats(conv, db)
+		mysql.SetRowStats(conv, db, tableSchema)
```
hengfengli, line 792: Same above.

main.go
```diff
@@ -797,10 +799,10 @@ func SetRowStats(driver string, conv *internal.Conv, db *sql.DB) error {
 }
 
 // ProcessSQLData invokes ProcessSQLData function from a sql package based on driver selected.
-func ProcessSQLData(driver string, conv *internal.Conv, db *sql.DB) error {
+func ProcessSQLData(driver string, conv *internal.Conv, db *sql.DB, tableSchema string) error {
 	switch driver {
 	case MYSQL:
-		mysql.ProcessSQLData(conv, db)
+		mysql.ProcessSQLData(conv, db, tableSchema)
```
hengfengli, line 805: Same above.

mysql/infoschema.go
```diff
@@ -28,8 +28,8 @@ import (
 // ProcessInfoSchema performs schema conversion for source database
 // 'db'. Information schema tables are a broadly supported ANSI standard,
 // and we use them to obtain source database's schema information.
-func ProcessInfoSchema(conv *internal.Conv, db *sql.DB) error {
-	tables, err := getTables(db)
+func ProcessInfoSchema(conv *internal.Conv, db *sql.DB, tableSchema string) error {
```
hengfengli, line 31: `tableSchema` is a very specific term related to `information_schema`. When I first look at this variable name, I have no idea what it should be. I think `dbName` is more clear under `mysql` context. This is just my personal opinion.

main.go
```diff
@@ -290,7 +291,8 @@ func dataFromSQL(driver string, config spanner.BatchWriterConfig, client *sp.Cli
 	if err != nil {
 		return nil, err
 	}
-	err = SetRowStats(driver, conv, sourceDB)
+	tableSchema := os.Getenv("MYSQLDATABASE")
```
hengfengli, line 294: Same here. `dataFromSQL` is a generic method.

main.go
nevinheintze: Yes, the concept of schema is differs significantly between postgres and mysql.

For postgres, it is an extra layer of namespace within a database. By default tables are created using the 'public' schema, but can they can also be defined within a schema (namespace) e.g. myschema.table. When I connect to a postgres server, I only have access to the database I connected to.

In contrast, for mysql, schema is the same as database name. When I connect to a mysql server, I actually have access to all databases on that server, not just to the specific database I used in the connection.

I agree with Hengfeng: I think this has to be handled at the mysql level. And probably deserves a comment along the lines of:

// In MySQL, schema is the same as database name.

mysql/infoschema.go
nevinheintze: Yes, I agree: that would be easier to follow.
Maybe even call getTables's parameter dbName, and then have the comment about what is going on within getTables, where we get to make the connection between dbName and MySQL schema.

Note that sql.DB already effectively has the dbName embedded with it (dbName is part of the stuff passed to Open), but unfortunately there is now way to extract it from sql.DB.

Dharmil-searce: I have updated parameter to `dbName` and have added a function comment in `getTables()` function.

main.go
Dharmil-searce: Agreed. Updated to code and changed parameter to `dbName` as discussed in comments below. 

I am not sure where to place '// In MySQL, schema is the same as database name.' comment. I have added it above the MySQL case in `ProcessInfoSchema` function.

mysql/infoschema.go
```diff
@@ -172,25 +172,22 @@ type schemaAndName struct {
...
-	}
-	q := "SELECT table_schema, table_name FROM information_schema.tables where table_type = 'BASE TABLE'"
-	rows, err := db.Query(q)
+// getTables return list of tables in the selected database.
+// Note that sql.DB already effectively has the dbName
+// embedded within it (dbName is part of the DSN passed to sql.Open),
+// but unfortunately there is no way to extract it from sql.DB.
+func getTables(db *sql.DB, dbName string) ([]schemaAndName, error) {
+	q := "SELECT table_name FROM information_schema.tables where table_type = 'BASE TABLE' and table_schema=?"
+	rows, err := db.Query(q, dbName)
```
hengfengli, line 181: I think `// In MySQL, schema is the same as database name.` needs to be put in front of this line instead of the previous place.

main.go
```diff
@@ -774,8 +774,9 @@ func ProcessDump(driver string, conv *internal.Conv, r *internal.Reader) error {
 // ProcessInfoSchema invokes process infoschema function from a sql package based on driver selected.
 func ProcessInfoSchema(driver string, conv *internal.Conv, db *sql.DB) error {
 	switch driver {
+	// In MySQL, schema is the same as database name.
```
hengfengli, line 777: Move this to `getTables()`.

mysql/infoschema.go
Dharmil-searce: Got it. Done.

PR #73, opened 2020-07-09
Fix : MySQL reserved keywords and spaces issue in schema and table name

Dharmil-searce:
To handle case when MySQL reserved keywords or spaces are used in schema name(database) or table name, we have quoted schema and table name with backticks (`).


Dharmil-searce:
We cannot use quoting as dome in postgres (\`select count(\*) from "%s"."%s"\`), because double quotes(") are not allowed in MySQL and only backticks(\`) can be used in column, schema or table name. Also, using quoting like \`select count(\*) from \`%s\`.\`%s\`\` is not possible.
Hence, only option is "select count(*) from \`%s\`.\`%s\`".

PR #74, opened 2020-07-10
Fix: issue in parsing spatial datatypes and added reporting of auto_increment option

Dharmil-searce:


internal/report.go
```diff
@@ -251,6 +253,7 @@ var issueDB = map[SchemaIssue]struct {
 	Decimal:               {brief: "Spanner does not support decimal. This type mapping could lose precision and is not recommended for production use", severity: warning},
 	DecimalThatFits:       {brief: "Spanner does not support decimal, but this type mapping preserves the decimal's specified precision", severity: note},
 	Serial:                {brief: "Spanner does not support autoincrementing types", severity: warning},
+	AutoIncrement:         {brief: "Spanner does not support auto_increment attribute", severity: warning},
```
nevinheintze, line 256: I'm happy with this as is for this PR.

But we should discuss how to rationalize SchemaIssues (and reporting of them) e.g. we should probably have a single generic autoincrement issue and message.

mysql/mysqldump.go
```diff
@@ -466,7 +468,8 @@ func handleSpatialDatatype(conv *internal.Conv, chunk string, l [][]byte) ([]ast
 		return nil, true
 	}
 	for _, spatial := range MysqlSpatialDataTypes {
-		chunk = strings.ReplaceAll(chunk, " "+spatial, " text")
+		spatialRegexp := regexp.MustCompile("(?i)" + " " + spatial)
```
hengfengli, line 471: Can we move this to a global var? E.g., 

```go
var MysqlSpatialDataTypes = []string{"geometrycollection", "multipoint", "multilinestring", "multipolygon", "point", "linestring", "polygon", "geometry"}

var spatialRegexps = func() []*regexp.Regexp {
	l := make([]*regexp.Regexp, len(MysqlSpatialDataTypes))
	for i, spatial := range MysqlSpatialDataTypes {
		l[i] = regexp.MustCompile("(?i)" + " " + spatial)
	}
	return l
}()
```

This would bring `regexp.MustCompile` to the build time instead of runtime and can early detect errors if exist.

Dharmil-searce: Sure, I have moved it to a global variable as per your suggestion.

internal/report.go
Dharmil-searce: Maybe we can remove `Serial` and do following changes:

- Change issue description as `AutoIncrement : {brief: "Spanner does not support auto_increment attribute/data types"}`

- Change `[]internal.SchemaIssue{internal.Serial}` in postgres/toddl.go --> `[]internal.SchemaIssue{internal.AutoIncrement}`

- Remove `Serial` from `schemaIssue` in convert.go

There would be one problem in final reporting because reporting the type mapping would not be required in MySQL:

```
case AutoIncrement:
	l = append(l, fmt.Sprintf("Column '%s' is an autoincrement column. %s", srcCol, issueDB[i].brief))
```
while reporting for serial currently is as follows :
```
default:
	l = append(l, fmt.Sprintf("Column '%s': type %s is mapped to %s. %s", srcCol, srcType, spType, issueDB[i].brief))
```

What are your thoughts?

PR #75, opened 2020-07-14
Fix: 'Delimiter' error and ignore stored program statements

Dharmil-searce:
Changes included in this PR:

- Trimmed `*ast.` prefix from StmtNodes.

- Reporting of View, Index and Sequence statement.

- Handle stored programs (`function`,`procedure` and `trigger`) and `DELIMITER` statement as they are not supported by Pingcap parser. Added `skipUnsupported()` function to skip these statements.

- Refactored `handleSpatialDatatype()` to handle 'SPATIAL INDEX/KEY' and 'SRID'(spatial reference identifier) attribute.

### DIFF
--- a/internal/report.go
+++ b/internal/report.go
@@ -393,15 +393,15 @@ func ignoredStatements(conv *Conv) (l []string) {
 		switch s {
 		case "CreateFunctionStmt":
 			l = append(l, "functions")
-		case "CreateSeqStmt":
+		case "CreateSeqStmt", "CreateSequenceStmt":
 			l = append(l, "sequences")
-		case "CreatePLangStmt":
+		case "CreatePLangStmt", "CreateProcedureStmt":
 			l = append(l, "procedures")
 		case "CreateTrigStmt":
 			l = append(l, "triggers")
-		case "IndexStmt":
+		case "IndexStmt", "CreateIndexStmt":
 			l = append(l, "(non-primary) indexes")
-		case "ViewStmt":
+		case "ViewStmt", "CreateViewStmt":
 			l = append(l, "views")
 		}
 	}
--- a/mysql/mysqldump.go
+++ b/mysql/mysqldump.go
@@ -33,6 +33,7 @@ import (
 
 var valuesRegexp = regexp.MustCompile("\\((.*?)\\)")
 var insertRegexp = regexp.MustCompile("INSERT\\sINTO\\s(.*?)\\sVALUES\\s")
+var unsupportedRegexp = regexp.MustCompile("function|procedure|trigger")
 
 // MysqlSpatialDataTypes is an array of all MySQL spatial data types.
 var MysqlSpatialDataTypes = []string{"geometrycollection", "multipoint", "multilinestring", "multipolygon", "point", "linestring", "polygon", "geometry"}
@@ -43,6 +44,8 @@ var spatialRegexps = func() []*regexp.Regexp {
 	}
 	return l
 }()
+var spatialIndexRegex = regexp.MustCompile("(?i)\\sSPATIAL\\s")
+var spatialSridRegex = regexp.MustCompile("(?i)\\sSRID\\s\\d*")
 
 // ProcessMySQLDump reads mysqldump data from r and does schema or data conversion,
 // depending on whether conv is configured for schema mode or data mode.
@@ -140,7 +143,7 @@ func processStatement(conv *internal.Conv, stmt ast.StmtNode) bool {
 		processInsertStmt(conv, s)
 		return true
 	default:
-		conv.SkipStatement(reflect.TypeOf(stmt).String())
+		conv.SkipStatement(NodeType(stmt))
 	}
 	return false
 }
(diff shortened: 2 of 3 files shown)


mysql/mysqldump.go
```diff
@@ -410,9 +408,26 @@ func getTypeModsAndID(conv *internal.Conv, columnType string) (string, []int64)
...
 // handleParseError handles error while parsing mysqldump
 // statements and attempts at creating parsable chunk.
-// Error can be due to insert statement or unsupported Spatial
-// datatypes in create statement.
+// Error can be due to insert statement, unsupported Spatial
+// datatypes in create statement or unsupported stored programs.
 func handleParseError(conv *internal.Conv, chunk string, err error, l [][]byte) ([]ast.StmtNode, bool) {
+	// Check error for statements that are not supported by Pingcap parser.
+	// It includes delimiter, function, procedures and triggers.
+	// In case it's due to 'DELIMITER', we raparse till the chunk
```
nevinheintze, line 416: Typo: raparse --> reparse

mysql/mysqldump.go
```diff
@@ -410,9 +408,26 @@ func getTypeModsAndID(conv *internal.Conv, columnType string) (string, []int64)
...
+	// contains 2 delimiters. Eg:
+	//  1) 'DELIMITER $$' - First one is to define a delimiter
+	//     for functions, stored procedures or triggers.
+	//  2) 'DELIMITER ;' - It is restored to default delimiter i.e. ';'
+	// OR,
+	// Case when it's just due to functions, procedures or triggers
+	// without a delimiter statement.
+	errMsg := strings.ToLower(err.Error())
+	if unsupportedRegexp.MatchString(errMsg) || strings.Contains(errMsg, "delimiter") {
+		if strings.Count(strings.ToLower(chunk), "delimiter") == 1 {
```
nevinheintze, line 426: So if there was only one delimiter statement, we'd skip all statements from the delimiter statement onwards to the end of the file, at which point we'll generate a "Error parsing last NNN lines of input".

I don't have any good suggestions for handling this better.

If I understand correctly, mysqldump doesn't generate such dump files: it only generates dump files with pairs of delimiter statements -- is that a reasonable to assume?

mysql/mysqldump.go
```diff
@@ -410,9 +408,26 @@ func getTypeModsAndID(conv *internal.Conv, columnType string) (string, []int64)
 
 // handleParseError handles error while parsing mysqldump
 // statements and attempts at creating parsable chunk.
-// Error can be due to insert statement or unsupported Spatial
-// datatypes in create statement.
+// Error can be due to insert statement, unsupported Spatial
+// datatypes in create statement or unsupported stored programs.
 func handleParseError(conv *internal.Conv, chunk string, err error, l [][]byte) ([]ast.StmtNode, bool) {
+	// Check error for statements that are not supported by Pingcap parser.
```
nevinheintze, line 414: Suggest rephrasing slightly:
        // Check error for statements that are not supported by Pingcap parser such
        // as delimiter, function, procedures and triggers.
	// If the error is due to a delimiter, we reparse till the chunk
	// contains 2 delimiters, which is the typical way delimiters
        // are used by mysqldump e.g.
	//    DELIMITER ;; - First one redefines delimiter to something unusual
	//    - What follows is the definition of a function, procedure
	//    - or trigger, which can freely use ';' in its body.
	//    DELIMITER ; - Second one restores default delimiter
	// We also handle the case of functions, procedures or triggers
	// without a delimiter statement.

mysql/mysqldump.go
Dharmil-searce: Yes correct, MySQL generates dump with pairs of delimiter only and is limited to functions, stored procedures and triggers. I have also confirmed this with our DBA team.

> So if there was only one delimiter statement, we'd skip all statements from the delimiter statement onwards to the end of the file, at which point we'll generate a "Error parsing last NNN lines of input".

If dump file is generated using `mysqldump` command or best practices are followed as recommended by MySQL, then this case is not possible. But if anyone has written a dump file manually then it could be possible. 

Overall it's fair to assume that the dump file will be as per the standards.

mysql/mysqldump.go
Dharmil-searce: Fixed.

mysql/mysqldump.go
nevinheintze: Thanks!

mysql/mysqldump.go
```diff
@@ -43,6 +44,8 @@ var spatialRegexps = func() []*regexp.Regexp {
 	}
 	return l
 }()
+var spatialIndexRegex = regexp.MustCompile("(?i)\\sSPATIAL\\s")
```
hengfengli, line 47: nits: consider to group these variables together:  

```
var (
    a =
    b =
    ...
)
```

internal/report.go
```diff
@@ -393,15 +393,15 @@ func ignoredStatements(conv *Conv) (l []string) {
...
-		case "CreatePLangStmt":
+		case "CreatePLangStmt", "CreateProcedureStmt":
 			l = append(l, "procedures")
 		case "CreateTrigStmt":
 			l = append(l, "triggers")
-		case "IndexStmt":
+		case "IndexStmt", "CreateIndexStmt":
 			l = append(l, "(non-primary) indexes")
-		case "ViewStmt":
+		case "ViewStmt", "CreateViewStmt":
```
hengfengli, line 404: `default` case is missing.

mysql/mysqldump.go
```diff
@@ -369,8 +369,6 @@ func updateColsByOption(conv *internal.Conv, tableName string, col *ast.ColumnDe
 			column.Ignored.Check = true
 		case ast.ColumnOptionReference:
 			column.Ignored.ForeignKey = true
-		default:
```
hengfengli, line 372: Do we intend to remove this default case?

PR #79, opened 2020-08-16
feat: support DynamoDB schema conversion

hengfengli:
This is the first part (schema conversion) of supporting DynamoDB as a source DB. 

## TODOs

* Schema conversion: 
    - [x] initial schema conversion
    - [x] unit tests
    - [x] add `tables` flag
    - [x] add `schema-sample-size` flag
    - [x] convert Number to Numeric if possible; otherwise, convert it to a String

## Idea

The main idea is to scan a certain number of sample data from the table and make a best guess of the type data. If the table is small, we will scan the entire table. 

The algorithm logic can be described as follows: 

1. Calculate the percentage for each data type = count of rows for the data type / count of total scanned rows. 
2. Use an error threshold `0.1%` to remove noise: 
    1. If the percentage is lower than or equal to the error threshold, then we skip this data type. 
    2. If the percentage of Null and NotPresent is lower than or equal to the error threshold, then this column is not nullable (Columns with the Null data type or not presented are considered as the same case). 
3. For the remaining data types (except Null or NotPresent), we use a conflicting threshold `5%`. The percentage of a data type = count of rows for the data type / count of rows after removing Null and NotPresent. 
    1. If there is only one candidate, then the candidate is the data type. 
    2. If there are no candidates or more than one candidate, then `String` would be the data type. 

## Examples

Assume that the count of total scanned rows is `1000`. 

1)

```golang
stats := {
    "a": {
        "NumberFloat": 1,
	"NumberInt":   999,
    },
}
```

In the above example, `NumberFloat` is considered as a noise and will not be considered in Step `3`. 

2)

```golang
stats := {
    "a": {
	"NumberInt": 999,
    },
}
```

In the above example, the column is not nullable because the count of rows that are not presented is equal to 0.1%. 

3)

```golang
stats := {
    "a": {
	"NumberInt": 900,
    },
}
```

Now, the column is nullable because we have enough rows that are not presented. 

4) 

```golang
stats := {
    "a": {
        "String":    40,
	"NumberInt": 760,
    },
}
```

The percentage of `String` (40) is 5% over the normalized count (800). So `String` is not a candidate. `NumberInt` will be the data type. 

5) 

```golang
stats := {
    "a": {
        "String":    41
	"NumberInt": 759,
    },
}
```

Now, the percentage of `String` (41) is more than 5%. So `String` is a possible candidate. Therefore, we have two conflicting candidates, so we choose `String` as the data type. 

**Note**: More examples can be found in `dynamodb/schema_test.go::TestInferDataTypes`. 

## How to run it? 

If you have AWS credentials (`~/.aws/credentials`) and config (`~/.aws/config`) in your home directory, then you need to enable config via:

```bash
export AWS_SDK_LOAD_CONFIG=true
```

Otherwise, you need to set env vars explicitly: 

```bash
export AWS_ACCESS_KEY_ID=xxxxxx
export AWS_SECRET_ACCESS_KEY=xxxxxx
export AWS_DEFAULT_REGION=xxxxxx
```

Then, you can run the following command:

```bash
go run github.com/cloudspannerecosystem/harbourbridge -driver=dynamodb

# Or, specify which tables you want to migrate: 
go run github.com/cloudspannerecosystem/harbourbridge -driver=dynamodb -tables=Table1,Table2,Table3

# Or, specify to use how many rows to determine the schema: 
go run github.com/cloudspannerecosystem/harbourbridge -driver=dynamodb -schema-sample-size=1000
```

You will see an exported schema file, like:

```sql
-- Schema generated 2020-08-16 11:35:13
--
-- Spanner schema for source table Migration
--
CREATE TABLE Migration (
    Bool BOOL,                     -- From: Bool Bool
    Bytes BYTES(MAX),              -- From: Bytes Binary
    BytesSet ARRAY<BYTES(MAX)>,    -- From: BytesSet BinarySet
    List STRING(MAX),              -- From: List List
    Map STRING(MAX),               -- From: Map Map
    Number INT64,                  -- From: Number NumberInt
    NumberSet ARRAY<INT64>,        -- From: NumberSet NumberIntSet
    String STRING(MAX),            -- From: String String
    StringSet ARRAY<STRING(MAX)>,  -- From: StringSet StringSet
    Username STRING(MAX) NOT NULL  -- From: Username String
) PRIMARY KEY (Username)
```


dynamodb/schema.go
```diff
@@ -0,0 +1,339 @@
...
+
+		// Sort the slice reversely so the most frequent data type will be
+		// placed first.
+		sort.Slice(statItems, func(i, j int) bool {
+			// If counts are equal, then sort by names in alphabetical order.
+			if statItems[i].Count == statItems[j].Count {
+				return statItems[i].Type < statItems[j].Type
+			}
+			return statItems[i].Count > statItems[j].Count
+		})
```
nevinheintze, line 263: I'm wondering if this heuristic is the right one.
For example, if we have 60% Float and 40% String would we want to use Float?
For some low-levels of conflicting types, we might want to consider them errors or random spurious events.
But if there is a significant level of conflicting types, we might want to consider using String.

dynamodb/schema.go
```diff
@@ -0,0 +1,358 @@
...
+	ListTables(input *dynamodb.ListTablesInput) (*dynamodb.ListTablesOutput, error)
+	DescribeTable(input *dynamodb.DescribeTableInput) (*dynamodb.DescribeTableOutput, error)
+	Scan(input *dynamodb.ScanInput) (*dynamodb.ScanOutput, error)
+}
+
+// ProcessSchema performs schema conversion for source tables in a DynamoDB
+// database. We use the standard APIs to obtain source table's schema
+// information, including primary keys and secondary indexes. DynamoDB is a
+// schemaless database that some optional attributes can be missed or has
+// different data types in some rows. Therefore, we have to sample a number of
```
nevinheintze, line 56: Suggest rephrasing: I think the comment is a bit too detailed and is likely to get out-of-date.

How about:

ProcessSchema performs schema conversion for source tables in a DynamoDB
database. Since DynamoDB is a schemaless database, this process is imprecise.
We obtain schema information from two sources: from the table's metadata,
and from analyzing a sample of the table's rows.

dynamodb/schema.go
```diff
@@ -0,0 +1,358 @@
...
+	schemaToDDL(conv)
+	conv.AddPrimaryKeys()
+	return nil
+}
+
+func listTables(client dynamoClient) ([]string, error) {
+	var tables []string
+	var lastEvaluatedTableName *string
+
+	for {
```
nevinheintze, line 83: Would this be simpler as:

	var tables []string
         input := &dynamodb.ListTablesInput{}
	for {
		result, err := client.ListTables(input)
		if err != nil {
			return nil, err
		}
		for _, t := range result.TableNames {
			tables = append(tables, *t)
		}

		if result.LastEvaluatedTableName == nil {
			return tables, nil
		}
		input.lastEvaluatedTableName = result.LastEvaluatedTableName
	}

i.e. remove lastEvaluatedTableName var and initialize input once before loop and reuse it.

dynamodb/schema.go
```diff
@@ -0,0 +1,358 @@
...
+	Name     string
+	Nullable bool
+}
+
+type index struct {
+	Name string
+	Keys []string
+}
+
+func (s *dynamoDBSchema) parseIndexes(client dynamoClient) error {
```
nevinheintze, line 137: Maybe analyzeMetadata instead of parseIndexes?
I'm not sure DynamoDB considers primary keys to be indexes?

dynamodb/schema.go
```diff
@@ -0,0 +1,358 @@
...
+	input := &dynamodb.DescribeTableInput{
+		TableName: aws.String(s.TableName),
+	}
+
+	result, err := client.DescribeTable(input)
+	if err != nil {
+		return fmt.Errorf("failed to make a DescribeTable API call for table %v: %v", s.TableName, err)
+	}
+
```
nevinheintze, line 146: Should we use AttributeDefinitions to get types for keys?

Also, if you give attribute definitions for other columns, how are they treated? Are the specified types enforced? (my guess is probably not, but I don't know for sure).
Should we use types for these addition attribute definitions to help build the schema?

dynamodb/schema.go
```diff
@@ -0,0 +1,358 @@
...
+	// The order of primary keys is important.
+	var schemaPKeys []schema.Key
+	for _, colName := range s.PrimaryKeys {
+		schemaPKeys = append(schemaPKeys, schema.Key{Column: colName})
+		colDef := colDefs[colName]
+		colDefs[colName] = schema.Column{
+			Name:    colName,
+			Type:    colDef.Type,
+			NotNull: true,
```
nevinheintze, line 319: Just checking: in dynamoDB, NULL is not a legal value for a key.

dynamodb/schema.go
```diff
@@ -0,0 +1,358 @@
...
+
+func (s *dynamoDBSchema) scanSampleData(client dynamoClient, sampleSize int64) (map[string]map[string]int64, int64, error) {
+	// A map from column name to a count map of possible data types.
+	stats := make(map[string]map[string]int64)
+	var lastEvaluatedKey map[string]*dynamodb.AttributeValue
+	var count int64
+
+	for {
+		// Build the query input parameters
+		params := &dynamodb.ScanInput{
```
nevinheintze, line 172: Do we need to create dynamodb.ScanInput for each iteration of the loop?
Could we just define one before the start of loop and re-use it.
And then get rid of the lastEvaluatedKey var (and line 175 to 177) and instead change line 203 to:

params.ExclusiveStartKey = result.LastEvaluatedKey

I think that would be less code and a bit easier to understand.

dynamodb/schema.go
```diff
@@ -0,0 +1,358 @@
...
+			return nil, 0, fmt.Errorf("failed to make Query API call for table %v: %v", s.TableName, err)
+		}
+
+		// Iterate the items returned
+		for _, attrsMap := range result.Items {
+			for attrName, attr := range attrsMap {
+				if _, ok := stats[attrName]; !ok {
+					stats[attrName] = make(map[string]int64)
+				}
+				incDyDataTypeCount(attrName, attr, stats[attrName])
```
nevinheintze, line 191: incDyDataTypeCount --> incTypeCount

dynamodb/schema.go
```diff
@@ -0,0 +1,358 @@
...
+
+func incDyDataTypeCount(attrName string, attr *dynamodb.AttributeValue, s map[string]int64) {
+	if attr.S != nil {
+		incCount(s, typeString)
+	} else if attr.BOOL != nil {
+		incCount(s, typeBool)
+	} else if attr.N != nil {
+		if int64Parsable(*attr.N) {
+			incCount(s, typeNumberInt)
+		} else {
```
nevinheintze, line 216: Maybe check that *attr.N is parseable as a float64?
And then use string if not.

dynamodb/schema.go
```diff
@@ -0,0 +1,358 @@
...
+			break
+		}
+		// If there are more rows, then continue.
+		lastEvaluatedKey = result.LastEvaluatedKey
+	}
+	return stats, count, nil
+}
+
+func incDyDataTypeCount(attrName string, attr *dynamodb.AttributeValue, s map[string]int64) {
+	if attr.S != nil {
```
nevinheintze, line 209: Maybe simple as switch statement?

switch {
  case attr.S != nil:
     ....
  case attr.BOOL != nil:
     ....
   etc

dynamodb/schema.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+func int64Parsable(n string) bool {
+	if _, err := strconv.ParseInt(n, 10, 64); err == nil {
+		return true
+	}
+	return false
+}
+
+func incCount(m map[string]int64, key string) {
```
nevinheintze, line 353: I don't think this function is needed.
You can just write:

    m[key]++

If a key isn't in the map, its default value is 0.

dynamodb/schema.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+func (s *dynamoDBSchema) inferDataTypes(stats map[string]map[string]int64, rows int64) {
+	if s.ColumnTypes == nil {
+		s.ColumnTypes = make(map[string]colType)
+	}
+
+	for col, countMap := range stats {
+		var statItems, candidates []statItem
+		var presentRows, normRows int64
```
nevinheintze, line 254: What's the reason to distinguish normRows and presentRows?
They can only differ by a very tiny amount.
Or am I missing some corner case?

dynamodb/schema.go
hengfengli: The `AttributeDefinitions` only returns the columns of a primary key, not any optional columns. 

When I try to create a table via: 

```bash
aws dynamodb create-table --table-name Migration \                                                                
                                  --attribute-definitions AttributeName=Username,AttributeType=S AttributeName=Unused,AttributeType=S \
                                  --key-schema AttributeName=Username,KeyType=HASH \
                                  --provisioned-throughput ReadCapacityUnits=5,WriteCapacityUnits=5
```

As you can see, `attribute-definitions` does not match `key-schema` and then I got the following error: 

```bash
An error occurred (ValidationException) when calling the CreateTable operation: One or more parameter values were invalid: Number of attributes in KeySchema does not exactly match number of attributes defined in AttributeDefinitions. 
```

What I mean is that `AttributeDefinitions` must match `KeySchema`, which is the primary key. No optional columns will appear in `AttributeDefinitions`. 

> Should we use AttributeDefinitions to get types for keys?

Yes, we can, but only for columns of a primary key. We can use the `AttributeDefinitions` as a base schema, which only contains the types of columns in a primary key. Then, use sampling to guess the types of optional columns. 

The current implementation didn't do this because the types of columns in a primary key are forced and they have an absolute win in the sampling process. I try to avoid the process to add optional columns to a base table. If you think using `AttributeDefinitions` as a base table is a better idea, then I can update it. 

BTW: worth to mention that for secondary indexes, there are no `AttributeDefinitions`: 

Local secondary index: 

```json
"LocalSecondaryIndexes": [
            {
                "IndexName": "AlbumTitleIndex",
                "KeySchema": [
                    {
                        "AttributeName": "Artist",
                        "KeyType": "HASH"
                    },
                    {
                        "AttributeName": "AlbumTitle",
                        "KeyType": "RANGE"
                    }
                ],
                "Projection": {
                    "ProjectionType": "INCLUDE",
                    "NonKeyAttributes": [
                        "Genre",
                        "Year"
                    ]
                },
                "IndexSizeBytes": 0,
                "ItemCount": 0,
                "IndexArn": "arn:aws:dynamodb:us-west-2:123456789012:table/MusicCollection/index/AlbumTitleIndex"
            }
        ]
```

Global secondary index: 

```json
"GlobalSecondaryIndexes": [
            {
                "IndexName": "GameTitleIndex",
                "KeySchema": [
                    {
                        "AttributeName": "GameTitle",
                        "KeyType": "HASH"
                    },
                    {
                        "AttributeName": "TopScore",
                        "KeyType": "RANGE"
                    }
                ],
                "Projection": {
                    "ProjectionType": "INCLUDE",
                    "NonKeyAttributes": [
                        "UserId"
                    ]
                },
                "IndexStatus": "CREATING",
                "ProvisionedThroughput": {
                    "NumberOfDecreasesToday": 0,
                    "ReadCapacityUnits": 10,
                    "WriteCapacityUnits": 5
                },
                "IndexSizeBytes": 0,
                "ItemCount": 0,
                "IndexArn": "arn:aws:dynamodb:us-west-2:123456789012:table/GameScores/index/GameTitleIndex"
            }
        ]
```

Ref: https://docs.aws.amazon.com/cli/latest/reference/dynamodb/create-table.html

dynamodb/schema.go
hengfengli: No. You can't use a column with NULL type for a primary key. 

```bash
aws dynamodb create-table --table-name Migration \                                                                             
                              --attribute-definitions AttributeName=Username,AttributeType=NULL \
                              --key-schema AttributeName=Username,KeyType=HASH \
                              --provisioned-throughput ReadCapacityUnits=5,WriteCapacityUnits=5

An error occurred (ValidationException) when calling the CreateTable operation: 1 validation error detected: Value 'NULL' at 'attributeDefinitions.1.member.attributeType' failed to satisfy constraint: Member must satisfy enum value set: [B, N, S]
```

It only allows String, Number, and Bytes as keys and I guess because they are hashable?

dynamodb/schema.go
hengfengli: Maybe the name is confusing. 

`rows`: the total number of sample rows. 
`presentRows`: the number of rows that have a value on this column. 
`rows-presentRows`: the number of rows that do not have a value on this column. It used to check whether or not this column can be nullable. 

`normRows`: the number of rows after removing erroneous ones (the percent of type is <= err threshold) and nullable rows. 

I will use an example to illustrate this: 

```bash
# assume error threshold = 0.1%
rows = 1000

stats = {
    "Username": {
        "String":    800,
        "NumberInt": 1,
        "Bytes": 1,
        "Bool": 1,
    },
}
```

Then, 

```
presentRows = 800 + 1 + 1 + 1 = 803
rows-presentRows = 1000 - 803 = 197 > 1  # This means that this column should be nullable. 

normRows = 800
```

If you think `normRows` and `presentRows` only have a very tiny difference, we can just use `presentRows`.

dynamodb/schema.go
hengfengli: Fixed.

dynamodb/schema.go
hengfengli: Yes, we can do this. Fixed.

dynamodb/schema.go
hengfengli: `presentRows` is tested by the test case: 

```bash
		"enough_null_row": {
			"NumberInt": 900,
		},
```

`normRows` is tested by the test case: 

```bash
		"equal_conflict_row_after_norm": {
			"String":    400,
			"NumberInt": 400,
		},
```

dynamodb/schema.go
hengfengli: Nice suggestion. Fixed.

dynamodb/schema.go
hengfengli: I'm not sure whether or not it considers `parsable`, even if the number cannot be represented accurately in float64. 

We also should consider whether or not this number can be converted to a Numeric. I'll give it a try.

dynamodb/schema.go
nevinheintze: Ok, got it.
Thanks for all of the clarifications.
Yes, that makes perfect sense now.
And agreed that if key type is enforced, then the sampled data is guaranteed to give us the right type.

I guess the only corner case is: what happens if the table is empty? Hmm, thinking about it, I guess all you could do in that case is use the meta data to add a Spanner table with just the key column and no other columns, which isn't very useful. Ok, let's go with what you have.

dynamodb/schema.go
nevinheintze: Thanks!

dynamodb/schema.go
nevinheintze: What puzzled me as I was looking over the code was why there were two counts. And it made the code harder to understand since there were two very similar concepts to keep track of.

At line 253, where we check the error threshold, the comment explains that there is a high chance they are mistakes and that we should discard them.

We then include them for the purposes of determining nullable, but ignore them for determining types.

If we justify ignoring them because they are likely mistakes, shouldn't we ignore them for both calculations?

I think either we should be consistent, or we should have a good reason (and a comment) to explain why we are doing something different for these two cases.

dynamodb/schema.go
```diff
@@ -0,0 +1,367 @@
...
+	return schema.Table{
+		Name:        s.TableName,
+		ColNames:    colNames,
+		ColDefs:     colDefs,
+		PrimaryKeys: schemaPKeys,
+		Indexes:     indexes,
+	}
+}
+
+func numericParsable(n string) bool {
```
nevinheintze, line 342: This is a bit subtle.
I think it would be worth a function-level comment along the lines of:

numericParsable determines whether its argument is a valid
Spanner numeric values. This is based on <link to Spanner documentation of Spanner numeric datatype>.

and maybe a comment in the code about what the calculation is doing.

dynamodb/schema.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+func incTypeCount(attrName string, attr *dynamodb.AttributeValue, s map[string]int64) {
+	switch {
+	case attr.S != nil:
+		s[typeString]++
+	case attr.BOOL != nil:
+		s[typeBool]++
+	case attr.N != nil:
+		if numericParsable(*attr.N) {
```
nevinheintze, line 203: Maybe leave a comment about our treatment of numeric and possibilities for future modifications?

Something like:

// We map the DynamoDB Number type into Spanner's NUMERIC type
// if it fits and STRING otherwise. Note that DyanamoDB's Number
// type has more precision/range than Spanner's NUMERIC.
// We could potentially do a more detailed analysis and see if
// the number fits in an INT64 or FLOAT64, but we've chosen to
// keep the analysis simple for the moment.

dynamodb/schema.go
```diff
@@ -0,0 +1,367 @@
...
+	stats, count, err := dySchema.scanSampleData(client, sampleSize)
+	if err != nil {
+		return err
+	}
+	dySchema.inferDataTypes(stats, count)
+	conv.SrcSchema[table] = dySchema.genericSchema()
+	return nil
+}
+
+type dynamoDBSchema struct {
```
nevinheintze, line 112: This datatype (along with colType and index) are very similar to schema.go.
Why not use schema.go directly?

Put another way, genericSchema() seems mostly a copy function - the only thing it seems to do is sort the list of column names. Could we just get rid of it?

main.go
```diff
@@ -78,6 +85,8 @@ func init() {
 	flag.StringVar(&instanceOverride, "instance", "", "instance: Spanner instance to use")
 	flag.StringVar(&filePrefix, "prefix", "", "prefix: file prefix for generated files")
 	flag.StringVar(&driverName, "driver", "pg_dump", "driver name: flag for accessing source DB or dump files (accepted values are \"pg_dump\", \"postgres\", \"mysqldump\", and \"mysql\")")
+	flag.StringVar(&tables, "tables", "", "tables: tables to use for migration (only for DynamoDB)")
```
nevinheintze, line 88: I think we should probably have this for all drivers, not just DynamoDB.

Is there some reason we specifically need this for Dynamo?

main.go
```diff
@@ -318,6 +334,21 @@ func dataFromSQL(driver string, config spanner.BatchWriterConfig, client *sp.Cli
 	return writer, nil
 }
 
+func schemaFromDynamoDB(t string, sampleSize int64) (*internal.Conv, error) {
```
nevinheintze, line 337: Maybe rename: t --> tables.
For consistency e.g. with schemaConv.

I know I have some single letter function args, but in retrospect that is probably a mistake.

My understanding is that function args should be somewhat descriptive (but brief).

main.go
```diff
@@ -186,12 +200,14 @@ func toSpanner(driver, projectID, instanceID, dbName string, ioHelper *ioStreams
 	return nil
 }
 
-func schemaConv(driver string, ioHelper *ioStreams) (*internal.Conv, error) {
+func schemaConv(driver, tables string, sampleSize int64, ioHelper *ioStreams) (*internal.Conv, error) {
```
nevinheintze, line 203: Maybe drop sampleSize since it is only relevant for DynamoDB, and have schemaFromDynamoDB access the flag directly?

spanner/ddl/ast.go
```diff
@@ -42,13 +42,15 @@ const (
...
+	// Numeric represent NUMERIC type.
+	Numeric string = "NUMERIC"
 	// MaxLength is a sentinel for Type's Len field, representing the MAX value.
 	MaxLength = math.MaxInt64
 )
 
 // Type represents the type of a column.
 //     type:
-//        { BOOL | INT64 | FLOAT64 | STRING( length ) | BYTES( length ) | DATE | TIMESTAMP }
+//        { BOOL | INT64 | FLOAT64 | STRING( length ) | BYTES( length ) | DATE | TIMESTAMP | NUMERIC }
```
nevinheintze, line 53: We should probably update postgres and mysql with NUMERIC.
Probably best to track that with an issue?

nevinheintze (approved):

dynamodb/schema.go
hengfengli: > If we justify ignoring them because they are likely mistakes, shouldn't we ignore them for both calculations?

Yes, we can. But in this case, we still need to keep `presentRows`, which is used to calculate how many rows are not given. If we ignore discarded rows when determining nullable, then the code would become: 

```go
nullable := float64(rows-presentRows)/float64(normRows) > errThreshold
```

We changed the denominator from `rows` to `normRows`. 

The simple change would be to remove `normRows` so we always use `rows` instead of `normRows`.

main.go
hengfengli: We don't have to specify this parameter for Dynamo. In that case, all Dynamo tables under the user account will be migrated. With this `tables` parameter, user can choose to only migrate a single table or a list of tables instead of all of them. 

Yeah, we can let all drivers to support this, but currently it only supports for DynamoDB driver.

spanner/ddl/ast.go
hengfengli: I created an issue #87 for this.

main.go
hengfengli: Fixed.

dynamodb/schema.go
hengfengli: Fixed. This is a good suggestion. Looks much better now.

main.go
nevinheintze: We probably need to give some thought to flag naming if we're going to have this for all drivers.
In general, we've tried to have our flags consistent with those used by pg_dump (and to a lesser extend, mysqldump).

In this case, pg_dump uses '-t' (and the argument is a pattern).
And mysqldump just uses command arguments i.e.

  mysqldump -u username -p db_name table1_name table2_name table3_name

but it also has a flag '-tables' to indicate that all arguments should be considered tables (first arg is no longer db_name).

A couple of options:
a) drop '--tables' for the time being (assuming it isn't fundamental for Dynamo), and add it as a separate feature across all drivers.

b) check-in current code. Main drawback is it is kind of weird to have a flag that is specific to one driver (all of our other flags are generic), and we might run into issues when we make it generic.

c) come up with a flag naming schema we can use for pg and mysql, and implementing it for dynamo in this PR (with a TODO to implement it for the others).

I have a slight preference for (a) because it is simple with no loose ends, but open to other options.

hengfengli: You're right. It deserves more considerations. I'll remove it from this change.

dynamodb/schema.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+
+		// If this column is in the primary key, then it cannot be null.
+		nullable := false
+		if !isPKey {
+			nullable = float64(rows-presentRows)/float64(rows) > errThreshold
+		}
+
+		for _, si := range statItems {
+			if float64(si.Count)/float64(rows) > conflictThreshold {
```
nevinheintze, line 275: rows --> presentRows

nevinheintze (approved):
Thanks for all of the changes!

main.go
```diff
@@ -60,6 +63,8 @@ const (
 	MYSQLDUMP string = "mysqldump"
 	// MYSQL is the driver name for MySQL.
 	MYSQL string = "mysql"
+	// DYNAMODB is the driver name for AWS DynamoDB.
```
nevinheintze, line 66: Maybe update comment:

// DYNAMODB is the driver name for AWS DynamoDB.
// This is an experimental driver; implementation in progress.

PR #80, opened 2020-09-02
feat: support DynamoDB data conversion

hengfengli:
This PR is based on #79. 

* Data conversion: 
    - [x] initial data conversion
    - [x] encode List & Map to json strings
    - [x] add a flag for ConsistentRead mode scan. By default, set `ConsistentRead=false`. 
    - [x] set total rows & report progress (use `describe_table` api to get an inaccurate number of total rows). 
    - [x] unit tests

### DIFF
--- a/dynamodb/data.go
+++ b/dynamodb/data.go
@@ -43,9 +43,8 @@ func ProcessData(conv *internal.Conv, client dynamoClient, consistentRead bool) 
 		}
 
 		var lastEvaluatedKey map[string]*dynamodb.AttributeValue
-
 		for {
-			// Build the query input parameters
+			// Build the query input parameters.
 			params := &dynamodb.ScanInput{
 				TableName:      aws.String(srcTable),
 				ConsistentRead: &consistentRead,
@@ -54,32 +53,39 @@ func ProcessData(conv *internal.Conv, client dynamoClient, consistentRead bool) 
 				params.ExclusiveStartKey = lastEvaluatedKey
 			}
 
-			// Make the DynamoDB Query API call
+			// Make the DynamoDB Query API call.
 			result, err := client.Scan(params)
 			if err != nil {
 				return fmt.Errorf("failed to make Query API call for table %v: %v", srcTable, err)
 			}
 
-			// Iterate the items returned
+			// Iterate the items returned.
 			for _, attrsMap := range result.Items {
+				var srcStrVals []string
 				var cvtVals []interface{}
+				var cvtErrs map[string]string
 				for i, srcColName := range srcSchema.ColNames {
 					var cvtVal interface{}
 					if attrsMap[srcColName] == nil {
 						cvtVal = nil
 					} else {
-						// Convert data to the target type
+						// Convert data to the target type.
 						cvtVal, err = cvtColValue(attrsMap[srcColName], srcSchema.ColDefs[srcColName].Type.Name, spSchema.ColDefs[spCols[i]].T.Name)
 						if err != nil {
-							return fmt.Errorf("failed to convert column: %v to %v", attrsMap[srcColName], spSchema.ColDefs[spCols[i]])
+							cvtErrs[srcColName] = fmt.Sprintf("failed to convert column: %v", srcColName)
 						}
 					}
+					srcStrVals = append(srcStrVals, attrsMap[srcColName].GoString())
 					cvtVals = append(cvtVals, cvtVal)
 				}
-
-				conv.WriteRow(srcTable, spTable, spCols, cvtVals)
+				if len(cvtErrs) == 0 {
+					conv.WriteRow(srcTable, spTable, spCols, cvtVals)
+				} else {
+					conv.Unexpected(fmt.Sprintf("Errors while converting data: %s\n", cvtErrs))
+					conv.StatsAddBadRow(srcTable, conv.DataMode())
+					conv.CollectBadRow(srcTable, srcSchema.ColNames, srcStrVals)
+				}
 			}
-
 			if result.LastEvaluatedKey == nil {
 				break
 			}
@@ -87,7 +93,6 @@ func ProcessData(conv *internal.Conv, client dynamoClient, consistentRead bool) 
 			lastEvaluatedKey = result.LastEvaluatedKey
 		}
 	}
-
 	return nil
 }
 
@@ -110,13 +115,13 @@ func cvtColValue(attrVal *dynamodb.AttributeValue, srcType string, spType string
 		case typeMap:
 			b, err := json.Marshal(attrVal)
 			if err != nil {
-				return nil, fmt.Errorf("failed to encode a map object: %v to a json string", attrVal.GoString())
+				return nil, fmt.Errorf("failed to convert a map object: %v to a json string", attrVal.GoString())
 			}
 			return string(b), nil
 		case typeList:
 			b, err := json.Marshal(attrVal)
 			if err != nil {
-				return nil, fmt.Errorf("failed to encode a list object: %v to a json string", attrVal.GoString())
+				return nil, fmt.Errorf("failed to convert a list object: %v to a json string", attrVal.GoString())
 			}
 			return string(b), nil
 		case typeString:
--- a/dynamodb/data.go
+++ b/dynamodb/data.go
@@ -30,7 +30,7 @@ import (
 // on the source and Spanner schemas), and write it to Spanner. If we can't
 // get/process data for a table, we skip that table and process the remaining
 // tables.
-func ProcessData(conv *internal.Conv, client dynamoClient) error {
+func ProcessData(conv *internal.Conv, client dynamoClient, consistentRead bool) error {
 	for srcTable, srcSchema := range conv.SrcSchema {
 		spTable, err1 := internal.GetSpannerTable(conv, srcTable)
 		spCols, err2 := internal.GetSpannerCols(conv, srcTable, srcSchema.ColNames)
@@ -47,7 +47,8 @@ func ProcessData(conv *internal.Conv, client dynamoClient) error {
 		for {
 			// Build the query input parameters
 			params := &dynamodb.ScanInput{
-				TableName: aws.String(srcTable),
+				TableName:      aws.String(srcTable),
+				ConsistentRead: &consistentRead,
 			}
 			if lastEvaluatedKey != nil {
 				params.ExclusiveStartKey = lastEvaluatedKey
--- a/dynamodb/data.go
+++ b/dynamodb/data.go
@@ -30,7 +30,7 @@ import (
 // on the source and Spanner schemas), and write it to Spanner. If we can't
 // get/process data for a table, we skip that table and process the remaining
 // tables.
-func ProcessData(conv *internal.Conv, client dynamoClient, consistentRead bool) error {
+func ProcessData(conv *internal.Conv, client dynamoClient) error {
 	for srcTable, srcSchema := range conv.SrcSchema {
 		spTable, err1 := internal.GetSpannerTable(conv, srcTable)
 		spCols, err2 := internal.GetSpannerCols(conv, srcTable, srcSchema.ColNames)
@@ -46,8 +46,7 @@ func ProcessData(conv *internal.Conv, client dynamoClient, consistentRead bool) 
 		for {
 			// Build the query input parameters.
 			params := &dynamodb.ScanInput{
-				TableName:      aws.String(srcTable),
-				ConsistentRead: &consistentRead,
+				TableName: aws.String(srcTable),
 			}
 			if lastEvaluatedKey != nil {
 				params.ExclusiveStartKey = lastEvaluatedKey
--- a/dynamodb/data.go
+++ b/dynamodb/data.go
@@ -59,7 +59,6 @@ func ProcessData(conv *internal.Conv, client dynamoClient) error {
 				return fmt.Errorf("failed to make Query API call for table %v: %v", srcTable, err)
 			}
 
-			var count int64
 			// Iterate the items returned
 			for _, attrsMap := range result.Items {
 				var cvtVals []interface{}
@@ -78,10 +77,7 @@ func ProcessData(conv *internal.Conv, client dynamoClient) error {
 				}
 
 				conv.WriteRow(srcTable, spTable, spCols, cvtVals)
-				count++
 			}
-
-			conv.Stats.Rows[srcTable] += count
 
 			if result.LastEvaluatedKey == nil {
 				break
--- a/dynamodb/data.go
+++ b/dynamodb/data.go
@@ -134,7 +134,7 @@ func cvtColValue(attrVal *dynamodb.AttributeValue, srcType string, spType string
 		case typeMap, typeList:
 			// For typeMap and typeList, attrVal is a very verbose data
 			// structure that contains null entries for unused type cases. We
-			// strip these out using marshallAttrValue. If it important that the
+			// strip these out using stripNull. If it important that the
 			// Spanner values can be easily unmarshalled back to
 			// dynamodb.AttributeValue types, then replace the following five
 			// lines with just:
--- a/dynamodb/data.go
+++ b/dynamodb/data.go
@@ -22,6 +22,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/cloudspannerecosystem/harbourbridge/internal"
+	"github.com/cloudspannerecosystem/harbourbridge/schema"
 	"github.com/cloudspannerecosystem/harbourbridge/spanner/ddl"
 )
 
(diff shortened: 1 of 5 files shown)


hengfengli:
@nevinheintze I have done the implementation. Please take a review. Two main questions: 

1. There are too many null fields when converting List and Map (dynamodb) to a String (spanner). You can take a look at [here](https://github.com/cloudspannerecosystem/harbourbridge/pull/80/files#diff-14c703eaae9731ccdc2115db6e01fa9cfb4073d50c332240974c018e3b7b6f22R149-R150). The good part is that it is easier for users to convert it back to a Map/List (a struct in dynamodb sdk). 
2. For the consistent read mode, do we need to enable it for both schema and data conversion? Because they are two different api calls and it means that we cannot guarantee a consistent view when sampling for schema and doing the actual data conversion.

dynamodb/data.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/cloudspannerecosystem/harbourbridge/spanner/ddl"
+)
+
+// ProcessData performs data conversion for DynamoDB database. For each table,
+// we extract data using Scan requests, convert the data to Spanner data (based
+// on the source and Spanner schemas), and write it to Spanner. If we can't
+// get/process data for a table, we skip that table and process the remaining
+// tables.
+func ProcessData(conv *internal.Conv, client dynamoClient, consistentRead bool) error {
+	for srcTable, srcSchema := range conv.SrcSchema {
```
nevinheintze, line 34: Quick question: we should change postgres/infoschema.go ProcessSQLData to use this same iteration structure https://github.com/cloudspannerecosystem/harbourbridge/blob/master/postgres/infoschema.go#L75. What do you think?
Not suggesting we should to it in this CL.

dynamodb/data.go
```diff
@@ -0,0 +1,161 @@
...
+		spSchema, ok := conv.SpSchema[spTable]
+		if err1 != nil || err2 != nil || !ok {
+			conv.Stats.BadRows[srcTable] += conv.Stats.Rows[srcTable]
+			conv.Unexpected(fmt.Sprintf("Can't get cols and schemas for table %s: err1=%s, err2=%s, ok=%t",
+				srcTable, err1, err2, ok))
+			continue
+		}
+
+		var lastEvaluatedKey map[string]*dynamodb.AttributeValue
+
```
nevinheintze, line 46: Maybe drop one of these blank lines?

dynamodb/data.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/cloudspannerecosystem/harbourbridge/internal"
+	"github.com/cloudspannerecosystem/harbourbridge/spanner/ddl"
+)
+
+// ProcessData performs data conversion for DynamoDB database. For each table,
+// we extract data using Scan requests, convert the data to Spanner data (based
+// on the source and Spanner schemas), and write it to Spanner. If we can't
+// get/process data for a table, we skip that table and process the remaining
+// tables.
+func ProcessData(conv *internal.Conv, client dynamoClient, consistentRead bool) error {
```
nevinheintze, line 33: This is getting to be a pretty long function. Maybe consider refactoring?

dynamodb/data.go
```diff
@@ -0,0 +1,161 @@
...
+			conv.Stats.BadRows[srcTable] += conv.Stats.Rows[srcTable]
+			conv.Unexpected(fmt.Sprintf("Can't get cols and schemas for table %s: err1=%s, err2=%s, ok=%t",
+				srcTable, err1, err2, ok))
+			continue
+		}
+
+		var lastEvaluatedKey map[string]*dynamodb.AttributeValue
+
+		for {
+			// Build the query input parameters
```
nevinheintze, line 48: Missing '.' at end of comment -- I think the convention we've adopted elsewhere is to add a dot.
Same for a number of other comments in this file.

dynamodb/data.go
```diff
@@ -0,0 +1,161 @@
...
+					} else {
+						// Convert data to the target type
+						cvtVal, err = cvtColValue(attrsMap[srcColName], srcSchema.ColDefs[srcColName].Type.Name, spSchema.ColDefs[spCols[i]].T.Name)
+						if err != nil {
+							return fmt.Errorf("failed to convert column: %v to %v", attrsMap[srcColName], spSchema.ColDefs[spCols[i]])
+						}
+					}
+					cvtVals = append(cvtVals, cvtVal)
+				}
+
```
nevinheintze, line 79: Drop blank line?

dynamodb/data.go
```diff
@@ -0,0 +1,161 @@
...
+				var cvtVals []interface{}
+				for i, srcColName := range srcSchema.ColNames {
+					var cvtVal interface{}
+					if attrsMap[srcColName] == nil {
+						cvtVal = nil
+					} else {
+						// Convert data to the target type
+						cvtVal, err = cvtColValue(attrsMap[srcColName], srcSchema.ColDefs[srcColName].Type.Name, spSchema.ColDefs[spCols[i]].T.Name)
+						if err != nil {
+							return fmt.Errorf("failed to convert column: %v to %v", attrsMap[srcColName], spSchema.ColDefs[spCols[i]])
```
nevinheintze, line 74: Do we want to return here?
That would mean we'd stop processing data on the first error we encountered.
Suggest add row to bad row stats and sampleBadRows and continuing.

dynamodb/data.go
```diff
@@ -0,0 +1,161 @@
...
+			return attrVal.B, nil
+		case typeBinarySet:
+			return attrVal.BS, nil
+		}
+	case ddl.String:
+		switch srcType {
+		case typeMap:
+			b, err := json.Marshal(attrVal)
+			if err != nil {
+				return nil, fmt.Errorf("failed to encode a map object: %v to a json string", attrVal.GoString())
```
nevinheintze, line 113: From a user point of view, we're doing data conversion in this code, so the use of "encoding" here might be confusing.
Maybe "convert"?
Same comment for the three other error messages.

nevinheintze, line 113: Assume we keep going on data conversion errors (see earlier comment), then we need to be careful of printing bad values into error messages.

The reason is that we put these errors into 'unexpected' (assuming we follow the convention in the mysql and postgres data conversion processing) and so if we convert a huge data set and encounter many errors, the unexpected map could get to be huge and we might OOM.

dynamodb/data_test.go
```diff
@@ -0,0 +1,177 @@
...
+		name    string
+		srcType string                   // Source DB type.
+		spType  string                   // Spanner DB type.
+		in      *dynamodb.AttributeValue // Input value for conversion.
+		want    interface{}              // Expected result.
+	}{
+		{"bool", typeBool, ddl.Bool, &dynamodb.AttributeValue{BOOL: &boolVal}, true},
+		{"binary", typeBinary, ddl.Bytes, &dynamodb.AttributeValue{B: binaryVal}, binaryVal},
+		{"binary set", typeBinarySet, ddl.Bytes, &dynamodb.AttributeValue{BS: binarySetVal}, binarySetVal},
+		{"map", typeMap, ddl.String, &dynamodb.AttributeValue{M: mapVal}, "{\"B\":null,\"BOOL\":null,\"BS\":null,\"L\":null,\"M\":{\"list\":{\"B\":null,\"BOOL\":null,\"BS\":null,\"L\":[{\"B\":null,\"BOOL\":null,\"BS\":null,\"L\":null,\"M\":null,\"N\":null,\"NS\":null,\"NULL\":null,\"S\":\"str-1\",\"SS\":null},{\"B\":null,\"BOOL\":null,\"BS\":null,\"L\":null,\"M\":null,\"N\":\"1234.56789\",\"NS\":null,\"NULL\":null,\"S\":null,\"SS\":null}],\"M\":null,\"N\":null,\"NS\":null,\"NULL\":null,\"S\":null,\"SS\":null}},\"N\":null,\"NS\":null,\"NULL\":null,\"S\":null,\"SS\":null}"},
```
nevinheintze, line 149: I see what you mean about how verbose the conversion is here.
Let's chat about this -- I'd like to understand this a bit more.
I guess one option could be to have a flag that suppresses all of the null fields to give a more useful Spanner representation.

dynamodb/schema.go
```diff
@@ -325,3 +326,30 @@ func numericParsable(n string) bool {
...
+	for _, t := range tables {
+		input := &dynamodb.DescribeTableInput{
+			TableName: aws.String(t),
+		}
+		result, err := client.DescribeTable(input)
+		if err != nil {
+			conv.Unexpected(fmt.Sprintf("failed to make a DescribeTable API call for table %v: %v", t, err))
+			return
+		}
+		conv.Stats.Rows[t] += *result.Table.ItemCount
```
nevinheintze, line 353: Why is this "+=" and not just "=" here?
We should just get one call for each table, so there shouldn't be anything we need to sum over, or am I missing something?

dynamodb/schema.go
```diff
@@ -325,3 +326,30 @@ func numericParsable(n string) bool {
 
 	return true
 }
+
+// SetRowStats populates conv with the number of rows in each table. In
+// DynamoDB, we use describe_table api to get the number of total rows, but this
+// number is updated approximately every six hours. Recent changes might not be
+// reflected in this value.
```
nevinheintze, line 333: Yup, understood. This is a tradeoff here in terms of having an efficient estimate for rows vs having a more exact/up-to-data number (which would never be exact since the number of rows potentially keeps changing).

I might be worth making the implications of this clearer, so if someone sees something odd in the behaviour of harbourbridge/dynamodb, they can track it to this point.  How about editing the comment along the following lines:

// SetRowStats populates conv with the number of rows in each table. In
// DynamoDB, we use describe_table api to get the number of total rows, but this
// number is updated approximately every six hours. This means that our row
// count could be out of date by up to six hours. One of the primary uses of
// row count is for calculating progress during data conversion. As a result, if
// there have been huge changes in the number of rows in a table over the last
// six hours, the progress calculation could be inaccurate.

main.go
```diff
@@ -91,6 +92,7 @@ func init() {
 	flag.StringVar(&filePrefix, "prefix", "", "prefix: file prefix for generated files")
 	flag.StringVar(&driverName, "driver", "pg_dump", "driver name: flag for accessing source DB or dump files (accepted values are \"pg_dump\", \"postgres\", \"mysqldump\", and \"mysql\")")
 	flag.Int64Var(&schemaSampleSize, "schema-sample-size", int64(100000), "schema-sample-size: the number of rows to use for inferring schema (only for DynamoDB)")
+	flag.BoolVar(&consistentRead, "consistent-read", false, "consistent-read: enable the 'consistent read' mode when scanning data but it causes more api quota usage (only for DynamoDB)")
```
nevinheintze, line 95: I'm not sure I understand the value of using strongly consistent reads here.

If I'm running HarbourBridge against a live dynamodb instance, then I'm going to effectively get a mix of data from different time instances: schema processing, and each table query for data conversion will be issued at different times. And if I'm reading a large table, then I might see inconsistencies across the contents of the same table in the sense that rows I get later could be from more recent writes than rows I read earlier. Using strongly consistent reads won't really help much.

Note that the time to reach consistency is typically about one second or less -- see https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadConsistency.html.

If my dynamodb instance isn't changing, then it probably doesn't matter what I do.

However, the cost of strongly consistent reads can be quite high.

I guess my concern with the flag is that it could give a false sense of security: I set the flag thinking it will help me get some more consistent view of the data, but it won't really help and it will cost me.

Let's chat more offline on this.

nevinheintze (approved):
Thanks Hengfeng.
Just a couple of minor comments -- let's chat and resolve the tomorrow.
I'll set up a quick sync on cal.

dynamodb/data.go
hengfengli: Yeah, this sounds good to me.

dynamodb/data.go
hengfengli: Is it possible to let `conv.Unexpected` to return error? If the `unexpected` map has too many entries, then we throw an error before OOM.

dynamodb/schema.go
hengfengli: You're right. I looked it in a wrong way (I overlooked `t` and thought this is the sum).

dynamodb/data.go
```diff
@@ -0,0 +1,237 @@
...
+		// If there are more rows, then continue.
+		lastEvaluatedKey = result.LastEvaluatedKey
+	}
+}
+
+func cvtRow(attrsMap map[string]*dynamodb.AttributeValue, srcSchema schema.Table, spSchema ddl.CreateTable, spCols []string) ([]interface{}, map[string]string, []string) {
+	var err error
+	var srcStrVals []string
+	var cvtVals []interface{}
+	var cvtErrs map[string]string
```
nevinheintze, line 97: Where is cvtErrs initialized?
I was expecting something like: cvtErrs = make(map[string]string) somewhere.
Won't this "panic: assignment to entry in nil map" if we get to line 108 and try to assign to the map?
Is there a test that covers the error handling in cvtRow?

dynamodb/data.go
```diff
@@ -0,0 +1,237 @@
...
+				srcTable, err1, err2, ok))
+			continue
+		}
+
+		err := scan(srcTable, client, func(m map[string]*dynamodb.AttributeValue) {
+			cvtVals, cvtErrs, srcStrVals := cvtRow(m, srcSchema, spSchema, spCols)
+			if len(cvtErrs) == 0 {
+				conv.WriteRow(srcTable, spTable, spCols, cvtVals)
+			} else {
+				conv.Unexpected(fmt.Sprintf("Errors while converting data: %s\n", cvtErrs))
```
nevinheintze, line 51: This will give a somewhat unintuitive error message.
I think it will show something like:
    Errors while converting data: map[col1:failed to convert column: col1 col2:failed to convert column: col2]

That's a little hard to parse. And col names without table names aren't so useful.

Maybe instead of a map, construct a list of bad columns in cvtRows, and print something like:
    Data conversion error for table T in column(s) [col1 col2]

Note that our error reporting for data conversion for mysql and postgres isn't great either -- there we just report what types we had problems with.

dynamodb/data.go
```diff
@@ -0,0 +1,237 @@
...
+		spSchema, ok := conv.SpSchema[spTable]
+		if err1 != nil || err2 != nil || !ok {
+			conv.Stats.BadRows[srcTable] += conv.Stats.Rows[srcTable]
+			conv.Unexpected(fmt.Sprintf("Can't get cols and schemas for table %s: err1=%s, err2=%s, ok=%t",
+				srcTable, err1, err2, ok))
+			continue
+		}
+
+		err := scan(srcTable, client, func(m map[string]*dynamodb.AttributeValue) {
+			cvtVals, cvtErrs, srcStrVals := cvtRow(m, srcSchema, spSchema, spCols)
```
nevinheintze, line 47: Maybe spVals instead of cvtVals for consistency?

dynamodb/data.go
```diff
@@ -0,0 +1,237 @@
...
+	case ddl.String:
+		switch srcType {
+		case typeMap, typeList:
+			val, err := marshalAttrValue(attrVal)
+			if err != nil {
+				return nil, fmt.Errorf("failed to convert %v to a go struct", attrVal.GoString())
+			}
+			// If the original representation is preferred, then comment the
+			// above code and change the following line to:
+			//   b, err := json.Marshal(attrVal)
```
nevinheintze, line 141: Suggest giving a little more context for the decision here and maybe moving the comment to the top of the case i.e. at line 135. Maybe:
			// For typeMap and typeList, attrVal is a very verbose datastructure
                         // that contains null entries for unused type cases. We strip
                         // these out using marshallAttrValue. If it important that the Spanner
                         // values can be easily unmarshalled back to dynamodb.AttributeValue
                         // types, then replace the following five lines with just: 
		        //    b, err := json.Marshal(attrVal)
                         // but note that this will consume extra Spanner storage.

dynamodb/data.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+	}
+	return nil, fmt.Errorf("can't convert value of type %s to Spanner type %s", attrVal.GoString(), spType)
+}
+
+// marshalAttrValue converts a dynamodb.AttributeValue to a Go struct which can
+// be easily encoded to a json string. If we use the normal json encoder, it
+// will have many null values. The purpose of this function is to remove the
+// null values in the json string.
+func marshalAttrValue(a *dynamodb.AttributeValue) (interface{}, error) {
```
nevinheintze, line 192: This doesn't really marshal.
Maybe rename to stripNull?

dynamodb/data_test.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+type spannerData struct {
+	table string
+	cols  []string
+	vals  []interface{}
+}
+
+func TestProcessData(t *testing.T) {
```
nevinheintze, line 36: Maybe add a row that generates an error to exercise the error handling in cvtRow?

nevinheintze (approved):
Thanks for the updates -- looks good.
A couple of minor suggestions -- most important is probably the error handling in cvtRow (where I think we could panic if we encounter an error).

dynamodb/data.go
hengfengli: You're right. I missed this. I added a test to cover this error.

dynamodb/data.go
hengfengli: If we only show a list of bad columns, then `cvtErrs` is not needed any more.

dynamodb/data.go
```diff
@@ -0,0 +1,242 @@
...
+			return attrVal.B, nil
+		case typeBinarySet:
+			return attrVal.BS, nil
+		}
+	case ddl.String:
+		switch srcType {
+		case typeMap, typeList:
+			// For typeMap and typeList, attrVal is a very verbose data
+			// structure that contains null entries for unused type cases. We
+			// strip these out using marshallAttrValue. If it important that the
```
nevinheintze, line 137: marshallAttrValue --> stripNull

dynamodb/data_test.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	for _, tc := range testcases {
+		cvtVal, err := cvtColValue(tc.in, tc.srcType, tc.spType)
+		assert.Nil(t, err, fmt.Sprintf("Failed to convert %v from %s to %s", tc.in, typeString, ddl.String))
+		assert.Equal(t, tc.want, cvtVal, tc.name)
+	}
+}
+
+func TestMarshalAttrValue(t *testing.T) {
```
nevinheintze, line 195: TestMarshalAttrValue --> TestStripNull?

nevinheintze (approved):
Thanks -- just a couple of very minor typos.
Ready to submit!

PR #81, opened 2020-09-08
refactor: update the Spanner data type

Dharmil-searce:
Refactored `ddl/ast.go` and all other files dependent on it to handle Spanner datatype such that we can remove usage of interface. Spanner column type is refactored and represented by :
```sh
type Type struct {
	Name string
	Len int64
        IsArray bool
}
```

Further, we want to support `schema-only` and `data-only` modes. In order to support it, we need a feature to extract all relevant information of source and spanner database in a JSON file. To facilitate the same, refactoring of code is done.


spanner/ddl/ast.go
```diff
@@ -20,110 +20,67 @@
...
-// and struct types for each case:
-//    type Int struct { }
-//    type String struct { length int64 }
-// and finally, we define functions:
-//    func (i Int) PrintScalarType() { ... }
-//    func (s String) PrintScalarType() { .. }
-//
-// The net result is that the only way to build a ScalarType
-// is using Int or String.
+// Package ddl provides a generic representation of Spanner DDL.
```
nevinheintze, line 23: Redundant comment -- looks like a repeat of line 15?

spanner/ddl/ast.go
```diff
@@ -20,110 +20,67 @@
...
 //     scalar_type:
 //        { BOOL | INT64 | FLOAT64 | STRING( length ) | BYTES( length ) | DATE | TIMESTAMP }
-type ScalarType interface {
-	PrintScalarType() string
+type Type struct {
+	Name string
+	// Len encodes the following Spanner DDL definition:
+	//     length:
+	//        { int64_value | MAX }
+	Len interface{}
```
nevinheintze, line 66: I'm ok with this encoding and I don't have anything better to suggest.
But just wanted to note that, in principle, we have three different things to represent:

a) an integer length
b) a MAX length
c) no length

The interface encoding is kind of convenient since it allows us to write:

  Type{Name: Int64}
  Type{Name: String, Len: MaxLength}
  Type{Name: Bytes, Len: int64(42)}

But it allows us to put anything as a Len, including an arbitrary string.

spanner/ddl/ast.go
Dharmil-searce: Removed it.

spanner/ddl/ast.go
Dharmil-searce: Yes, that's correct. `Type.Len` could be `int64(value)`, `MaxLength` or `NULL` value. But that allows us to put other arbitrary strings as well. This is a trade off for such refactoring.

spanner/ddl/ast.go
```diff
@@ -18,113 +18,65 @@
...
+	// IsArray represents if Type is an array_type or not
+	// When false, column has type T; when true, it is an array of type T.
+	IsArray bool
```
Dharmil-searce, line 63: Moved IsArray from ColumnDef to Type. Also, removed all declarations of ScalarType.

spanner/ddl/ast.go
hengfengli: I would recommend you to take a look at this link: https://github.com/googleapis/google-cloud-go/blob/005eb3d360540e4747b607cc681950fa2ebb2d82/spanner/spansql/types.go#L245-L253. 

This is the existing code in `spanner/spansql`. I would prefer to define `Len` as `Int64`.

hengfengli: Probably hold on. The only concern of my suggestion is that if we encode it into json, it will become a very large number in the file everywhere when there is a STRING(MAX). The good part is that the Go type is fixed.

hengfengli (commented):
Thanks for making the changes. It looks good to me. The only concern is about the type of `Len`.

spanner/ddl/ast.go
nevinheintze: This isn't too bad.
STRING(MAX) would get encoded as:
{"Name":"STRING","Len":9223372036854775807}
Its isn't likely to cause a huge increase in json for our configs, and the size isn't really an issue in any case, since schemas are pretty small.

Also, I like the idea of consistency with google-cloud-go/spanner/spansql/types.go.
At some point we may want to consider replacing ast.go with those definitions (but that's a discussion for another day).

nevinheintze (approved):
Looks good.
One issue to resolve is Hengfeng's suggestion.

spanner/ddl/ast.go
hengfengli: @Dharmil-searce Can you please make the change according to our discussion here? Thanks.

hengfengli:
A recommendation for the PR title. There is a convention to follow: https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716.

Dharmil-searce:
> A recommendation for the PR title. There is a convention to follow: https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716.

@hengfengli Sure, I will keep it in mind from next time onwards.

spanner/ddl/ast.go
Dharmil-searce: Done. I have updated `Len` to `Int64` and tried to keep `PrintColumnDefType` consistent with `spanner/spansql` code.

hengfengli (approved):
Thanks for the change. LGTM.

PR #83, opened 2020-09-14
feature: schema-only conversion mode

Dharmil-searce:
Changes included in this PR:

- Support for schema-only conversion mode by specifying a `-schema-only` flag while running the tool. 

- Generated a schema conversion JSON file for storing the state of the conversion. This file will be generated only when tool is run using `-schema-only` flag.

- Updated `report.go` file such that rating of data conversion (`rateData()`) and `fillRowStats()` will be skipped when running `-schema-only` mode.

- Exported `schema`, `data` and `skip` fields of `statementStat` struct used in `Conv` (internal/convert.go). Because, only exported fields can be encoded/decoded in JSON.


nevinheintze:
I'm still collecting my thoughts on this, and in particular, how all of the modes play together.
Here are some initial comments.

We want harbourbridge to operate in several modes:

1) turn-key evaluation mode: this is the current behaviour and the default behaviour. You get pg_dump by default and you can use -driver to specify other sources.

2) guided-schema mode (or maybe customized-mode): When you run in this mode, we start an http server and you get the UI running in a browser. You connect to a database, and basically run in schema mode. You use the UI to configure the source database (we ignore -driver). We run schema mode, and write out the JSON file, and then the UI shows the result (using the JSON file) and you can tweak the schema and mappings (and these tweaks update the JSON file).

3) data mode: When you've finished with guided-schema mode, you use this mode to convert data and write it to Spanner. The schema information and mappings are extracted from the JSON file (you have to specify a JSON file for this mode).

We could also consider an optional fourth mode:

4) Turn-key, but don't do data conversion. Kind of a schema-only mode, but without UI or customization. I'm not sure we need this mode, and it may be simpler not to offer it (if you really need the functionality, you can use guided-schema mode, and just do no customization).

What I'm wondering is what the flags will look like for controlling these options e.g. we could do it with just one 
 '-mode' flag. And how the schema-only flag added by this PR fits into the end-picture.

Apologies this is a bit vague. I want to think more about this overnight.
But I'd also like to hear your thoughts.

Dharmil-searce:
The idea is that harbourbridge could run as a CLI tool and extend the functionalities to run it as a WEB tool as well. We can run WEB interface tool by the following command. A http server would be up and running. 
```sh
$ harbourbridge --web
``` 
Talking about CLI tool, it will support 3 different modes:

1. `--schema-only` : This mode can be used by a user who wants to only evaluate the schema conversion and get a report for just the schema conversion assessment. He do not need to configure a Gcloud project and a Spanner instance for running this mode. The command to run using this mode would look like this : 
   ```sh 
    $ harbourbridge --schema-only --driver=mysql
   ``` 
     That's it! 3 files will be generated at the end : 1) schema.conv.json file (JSON file storing the state of the schema conversion) 2) schema.ddl.txt file (Spanner DDL definitions with comments) and 3) Summary report file (It will contain summary related to schema conversion only).

2. `--data-only` : This mode can be used by a user when he has a schema conversion JSON file either from CLI tool or WEB tool. The command to run this mode would look like this : 
     ```sh 
      $ harbourbridge --data-only --schemaFile=xxx.schema.conv.json --driver=mysql --instance=instanceName
     ```
    Running this mode will create a DB and migrate data according to the schema conversion defined in the JSON file. It will generate 2 files: 1) Bad Data file and 2) Summary report file  [Both this files would be as per the current tool]

3. Default mode (turn-key evaluation mode) : A user would not need to specify any mode to run it. The default mode behaviour would be to convert schema, create DB and migrate data.

My thoughts on some of your other questions:

> What I'm wondering is what the flags will look like for controlling these options e.g. we could do it with just one
> '-mode' flag

Usage of `--schema-only` and `--data-only` flags is pretty common for MySQL and PostgreSQL users. If none of these 2 flags are specified, then the tool would run in default mode i.e. schema+data. Thus, I don't think we need a `-mode` flag and then specify the mode. 

> I'm not sure we need this mode, and it may be simpler not to offer it (if you really need the functionality, you can use guided-schema mode, and just do no customization)

For the users who are more comfortable with CLI itself and need to quickly test out the schema conversion assessment without actually running the http server. They do not need to provide a Gcloud project and Spanner instance details as well. We would anyways need to split the tool to support `--data-only` mode, then why not provide a mode for user to run `--schema-only` from CLI directly. It's also pretty common mode in MySQL and PostgreSQL.

@nevinheintze Let me know your views on this and I will then work on it accordingly.

hengfengli:
It sounds reasonable to me to have 4 modes: 

1. turn-key evaluation mode (default): only exports a report and a DDL-based schema file. 
2. `harbourbridge --guided-schema` or `harbourbridge --web`: starts a web server for GUI and outputs a report, a DDL-based schema file, a schema json file. 
3. `harbourbridge --data-only --schema-file=xxx.schema.conv.json`: takes a schema json file as input. 
4. `harbourbridge --schema-only`: outputs a report, a DDL-based schema file, a schema json file. 

1 is a combination of 3 and 4 and does not export a schema json file. 

2 is an improvement for 4 with nice user-friendly GUI.

nevinheintze:
What's the relationship between --web and {--schema-only, --data-only}?

Do we view them as orthogonal (which allows us to add more features to web mode), or do we view them as dependent (which reflects the functionality we're currently adding, where --web is only for schema)?

Dharmil-searce:
We view them as independent. User would just be running a `--web` flag and http server would be up. User would do the schema conversion in UI and then after the schema conversion, for now we would be providing them instructions on how to migrate the data in UI. `--schema-only` and `--data-only` would work independently from `--web`. This will allow any new feature to be added to `--web` tool without worrying or disturbing the CLI tool which would be supporting 3 modes : default, `--schema-only` and `--data-only`.

hengfengli:
@Dharmil-searce But `--web` and `--schema-only` will share some common underlying code, right?

Dharmil-searce:
> @Dharmil-searce But `--web` and `--schema-only` will share some common underlying code, right?

Yes correct, they would share many things in common but also there would be some other functions and APIs as well that would be added to `--web` version to enable editing datatypes, editing conversion tables, resuming a session etc.

nevinheintze:
Thanks for the discussion and clarifications.

Here's my understanding of what's being proposed:

1) The tool either runs in web-mode or commandline-mode.

2) In web-mode (controlled by a flag --web or --ui or something similar) we start the http server and control the tool via the web. We will ignore other commandline flags e.g. we will ignore --driver, --schema-only, --data-only etc. In web-mode, the ui determines how the tool runs.

3) In commandline-mode, we use commandline flags to determine how the tool runs.

4) We will add flags --data-only and --schema-only to control command-line mode. We use these specific flag names because that's what is already used by pg_dump (and mysqldump has something similar, although they use --no-data and --no-create-info). Consistency is good.

So, back to the initial comment on this thread, the command-line options for the four use cases would be:

1. turn-key evaluation mode:
  harbourbridge --driver=...

2. guided-schema mode
  harbourbridge --web // other flags ignored

3. data mode
  harbourbridge --data-only --driver=...

4. Turn-key, but don't do data conversion
  harbourbridge --schema-only --driver=...

README.md
```diff
@@ -230,11 +231,14 @@ run HarbourBridge on the modified pg_dump/mysqldump output.
 
 HarbourBridge generates several files as it runs:
 
-- Schema file (ending in `schema.txt`): contains the generated Spanner
-  schema, interspersed with comments that cross-reference to the relevant
+- Schema DDL file (ending in `schema.ddl.txt`): contains the generated Spanner
```
nevinheintze, line 234: Let's keep schema.txt.
I don't see any strong reason to change.

See comments on changing conversion file name.

README.md
```diff
@@ -230,11 +231,14 @@ run HarbourBridge on the modified pg_dump/mysqldump output.
...
+  schema DDL, interspersed with comments that cross-reference to the relevant
   PostgreSQL/MySQL schema definitions.
 
-- Report file (ending in `report.txt`) contains a detailed analysis of the
+- Schema conversion file (ending in `schema.conv.json`): contains all schema
```
nevinheintze, line 238: The `schema.conv.json` file has a lot more than just schema. It has everything on the conv datastructure, including schema, mappings, stats, errors, statements etc -- including everything needed to generate report.txt.

It is basically the snapshot of the state of the tool -- kind of a record of the session.

How about session.json?

README.md
```diff
@@ -230,11 +231,14 @@ run HarbourBridge on the modified pg_dump/mysqldump output.
...
 
-- Schema file (ending in `schema.txt`): contains the generated Spanner
-  schema, interspersed with comments that cross-reference to the relevant
+- Schema DDL file (ending in `schema.ddl.txt`): contains the generated Spanner
+  schema DDL, interspersed with comments that cross-reference to the relevant
   PostgreSQL/MySQL schema definitions.
 
-- Report file (ending in `report.txt`) contains a detailed analysis of the
+- Schema conversion file (ending in `schema.conv.json`): contains all schema 
+conversion state endcoded as JSON. This file is written for `-schema-only` mode.
```
nevinheintze, line 239: I think we should write this file all of the time.
It captures the state/session of the tool.
I don't see a strong reason to just do it for schema-mode, and the fewer special-cases the better.

Moreover, I could see it being useful in turn-key mode:
a) run in turn-key mode and get a session.json
b) add some more data using data-only mode, and use the session.json from step (a) to get the schema and table/col mappings.

README.md
Dharmil-searce: We are providing 2 options to view the schema conversion in the UI i.e. `schema conversion assessment` and `DDL`. 
When a user exports the report, then exported file from `DDL` would actually be `schema.txt` as it is in current tool. 

I thought of changing it to avoid the confusion that the file downloaded from `DDL` view in UI is a `schema.txt` file. Other reason was: we would need to provide a flag for input of the JSON file in `--data-only` mode. If we assign the flag as `--schema` or `--schemaFile` then it would be confusing. 

But surely, I got your point and I will revoke the change.

README.md
Dharmil-searce: Yup, sounds good. I will make that to `session.json`.

README.md
Dharmil-searce: Got it ! I will make the change to generate this file all the time.

README.md
```diff
@@ -234,7 +234,10 @@ HarbourBridge generates several files as it runs:
   schema, interspersed with comments that cross-reference to the relevant
   PostgreSQL/MySQL schema definitions.
 
-- Report file (ending in `report.txt`) contains a detailed analysis of the
+- Session file (ending in `session.json`): contains all schema and data
+conversion state endcoded as JSON. It is basically a snapshot of the session.
```
nevinheintze, line 238: Indent to match previous line?

README.md
```diff
@@ -271,6 +274,9 @@ database (plus a '.') is used.
 `-v` Specifies verbose mode. This will cause HarbourBridge to output detailed
 messages about the conversion.
 
+`-schema-only` Specifies that conversion would be schema-only and will
+exclude data conversion.
```
nevinheintze, line 278: Suggest rewording:

Specifies that only schema processing will be performed. Any data in the source database
will be ignored.

internal/report.go
```diff
@@ -355,9 +357,12 @@ func ok(total, badCount int64) bool {
 	return badCount < total/3
 }
 
-func rateConversion(rows, badRows, cols, warnings int64, missingPKey, summary bool) string {
-	return fmt.Sprintf("Schema conversion: %s.\n", rateSchema(cols, warnings, missingPKey, summary)) +
-		fmt.Sprintf("Data conversion: %s.\n", rateData(rows, badRows))
+func rateConversion(rows, badRows, cols, warnings int64, missingPKey, summary bool, schemaOnly bool) string {
```
nevinheintze, line 360: I'm fine with this at the moment, but just noting that it'll probably need to change when we add support for --data-only.

The use of conv.SchemaMode here is a bit fragile here (in the arg passed to rateConversion) since it reflects the last mode used, not how the tool was configured to run.

For example, if we do both schema and data mode, it will be data mode.
If we just do data mode, it will be data mode.
If we just to schema mode, it will schema mode.

Longer term, I think what we really want is to pass in how the tool was configured.
If we did schema processing, we should print schema processing stats.
If we did data processing, we should print data processing stats.

main.go
```diff
@@ -567,6 +580,26 @@ func writeSchemaFile(conv *internal.Conv, now time.Time, name string, out *os.Fi
...
 
+func writeSessionFile(conv *internal.Conv, now time.Time, name string, out *os.File) {
+	f, err := os.Create(name)
+	if err != nil {
+		fmt.Fprintf(out, "Can't create session file %s: %v\n", name, err)
+		return
+	}
+	// Session file will basically contain 'conv' struct in JSON format.
+	// It contains all the information for schema and data conversion state.
+	convJSON, err := json.MarshalIndent(conv, "", " ")
```
nevinheintze, line 591: conv has a lot of stuff.
Most of it is small.
But sampleBadRows could be fairly big (we have a hard-coded 10MB limit).

Maybe worth a comment in the conv definition that this datastructure is serialized into session.json, so care should be take adding large new datastructures into conv?
We could also consider dropping some parts of conv.

nevinheintze (approved):
A couple of minor comments.
Looks good -- thanks!

internal/report.go
nevinheintze: As discussed, we'd like to generate a full report (schema and data stats) when we run in data mode.

Moreover, we always have to do data processing last, so even though conv.SchemaMode reflects the last mode we ran int, it always tells us whether we did data processing.

In short, let's leave this as is.

main.go
nevinheintze: As noted during the discussion earlier today, we write out conv before we do data conversion.
So sampleBadRows will always be empty.
Also, sampleBadRows is lower case, so it won't be written to JSON.

PR #84, opened 2020-09-17
fix: fix handling of pg_dump inserts when columns not specified

nevinheintze:
Handle pg_dump inserts of the form:

 INSERT INTO test VALUES (1, "hello", "bye");

When columns are not specified, we should use all columns in the table.


hengfengli (approved):
LGTM.

PR #88, opened 2020-10-04
feature: data-only conversion mode

Dharmil-searce:
Changes included in this PR:

- Support for data-only conversion mode by specifying a -data-only flag while running the tool.

- A spanner database will be created based on the schema state provided by a session file (`-session`) and then data will be migrated.

Sample Usage:
```sh
$ harbourbridge --data-only --session=sessionFile.json --instance=test-instance --driver=mysqldump < mysqlDumpFile.sql
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -92,9 +92,9 @@ func init() {
 	flag.StringVar(&driverName, "driver", "pg_dump", "driver name: flag for accessing source DB or dump files (accepted values are \"pg_dump\", \"postgres\", \"mysqldump\", and \"mysql\")")
 	flag.Int64Var(&schemaSampleSize, "schema-sample-size", int64(100000), "schema-sample-size: the number of rows to use for inferring schema (only for DynamoDB)")
 	flag.BoolVar(&verbose, "v", false, "verbose: print additional output")
-	flag.BoolVar(&schemaOnly, "schema-only", false, "schema-only: mode for schema only conversion to Spanner")
-	flag.BoolVar(&dataOnly, "data-only", false, "data-only: mode for data only conversion to Spanner")
-	flag.StringVar(&sessionJSON, "session", "", "session: session file generated through schema-only or web mode")
+	flag.BoolVar(&schemaOnly, "schema-only", false, "schema-only: in this mode we do schema conversion, but skip data conversion")
+	flag.BoolVar(&dataOnly, "data-only", false, "data-only: in this mode we skip schema conversion and just do data conversion (use the session flag to specify the session file for schema and data mapping)")
+	flag.StringVar(&sessionJSON, "session", "", "session: specifies the file we restore session state from (used in schema-only to provide schema and data mapping)")
 }
 
 func usage() {
@@ -156,6 +156,14 @@ func main() {
 		filePrefix = dbName + "."
 	}
 
+	if schemaOnly && dataOnly {
+		panic(fmt.Errorf("can't use both schema-only and data-only modes at once"))
+	}
+
+	if dataOnly && sessionJSON == "" {
+		panic(fmt.Errorf("when using data-only mode, the session must specify the session file to use"))
+	}
+
 	err = toSpanner(driverName, project, instance, dbName, ioHelper, filePrefix, now)
 	if err != nil {
 		panic(err)
@@ -170,15 +178,13 @@ func main() {
 //   3. Run data conversion
 //   4. Generate report
 func toSpanner(driver, projectID, instanceID, dbName string, ioHelper *ioStreams, outputFilePrefix string, now time.Time) error {
-	conv := internal.MakeConv()
+	var conv *internal.Conv
 	var err error
 	if !dataOnly {
 		conv, err = schemaConv(driver, ioHelper)
 		if err != nil {
 			return err
 		}
-		// close the seekable file
-
 		if ioHelper.seekableIn != nil {
 			defer ioHelper.in.Close()
 		}
@@ -190,13 +196,8 @@ func toSpanner(driver, projectID, instanceID, dbName string, ioHelper *ioStreams
 			return nil
 		}
 	} else {
-		// read a session JSON file and unmarshal it's content into
-		// *internal.Conv when running in 'data-only' mode.
-		s, err := ioutil.ReadFile(sessionJSON)
-		if err != nil {
-			return err
-		}
-		err = json.Unmarshal(s, &conv)
+		conv = internal.MakeConv()
+		err = readSessionFile(conv)
 		if err != nil {
 			return err
 		}
@@ -403,6 +404,8 @@ func schemaFromDump(driver string, ioHelper *ioStreams) (*internal.Conv, error) 
 }
 
 func dataFromDump(driver string, config spanner.BatchWriterConfig, ioHelper *ioStreams, client *sp.Client, conv *internal.Conv) (*spanner.BatchWriter, error) {
+	// TODO: refactor of the way we handle getSeekable
+	// to avoid the code duplication here
 	if !dataOnly {
 		_, err := ioHelper.seekableIn.Seek(0, 0)
 		if err != nil {
@@ -654,6 +657,20 @@ func writeSessionFile(conv *internal.Conv, now time.Time, name string, out *os.F
 		return
 	}
 	fmt.Fprintf(out, "Wrote session to file '%s'.\n", name)
+}
+
+// readSessionFile reads a session JSON file and
+// unmarshal it's content into *internal.Conv.
+func readSessionFile(conv *internal.Conv) error {
+	s, err := ioutil.ReadFile(sessionJSON)
+	if err != nil {
+		return err
+	}
+	err = json.Unmarshal(s, &conv)
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
 // writeBadData prints summary stats about bad rows and writes detailed info
--- a/main.go
+++ b/main.go
@@ -81,6 +81,8 @@ var (
 	schemaSampleSize = int64(0)
 	verbose          bool
 	schemaOnly       bool
+	dataOnly         bool
+	sessionJSON      string
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -167,21 +170,36 @@ func main() {
...
-	}
-	// close the seekable file
+	conv := internal.MakeConv()
+	var err error
+	if !dataOnly {
+		conv, err = schemaConv(driver, ioHelper)
+		if err != nil {
+			return err
+		}
+		// close the seekable file
```
nevinheintze, line 180: Looks like a previous PR introduced a spurious new line between the comment and the code.
Suggest removing the new line.
Or we could just remove the comment -- I think it is pretty self-evident.

main.go
```diff
@@ -167,21 +170,36 @@ func main() {
...
+		writeSchemaFile(conv, now, outputFilePrefix+schemaFile, ioHelper.out)
+		writeSessionFile(conv, now, outputFilePrefix+sessionFile, ioHelper.out)
+		if schemaOnly {
+			report(driver, nil, ioHelper.bytesRead, "", conv, outputFilePrefix+reportFile, ioHelper.out)
+			return nil
+		}
+	} else {
+		// read a session JSON file and unmarshal it's content into
+		// *internal.Conv when running in 'data-only' mode.
+		s, err := ioutil.ReadFile(sessionJSON)
```
nevinheintze, line 195: We should probably check that sessionJSON is set, and raise an appropriate error if not.
e.g. something like: when using data-only mode, the session must specify the session file to use.

main.go
```diff
@@ -167,21 +170,36 @@ func main() {
...
+		}
 
-	writeSchemaFile(conv, now, outputFilePrefix+schemaFile, ioHelper.out)
-	writeSessionFile(conv, now, outputFilePrefix+sessionFile, ioHelper.out)
-	if schemaOnly {
-		report(driver, nil, ioHelper.bytesRead, "", conv, outputFilePrefix+reportFile, ioHelper.out)
-		return nil
+		writeSchemaFile(conv, now, outputFilePrefix+schemaFile, ioHelper.out)
+		writeSessionFile(conv, now, outputFilePrefix+sessionFile, ioHelper.out)
+		if schemaOnly {
```
nevinheintze, line 188: The current code leads to the following confusing behaviour:
If I set both schemaOnly and dataOnly, then we will just get data conversion (i.e. same behaviour as dataOnly).

I think it would be better to simply exit if both flags are set and give an error.

Code-wise, I'm wondering if we should do some flag validation at the top of toSpanner, and check for consistency of dataOnly, schemaOnly and session.
i.e. have one block of code that checks for consistency and generates the right errors.

main.go
```diff
@@ -167,21 +170,36 @@ func main() {
...
+			report(driver, nil, ioHelper.bytesRead, "", conv, outputFilePrefix+reportFile, ioHelper.out)
+			return nil
+		}
+	} else {
+		// read a session JSON file and unmarshal it's content into
```
nevinheintze, line 193: Suggest factoring out this block (line 193-202) into a separate function, say readSessionFile.
toSpanner is getting quite long now.

main.go
```diff
@@ -385,10 +403,22 @@ func schemaFromDump(driver string, ioHelper *ioStreams) (*internal.Conv, error)
...
-		fmt.Printf("\nCan't seek to start of file (preparation for second pass): %v\n", err)
-		return nil, fmt.Errorf("can't seek to start of file")
+	if !dataOnly {
+		_, err := ioHelper.seekableIn.Seek(0, 0)
+		if err != nil {
+			fmt.Printf("\nCan't seek to start of file (preparation for second pass): %v\n", err)
+			return nil, fmt.Errorf("can't seek to start of file")
+		}
+	} else {
+		// Note: input file is kept seekable to plan for future
```
nevinheintze, line 413: At some point we might need to do a refactor of the way we handle getSeekable to avoid the code duplication here: potentially we could do this right at the top of toSpanner i.e. before we call schemaConv. We could look at the driver and call getSeekable if we have pg_dump or mysqldump.

But let's leave as is for the moment.

main.go
```diff
@@ -91,7 +93,8 @@ func init() {
 	flag.Int64Var(&schemaSampleSize, "schema-sample-size", int64(100000), "schema-sample-size: the number of rows to use for inferring schema (only for DynamoDB)")
 	flag.BoolVar(&verbose, "v", false, "verbose: print additional output")
 	flag.BoolVar(&schemaOnly, "schema-only", false, "schema-only: mode for schema only conversion to Spanner")
-
+	flag.BoolVar(&dataOnly, "data-only", false, "data-only: mode for data only conversion to Spanner")
+	flag.StringVar(&sessionJSON, "session", "", "session: session file generated through schema-only or web mode")
```
nevinheintze, line 97: This comment is a little confusing: not sure if it is an input or name of file we will generate.
Maybe drop the comment about web mode since we don't have that yet?

Suggest rephrasing as:
"session: specifies the file we restore session state from (used in schema-only to provide schema and mapping configuration)"

main.go
```diff
@@ -91,7 +93,8 @@ func init() {
 	flag.Int64Var(&schemaSampleSize, "schema-sample-size", int64(100000), "schema-sample-size: the number of rows to use for inferring schema (only for DynamoDB)")
 	flag.BoolVar(&verbose, "v", false, "verbose: print additional output")
 	flag.BoolVar(&schemaOnly, "schema-only", false, "schema-only: mode for schema only conversion to Spanner")
-
+	flag.BoolVar(&dataOnly, "data-only", false, "data-only: mode for data only conversion to Spanner")
```
nevinheintze, line 96: How about:

"data-only mode: in this mode we skip schema conversion and just do data conversion (use the session flag to specify the session file for schema and data mapping)"

and maybe something similar for schema-only:

"schema-only: in this mode we do schema conversion, but skip data conversion"

nevinheintze (commented):
Looks good -- just a few minor comments/clarifications.

main.go
```diff
@@ -167,21 +170,36 @@ func main() {
 //   3. Run data conversion
 //   4. Generate report
 func toSpanner(driver, projectID, instanceID, dbName string, ioHelper *ioStreams, outputFilePrefix string, now time.Time) error {
-	conv, err := schemaConv(driver, ioHelper)
-	if err != nil {
-		return err
-	}
-	// close the seekable file
+	conv := internal.MakeConv()
```
hengfengli, line 173: Can we declare a `conv` variable here instead of making an actual instance? E.g., 

```go
var conv *internal.Conv
if !dataOnly {
    conv, err = schemaConv(driver, ioHelper)
} else {
    conv = internal.MakeConv()
}
```

main.go
```diff
@@ -167,21 +170,36 @@ func main() {
...
+			return nil
+		}
+	} else {
+		// read a session JSON file and unmarshal it's content into
+		// *internal.Conv when running in 'data-only' mode.
+		s, err := ioutil.ReadFile(sessionJSON)
+		if err != nil {
+			return err
+		}
+		err = json.Unmarshal(s, &conv)
```
hengfengli, line 199: If I understand it correctly, the content of session json file is the `conv` struct, is this correct?

main.go
Dharmil-searce: Correct. Removed comment.

main.go
Dharmil-searce: Done

main.go
Dharmil-searce: Yes correct. For now, I will add it as TODO in this block.

main.go
Dharmil-searce: Sure. Done.

main.go
Dharmil-searce: Yup, sounds good. Updated.

main.go
Dharmil-searce: Yeah sure. That would be a better way. Done.

main.go
Dharmil-searce: Yes correct. That's the reason we had refactored the `conv` struct previously.

main.go
Dharmil-searce: Added check condition before the call to toSpanner() as you suggested in other comment.

main.go
Dharmil-searce: We are using dataOnly and schemaOnly as modes (main.go file) as well as we are using them while conversion (conv.go file). Hence, I think processSchema and processData would add complexity in understanding. Please review the checks that I have added before the call to toSpanner():
https://github.com/cloudspannerecosystem/harbourbridge/blob/20a5a88a6ce89b030ddd0a024800f3eb1a174526/main.go#L158-L167

main.go
```diff
@@ -167,21 +178,29 @@ func main() {
 //   3. Run data conversion
 //   4. Generate report
 func toSpanner(driver, projectID, instanceID, dbName string, ioHelper *ioStreams, outputFilePrefix string, now time.Time) error {
```
nevinheintze, line 180: We're developing an inconsistency in the way that we use vars like driverName, schemaOnly, dataOnly and filePrefix.
Some we only access in the main function, and pass as function args e.g. driverName.
But for others, we access directly in code outside of main.

We already had this inconsistency before this PR, but adding dataOnly and schemaOnly expands it a bit.

What do you think about the following convention:
- any access of vars in main.go are direct accesses
- we no longer pass vars as args to functions

That would mean dropping the driver and outputFilePrefix args for toSpanner.

hengfengli: Sounds good to me. The integration tests may need to be refactored accordingly.

main.go
nevinheintze: Change looks good -- it addresses my concerns.

I agree with you about the confusion of convert.go.
We should probably rename dataOnly and schemaOnly modes in that file.
Having flags dataOnly and schemaOnly and modes dataOnly and schemaOnly that mean slightly different things is really confusing!

But let's not solve the issue in this PR.

nevinheintze (approved):
Thanks!

main.go
Dharmil-searce: @nevinheintze @hengfengli Agreed. For consistency, I will make changes as suggested and also update the integration tests. I will raise a separate PR for that very soon. Let's go ahead and merge the current one, if that's fine.

nevinheintze: Yes, separate PR is good!

PR #90, opened 2020-11-09
docs: add dynamodb readme

hengfengli:


dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+using data from an existing PostgreSQL or MySQL or DynamoDB database. This
+README provides details of the tool's DynamoDB capabilities. For general
+HarbourBridge information see this [README](https://github.com/cloudspannerecosystem/harbourbridge#harbourbridge-turnkey-spanner-evaluation).
+
+## Example DynamoDB Usage
+
+The following examples assume `harbourbridge` has been added to your PATH
+environment variable.
+
+HarbourBridge can be run directly on a DynamoDB database (via the aws go sdk
```
nevinheintze, line 13: This sentence doesn't add too much -- it was originally there to explain the dump vs direction functionality, which isn't relevant to DynamoDB.

Suggest either just deleting it.
Or may preserving the comment in parenthesis (since that is kind of useful).
Maybe add a short sentence to line 18: "HarbourBridge accesses your DynamoDB database via the aws go sdk package."

dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+package).
+
+Before running HarbourBridge, make sure that you have
+[set up your AWS credentials/region](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html)
+correctly.
+
+To use the tool directly on a DynamoDB database (it will migrate all tables),
```
nevinheintze, line 20: Drop "directly". I don't think it is needed in this context.

dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+run
+
+```sh
+harbourbridge -driver=dynamodb
+```
+
+Due the schemaless nature of DynamoDB, the tool infers the schema based on a
+certain amount of sampled data, by default, 100000 rows. If a table has more
+rows than the default value, we only use 100000 rows for estimating the schema.
+You can change this value, run
```
nevinheintze, line 30: You can change this value via the flag `schema-sample-size`. For example

dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+```sh
+harbourbridge -driver=dynamodb -schema-sample-size=1000000
```
nevinheintze, line 33: It'd hard to see the difference between 100000 and 1000000.
Suggest changing 1000000 to 500000 to make this clearer.

dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Schema Inference
+
+DynamoDB is schemaless and it is impossible to get the full information of the
+schema. So we sample a set of rows and use them to model the schema in Cloud
+Spanner. For a large-scale table, scanning the entire table would be extremely
+slow. Therefore, we can only scan a part of it. We provide a flag
+`schema-sample-size` to specify how many rows to use for schema inference.
+
+In addition, for the performance purpose, we choose to call Scan on the table
```
nevinheintze, line 64: purpose --> reasons

dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+| `StringSet`        | `ARRAY<STRING>`            |                                           |
+| `NumberSet`        | `ARRAY<NUMERIC or STRING>` |                                           |
+| `BinarySet`        | `ARRAY<BYTES>`             |                                           |
+
+We discuss these, as well as other limits and notes on schema conversion, in the
+following sections.
+
+### Schema Inference
+
+DynamoDB is schemaless and it is impossible to get the full information of the
```
nevinheintze, line 58: Suggest rewording:

DynamoDB is schemaless database: other than a primary index and optional
secondary index, column names and types are essentially unconstrained
and can vary from one row to the next.

However, many customers use DynamoDB in a consistent, structured way
with a fairly well defined set of columns and types. Our HarbourBridge support
for DynamoDB focuses on this use-case, and we construct a Spanner schema
by inspecting table data.

For small tables, we inspect all rows of the table. For large tables, scanning the
entire table would be extremely expensive and slow, and so we only inspect the
first N rows (defined by the flag `schema-sample-size`) from the table scan.
While DynamoDB doesn't return scan results in order, they might not be a truly
random sample of rows. However, the alternative of randomly sampling rows
would be much more expensive. 

Columns with consistent types are assigned Spanner types as detailed below.
Columns without a consistent type are mapped to STRING.

dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+Spanner. For a large-scale table, scanning the entire table would be extremely
+slow. Therefore, we can only scan a part of it. We provide a flag
+`schema-sample-size` to specify how many rows to use for schema inference.
+
+In addition, for the performance purpose, we choose to call Scan on the table
+instead of randomly sample rows. This might be biased but we believe that the 
+result of Scan is not order-guaranteed and it can represent a sampled part
+of the table.
+
+### `Number`
```
nevinheintze, line 69: Maybe make these subsections.
I think #### is ok?

dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+The Number type in DynamoDB is encoded as a string, which can represent an
```
nevinheintze, line 71: Suggest rephrasing:

In most cases, we map the Number type in DynamoDB to Spanner's Numeric type.
However, since the [range of Numeric](https://cloud.google.com/spanner/docs/storing-numeric-data)
in Cloud Spanner is smaller than the [range of Number](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html)
in DynamoDB, this conversion could result in out of range with potential
precision loss. To address this possibility, we try to convert the sample data, and if it
consistently fails, we choose STRING type for the column.

dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+[Numeric type](https://cloud.google.com/spanner/docs/storing-numeric-data)
+in Cloud Spanner is smaller than the range of the
+[Number type](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html)
+in DynamoDB. The conversion would result in out of range with potential
+precision loss. For simplicity, we will first try to convert it into a NUMERIC
+type. If it fails, we convert it into a STRING type. 
+
+### `Null` Data Type
+
+In DynamoDB, an attribute can have a Null data type that represents an unknown
```
nevinheintze, line 82: This is the first time you've used attribute.
Maybe use it earlier when talking about schema i.e. say something "DynamoDB attributes are mapped to Spanner columns" and then use that terminology consistently when referring the DynamoDB columns.
Or just use column consistently throughout (including here).

dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+type as an occasional error. It would not be considered as a candidate type for
+the column. In this case, we can filter a certain amount of noise when modeling
+the real schema. 
+
+### Multi-type Columns
+
+For a special scenario, we may get a column that has equal distribution of two
+data types. E.g., a column has 40% rows in String and 60% rows in Number. If we
+choose Number as its type, it may lead to 40% data loss in the data conversion.
+In the migration, we define a conflicting threshold on normalized rows (after
```
nevinheintze, line 114: Drop normalized. Reader won't know what that means.
And I think the comment in parenthesis explains the point in any case.

dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+default, the conflicting threshold is 5% and if the percentages of two or more
+data types are greater than it, we would consider that the column has
+conflicting data types. As a safe choice, we define this column as a STRING type
+in Cloud Spanner. 
+
+## Data Conversion
+
+### A Scan for Entire Table
+
+A pass for the entire table is required. We will use Scan API to read data. Each
```
nevinheintze, line 125: This explanation seems to indicate we are only converting one table?

Maybe start by saying something like:

Data conversion proceeds table by table. For each table, we use the Scan API to read data.

dynamodb/README.md
```diff
@@ -0,0 +1,134 @@
...
+read has a size limit up to 1MB. By using the returned token, we make a
+subsequent call to continue retrieving data from the table.
+
+The row result contains the data type and data itself. According to our
+[inferred schema](#schema-inference), we will parse the row to a format that
+Cloud Spanner can support. If the value parsing fails, we would drop the entire
+row and record it as bad data in the report. If a column does not appear or 
+column has a NULL data type, we would process this as a NULL value in Cloud Spanner. 
+The job is done when all rows are inserted into the table in Cloud Spanner.
```
nevinheintze, line 134: Maybe drop this last sentence? I'm not sure it is needed.
Also, as it stands, it kind of suggests we only process one table.

nevinheintze (approved):
Looks good. Added a number of comments.

At some point we should modify the main HarbourBridge README to include DynamoDB support. But happy to defer that till we announce DynamoDB support.

hengfengli:
@nevinheintze Thanks for the comments.

dynamodb/README.md
```diff
@@ -0,0 +1,136 @@
...
+| `StringSet`        | `ARRAY<STRING>`            |                                           |
+| `NumberSet`        | `ARRAY<NUMERIC or STRING>` |                                           |
+| `BinarySet`        | `ARRAY<BYTES>`             |                                           |
+
+We discuss these, as well as other limits and notes on schema conversion, in the
+following sections.
+
+### Schema Inference
+
+DynamoDB is schemaless database: other than a primary index and optional
```
nevinheintze, line 56: Missing "a" i.e. "DynamoDB is a schemaless"

I think this was probably my typo -- sorry!

dynamodb/README.md
```diff
@@ -0,0 +1,136 @@
...
+#### `List` and `Map`
+
+In Cloud Spanner, the most similar type to List and Map is
+[STRUCT](https://cloud.google.com/spanner/docs/data-types#struct_type), but it
+is not a valid column type (available for query but not for storage).
+Therefore, we encode them into a json string. 
+
+#### Occasional Errors
+
+As no schema is forced when writing to a DynamoDB table, it can happen that a
```
nevinheintze, line 103: Suggest rephrasing:

To prevent a few spurious rows from disrupting our schema construction
for an otherwise consistently typed table, we define an error threshold: if
the percentage of the data type is lower than or equal to an extremely
low value (0.1%), then we consider the data type as an occasional error.
Such rows are not considered as a candidate type for the column.

nevinheintze (approved):

Issue #21, opened 2020-01-27
Consider support for indexes


nevinheintze:
HarbourBridge maps PostgreSQL primary keys to Spanner primary keys. But it ignores other indexing. These could be mapped to Spanner secondary indexes.

PR #95, opened 2020-12-15
feature: secondary index

Pradip97:
Index key PR comment:  > This PR is for the new feature: Index keys addressing issue https://github.com/cloudspannerecosystem/harbourbridge/issues/21 , changes included in this PR:
> *`internal/convert.go` - print indexes along with create table.
> *`internal/mapping.go` - added GetSpannerIndexKeyName to convert source key name to spanner key name avoiding collision.  
> * `mysql/infoschema.go` - support for index key getIndexes function
> * `mysql/mysqldump.go` - support for index key, we process CREATE INDEX statements, along with indexes mentioned in CREATE TABLE and ALTER TABLE, we also convert UNIQUE constraint to UNIQUE index.
> * `mysql/toddl.go` - support for index key
> * `postgres/infoschema.go` - support for index key getIndexes function
> * `postgres/pg_dump.go` - support for index key, we process CREATE INDEX statements, we also convert UNIQUE constraint to UNIQUE index. 
> * `postgres/toddl.go` - support for index key

> Note: we convert all unique constraint to unique secondary index in spanner as mentioned [here](https://cloud.google.com/spanner/docs/migrating-postgres-spanner#indexes) . this applies to both postgres and mysql.
> Note: we skip ordering of keys in mysqldump (i.e, ascending/descending) as mysqldump parser is not able to parse this order.
> This PR is completed.


mysql/toddl.go
```diff
@@ -228,3 +232,31 @@ func cvtForeignKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Forei
...
+				continue
+			}
+			srcIndex.Name = fmt.Sprintf("Index_%s_%x-%x", srcTable, b[0:2], b[2:4])
+		}
+
+		spKeyName := internal.GetSpannerIndexKeyName(srcIndex.Name, schemaIndexKeys)
+		spIndexes = append(spIndexes, ddl.CreateIndex{Name: spKeyName, Unique: srcIndex.Unique, Keys: spKeys})
+	}
+	return spIndexes
+}
```
agasheesh, line 262: this function seems to be duplicated between mysql and postgresql - one way to remove this duplication is to define a function on schema.Index to produce the corresponding ddl.CreateIndex statement since a given schema.Index struct contains all (but srcTable) fields required to produce the corresponding ddl.CreateIndex struct

internal/convert.go
```diff
@@ -179,6 +179,9 @@ func (conv *Conv) GetDDL(c ddl.Config) []string {
 	var ddl []string
 	for _, t := range tables {
 		ddl = append(ddl, conv.SpSchema[t].PrintCreateTable(c))
+		for _, i := range conv.SpSchema[t].Indexes {
```
agasheesh, line 182: just a minor nit pick, but `index` would be more readable than `i` in this case

internal/mapping.go
```diff
@@ -164,3 +164,36 @@ func GetSpannerKeyName(srcKeyName string, schemaForeignKeys map[string]bool) str
...
+		id := len(schemaIndexKeys)
+		for {
+			c := spKeyName + "_" + strconv.Itoa(id)
+			if _, found := schemaIndexKeys[c]; !found {
+				spKeyName = c
+				break
+			}
+			id++
+		}
+	}
```
agasheesh, line 196: This pattern is now duplicated quite often in the code, consider moving this to a common function 

func findNextUniqueKey(key string, existingKeys) string {
 ...
}

agasheesh:
Just FYI, this pull request seem to be fixing the issue https://github.com/cloudspannerecosystem/harbourbridge/issues/21. Please feel free to link the same here to linked issues.

mysql/README.md
```diff
@@ -197,6 +197,10 @@ and `ON UPDATE` actions, we drop them.
 Spanner does not currently support default values. We drop these
 MySQL features during conversion.
 
+### Secondary Indexes
+
+The tool maps MySQL secondary indexes to spanner secondary indexes, and preserves constraint names where possible. Note that Spanner requires index key constraint names to be globally unique (within a database), but in MySQL they only have to be unique for a table, so we add a uniqueness suffix to a name if needed.The tool also maps `UNIQUE` constraint into `UNIQUE` secondary index. Note that for MySQL dump we skip ordering of the key column (i.e, ascending/descending) as mysqldump parser is not able to parse those values.
+
```
agasheesh, line 203: > Note that for MySQL dump we skip ordering of the key column (i.e, ascending/descending) as mysqldump parser is not able to parse those values.

Can you please open an issue and add a corresponding TODO in the code so that we don't forget fixing the same in the future.

agasheesh (changes requested):
Thanks for making these changes. This is a first pass on the pull request.

spanner/ddl/ast.go
```diff
@@ -167,6 +167,7 @@ type CreateTable struct {
 	ColDefs  map[string]ColumnDef // Provides definition of columns (a map for simpler/faster lookup during type processing)
 	Pks      []IndexKey
 	Fks      []Foreignkey
+	Indexes  []CreateIndex
```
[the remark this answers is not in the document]
Pradip97: To be consistent with foreign key and primary key, i think keeping indexes as an array is better option. please let me know if there is any issue in keeping it as an array, i'll update it.

postgres/toddl.go
```diff
@@ -216,3 +220,30 @@ func cvtForeignKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Forei
...
+			spCol, err := internal.GetSpannerCol(conv, srcTable, k.Column, true)
+			if err != nil {
+				conv.Unexpected(fmt.Sprintf("Can't map index for table %s", srcTable))
+				continue
+			}
+			spKeys = append(spKeys, ddl.IndexKey{Col: spCol, Desc: k.Desc})
+		}
+		if srcIndex.Name == "" {
+			b := make([]byte, 4)
+			_, err := rand.Read(b)
```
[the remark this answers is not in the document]
Pradip97: its an inbuilt function and ideally it won't throw any error, but as it is returning an error object, we are checking for possibility.

mysql/toddl.go
Pradip97: if we make this a function on schema.Index, we need to put this in schema package, but in that case we will get cyclic import issue, as schema will be importing from internal and internal is importing from schema.

spanner/ddl/ast.go
```diff
@@ -207,20 +208,27 @@ func (ct CreateTable) PrintCreateTable(config Config) string {
...
-	Name  string
-	Table string
-	Keys  []IndexKey
+	Name string
+	/* Since CreateIndex is printed along with CreateTable,
+	storing Table name is not required */
+	// Table string
+	Unique bool
+	Keys   []IndexKey
 	// We have no requirements for unique and null-filtered options and
```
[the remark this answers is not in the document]
Pradip97: updated comment

spanner/ddl/ast.go
```diff
@@ -207,20 +208,27 @@ func (ct CreateTable) PrintCreateTable(config Config) string {
 // CreateIndex encodes the following DDL definition:
 //     create index: CREATE [UNIQUE] [NULL_FILTERED] INDEX index_name ON table_name ( key_part [, ...] ) [ storing_clause ] [ , interleave_clause ]
 type CreateIndex struct {
-	Name  string
-	Table string
-	Keys  []IndexKey
+	Name string
+	/* Since CreateIndex is printed along with CreateTable,
```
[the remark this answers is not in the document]
Pradip97: makes sense, i have converted back to original.

spanner/ddl/ast.go
```diff
@@ -207,20 +208,27 @@ func (ct CreateTable) PrintCreateTable(config Config) string {
...
+	// Table string
+	Unique bool
+	Keys   []IndexKey
 	// We have no requirements for unique and null-filtered options and
 	// storing/interleaving clauses yet, so we omit them for now.
 }
 
 // PrintCreateIndex unparses a CREATE INDEX statement.
-func (ci CreateIndex) PrintCreateIndex(c Config) string {
+func (ci CreateIndex) PrintCreateIndex(Table string, c Config) string {
```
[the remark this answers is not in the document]
Pradip97: i have converted back to original.

spanner/ddl/ast.go
```diff
@@ -207,20 +208,27 @@ func (ct CreateTable) PrintCreateTable(config Config) string {
...
 	var keys []string
 	for _, p := range ci.Keys {
 		keys = append(keys, p.PrintIndexKey(c))
 	}
-	return fmt.Sprintf("CREATE INDEX %s ON %s (%s)", c.quote(ci.Name), c.quote(ci.Table), strings.Join(keys, ", "))
+	var unique string
+	if ci.Unique == true {
+		unique = "UNIQUE"
+	}
+	return fmt.Sprintf("CREATE %s INDEX %s ON %s (%s)", unique, c.quote(ci.Name), c.quote(Table), strings.Join(keys, ", "))
```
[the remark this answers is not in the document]
Pradip97: updated syntax.

spanner/ddl/ast_test.go
```diff
@@ -101,21 +102,30 @@ func TestPrintCreateTable(t *testing.T) {
...
 func TestPrintCreateIndex(t *testing.T) {
-	ci := CreateIndex{
-		"myindex",
-		"mytable",
-		[]IndexKey{IndexKey{Col: "col1", Desc: true}, IndexKey{Col: "col2"}},
-	}
+	ci := []CreateIndex{
+		CreateIndex{
+			"myindex",
+			false,
```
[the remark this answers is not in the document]
Pradip97: added comment

spanner/ddl/ast_test.go
```diff
@@ -101,21 +102,30 @@ func TestPrintCreateTable(t *testing.T) {
...
+		CreateIndex{
+			"myindex2",
+			true,
+			[]IndexKey{IndexKey{Col: "col1", Desc: true}, IndexKey{Col: "col2"}},
+		}}
 	tests := []struct {
 		name       string
+		table      string
 		protectIds bool
 		expected   string
```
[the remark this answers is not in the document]
Pradip97: updated, moved expected to last.

postgres/toddl.go
```diff
@@ -216,3 +220,30 @@ func cvtForeignKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Forei
...
 }
+
+func cvtIndexes(conv *internal.Conv, srcTable string, srcIndexes []schema.Index, schemaIndexKeys map[string]bool) []ddl.CreateIndex {
+	var spIndexes []ddl.CreateIndex
+	for _, srcIndex := range srcIndexes {
+		var spKeys []ddl.IndexKey
+		for _, k := range srcIndex.Keys {
+			spCol, err := internal.GetSpannerCol(conv, srcTable, k.Column, true)
+			if err != nil {
+				conv.Unexpected(fmt.Sprintf("Can't map index for table %s", srcTable))
```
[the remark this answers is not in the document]
Pradip97: updated error message

postgres/toddl_test.go
```diff
@@ -43,6 +43,7 @@ func TestToSpannerType(t *testing.T) {
 		},
 		PrimaryKeys: []schema.Key{schema.Key{Column: "a"}},
 		ForeignKeys: []schema.ForeignKey{schema.ForeignKey{Name: "fk_test", Columns: []string{"d"}, ReferTable: "ref_table", ReferColumns: []string{"b"}}},
+		Indexes:     []schema.Index{schema.Index{Name: "index1", Unique: true, Keys: []schema.Key{schema.Key{Column: "a", Desc: false}, schema.Key{Column: "d", Desc: true}}}},
```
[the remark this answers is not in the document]
Pradip97: added another index with Unique: false.

internal/convert.go
Pradip97: updated from i to index

internal/mapping.go
Pradip97: separated common code into a function.

mysql/README.md
Pradip97: opened an issue https://github.com/cloudspannerecosystem/harbourbridge/issues/96

internal/mapping.go
```diff
@@ -128,6 +128,27 @@ func GetSpannerCols(conv *Conv, srcTable string, srcCols []string) ([]string, er
 	return spCols, nil
 }
 
+func getUniqueKeyName(srcKeyName string, schemaKeys map[string]bool) string {
```
agasheesh, line 131: Consider renaming this as following:

func getUniqueId(srcId string, spannerIds map[string]bool) string

This will allow you to re-use this function further for tableName, columnName in addition to indexKeyName and foreignKeyName.

Also, it's a good idea to move this utility function to the end of the file, such that public functions are listed first in the file.

internal/mapping.go
```diff
@@ -145,22 +166,20 @@ func GetSpannerKeyName(srcKeyName string, schemaForeignKeys map[string]bool) str
...
+// We need to make sure
+// of the following things:
+// a) the new index key name is legal
+// b) the new index key name doesn't clash with other Spanner
+//    index key names
+// Note that index key constraint names in Spanner have to be globally unique
+// (across the database). But in some source databases, such as MySQL,
+// they only have to be unique for a table. Hence we must map each source
+// constraint name to a unique spanner constraint name.
+func GetSpannerIndexKeyName(srcKeyName string, schemaIndexKeys map[string]bool) string {
```
agasheesh, line 183: Consider renaming these functions as following:

func GetSpannerKeyName(srcKeyName string, schemaForeignKeys map[string]bool) string
->
func ToSpannerForeignKey(srcForeignKey string, spannerForeignKeys map[string]bool) string

func GetSpannerIndexKeyName(srcKeyName string, schemaIndexKeys map[string]bool) string
->
func ToSpannerIndexKey(srcIndexKey string, spannerIndexKeys map[string]bool) string

func GetSpannerTable(conv *Conv, srcTable string) (string, error) 
->
func ToSpannerTable(srcTable string, conv *Conv) (string, error) {

func GetSpannerCol(conv *Conv, srcTable, srcCol string, mustExist bool) (string, error) {
->
func ToSpannerCol(srcTable, srcCol string, mustExist bool, conv *Conv) (string, error)

Motivation:
1. You can drop Name suffix e.g., srcKeyName -> srcIndexKey, since it's clear from the string type that it's the name and is consistent with the other argument i.e., schemaIndexKeys
2. srcIndexKey provides more info about the input argument compared to srcKeyName
3. To make function definitions consistent in this file.
4. Good idea to move conv *Conv to being the last argument to signal that it may be updated within the function.

internal/mapping_test.go
```diff
@@ -119,3 +119,30 @@ func TestGetSpannerKeyName(t *testing.T) {
 		assert.Equal(t, tc.spKeyName, spKeyName, tc.name)
 	}
 }
+
+func TestGetSpannerIndexKeyName(t *testing.T) {
```
agasheesh, line 123: Instead of duplicating these tests, consider testing the util function `func getUniqueId(srcId string, spannerIds map[string]bool) string` itself. It will make it far easier to update tests in case this function changes in the future.

mysql/infoschema.go
```diff
@@ -375,6 +380,47 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+			conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
+			continue
+		}
+		if _, found := m[name]; !found {
+			keyNames = append(keyNames, name)
+		}
+		isDesc := false
+		if collation == "D" {
+			isDesc = true
+		}
```
agasheesh, line 412: consider rewriting this as for brevity

isDesc := (collation == "D"), and similarly below
unique[name] = (nonUnique == "0")

mysql/infoschema.go
```diff
@@ -375,6 +380,47 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+	var name, col, seq, collation, nonUnique string
+	m := make(map[string][]schema.Key)
+	unique := make(map[string]bool)
+	var keyNames []string
+	for rows.Next() {
+		err := rows.Scan(&name, &col, &seq, &collation, &nonUnique)
+		if err != nil {
+			conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
+			continue
+		}
```
agasheesh, line 405: It's more typical to write this in Go as

if err := rows.Scan(&name, &col, &seq, &collation, &nonUnique); err != nil {
   conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
   continue
}

similarly elsewhere

mysql/infoschema.go
```diff
@@ -375,6 +380,47 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+			AND TABLE_NAME = ?
+			AND INDEX_NAME != 'PRIMARY' 
+		ORDER BY INDEX_NAME, SEQ_IN_INDEX;`
+	rows, err := db.Query(q, table.schema, table.name)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	var name, col, seq, collation, nonUnique string
+	m := make(map[string][]schema.Key)
```
agasheesh, line 397: considering using a more descriptive variable name for this map since it's difficult to read in the current form

also, rather than create maps for keys i.e., m and for unique constraint i.e., unique, it might be easier to create a map from indexKey name to schema.Key, and populate it within rows.Next() loop; this way you may be able to catch 

at the very end before returning you can then create an array from map with sorted indexKey names

agasheesh: I am also wondering whether indexKey names are unique in mysql db globally or per table?

agasheesh (changes requested):

internal/mapping_test.go
Pradip97: Actually ToSpannerIndexKey and ToSpannerForeignKey behaves differently when source key name is empty string, considering that we need to test these functions separately

mysql/infoschema.go
Pradip97: yeah its much cleaner, thanks

mysql/infoschema.go
Pradip97: yeah this looks cleaner, thanks

internal/mapping.go
Pradip97: updated names for GetSpannerKeyName and GetSpannerIndexKeyName, but skipping renaming other functions at the moment, as those are used at lot of places. keeping this comment open so that we can revisit this later.

internal/mapping.go
Pradip97: updated name of getUniqueKeyName

mysql/infoschema.go
Pradip97: IndexKey names are unique per table only in mysql, while in postgres it has to be unique globally

Pradip97: updated variable names and updated approach as suggested for populating indexes.

agasheesh (changes requested):
Still reviewing test cases. Sending out the comments I have so far. I may have left comments only in mysql code for some common suggestions, please fix the same in corresponding pgsql code as well where applicable.

Thanks for sticking through all the changes, I think we are very close to submitting this PR.

mysql/infoschema.go
```diff
@@ -375,6 +380,42 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+	for rows.Next() {
+		if err := rows.Scan(&indexName, &column, &sequence, &collation, &nonUnique); err != nil {
+			conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
+			continue
+		}
+		if _, found := mIndexes[indexName]; !found {
+			keyNames = append(keyNames, indexName)
+			mIndexes[indexName] = schema.Index{Name: indexName, Unique: (nonUnique == "0"), Keys: []schema.Key{schema.Key{Column: column, Desc: (collation == "D")}}}
+			continue
+		}
```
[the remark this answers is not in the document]
Pradip97: this is not unexpected situation, its just checking if it is new index.

mysql/infoschema.go
```diff
@@ -375,6 +380,42 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+			AND TABLE_NAME = ?
+			AND INDEX_NAME != 'PRIMARY' 
+		ORDER BY INDEX_NAME, SEQ_IN_INDEX;`
+	rows, err := db.Query(q, table.schema, table.name)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	var indexName, column, sequence, collation, nonUnique string
+	mIndexes := make(map[string]schema.Index)
```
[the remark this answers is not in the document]
Pradip97: Because it is a map of indexes, mapping from indexName to schema.Index, and in the end we are appending all the indexes in indexes array.

mysql/mysqldump.go
```diff
@@ -142,12 +142,35 @@ func processStatement(conv *internal.Conv, stmt ast.StmtNode) bool {
 	case *ast.InsertStmt:
 		processInsertStmt(conv, s)
 		return true
+	case *ast.CreateIndexStmt:
+		if conv.SchemaMode() {
+			processCreateIndex(conv, s)
+		}
```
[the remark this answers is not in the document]
Pradip97: i'm not sure i'm understanding it correctly, why there is processInsertStmt if s==*ast.CreateIndexStmt, and why do we need to process it in DataMode().

mysql/infoschema.go
```diff
@@ -375,6 +380,42 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+	var keyNames []string
```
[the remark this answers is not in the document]
Pradip97: changed from indexName to index for single index name, 
and changed from keyNames to indexNames for list of index names.

mysql/infoschema.go
```diff
@@ -375,6 +380,42 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
 	return foreignKeys, nil
 }
 
+// getIndexes return list all the indexes.
+func getIndexes(conv *internal.Conv, db *sql.DB, table schemaAndName) (indexes []schema.Index, err error) {
```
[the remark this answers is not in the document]
Pradip97: got it, using unnamed return values for consistency.

mysql/mysqldump.go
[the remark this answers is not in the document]
agasheesh: Sorry about the typo, it should be `*ast.InsertStmt` instead.

```
if conv.DataMode() {
    if s == *ast.InsertStmt {
        processInsertStmt(conv, s)
        return true
    }
} else { // conv.SchemaMode()
    switch s := stmt.(type) {
        ....
    }
}
return false
```

mysql/toddl.go
agasheesh: I see, we will probably get to it as part of larger refactoring. Sounds good to keep at as it is. Resolving.

postgres/toddl.go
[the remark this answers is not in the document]
agasheesh: SGTM, let's change to `if _, err := ...; err != nil` style here though. And add a comment after `if srcIndex.Name == ""` as following:

// Generate a random name if index name is empty in postgres since spanner requires globally unique non-empty index names.

spanner/ddl/ast_test.go
```diff
@@ -101,21 +102,31 @@ func TestPrintCreateTable(t *testing.T) {
...
 	}{
-		{"no quote", false, "CREATE INDEX myindex ON mytable (col1 DESC, col2)"},
-		{"quote", true, "CREATE INDEX `myindex` ON `mytable` (`col1` DESC, `col2`)"},
+		{"no quote non unique", false, ci[0], "CREATE INDEX myindex ON mytable (col1 DESC, col2)"},
+		{"quote non unique", true, ci[0], "CREATE INDEX `myindex` ON `mytable` (`col1` DESC, `col2`)"},
+		{"unique key", true, ci[1], "CREATE UNIQUE INDEX `myindex2` ON `mytable` (`col1` DESC, `col2`)"},
 	}
 	for _, tc := range tests {
-		assert.Equal(t, normalizeSpace(tc.expected), normalizeSpace(ci.PrintCreateIndex(Config{ProtectIds: tc.protectIds})))
+		assert.Equal(t, normalizeSpace(tc.expected), normalizeSpace(tc.index.PrintCreateIndex(Config{ProtectIds: tc.protectIds})))
```
agasheesh, line 129: [optional] while we are here, let's get rid of normalizeSpace if possible - if it turns to be rather a lot of work, then let's leave it as it is since this is existing code

postgres/toddl.go
```diff
@@ -216,3 +220,30 @@ func cvtForeignKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Forei
...
+			b := make([]byte, 4)
+			_, err := rand.Read(b)
+			if err != nil {
+				conv.Unexpected(fmt.Sprintf("Can't map index name for table %s", srcTable))
+				continue
+			}
+			srcIndex.Name = fmt.Sprintf("Index_%s_%x-%x", srcTable, b[0:2], b[2:4])
+		}
+		spKeyName := internal.ToSpannerIndexKey(srcIndex.Name, schemaIndexKeys)
+		spIndexes = append(spIndexes, ddl.CreateIndex{Name: spKeyName, Table: spTableName, Unique: srcIndex.Unique, Keys: spKeys})
```
agasheesh, line 246: it's easier to read it as following - 

```
spIndex := ddl.CreateIndex{
     Name: spKeyName,
     Table: spTableName,
     Unique: srcIndex.Unique,
     Keys: spKeys
}

spIndexes = append(spIndexes,spIndex)
```

mysql/mysqldump.go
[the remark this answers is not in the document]
agasheesh: On second thought, let's leave it as it is since this is existing code and not related to this PR.

postgres/pgdump.go
```diff
@@ -185,13 +185,48 @@ func processStatements(conv *internal.Conv, statements []nodes.Node) *copyOrInse
...
+	var schemaKeys []schema.Key
+	for _, ip := range indexParams.Items {
+		schemaKeys = append(schemaKeys, schema.Key{Column: *ip.(nodes.IndexElem).Name})
+	}
+	return schemaKeys
+}
+
+func processIndexStmt(conv *internal.Conv, n nodes.IndexStmt) {
+	if n.Relation == nil {
+		logStmtError(conv, n, fmt.Errorf("relation is nil"))
```
agasheesh, line 209: We should return more descriptive errors e.g., `fmt.Errorf("Cannot process index statement %q with nil relation.")`

postgres/pgdump.go
```diff
@@ -185,13 +185,48 @@ func processStatements(conv *internal.Conv, statements []nodes.Node) *copyOrInse
...
+		ctable := conv.SrcSchema[tableName]
+		var cols []string
+		for _, ip := range n.IndexParams.Items {
+			cols = append(cols, *ip.(nodes.IndexElem).Name)
+		}
+		ctable.Indexes = append(ctable.Indexes, schema.Index{Name: *n.Idxname, Unique: *&n.Unique, Keys: toIndexKeys(conv, tableName, n.IndexParams.Items)})
+		conv.SrcSchema[tableName] = ctable
+	} else {
+		conv.SkipStatement(prNodes([]nodes.Node{n}))
+	}
```
agasheesh, line 227: I wonder if this is an expected error scenario? If not, should be use `conv.Unexpected` instead with appropriate message?

postgres/pgdump.go
```diff
@@ -585,6 +620,10 @@ func updateSchema(conv *internal.Conv, table string, cs []constraint, stmtType s
 			ct := conv.SrcSchema[table]
 			ct.ForeignKeys = append(ct.ForeignKeys, toForeignKeys(c)) // Append to previous foreign keys.
 			conv.SrcSchema[table] = ct
+		case nodes.CONSTR_UNIQUE:
+			ct := conv.SrcSchema[table]
+			ct.Indexes = append(ct.Indexes, schema.Index{Name: "", Unique: true, Keys: toSchemaKeys(conv, table, c.cols)})
+			conv.SrcSchema[table] = ct
```
agasheesh, line 626: This is non-trivial code, and deserves a comment on the line of -

```
\\ Convert unique column constraint in postgres to a corresponding unique index in Spanner since Spanner doesn't support unique constraints on columns.
```

postgres/pgdump.go
```diff
@@ -620,6 +659,19 @@ func toSchemaKeys(conv *internal.Conv, table string, s []string) (l []schema.Key
 	return l
 }
 
+// toIndexKeys converts a string list of PostgreSQL primary keys to
+// schema Index keys.
+func toIndexKeys(conv *internal.Conv, table string, s []nodes.Node) (l []schema.Key) {
```
agasheesh, line 664: Let's get rid of named return parameter as discussed before.

postgres/pgdump.go
```diff
@@ -620,6 +659,19 @@ func toSchemaKeys(conv *internal.Conv, table string, s []string) (l []schema.Key
...
+	for _, k := range s {
```
agasheesh, line 665: not sure what do k and s stand here for - why not something like `for _, node := range nodes`

postgres/pgdump.go
```diff
@@ -185,13 +185,48 @@ func processStatements(conv *internal.Conv, statements []nodes.Node) *copyOrInse
...
+		logStmtError(conv, n, fmt.Errorf("can't get table name: %w", err))
+		return
+	}
+	if _, ok := conv.SrcSchema[tableName]; ok {
+		ctable := conv.SrcSchema[tableName]
+		var cols []string
+		for _, ip := range n.IndexParams.Items {
+			cols = append(cols, *ip.(nodes.IndexElem).Name)
+		}
+		ctable.Indexes = append(ctable.Indexes, schema.Index{Name: *n.Idxname, Unique: *&n.Unique, Keys: toIndexKeys(conv, tableName, n.IndexParams.Items)})
```
agasheesh, line 223: it's easier to read when each field in a struct is defined on a new line -

e.g., 

```
ctable.Indexes = append(ctable.Indexes, schema.Index{
    Name: *n.Idxname,
    Unique: *&n.Unique,
    Keys: toIndexKeys(conv, tableName, n.IndexParams.Items)
})
```

postgres/pgdump.go
agasheesh, line 626: I wonder why do we need `conv.SrcSchema[table] = ct` here (and similarly in the existing code in the cases above since points to the element in the map and isn't copied by value simply)

postgres/pgdump.go
```diff
@@ -585,6 +620,10 @@ func updateSchema(conv *internal.Conv, table string, cs []constraint, stmtType s
 			ct := conv.SrcSchema[table]
 			ct.ForeignKeys = append(ct.ForeignKeys, toForeignKeys(c)) // Append to previous foreign keys.
 			conv.SrcSchema[table] = ct
+		case nodes.CONSTR_UNIQUE:
```
agasheesh, line 623: I think we are missing tests for this case.

postgres/infoschema.go
```diff
@@ -415,6 +420,67 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+		}
+		if _, found := mIndexes[index]; !found {
+			indexNames = append(indexNames, index)
+			mIndexes[index] = schema.Index{Name: index, Unique: (isUnique == "t"), Keys: []schema.Key{schema.Key{Column: column, Desc: (collation == "DESC")}}}
+			continue
+		}
+		indexKey := mIndexes[index]
+		indexKey.Keys = append(indexKey.Keys, schema.Key{Column: column, Desc: (collation == "DESC")})
+		mIndexes[index] = indexKey
+	}
```
agasheesh, line 476: most of the comments in mysql/inforschema.go related to naming also apply here

mysql/mysqldump.go
```diff
@@ -321,6 +360,11 @@ func processAlterTable(conv *internal.Conv, stmt *ast.AlterTableStmt) {
 					ctable.ForeignKeys = append(ctable.ForeignKeys, constraint.fk)
 					conv.SrcSchema[tableName] = ctable
 				}
+				if constraint.isUniqueKey {
+					ctable := conv.SrcSchema[tableName]
+					ctable.Indexes = append(ctable.Indexes, schema.Index{Name: "", Unique: true, Keys: []schema.Key{schema.Key{Column: colname, Desc: false}}})
+					conv.SrcSchema[tableName] = ctable
+				}
```
agasheesh, line 367: This is non-trivial code, and deserves a comment on the line of -

```
\\ Convert unique column constraint in mysql to a corresponding unique index in Spanner since Spanner doesn't support unique constraints on columns.
```

agasheesh (changes requested):

mysql/mysqldump.go
```diff
@@ -321,6 +360,11 @@ func processAlterTable(conv *internal.Conv, stmt *ast.AlterTableStmt) {
 					ctable.ForeignKeys = append(ctable.ForeignKeys, constraint.fk)
 					conv.SrcSchema[tableName] = ctable
 				}
+				if constraint.isUniqueKey {
```
agasheesh, line 363: I think we are missing test cases for when ALTER TABLE adds a uniqueness constraint.

postgres/toddl.go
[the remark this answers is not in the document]
Pradip97: added comment and updated err statement

internal/mapping_test.go
Pradip97: updated this, now testing getUniqueId and empty string case separately

mysql/infoschema.go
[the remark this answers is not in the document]
Pradip97: updated variable names

mysql/infoschema.go
```diff
@@ -375,6 +380,43 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+		WHERE TABLE_SCHEMA = ?
+			AND TABLE_NAME = ?
+			AND INDEX_NAME != 'PRIMARY' 
+		ORDER BY INDEX_NAME, SEQ_IN_INDEX;`
+	rows, err := db.Query(q, table.schema, table.name)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	var index, column, sequence, collation, nonUnique string
```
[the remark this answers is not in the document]
Pradip97: updated variable names

mysql/infoschema.go
```diff
@@ -375,6 +380,43 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+	mIndexes := make(map[string]schema.Index)
+	var indexNames []string
+	var indexes []schema.Index
```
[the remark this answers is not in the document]
Pradip97: updated variable name

mysql/infoschema.go
```diff
@@ -375,6 +380,43 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+	for rows.Next() {
+		if err := rows.Scan(&index, &column, &sequence, &collation, &nonUnique); err != nil {
+			conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
+			continue
+		}
+		if _, found := mIndexes[index]; !found {
+			indexNames = append(indexNames, index)
+			mIndexes[index] = schema.Index{Name: index, Unique: (nonUnique == "0"), Keys: []schema.Key{schema.Key{Column: column, Desc: (collation == "D")}}}
+			continue
```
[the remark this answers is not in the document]
Pradip97: updated code as suggested

postgres/toddl.go
Pradip97: updated code as suggested

postgres/pgdump.go
Pradip97: updated error statement as suggested, but not printing  statement as it is not in readable format

postgres/pgdump.go
Pradip97: whenever there is a situation when we can not read a statement we report skipStatement for that. but we also should report unexpected for that, i'll add it.

postgres/pgdump.go
Pradip97: added comment

postgres/pgdump.go
Pradip97: updated return type

postgres/pgdump.go
Pradip97: as nodes is an imported library name, we can not use it here.

postgres/pgdump.go
Pradip97: updated code as mentioned

postgres/pgdump.go
Pradip97: as ct is not a pointer, after changing actual content inside ct, it doesn't actually update conv.SrcSchema[table], so we have to reasign it to conv.SrcSchema[table].

postgres/pgdump.go
Pradip97: added tests for this case

mysql/mysqldump.go
Pradip97: added comment

mysql/mysqldump.go
Pradip97: added tests for ALTER TABLE statement

internal/mapping.go
```diff
@@ -141,26 +141,45 @@ func GetSpannerCols(conv *Conv, srcTable string, srcCols []string) ([]string, er
...
+//    index key names
+// Note that index key constraint names in Spanner have to be globally unique
+// (across the database). But in some source databases, such as MySQL,
+// they only have to be unique for a table. Hence we must map each source
+// constraint name to a unique spanner constraint name.
+func ToSpannerIndexKey(srcKeyName string, schemaIndexKeys map[string]bool) string {
+	return getUniqueId(srcKeyName, schemaIndexKeys)
+}
+
+func getUniqueId(srcKeyName string, schemaKeys map[string]bool) string {
```
nevinheintze, line 166: Thanks for abstracting this functionality out into a separate function!
I think id is the right concept to use here.

Suggest some slight tweaking:
getUniqueId -> getSpannerId  (since it both maps to spanner and manages uniqueness)
srcKeyName -> srcId
schemaKeys -> used

and then in the code, systematically use Id for KeyName.

mysql/README.md
```diff
@@ -197,6 +197,10 @@ and `ON UPDATE` actions, we drop them.
 Spanner does not currently support default values. We drop these
 MySQL features during conversion.
 
+### Secondary Indexes
+
+The tool maps MySQL secondary indexes to spanner secondary indexes, and preserves constraint names where possible. Note that Spanner requires index key constraint names to be globally unique (within a database), but in MySQL they only have to be unique for a table, so we add a uniqueness suffix to a name if needed.The tool also maps `UNIQUE` constraint into `UNIQUE` secondary index. Note that for MySQL dump we skip ordering of the key column (i.e, ascending/descending) as mysqldump parser is not able to parse those values.
```
nevinheintze, line 202: 1) This looks like one long line? Please break into separate lines (approx 70-80 characters),  as done elsewhere in this file. 
2) spanner --> Spanner
3) Missing space before new sentence "The tool ..."

mysql/README.md
```diff
@@ -197,6 +197,16 @@ and `ON UPDATE` actions, we drop them.
...
 MySQL features during conversion.
 
+### Secondary Indexes
+
+The tool maps MySQL secondary indexes to Spanner secondary indexes, and preserves
+constraint names where possible. Note that Spanner requires index key constraint
+names to be globally unique (within a database), but in MySQL they only have to be
+unique for a table, so we add a uniqueness suffix to a name if needed. The tool also
+maps `UNIQUE` constraint into `UNIQUE` secondary index. Note that for MySQL dump we
+skip ordering of the key column (i.e, ascending/descending) as mysqldump parser is
```
nevinheintze, line 207: Suggest rephrasing this last sentence:

Note that due to limitations of our
mysqldump parser, we are not able to handle key column ordering (i.e. ASC/DESC) in
mysqldump files. All key columns will be treated as ASC.

mysql/infoschema.go
```diff
@@ -375,6 +380,42 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
 	return foreignKeys, nil
 }
 
+// getIndexes return list all the indexes.
```
nevinheintze, line 383: Suggest:

// getIndexes return a list all indexes for the specified table.

mysql/infoschema.go
```diff
@@ -375,6 +380,42 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+		ORDER BY INDEX_NAME, SEQ_IN_INDEX;`
+	rows, err := db.Query(q, table.schema, table.name)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	var name, column, sequence, collation, nonUnique string
+	indexes := make(map[string]schema.Index)
+	var indexNames []string
+	var indexArr []schema.Index
```
nevinheintze, line 399: Is Arr short for array?

Suggest: indexArr --> indexes
since this isn't an array (it's a slice).
And maybe using indexMap for the current indexes variable?

mysql/infoschema.go
```diff
@@ -375,6 +380,42 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+		}
+		if _, found := indexes[name]; !found {
+			indexNames = append(indexNames, name)
+			indexes[name] = schema.Index{Name: name, Unique: (nonUnique == "0")}
+		}
+		index := indexes[name]
+		index.Keys = append(index.Keys, schema.Key{Column: column, Desc: (collation == "D")})
+		indexes[name] = index
+	}
+	sort.Strings(indexNames)
```
nevinheintze, line 413: Why do we need to sort this, since the query already returns rows sorted by INDEX_NAME?

mysql/README.md
Pradip97: updated content

mysql/infoschema.go
Pradip97: updated comment

mysql/infoschema.go
Pradip97: updated variable names

mysql/infoschema.go
Pradip97: removed sorting, sorting was required for previous implementation, missed to remove the line.

mysql/mysqldump.go
```diff
@@ -228,6 +263,12 @@ func processConstraint(conv *internal.Conv, table string, constraint *ast.Constr
 		updateCols(conv, ast.ConstraintPrimaryKey, constraint.Keys, st.ColDefs, table)
 	case ast.ConstraintForeignKey:
 		st.ForeignKeys = append(st.ForeignKeys, toForeignKeys(conv, constraint))
+	case ast.ConstraintIndex:
+		st.Indexes = append(st.Indexes, schema.Index{Name: constraint.Name, Keys: toSchemaKeys(constraint.Keys)})
+	case ast.ConstraintUniq:
+		// Convert unique column constraint in postgres to a corresponding unique index in Spanner since
```
nevinheintze, line 269: postgres --> mysql

mysql/mysqldump.go
```diff
@@ -201,14 +230,20 @@ func processCreateTable(conv *internal.Conv, stmt *ast.CreateTableStmt) {
 		if constraint.fk.Columns != nil {
 			fkeys = append(fkeys, constraint.fk)
 		}
+		if constraint.isUniqueKey {
```
nevinheintze, line 233: As discussed, we'd like to avoid doing spanner-specific transformations when transformation from source database to schema.go.

However, there is no simple way to do this right now. Let's keep as is. I've filed an issue to improve this part of HarbourBridge (see https://github.com/cloudspannerecosystem/harbourbridge/issues/105).

But let's add a TODO:

// TODO: Avoid Spanner-specific schema transformations in this file -- they should be
// only appear in toddl.go. This file should focus on generic transformation from source
// database schemas into schema.go.
//

mysql/mysqldump.go
```diff
@@ -236,10 +277,17 @@ func processConstraint(conv *internal.Conv, table string, constraint *ast.Constr
 
 // toSchemaKeys converts a string list of MySQL primary keys to
```
nevinheintze, line 278: Suggest changing the description of the function rather than adding and "also used for " comment.

How about:

// toSchemaKeys converts a string list of MySQL keys to schema keys.
// Note that we map all MySQL keys to ascending ordered schema keys.
// For primary keys: this is fine because MySQL primary keys are always ascending.
// However, for non-primary keys (aka indexes) this is incorrect: we are dropping
// the MySQL key order specification.
// TODO: Resolve ordering issue for non-primary keys.

mysql/mysqldump.go
```diff
@@ -236,10 +277,17 @@ func processConstraint(conv *internal.Conv, table string, constraint *ast.Constr
...
 // schema primary keys.
+// This function is also used for converting MySQL index key to
+// schema index keys, mysqldump parser is not able to parse order
+// of the key (i.e, ascending/descending), so we map all keys to
+// ascending.
 func toSchemaKeys(columns []*ast.IndexPartSpecification) (keys []schema.Key) {
 	for _, colname := range columns {
 		// MySQL primary keys have no notation of ascending/descending.
```
nevinheintze, line 286: Suggested deleting the comment here, since it is now covered in function comment.

Pradip97: deleted comment

mysql/mysqldump.go
Pradip97: added function comment

mysql/toddl.go
```diff
@@ -228,3 +232,38 @@ func cvtForeignKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Forei
...
+		for _, k := range srcIndex.Keys {
+			spCol, err := internal.GetSpannerCol(conv, srcTable, k.Column, true)
+			if err != nil {
+				conv.Unexpected(fmt.Sprintf("Can't map index key column name for table %s", srcTable))
+				continue
+			}
+			spKeys = append(spKeys, ddl.IndexKey{Col: spCol, Desc: k.Desc})
+		}
+		if srcIndex.Name == "" {
+			// Generate a random name if index name is empty in postgres
```
nevinheintze, line 249: postgres --> MySQL
spanner --> Spanner

mysql/toddl.go
```diff
@@ -228,3 +232,38 @@ func cvtForeignKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Forei
...
+			// since spanner requires globally unique non-empty index names.
+			b := make([]byte, 4)
+			if _, err := rand.Read(b); err != nil {
+				conv.Unexpected(fmt.Sprintf("Can't map index name for table %s", srcTable))
+				continue
+			}
+			srcIndex.Name = fmt.Sprintf("Index_%s_%x-%x", srcTable, b[0:2], b[2:4])
```
nevinheintze, line 256: I'm wondering if we need the complexity of random generation here.
Couldn't we just use 
```
srcIndex.Name = fmt.Sprintf("Index_%s", srcTable)
```
and then rely on ToSpannerIndexKey to add a unique id if needed?

postgres/README.md
```diff
@@ -202,6 +202,13 @@ Spanner doesn't support `ON DELETE` and `ON UPDATE` actions, so we drop these.
 Spanner does not currently support default values. We drop these
 PostgreSQL features during conversion.
 
+### Secondary Indexes
+
+The tool maps PostgresSQL secondary indexes to Spanner secondary indexes, and preserves
+constraint names where possible. The tool also maps `UNIQUE` constraint into `UNIQUE`
+secondary index, Check [here](https://cloud.google.com/spanner/docs/migrating-postgres-spanner#indexes)
+for more details.
```
nevinheintze, line 210: Suggest minor reword:

The tool maps PostgresSQL secondary indexes to Spanner secondary indexes, preserving
constraint names where possible. The tool also maps PostgreSQL `UNIQUE` constraints to
Spanner `UNIQUE` secondary indexes. Check [here](https://cloud.google.com/spanner/docs/migrating-postgres-spanner#indexes) for more details.

postgres/pgdump.go
```diff
@@ -185,13 +185,49 @@ func processStatements(conv *internal.Conv, statements []nodes.Node) *copyOrInse
...
+	if n.Relation == nil {
+		logStmtError(conv, n, fmt.Errorf("cannot process index statement with nil relation."))
+		return
+	}
+	tableName, err := getTableName(conv, *n.Relation)
+	if err != nil {
+		logStmtError(conv, n, fmt.Errorf("can't get table name: %w", err))
+		return
+	}
+	if _, ok := conv.SrcSchema[tableName]; ok {
```
nevinheintze, line 217: Combine this line and the next i.e.

if ctable, ok := conv.SrcSchema[tableName]; ok {

postgres/pgdump.go
```diff
@@ -185,13 +185,49 @@ func processStatements(conv *internal.Conv, statements []nodes.Node) *copyOrInse
...
+		ctable := conv.SrcSchema[tableName]
+		ctable.Indexes = append(ctable.Indexes, schema.Index{
+			Name:   *n.Idxname,
+			Unique: *&n.Unique,
```
nevinheintze, line 221: *&n.Unique -- isn't this just the same as n.Unique?

postgres/pgdump.go
```diff
@@ -620,6 +668,20 @@ func toSchemaKeys(conv *internal.Conv, table string, s []string) (l []schema.Key
 	return l
 }
 
+// toIndexKeys converts a string list of PostgreSQL primary keys to
```
nevinheintze, line 671: Replace "string list" with just list?
Since this isn't a string list.

postgres/toddl.go
```diff
@@ -216,3 +220,37 @@ func cvtForeignKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Forei
...
+		for _, k := range srcIndex.Keys {
+			spCol, err := internal.GetSpannerCol(conv, srcTable, k.Column, true)
+			if err != nil {
+				conv.Unexpected(fmt.Sprintf("Can't map index key column name for table %s", srcTable))
+				continue
+			}
+			spKeys = append(spKeys, ddl.IndexKey{Col: spCol, Desc: k.Desc})
+		}
+		if srcIndex.Name == "" {
+			// Generate a random name if index name is empty in postgres
```
nevinheintze: Same comment here as for mysql/toddl.go: not need for the random generation since we call ToSpannerIndexKey just below.
Just using:

```
	srcIndex.Name = fmt.Sprintf("Index_%s", srcTable)
```
would be sufficient.

postgres/infoschema.go
```diff
@@ -415,6 +420,65 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
 	return foreignKeys, nil
 }
 
+// getIndexes return a list of all indexes for the specified table.
+func getIndexes(conv *internal.Conv, db *sql.DB, table schemaAndName) ([]schema.Index, error) {
+	q := `SELECT
+			irel.relname AS index_name,
```
nevinheintze, line 426: Wow!
I've been starting at this for a while now, and playing with it in psql.
It looks solid, but it's much more complex that I'd expected.

Any insights you can offer for how you constructed this?
I'm wondering if we should add a comment to help with understanding/maintaining this.

mysql/toddl.go
Pradip97: updated comment

mysql/toddl.go
Pradip97: yes make sense to simplify this

postgres/README.md
Pradip97: updated readme

postgres/pgdump.go
Pradip97: updated

postgres/pgdump.go
Pradip97: yes its same, updated to n.Unique

postgres/pgdump.go
Pradip97: updated

postgres/toddl.go
Pradip97: updated this, removed random generation

postgres/toddl.go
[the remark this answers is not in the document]
Pradip97: not relevant anymore

mysql/infoschema.go
Pradip97: updated variable names as part of other comments

postgres/infoschema.go
Pradip97: Yeah, i took major help from our DBA team and some from this stackoverflow answer: https://stackoverflow.com/questions/6777456/list-all-index-names-column-names-and-its-table-name-of-a-postgresql-database/44460269#44460269

mysql/mysqldump.go
Pradip97: added TODO, and this can be resolved separately

spanner/ddl/ast_test.go
Pradip97: Yeah, better to handle that in separate PR

postgres/pgdump.go
Pradip97: unexpected statement also added

mysql/infoschema.go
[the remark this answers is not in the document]
Pradip97: this is updated

internal/mapping.go
```diff
@@ -144,27 +144,46 @@ func GetSpannerCols(conv *Conv, srcTable string, srcCols []string) ([]string, er
...
+func ToSpannerForeignKey(srcId string, used map[string]bool) string {
+	if srcId == "" {
 		return ""
 	}
-	spKeyName, _ := FixName(srcKeyName)
-	if _, found := schemaForeignKeys[spKeyName]; found {
+	return getSpannerId(srcId, used)
+}
+
+// ToSpannerIndexKey maps source index key name to
```
nevinheintze, line 154: Can we clean up the comment a little.
- should use full sentences and punctuation e.g. add full stop after first sentence.
- format so its more readable e.g. join the couple of short lines of text together such as lines 156 and 157.
- be consistent about terminology e.g. index key vs index key name vs constraint name; maybe we should use index name everywhere (including the function name).

postgres/infoschema.go
nevinheintze: How about adding a comment:

// Extracting index definitions from PostgreSQL information schema tables is complex.
// See https://stackoverflow.com/questions/6777456/list-all-index-names-column-names-and-its-table-name-of-a-postgresql-database/44460269#44460269
// for background.

postgres/toddl.go
```diff
@@ -225,3 +228,32 @@ func cvtForeignKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Forei
...
+				continue
+			}
+			spKeys = append(spKeys, ddl.IndexKey{Col: spCol, Desc: k.Desc})
+		}
+		if srcIndex.Name == "" {
+			// Generate a name if index name is empty in Postgres.
+			// Collision of index name will be handled by ToSpannerIndexKey.
+			srcIndex.Name = fmt.Sprintf("Index_%s", srcTable)
+		}
+		spKeyName := internal.ToSpannerIndexKey(srcIndex.Name, schemaIndexKeys)
```
nevinheintze, line 249: As noted in mapping.go, how about we make the naming a bit more precise.
Suggest ToSpannerIndexName, and changing the var name to spIndexName.
And maybe schemaIndexKeys to spIndexNames?

Or is there some reason to keep using "Keys" here?
I find it confusing/misleading, since we have other kinds of keys here (the columns themselves).


be consistent about terminology e.g. index key vs index key name vs constraint name; maybe we should use index name everywhere, including the function name, the comment and in README.md?

nevinheintze (commented):
Finished review (after conflict resolution). Other than the issue noted below, I'm ready to approve this PR.

As noted in the chat room, I ran into an issue: I think we need to ensure that index names and foreign key constraint names don't overlap. In other words, it looks like index names and foreign key constraints share the same namespace in Spanner.

I ran into this while doing a final round of testing on sakila. Here's a cut down version:

CREATE TABLE film_category (
    film_id INT64 NOT NULL,
    category_id INT64 NOT NULL,
    last_update TIMESTAMP NOT NULL
) PRIMARY KEY (film_id, category_id);

CREATE TABLE category (
    category_id INT64 NOT NULL,
    name STRING(25) NOT NULL,
    last_update TIMESTAMP NOT NULL
) PRIMARY KEY (category_id);

CREATE INDEX fk_film_category_category ON film_category (category_id);

ALTER TABLE film_category ADD CONSTRAINT fk_film_category_category FOREIGN KEY (category_id) REFERENCES category (category_id);

This last statement fails with "Duplicate name in schema: fk_film_category_category".

postgres/infoschema.go
Pradip97: added comment

internal/mapping.go
Pradip97: sure, updated comment

postgres/toddl.go
Pradip97: updated function name as mentioned and parameter name as well

Pradip97:
> Finished review (after conflict resolution). Other than the issue noted below, I'm ready to approve this PR.
> 
> As noted in the chat room, I ran into an issue: I think we need to ensure that index names and foreign key constraint names don't overlap. In other words, it looks like index names and foreign key constraints share the same namespace in Spanner.
> 
> I ran into this while doing a final round of testing on sakila. Here's a cut down version:
> 
> CREATE TABLE film_category (
> film_id INT64 NOT NULL,
> category_id INT64 NOT NULL,
> last_update TIMESTAMP NOT NULL
> ) PRIMARY KEY (film_id, category_id);
> 
> CREATE TABLE category (
> category_id INT64 NOT NULL,
> name STRING(25) NOT NULL,
> last_update TIMESTAMP NOT NULL
> ) PRIMARY KEY (category_id);
> 
> CREATE INDEX fk_film_category_category ON film_category (category_id);
> 
> ALTER TABLE film_category ADD CONSTRAINT fk_film_category_category FOREIGN KEY (category_id) REFERENCES category (category_id);
> 
> This last statement fails with "Duplicate name in schema: fk_film_category_category".

mysql/toddl.go
```diff
@@ -32,7 +32,7 @@ import (
 // Spanner. It uses the source schema in conv.SrcSchema, and writes
 // the Spanner schema to conv.SpSchema.
 func schemaToDDL(conv *internal.Conv) error {
-	schemaForeignKeys := make(map[string]bool)
+	spKeyNames := make(map[string]bool)
```
nevinheintze, line 35: spKeyNames is kind of confusing (since we use it for foreign key constraint names and index names) -- I think key is misleading.
And maybe we should emphasize that this is a set of used names (to manage conflicts).

How about naming this usedNames

I think this is worth a comment. Suggest something like:

	// Tracks Spanner names that have been used for foreign key constraints
	// and indexes. We use this to ensure we generate unique names when
	// we mapping from MySQL to Spanner since Spanner requires all foreign
	// key and index names to be distinct (you can't use the same name 
	// for a foreign key constraint and an index).
	usedNames := make(map[string]bool)

mysql/toddl.go
```diff
@@ -229,3 +230,32 @@ func cvtForeignKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Forei
 	}
 	return spKeys
 }
+
+func cvtIndexes(conv *internal.Conv, spTableName string, srcTable string, srcIndexes []schema.Index, spIndexNames map[string]bool) []ddl.CreateIndex {
```
nevinheintze, line 234: Suggest renaming: 
  spIndexNames -> usedNames
to better reflect what this parameter does (and for consistency with callsite).

postgres/toddl.go
```diff
@@ -28,7 +28,7 @@ import (
 // Spanner. It uses the source schema in conv.SrcSchema, and writes
 // the Spanner schema to conv.SpSchema.
 func schemaToDDL(conv *internal.Conv) error {
-	schemaForeignKeys := make(map[string]bool)
+	spKeyNames := make(map[string]bool)
```
nevinheintze, line 31: Suggest same name change (usedNames) and added comment as in mysql/toddl.go.

postgres/toddl.go
```diff
@@ -225,3 +226,32 @@ func cvtForeignKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Forei
 	}
 	return spKeys
 }
+
+func cvtIndexes(conv *internal.Conv, spTableName string, srcTable string, srcIndexes []schema.Index, spIndexNames map[string]bool) []ddl.CreateIndex {
```
nevinheintze, line 230: Suggest renaming: 
  spIndexNames -> usedNames
to better reflect what this parameter does (and for consistency with callsite).

And similar comment for cvtForeignKeys.

internal/mapping.go
```diff
@@ -144,27 +144,44 @@ func GetSpannerCols(conv *Conv, srcTable string, srcCols []string) ([]string, er
 // (across the database). But in some source databases, such as PostgreSQL,
 // they only have to be unique for a table. Hence we must map each source
 // constraint name to a unique spanner constraint name.
```
nevinheintze, line 146: As discussed in the group chat, the name returned here must not conflict with a Spanner table name.

What I suggest is that we just add a TODO for this PR so we can get it in. I think the likelihood of mysql and postgres databases using the same name for tables, indexes or foreign key constraints is fairly small. We’ll handle it with a follow up PR where we can focus on doing the refactoring.

Suggest adding a TODO:

// TODO: ensure that the name we return does not conflict with a Spanner table name.

internal/mapping.go
```diff
@@ -144,27 +144,44 @@ func GetSpannerCols(conv *Conv, srcTable string, srcCols []string) ([]string, er
...
+
+// ToSpannerIndexName maps source index name to legal Spanner index name.
+// We need to make sure of the following things:
+// a) the new index name is legal
+// b) the new index name doesn't clash with other Spanner
+//    index names
+// Note that index key constraint names in Spanner have to be globally unique
+// (across the database). But in some source databases, such as MySQL,
+// they only have to be unique for a table. Hence we must map each source
+// constraint name to a unique spanner constraint name.
```
nevinheintze, line 162: As discussed in the group chat, the name returned here must not conflict with a Spanner table name.

What I suggest is that we just add a TODO for this PR so we can get it in. I think the likelihood of mysql and postgres databases using the same name for tables, indexes or foreign key constraints is fairly small. We’ll handle it with a follow up PR where we can focus on doing the refactoring.

Suggest adding a TODO:

// TODO: ensure that the name we return does not conflict with a Spanner table name.

mysql/infoschema.go
```diff
@@ -375,6 +380,41 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+		WHERE TABLE_SCHEMA = ?
+			AND TABLE_NAME = ?
+			AND INDEX_NAME != 'PRIMARY' 
+		ORDER BY INDEX_NAME, SEQ_IN_INDEX;`
+	rows, err := db.Query(q, table.schema, table.name)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	var name, column, sequence, collation, nonUnique string
```
nevinheintze, line 396: The type of collation should be something that handles NULL values because the COLLATION column can be NULL (and is for some sakila tables).
Suggest using database/sql's NullString type. And use the Valid method to check the value when reading it.

We should do a quick audit of other information schema queries to check that when we scan we have the appropriate type, both for mysql and postgres. I'll look into this on Monday.

postgres/toddl.go
Pradip97: updated this

postgres/toddl.go
Pradip97: updated as mentioned

mysql/toddl.go
Pradip97: updated as mentioned

mysql/toddl.go
Pradip97: added comment

internal/mapping.go
Pradip97: Added a quick fix and a background comment on it. we will pre populate usedNames with Spanner table names to handle collision with foreign key constraint name and index name.

internal/mapping.go
Pradip97: Added a quick fix in toddl.go and a background comment on it. we will pre populate usedNames with Spanner table names to handle collision with foreign key constraint name and index name.

mysql/mysqldump.go
```diff
@@ -228,18 +266,29 @@ func processConstraint(conv *internal.Conv, table string, constraint *ast.Constr
...
+// toSchemaKeys converts a string list of MySQL keys to schema keys.
+// Note that we map all MySQL keys to ascending ordered schema keys.
+// For primary keys: this is fine because MySQL primary keys are always ascending.
+// However, for non-primary keys (aka indexes) this is incorrect: we are dropping
+// the MySQL key order specification.
+// TODO: Resolve ordering issue for non-primary keys.
 func toSchemaKeys(columns []*ast.IndexPartSpecification) (keys []schema.Key) {
 	for _, colname := range columns {
-		// MySQL primary keys have no notation of ascending/descending.
-		// We map them all into ascending primary keys.
```
agasheesh, line 242: We have updated the comment but I don't see handling of ASC/DESC order in the code below; we are also missing the corresponding test cases as well as to handle ordering.

agasheesh: I remember now that we are not able to parse ASC/DESC information for mysqldump parser. Can you please add a TODO here pointing to the issue that you filed - https://github.com/cloudspannerecosystem/harbourbridge/issues/96?

agasheesh (dismissed):

postgres/infoschema.go
```diff
@@ -415,6 +420,68 @@ func getForeignKeys(conv *internal.Conv, db *sql.DB, table schemaAndName) (forei
...
+	var indexNames []string
+	var indexes []schema.Index
+	for rows.Next() {
+		if err := rows.Scan(&name, &column, &sequence, &isUnique, &collation); err != nil {
+			conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
+			continue
+		}
+		if _, found := indexMap[name]; !found {
+			indexNames = append(indexNames, name)
+			indexMap[name] = schema.Index{Name: name, Unique: (isUnique == "t")}
```
nevinheintze, line 473: I think that should be: isUnique == "true"
Can you double check?

I ran into this when testing sakila, and saw that unique indexes were not being respected.

nevinheintze (approved):
Finished reviewing and testing. Found a few minor issues. I've filed issues for two of them (one which isn't related to this PR, and one is FULLTEXT indexes, which I think we should defer). The other issues is for unique indexes (infoschema), which we should confirm and fix.

Other than that, looks good -- thanks!

postgres/infoschema.go
Pradip97: Yes, i have checked this already, it can be 't' or 'f' only.

mysql/mysqldump.go
Pradip97: Yes added todo and linked the issue.

postgres/pgdump.go
```diff
@@ -620,6 +667,20 @@ func toSchemaKeys(conv *internal.Conv, table string, s []string) (l []schema.Key
 	return l
 }
 
+// toIndexKeys converts a list of PostgreSQL primary keys to
+// schema Index keys.
```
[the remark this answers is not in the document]
Pradip97: updated comment

postgres/pgdump.go
```diff
@@ -185,13 +185,48 @@ func processStatements(conv *internal.Conv, statements []nodes.Node) *copyOrInse
...
 	return nil
 }
 
+func toSchemaIndexKeys(indexParams nodes.List) (keys []schema.Key) {
+	var schemaKeys []schema.Key
+	for _, ip := range indexParams.Items {
+		schemaKeys = append(schemaKeys, schema.Key{Column: *ip.(nodes.IndexElem).Name})
+	}
+	return schemaKeys
+}
```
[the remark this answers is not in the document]
Pradip97: yes, removed

postgres/pgdump.go
```diff
@@ -620,6 +667,20 @@ func toSchemaKeys(conv *internal.Conv, table string, s []string) (l []schema.Key
 	return l
 }
 
+// toIndexKeys converts a list of PostgreSQL primary keys to
+// schema Index keys.
+func toIndexKeys(conv *internal.Conv, table string, s []nodes.Node) []schema.Key {
```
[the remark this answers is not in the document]
Pradip97: removed unused parameters

postgres/toddl.go
```diff
@@ -225,3 +242,32 @@ func cvtForeignKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Forei
...
+			spCol, err := internal.GetSpannerCol(conv, srcTable, k.Column, true)
+			if err != nil {
+				conv.Unexpected(fmt.Sprintf("Can't map index key column name for table %s", srcTable))
+				continue
+			}
+			spKeys = append(spKeys, ddl.IndexKey{Col: spCol, Desc: k.Desc})
+		}
+		if srcIndex.Name == "" {
+			// Generate a name if index name is empty in Postgres.
+			// Collision of index name will be handled by ToSpannerIndexKey.
```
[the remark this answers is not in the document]
Pradip97: updated comment

postgres/infoschema.go
Pradip97: Yes Nevin, i double checked, it is either "true" or "false", apologies i missed to check in testing.
But strangely in postgres console it is either "t" or "f".
Screen Shot 2021-02-01 at 10 09 21 AM

Pradip97: Fixed this issue and corresponding tests.

nevinheintze: Thanks!

mysql/infoschema.go
nevinheintze: This is now done. Resolving comment.

spanner/ddl/ast_test.go
Pradip97: ok, sounds good

PR #93, opened 2020-12-12
feature: web-api

Pradip97:
> This PR is for the new feature: Web APIs, changes included in this PR:
> *`web` - added a new package which includes all APIs and tests related to it.
> *`conversion` - added a new package to duplicate some functions from `main.go` which needs to be used in web apis. they are duplicated so that main flow doesn't get affected.  
> * `main.go` - added a new command line flag `-w` to start the web server. command to start the server : `go run main.go --web`
> * `report.go` - We need to make few of the functions and struct variables public as they need to be accessed from web apis.

> This PR is completed except README
> 
> 1. MySQL
>    
>      * [x]  Handle mysql direct connection and mysqldump.
>      * [x]  Tests for mysql APIs

> 2. PostgreSQL
>        
>      * [x]  Handle postgres direct connection and pg_dump.
>      * [x]  Tests for postgres APIs.
> 3. Interleave tabels
>
>       * [x]  Support for interleave tables 
>       * [x]  Tests for interleave tables 
> * [x]  Update README

PR #97, opened 2020-12-25
feature: front end

13shreya:
This PR is for front end code review, new files/folder included in this PR are: 

> -  `frontend/index.html` - index.html is the starting point to render UI and it includes all js, css files and third party plugins
> 
> -  `frontend/js/app.js` - app.js handles overall routing between different components based on load or hash change events
> - `frontend/js/utility.js` - utility.js includes all the common functions and global variables which are used in other js files
> - `frontend/js/main.js` - main.js returns the html code for main screen and consists of all event handler functions required for main screen
> - `frontend/js/schema-conversion.js` - schema-conversion.js returns the html code for schema-conversion screen with all event handler functions required for schema-conversion screen
> - `frontend/js/instructions.js` - instructions.js returns the html code for instructions screen
> - `frontend/css/style.css and frontend/css/report.css` - contains css styles for complete project

Note: This code will not run because [feature-webapi](https://github.com/cloudspannerecosystem/harbourbridge/pull/93) is not merged to the master branch


13shreya:
@googlebot I signed it!

frontend/css/report.css
```diff
@@ -0,0 +1,1049 @@
+.report-header{
```
jereneyz, line 1: Was this file originally written using css? Instead of sass / less?

There are a lot of comments in this css file. Please clean it up and leave only the useful ones.

jereneyz: For all color codes, either go all caps or all small.

frontend/index.html
```diff
@@ -0,0 +1,64 @@
...
+</head>
+
+<body>
+  
+  <header class="main-header">
+    <nav class="navbar navbar-static-top">
+      <img src="Icons/Icons/google-spanner-logo.png" class="logo">
+    </nav>
+    <nav class="navbar navbar-static-top">
+      <div class="header-topic"><a name='headerMenu' href='#/' id='homeScreen' class='inactive'>Home</a></div>
```
jereneyz, line 30: Let's have this as a separate line. Not good to inline multiple elements.

frontend/js/app.js
```diff
@@ -0,0 +1,54 @@
...
+  { path: '/', component: HomeComponent, },
+  { path: '/schema-report-connect-to-db', component: SchemaComponent, },
+  { path: '/schema-report-load-db-dump', component: SchemaComponent, },
+  { path: '/schema-report-import-db', component: SchemaComponent, },
+  { path: '/schema-report-resume-session', component: SchemaComponent, },
+  { path: '/instructions', component: InstructionsComponent, }
+];
+
+// function to fetch browser url
+const parseLocation = () => location.hash.slice(1).toLowerCase() || '/';
```
jereneyz, line 38: Recommend bracketing inline function definitions i.e. (() => location... ;)

frontend/js/app.js
```diff
@@ -0,0 +1,54 @@
...
+
+// function to find component based on browser url
+const findComponentByPath = (path, routes) => routes.find(r => r.path.match(new RegExp(`^\\${path}$`, 'gm'))) || undefined;
```
jereneyz, line 41: This is quite difficult to read. Recommend breaking it up into a more verbose function
const findComponentByPath = (path, routes) => {
  routes.find(r => {
    r.path.match(new RegExp(...));
  })
} || undefined;

frontend/js/app.js
```diff
@@ -0,0 +1,54 @@
...
+        <h1>Error</h1>
+      </section>
+    `;
+  }
+}
+
+// Pre defined routes 
+const routes = [
+  { path: '/', component: HomeComponent, },
+  { path: '/schema-report-connect-to-db', component: SchemaComponent, },
```
jereneyz, line 30: These cannot be strings. These need to be variables because you are checking against them later. You can use an enum or const struct(for js) to represent these. For example 
`const paths = {
  connectToDb: "/schema-report-connect-to-db",
  ...
}`
The string "schema-report-connect-to-db" can only appear once in the entire codebase. This is to allow for changes to be made easily to the URL in the future. For example, let's say we don't want to use the term "session" anymore we just have to change this enum and that's it.

frontend/js/main.js
```diff
@@ -0,0 +1,1910 @@
...
+    sourceTableFlag = 'Postgres';
+    localStorage.setItem('sourceDbName', sourceTableFlag);
+  }
+}
+
+/**
+ * Function to get paths and events generated from window
+ *
+ * @param {object} params object containing path and event as keys
+ * @return {null}
```
jereneyz, line 1632: Uhhh this returns a boolean right?

frontend/js/app.js
```diff
@@ -0,0 +1,54 @@
...
+
+// function to find component based on browser url
+const findComponentByPath = (path, routes) => routes.find(r => r.path.match(new RegExp(`^\\${path}$`, 'gm'))) || undefined;
+
+// function to render the html based on path
+const router = () => {
+  const path = parseLocation();
+  const { component = ErrorComponent } = findComponentByPath(path, routes) || {};
+  getComponentFlag = getComponent({'path': path, 'event': window.event.type});
+  if (!getComponentFlag) {
```
jereneyz, line 48: This function isn't clean. getComponent does rendering for some paths, but returns a boolean for other (error) paths and leaves this function to render? It doesn't really make sense. getComponent should do the rendering for all paths, error or otherwise. Perhaps rename it to renderComponent since getComponent is supposed to be an accessor, but it's currently mutating the dom.

frontend/js/instructions.js
```diff
@@ -0,0 +1,54 @@
+const renderInstructionsHtml = (params) => {
+  setActiveSelectedMenu('instructions');
+  return (
+    `
```
jereneyz, line 4: In general, it's a very very bad idea to have HTML components embed inside a js file because it is using a completely different engine to render this at runtime. You are much more susceptible to bugs. There are some ways to avoid this:
1. Separate .html file that you can read and embed into the main dom using javascript
2. HTML component inside the original html file with display: none. Use javascript to toggle css visibility. (I recommend doing this)

frontend/js/lib/vanillaSelectBox.js
```diff
@@ -0,0 +1,675 @@
+/*
```
jereneyz, line 1: For 3rd party files, it's best to put them all in one folder called third_party and be sure to add any necessary licenses.

frontend/js/main.js
```diff
@@ -0,0 +1,1910 @@
+/**
+ * Function to initiate home screen tasks like validating form input fields
+ *
+ * @return {null}
+ */
+const initHomeScreenTasks = () => {
+  jQuery(document).ready(function () {
+    setActiveSelectedMenu('homeScreen');
+    jQuery('#loadDbForm > div > input').keyup(function () {
```
jereneyz, line 9: I need a working version of this to test this, but I'm not sure if this works for the paste case (right click -> paste). I don't think that triggers the keyup function. Not too sure. If it doesn't trigger the keyup case, then the button will still be disabled even when there are contents pasted.

frontend/js/main.js
```diff
@@ -0,0 +1,1910 @@
...
+ *
+ * @return {null}
+ */
+const uploadFileHandler = (e) => {
+  e.preventDefault();
+  jQuery("#upload:hidden").trigger('click');
+}
+
+/**
+ * Function to update selected file name and to read the json content of selected file while file uploading while file uploading
```
jereneyz, line 50: Typo

frontend/css/report.css
13shreya: Yes this file was originally written using css
- unnecessary comments are removed now
- all color codes are now changed to caps
- now file includes only #rgba codes

frontend/index.html
13shreya: Agreed.. changes are done accordingly and code is pushed

frontend/js/app.js
13shreya: Agreed.. this function is now written with improved readability

frontend/js/app.js
13shreya: `const paths` is now declared in app.js file which contains variable names corresponding to each possible path. And in all other places strings are replaced with these variable names

frontend/js/main.js
13shreya: Yes, it is now modified to boolean return type

frontend/js/lib/vanillaSelectBox.js
13shreya: a new `third_party folder` is created which contains all third_party files. Paths in index.html file are modified accordingly

frontend/js/main.js
13shreya: Currently there are few dependencies from backend code to merge in master branch therefore working version is not available to test this.
- Checked paste case scenario (right click -> paste) and it is triggering the keyup function

frontend/js/main.js
13shreya: Comment is modified now

jereneyz (changes requested):
Schema conversion screen:
1. For red boxes, it is perhaps better UX to by default open the summary box. I was looking at the red box and I could not figure out why it was red, until I clicked on the View Summary button
2. I was clicking around a lot and I cannot figure out why I wasn't able to change anything even though the cursor was a pointer. After much clicking around, I found that I was supposed to click on Edit Spanner Schema. A couple of possible fixes here: a) we make the button much more obvious. b) We had a huge text to the left of Edit Spanner Schema that says "Schema locked for editing. Unlock to change ->"
3.  When editing schema, originally the spanner constraints say 1 constraint selected. When I unselect and reselect, it says all. I think we need to be consistent here.
4. Styling issue. Please see screenshot 1 in chat
5. BUG: it says 5 entires when I only have 4. Please see screenshot 2 in chat
6. BUG: After I save the editted schema, everything disappears. Please see screenshot 3 in chat.
7. BUG:  I changed all of the spanner types to STRING. The schema file downloaded reflects the change. However, when I click on DDL statements, I still see the old schema. Please see screenshot 4 in chat.
8. BUG: I cannot change the number of characters in Cloud Spanner's STRING type. After a global change of all types to strings, I end up with this really weird schema file:
9. CREATE TABLE products (
    product_id STRING(20) NOT NULL, -- From: product_id varchar(20)
    description STRING(1000),       -- From: description varchar(1000)
    price STRING(MAX),              -- From: price decimal(10,2)
    date_added STRING(MAX)          -- From: date_added date
) PRIMARY KEY (product_id)
9. BUG: The summary report also does not reflect the changes that i made in the conversion report. I have set everything to string, but I see some weird synth_id column that I don't know where it comes from for the cart table, and for products table, it is complaining about float. See screenshot 5.
10. BUG: in the middle of a session, I  clicked on the instructions, and then I clicked on Schema Conversion to go back. But it does nothing. (screenshot 6)
11. It is non-obvious what the goal is of this UI. The user has no idea what to do next. We should rewrite the instructions. The google team will provide you with the new instructions. 
12. At the bottom of the schema conversion page, we should include instructions on what to do as well.

13shreya:
> Schema conversion screen:
> 
> 1. For red boxes, it is perhaps better UX to by default open the summary box. I was looking at the red box and I could not figure out why it was red, until I clicked on the View Summary button
> 2. I was clicking around a lot and I cannot figure out why I wasn't able to change anything even though the cursor was a pointer. After much clicking around, I found that I was supposed to click on Edit Spanner Schema. A couple of possible fixes here: a) we make the button much more obvious. b) We had a huge text to the left of Edit Spanner Schema that says "Schema locked for editing. Unlock to change ->"
> 3. When editing schema, originally the spanner constraints say 1 constraint selected. When I unselect and reselect, it says all. I think we need to be consistent here.
> 4. Styling issue. Please see screenshot 1 in chat
> 5. BUG: it says 5 entires when I only have 4. Please see screenshot 2 in chat
> 6. BUG: After I save the editted schema, everything disappears. Please see screenshot 3 in chat.
> 7. BUG:  I changed all of the spanner types to STRING. The schema file downloaded reflects the change. However, when I click on DDL statements, I still see the old schema. Please see screenshot 4 in chat.
> 8. BUG: I cannot change the number of characters in Cloud Spanner's STRING type. After a global change of all types to strings, I end up with this really weird schema file:
> 9. CREATE TABLE products (
>    product_id STRING(20) NOT NULL, -- From: product_id varchar(20)
>    description STRING(1000),       -- From: description varchar(1000)
>    price STRING(MAX),              -- From: price decimal(10,2)
>    date_added STRING(MAX)          -- From: date_added date
>    ) PRIMARY KEY (product_id)
> 10. BUG: The summary report also does not reflect the changes that i made in the conversion report. I have set everything to string, but I see some weird synth_id column that I don't know where it comes from for the cart table, and for products table, it is complaining about float. See screenshot 5.
> 11. BUG: in the middle of a session, I  clicked on the instructions, and then I clicked on Schema Conversion to go back. But it does nothing. (screenshot 6)
> 12. It is non-obvious what the goal is of this UI. The user has no idea what to do next. We should rewrite the instructions. The google team will provide you with the new instructions.
> 13. At the bottom of the schema conversion page, we should include instructions on what to do as well.



>I have fixed most of these issues and pushed the latest code. 
>For the bug mentioned in point 6:
>- if we uncheck any checkbox (on the leftmost side of each row), it indicates that column will be removed from spanner. Similarly if we uncheck all the checkboxes and then click on save changes button , then resultant spanner schema will have 0 rows

>- Also in case if all the columns are removed from spanner, it will add a default column which is 'synth_id' because spanner requires a primary key for every table. This 'synth_id' will then appear in DDL statements tab as well

jereneyz:
Replying to the comments on Google Chat:
Re Screenshot 3: Yes. But how do I get it back? I accidentally selected none of the boxes, and clicked on save. But I don't know how to get the columns back if I change my mind.
Re Screenshot 4: I'm just using the one in examples cart.mysqldump

New issues:
14. BUG: How do I not export a table? I want to be able to skip a table
15. BUG: When I edit the Global Data Type, I don't see the schema updates for the first table, only the second table is updated in the DDL Statements panel.
16. BUG: After I do a global edit, When I open up the individual table to make a change, suddenly all the changes from the global edit is gone. Try using global to change all columns to String, then open up one of the tables and edit that. You will see that all columns go back.
17. BUG: The download schema file button on the Conversion Report is actually downloading the SQL Schema, not the json session. I also don't think we should call it the schema file. We should call it something else. (Please discuss with Nevin and Asheesh)

agasheesh:
Most of the comments here are addressed or made obsolete by the redesign of UI code and thus closing this PR in favor of https://github.com/cloudspannerecosystem/harbourbridge/pull/130.

PR #98, opened 2020-12-28
#87: feat: support NUMERIC

### LINKED ISSUE #87: Update postgres and mysql with NUMERIC
In spanner/v1.10.0, we have supported NUMERIC type. So we would like to update postgres and mysql with NUMERIC type.

nevinheintze:
Fixes #87. Add support for Spanner NUMERIC to schema and data conversion.

Before this PR, HarbourBridge mapped PostgreSQL and MySQL NUMERIC/DECIMAL types to Spanner's FLOAT64 because Spanner did not originally support NUMERIC. However, Spanner has started supporting NUMERIC type from v1.10.0 (around Sept/2020). We've been planning to support it in HarbourBridge since then.

We'd be holding off on doing so while all of the work for user-guided schema migration was ongoing (aka webui work). We've now reached a point where most of the potentially conflicting changes from that work have landed. So it's time to push ahead with NUMERIC support.


mysql/toddl_test.go
```diff
@@ -36,7 +36,7 @@ func TestToSpannerType(t *testing.T) {
 			"b": schema.Column{Name: "b", Type: schema.Type{Name: "float"}},
 			"c": schema.Column{Name: "c", Type: schema.Type{Name: "tinyint", Mods: []int64{1}}},
 			"d": schema.Column{Name: "d", Type: schema.Type{Name: "varchar", Mods: []int64{6}}},
-			"e": schema.Column{Name: "e", Type: schema.Type{Name: "double"}},
+			"e": schema.Column{Name: "e", Type: schema.Type{Name: "numeric"}},
```
nevinheintze, line 39: This test didn't originally test numeric (it was based on postgres/toddl_test.go, but changed the numeric case to double).
Changing it back to numeric to test generation of Spanner NUMERIC.

mysql/mysqldump_test.go
```diff
@@ -90,6 +91,8 @@ func TestProcessMySQLDump_SingleCol(t *testing.T) {
 
 func TestProcessMySQLDump_MultiCol(t *testing.T) {
 	// Next test more general cases: multi-column schemas and data conversion.
+	numericData := new(big.Rat)
+	numericData.SetString("5.44444")
```
agasheesh, line 95: nit; if we don't expect an error here (and don't need to catch the same), then it may make sense to use `numericData.SetString("5.44444")` directly where it's used since the return type is big.Rat itself

postgres/data.go
```diff
@@ -177,6 +180,15 @@ func convInt64(val string) (int64, error) {
...
 }
 
+func convNumeric(val string) (*big.Rat, error) {
+	r := new(big.Rat)
+	_, ok := r.SetString(val)
+	if !ok {
+		return r, fmt.Errorf("can't convert to big.Rat")
+	}
+	return r, nil
+}
```
agasheesh, line 190: Looking forward to get rid of this kind of duplication soon after the pending PRs are merged :) - same comments apply though here as the one in mysql/data.go

agasheesh (approved):
LGTM - left minor nits.

mysql/data.go
```diff
@@ -161,6 +164,15 @@ func convInt64(val string) (int64, error) {
 	return i, err
 }
 
+func convNumeric(val string) (*big.Rat, error) {
+	r := new(big.Rat)
+	_, ok := r.SetString(val)
+	if !ok {
+		return r, fmt.Errorf("can't convert to big.Rat")
```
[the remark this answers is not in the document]
nevinheintze: Done.

mysql/data.go
```diff
@@ -161,6 +164,15 @@ func convInt64(val string) (int64, error) {
 	return i, err
 }
 
+func convNumeric(val string) (*big.Rat, error) {
+	r := new(big.Rat)
+	_, ok := r.SetString(val)
+	if !ok {
```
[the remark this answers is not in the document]
nevinheintze: Done.

mysql/mysqldump_test.go
nevinheintze: Yes, but numericData.SetString returns two values, so can't use directly.

But you are right, this is kind of messy/ugly. Added mkRat helper function instead (and it checks for failed conversion and triggers test failure if it happens).

postgres/data.go
nevinheintze: Done.
Yes, agreed.

PR #101, opened 2020-12-28
fix: foreign key data dependency issue

### LINKED ISSUE #99: Foreign key reference is case sensitive in spanner
## Expected Behavior

The tool should be able to migrate foreign keys if referenced columns are not case sensitive, as mysql and postgres allows case insensitive refered column names.

## Actual Behavior

Spanner throws error on migrating foreign key with different case in refered column.

Fix needed in toddl to check once if the refered column actually exists, if not we can compare all columns with refered column with lower case and identify which column it is actually refering.

### LINKED ISSUE #100: Data dependency in foreign key
## Expected Behavior

The tool should be able to migrate data with depended values in foreign key.

## Actual Behavior

The spanner throws reference error on migrating data which is refering some other table without having refered row.
```
Error while writing 833 rows to Spanner: spanner: code = "FailedPrecondition", desc = "Foreign key constraint my_fk_83 is violated on table workorderrouting. Cannot find referenced values in location(LocationID)."
```

Here location is being filled after workorderrouting table, so its not able to find the value, as insert statements are executed in order defined in the dump file.

Solution: first transfer schema without foreign key, then migrate data, then update schema with ALTER TABLE statements.

Pradip97:
Fixes #100

This PR is fixing the above mentioned issue, and includes following changes:

(1)  internal/convert.go: created function GetFK , this function returns list of ALTER TABLE statements for foreign keys
(2) main.go: created function UpdateDDLForeignKeys, this function updates existing schema in spanner with ALTER TABLE statements.

### DIFF
--- a/internal/convert.go
+++ b/internal/convert.go
@@ -170,6 +170,7 @@ func (conv *Conv) SetDataMode() {
 
 // GetDDL Schema returns the Spanner schema that has been constructed so far.
 // Return DDL in alphabetical table order.
+// TODO: Move GetDDL function to ddl/ast.go.
 func (conv *Conv) GetDDL(c ddl.Config) []string {
 	var tables []string
 	for t := range conv.SpSchema {
--- a/internal/convert.go
+++ b/internal/convert.go
@@ -177,8 +177,10 @@ func (conv *Conv) GetDDL(c ddl.Config) []string {
 	}
 	sort.Strings(tables)
 	var ddl []string
-	for _, t := range tables {
-		ddl = append(ddl, conv.SpSchema[t].PrintCreateTable(c))
+	if c.Tables {
+		for _, t := range tables {
+			ddl = append(ddl, conv.SpSchema[t].PrintCreateTable(c))
+		}
 	}
 
 	// Append foreign key constraints to DDL.
@@ -192,29 +194,6 @@ func (conv *Conv) GetDDL(c ddl.Config) []string {
 			for _, fk := range conv.SpSchema[t].Fks {
 				ddl = append(ddl, fk.PrintForeignKeyAlterTable(c, t))
 			}
-		}
-	}
-	return ddl
-}
-
-// GetFK returns the list of ALTER TABLE statements for adding
-// foreign key constraints.
-func (conv *Conv) GetFK(c ddl.Config) []string {
-	var tables []string
-	for t := range conv.SpSchema {
-		tables = append(tables, t)
-	}
-	sort.Strings(tables)
-	var ddl []string
-	// Append foreign key constraints to DDL.
-	// We always use alter table statements for foreign key constraints.
-	// The alternative of putting foreign key constraints in-line as part of create
-	// table statements is tricky because of table order (need to define tables
-	// before they are referenced by foreign key constraints) and the possibility
-	// of circular foreign keys definitions. We opt for simplicity.
-	for _, t := range tables {
-		for _, fk := range conv.SpSchema[t].Fks {
-			ddl = append(ddl, fk.PrintForeignKeyAlterTable(c, t))
 		}
 	}
 	return ddl
--- a/internal/convert.go
+++ b/internal/convert.go
@@ -197,6 +197,29 @@ func (conv *Conv) GetDDL(c ddl.Config) []string {
 	return ddl
 }
 
+// GetFK returns the list of ALTER TABLE statements for adding
+// foreign key constraints.
+func (conv *Conv) GetFK(c ddl.Config) []string {
+	var tables []string
+	for t := range conv.SpSchema {
+		tables = append(tables, t)
+	}
+	sort.Strings(tables)
+	var ddl []string
+	// Append foreign key constraints to DDL.
+	// We always use alter table statements for foreign key constraints.
+	// The alternative of putting foreign key constraints in-line as part of create
+	// table statements is tricky because of table order (need to define tables
+	// before they are referenced by foreign key constraints) and the possibility
+	// of circular foreign keys definitions. We opt for simplicity.
+	for _, t := range tables {
+		for _, fk := range conv.SpSchema[t].Fks {
+			ddl = append(ddl, fk.PrintForeignKeyAlterTable(c, t))
+		}
+	}
+	return ddl
+}
+
 // WriteRow calls dataSink and updates row stats.
 func (conv *Conv) WriteRow(srcTable, spTable string, spCols []string, spVals []interface{}) {
 	if conv.dataSink == nil {
--- a/internal/convert_test.go
+++ b/internal/convert_test.go
@@ -15,7 +15,6 @@
 package internal
 
 import (
-	"strings"
 	"testing"
 
 	pg_query "github.com/lfittl/pg_query_go"
@@ -53,7 +52,7 @@ func TestGetDDL(t *testing.T) {
 			"b": ddl.ColumnDef{Name: "b", T: ddl.Type{Name: ddl.Int64}},
 		},
 		Pks: []ddl.IndexKey{ddl.IndexKey{Col: "a"}},
-		Fks: []ddl.Foreignkey{ddl.Foreignkey{Name: "fk1", Columns: []string{"b"}, ReferTable: "ref_table1", ReferColumns: []string{"ref_c"}}},
+		Fks: []ddl.Foreignkey{ddl.Foreignkey{Name: "fk1", Columns: []string{"b"}, ReferTable: "ref_table1", ReferColumns: []string{"ref_b"}}},
 	}
 	conv.SpSchema["table2"] = ddl.CreateTable{
 		Name:     "table2",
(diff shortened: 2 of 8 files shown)


agasheesh:
Can you please separate the PRs for 2 fixes such that each PR fixes a given issue. We are also missing test cases for case sensitivity changes.

internal/convert_test.go
```diff
@@ -74,6 +74,43 @@ func TestGetDDL(t *testing.T) {
...
+		Pks: []ddl.IndexKey{ddl.IndexKey{Col: "a"}},
+		Fks: []ddl.Foreignkey{ddl.Foreignkey{Name: "fk2", Columns: []string{"b", "c"}, ReferTable: "ref_table2", ReferColumns: []string{"ref_b", "ref_c"}}},
+	}
+	ddl := conv.GetFK(ddl.Config{ForeignKeys: true})
+	normalize := func(l []string) (nl []string) {
+		for _, s := range l {
+			nl = append(nl, strings.Join(strings.Fields(s), " "))
+		}
+		return nl
+	}
```
agasheesh, line 106: Instead of normalizing whitespaces in the test, should we just fix the PrintForeignKeyAlterTable function?

internal/convert_test.go
```diff
@@ -74,6 +74,43 @@ func TestGetDDL(t *testing.T) {
...
 
+func TestGetFK(t *testing.T) {
+	conv := MakeConv()
+	conv.SpSchema["table1"] = ddl.CreateTable{
+		Name:     "table1",
+		ColNames: []string{"a", "b"},
+		ColDefs: map[string]ddl.ColumnDef{
+			"a": ddl.ColumnDef{Name: "a", T: ddl.Type{Name: ddl.Int64}},
+			"b": ddl.ColumnDef{Name: "b", T: ddl.Type{Name: ddl.Int64}},
+		},
```
agasheesh, line 85: Consider dropping these extra fields from test struct since these are not relevant for the test at hand.

internal/convert.go
```diff
@@ -197,6 +197,29 @@ func (conv *Conv) GetDDL(c ddl.Config) []string {
...
+	// We always use alter table statements for foreign key constraints.
+	// The alternative of putting foreign key constraints in-line as part of create
+	// table statements is tricky because of table order (need to define tables
+	// before they are referenced by foreign key constraints) and the possibility
+	// of circular foreign keys definitions. We opt for simplicity.
+	for _, t := range tables {
+		for _, fk := range conv.SpSchema[t].Fks {
+			ddl = append(ddl, fk.PrintForeignKeyAlterTable(c, t))
+		}
+	}
```
agasheesh, line 219: This piece of code is also duplicated in GetDDL, consider removing that duplication from there.

main.go
```diff
@@ -576,6 +582,38 @@ func createDatabase(project, instance, dbName string, conv *internal.Conv, out *
 	return fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName), nil
 }
 
+func UpdateDDLForeignKeys(project, instance, dbName string, conv *internal.Conv, out *os.File) error {
```
agasheesh, line 585: I am wondering why is this function exported?

agasheesh (changes requested):
Looks mostly good to me. Only minor comments regarding code style.

internal/convert.go
nevinheintze: Yes, agreed, we should avoid this duplication.

I think the simplest way is to follow what we did when adding support for foreign keys: we added a ForeignKeys field to ddl.Config control printing of foreign keys. Let's add a Table field as well, and use it to control table.s

Side discussion: the existing ForeignKey field in ddl.Config kind of breaks the ddl.go abstraction because we define a field in ddl.Config that isn't actually used/referenced in ddl.go. I think we need to do a bit of refactoring of ddl.go (in a separate PR) to resolve this. The simplest option is probably to move the GetDDL function from convert.go to ddl.go i.e. consolidate all DDL functionality in ddl.go. Asheesh -- want do you think? In any case, let's not worry about this for the moment, and we should go ahead and add a Table field to ddl.Config and resolve this PR.

internal/convert_test.go
nevinheintze: Or did you intend to add a test with cyclic  foreign key dependencies so that table1 refers to table2 and vice versa?

I'm happy either way, but agree with Asheesh to trim the parts of the test that aren't relevant to what's being tested.

internal/convert_test.go
nevinheintze: Yes, agreed.
The issue is that we add an extra space between ADD and FOREIGN KEY.
e.g. we print
```
ALTER TABLE table1 ADD CONSTRAINT fk1  FOREIGN KEY ...
```

main.go
```diff
@@ -576,6 +582,38 @@ func createDatabase(project, instance, dbName string, conv *internal.Conv, out *
...
+	fmt.Fprintf(out, "Updating database %s in instance %s with default permissions ...\n", dbName, instance)
```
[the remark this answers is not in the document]
nevinheintze: Agreed. And probably delete the "default permissions" -- it isn't relevant here.

How about:
fmt.Fprintf(out, "Updating schema of database %s in instance %s with foreign key constraints ...\n", dbName, instance)

main.go
```diff
@@ -560,7 +566,7 @@ func createDatabase(project, instance, dbName string, conv *internal.Conv, out *
 	// Spanner DDL doesn't accept them), and protects table and col names
 	// using backticks (to avoid any issues with Spanner reserved words).
 	// We also exclude foreign keys from the schema sent to Spanner.
-	schema := conv.GetDDL(ddl.Config{Comments: false, ProtectIds: true, ForeignKeys: true})
+	schema := conv.GetDDL(ddl.Config{Comments: false, ProtectIds: true, ForeignKeys: false})
```
nevinheintze, line 569: I think this is worth a comment. How about:

// The database is created without foreign key constraints. We will add them later once
// data has been written to the database to avoid foreign key data dependencies during
// data writes.

main.go
```diff
@@ -576,6 +582,38 @@ func createDatabase(project, instance, dbName string, conv *internal.Conv, out *
...
+		op, err := adminClient.UpdateDatabaseDdl(ctx, &adminpb.UpdateDatabaseDdlRequest{
+			Database:   fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName),
+			Statements: []string{fkStmt},
+		})
+		if err != nil {
+			fmt.Printf("Can't add foreign key with statement %s: %s\n", fkStmt, err)
+			conv.Unexpected(fmt.Sprintf("Can't add foreign key with statement %s: %s", fkStmt, err))
+			continue
+		}
+		if err := op.Wait(ctx); err != nil {
```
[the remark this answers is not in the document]
nevinheintze: Agreed with Asheesh: we probably need to do more here to communicate what is happening.

I think there are two separate issues:
a) the UpdateDatabaseDdl operations themselves will take some time.
b) these operations will trigger long-running schema operations e.g. for backfill of referencing and referenced indexes.

I just did some testing for (a): it looks like each UpdateDatabaseDdl operation can take from 15 seconds to 100 seconds each (on average about 25 seconds).
This was for empty tables, where each UpdateDatabaseDdl operation was very simple (a single constraint with one col), and essentially identical (just for a different tables).

We already have a progress.go module for reporting progress which we use elsewhere e.g. for reporting progress on loading data. Let's just use that: its internal/progress.go.  Here's how we could use it:

```
func UpdateDDLForeignKeys(project, instance, dbName string, conv *internal.Conv, out *os.File) error {
	ctx := context.Background()
	adminClient, err := database.NewDatabaseAdminClient(ctx)
	if err != nil {
		return fmt.Errorf("can't create admin client: %w\n", analyzeError(err, project, instance))
	}
	defer adminClient.Close()
	// The schema we send to Spanner excludes comments (since Cloud
	// Spanner DDL doesn't accept them), and protects table and col names
	// using backticks (to avoid any issues with Spanner reserved words).
	fkStmts := conv.GetFK(ddl.Config{Comments: false, ProtectIds: true, ForeignKeys: true})
	msg := fmt.Sprintf("Updating schema of database %s in instance %s with foreign key constraints ...", dbName, instance)
	p := internal.NewProgress(int64(len(fkStmts)), msg, internal.Verbose())
	for i, fkStmt := range fkStmts {
		op, err := adminClient.UpdateDatabaseDdl(ctx, &adminpb.UpdateDatabaseDdlRequest{
			Database:   fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName),
			Statements: []string{fkStmt},
		})
		if err != nil {
			fmt.Printf("Can't add foreign key with statement %s: %s\n", fkStmt, err)
			conv.Unexpected(fmt.Sprintf("Can't add foreign key with statement %s: %s", fkStmt, err))
			continue
		}
		if err := op.Wait(ctx); err != nil {
			fmt.Printf("Can't add foreign key with statement %s: %s\n", fkStmt, err)
			conv.Unexpected(fmt.Sprintf("Can't add foreign key with statement %s: %s", fkStmt, err))
		}
		internal.VerbosePrintln("Updated schema with statement: " + fkStmt)
		p.MaybeReport(int64(i + 1))
	}
	p.Done()
	return nil
}
```



// TODO:

main.go
```diff
@@ -560,7 +566,7 @@ func createDatabase(project, instance, dbName string, conv *internal.Conv, out *
 	// Spanner DDL doesn't accept them), and protects table and col names
 	// using backticks (to avoid any issues with Spanner reserved words).
 	// We also exclude foreign keys from the schema sent to Spanner.
```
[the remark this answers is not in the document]
Pradip97: added comment for more info

main.go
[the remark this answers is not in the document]
Pradip97: updated print statement

main.go
```diff
@@ -220,6 +220,12 @@ func toSpanner(driver, projectID, instanceID, dbName string, ioHelper *ioStreams
 		fmt.Printf("\nCan't finish data conversion for db %s: %v\n", db, err)
 		return fmt.Errorf("can't finish data conversion")
 	}
+
+	err = UpdateDDLForeignKeys(projectID, instanceID, dbName, conv, ioHelper.out)
+	if err != nil {
```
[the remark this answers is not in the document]
Pradip97: sure updated function call

main.go
```diff
@@ -220,6 +220,12 @@ func toSpanner(driver, projectID, instanceID, dbName string, ioHelper *ioStreams
...
+		fmt.Printf("\nCan't perform update operation on db %s: %v\n", db, err)
+		return fmt.Errorf("can't perform update database")
```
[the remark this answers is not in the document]
Pradip97: updated error message with relevant info

main.go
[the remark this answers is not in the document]
Pradip97: Sure nevin, makes sense to show user what is happening, updated code as mentioned.

internal/convert.go
```diff
@@ -197,6 +197,29 @@ func (conv *Conv) GetDDL(c ddl.Config) []string {
 	return ddl
 }
 
+// GetFK returns the list of ALTER TABLE statements for adding
+// foreign key constraints.
+func (conv *Conv) GetFK(c ddl.Config) []string {
```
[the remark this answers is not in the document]
Pradip97: not relevant, as now we are using Tables flag

internal/convert_test.go
Pradip97: updated ALTER TABLE statement

internal/convert_test.go
Pradip97: merged foreign key test with GetDDL test as now we are using Tables flag

internal/convert.go
Pradip97: to avoid duplication, added Tables flag in Config now.

main.go
Pradip97: this function will be moved to conversion package once merged with web api branch, but for now it makes to be consistent with current code so i updated it to lower case function.

main.go
Pradip97: updated comment as mentioned

internal/convert_test.go
agasheesh: I see that we are still using normalize function for ASSERT statements. We should fix the PrintCreateTable and/or the unit test to remove the need for the same.

internal/convert.go
agasheesh: Resolving this for now temporarily by introducing Tables field seems reasonable for me. Yes, GetDDL should be a function on ddl itself. Pradip, can you please add a TODO where Tables field is defined in ddl.Config to clean this up such that this doesn't fall through the cracks.

The way that I imagined this to work though initially was to call GetFkDDL from GetDDL if ForeignKey field was true in ddl.Config.

internal/convert_test.go
```diff
@@ -50,28 +50,47 @@ func TestGetDDL(t *testing.T) {
 		ColNames: []string{"a", "b"},
 		ColDefs: map[string]ddl.ColumnDef{
 			"a": ddl.ColumnDef{Name: "a", T: ddl.Type{Name: ddl.Int64}},
-			"b": ddl.ColumnDef{Name: "b", T: ddl.Type{Name: ddl.Float64}},
+			"b": ddl.ColumnDef{Name: "b", T: ddl.Type{Name: ddl.Int64}},
 		},
-		Pks: []ddl.IndexKey{ddl.IndexKey{Col: "a"}}}
+		Pks: []ddl.IndexKey{ddl.IndexKey{Col: "a"}},
+		Fks: []ddl.Foreignkey{ddl.Foreignkey{Name: "fk1", Columns: []string{"b"}, ReferTable: "ref_table1", ReferColumns: []string{"ref_c"}}},
```
agasheesh, line 56: just a very minor nit, why do we refer to ref_c and not ref_b here for the sake of consistency?

main.go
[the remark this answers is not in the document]
agasheesh: Hi Pradip, can you please add the TODO as Nevin mentioned in his comment. This is definitely something we would look to fix in future.

For reference, the TODO mentioned above was:

// TODO: Improve performance of the foreign key constraint updates.
// For example, issue all of the update ops first before waiting for them to complete
// so that that can execute in parallel. We could also print out ids of the
// long-running operations.

agasheesh (changes requested):

internal/convert.go
Pradip97: I'll add TODO on GetDDL function to move it to ddl package in future.

internal/convert_test.go
Pradip97: renamed to ref_b

internal/convert_test.go
Pradip97: removed normalize function and updated expected statements in the tests

main.go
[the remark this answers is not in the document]
agasheesh: I see that the TODO was added, resolving the comment.

agasheesh (approved):

agasheesh:
LGTM, have approved the changes from my end. But please wait for a final approval from Nevin before submitting.

nevinheintze (approved):
Many thanks for all of the work on these changes!
Looks good.
Please go ahead and submit!

PR #102, opened 2020-12-29
Revert "#87: feat: support NUMERIC"

nevinheintze:
Reverts cloudspannerecosystem/harbourbridge#98

HarbourBridge is using an older version of the Cloud Spanner client library -- in particular, a version that doesn't support NUMERIC. This means that after #98, any source tables with NUMERIC will be converted to Spanner tables with NUMERIC, and then have all of their data rejected.

Reverting for the moment while I resolve the client version issue.


Pradip97 (approved):
Looks good to me, its a complete reverse of PR #87

PR #104, opened 2020-12-30
#87: feat: support NUMERIC (roll forward)

### LINKED ISSUE #87: Update postgres and mysql with NUMERIC

nevinheintze:
Fixes #87. Add support for Spanner NUMERIC to schema and data conversion (roll forward).

This was first submitted as PR98, but was rolled back because of client issues with big.Rat.

This PR consists of:
- a clean revert of cloudspannerecosystem/harbourbridge#102 (revert of #87)
- a fix for the client issue (numeric values now passed as a string via spanner.NumericString)
- integration test has been expanded to include numeric (it now includes all core types)
- updates github workflow to use recent version of emulator (to get numeric support)

Original PR description:
Fixes #87. Add support for Spanner NUMERIC to schema and data conversion.

Before this PR, HarbourBridge mapped PostgreSQL and MySQL NUMERIC/DECIMAL types to Spanner's FLOAT64 because Spanner did not originally support NUMERIC. However, Spanner has started supporting NUMERIC type from v1.10.0 (around Sept/2020). We've been planning to support it in HarbourBridge since then.

We'd be holding off on doing so while all of the work for user-guided schema migration was ongoing (aka webui work). We've now reached a point where most of the potentially conflicting changes from that work have landed. So it's time to push ahead with NUMERIC support.


.github/workflows/integration-tests-against-emulator.yaml
```diff
@@ -17,10 +17,13 @@ jobs:
 
     services:
       emulator:
-        image: gcr.io/cloud-spanner-emulator/emulator:0.8.0
+        image: gcr.io/cloud-spanner-emulator/emulator:1.1.1
         ports:
           - 9010:9010
           - 9020:9020
+        # needed because numeric support isn't on by default in emulator:1.1.1
+        # note: future versions will default enable this and deprecate the flag
```
agasheesh, line 25: nit; should we leave a TODO here instead of a node for the ease of finding this in future

agasheesh (approved):

.github/workflows/integration-tests-against-emulator.yaml
nevinheintze: DONE -- added TODO.

postgres/data.go
```diff
@@ -177,6 +180,18 @@ func convInt64(val string) (int64, error) {
 	return i, err
 }
 
+// convNumeric maps a source database string value (representing a numeric)
+// into a string representing a valid Spanner numeric.
+// Ideally we would just return a *big.Rat, but spanner.Mutation
```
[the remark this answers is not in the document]
nevinheintze: Done -- added TODO here and in mysql.

PR #114, opened 2021-03-09
Upgrade emulator version to 1.2.0 and remove the args from github act…

agasheesh:
Fixes the issue with integration tests not running due to following failure:

The workflow is not valid. .github/workflows/integration-tests-against-emulator.yaml (Line: 26, Col: 9): Unexpected value 'args'
.github/workflows/integration-tests-against-emulator.yaml#L26


nevinheintze (approved):
Thanks for tracking this down!

PR #115, opened 2021-03-09
Fix handling of numeric data for driver=postgres.

nevinheintze:
PR#104 (support NUMERIC) missed a data conversion case. Specifically, it did not cover the case when driver=postgres (aka infoschema processing with PostgreSQL).

This PR fixes the issue and adds a test case.

Notes:
1) The missed case was only for Postgres. It didn't occur for MySQL because for MySQL we use a simpler (but slightly more expensive) string-based-processing that allows us to directly call ProcessDataRow.
1) We have a long-term plan to consolidate the processing of postgres and mysql information schema processing, and in particular, to use the simpler string-based processing of mysql everywhere. This would have avoided this issue!

### DIFF
--- a/postgres/infoschema.go
+++ b/postgres/infoschema.go
@@ -581,6 +581,11 @@ func cvtSQLScalar(conv *internal.Conv, srcCd schema.Column, spCd ddl.ColumnDef, 
 		case string:
 			return convFloat64(v)
 		}
+	case ddl.Numeric:
+		switch v := val.(type) {
+		case []byte: // Note: PostgreSQL uses []byte for numeric.
+			return convNumeric(string(v))
+		}
 	case ddl.String:
 		switch v := val.(type) {
 		case bool:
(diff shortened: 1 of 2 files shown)


agasheesh (approved):
Thanks Nevin for digging into this and unblocking the integration tests. I was able to patch this PR into https://github.com/cloudspannerecosystem/harbourbridge/pull/114 and the integration tests did pass end-to-end.

Noted the feedback about the refactoring, I will try to get to the same soon given that PR#93 is landed now.

nevinheintze:
PR#104 (support NUMERIC) missed a data conversion case. Specifically, it did not cover the case when driver=postgres (aka infoschema processing with PostgreSQL).

This PR fixes the issue and adds a test case.

Notes:
1) The missed case was only for Postgres. It didn't occur for MySQL because for MySQL we use a simpler (but slightly more expensive) string-based-processing that allows us to directly call ProcessDataRow.
2) We have a long-term plan to consolidate the processing of postgres and mysql information schema processing, and in particular, to use the simpler string-based processing of mysql everywhere. This would have avoided this issue!

PR #120, opened 2021-03-17
Add .gitignore to ignore files intended to be untracked.

agasheesh:


nevinheintze (approved):
I was sure we already had a .gitignore file.
But we don't, and I checked the full history and we've never had one!
So much for my memory of these things ...

Thanks for adding!

PR #121, opened 2021-03-17
Add basic end to end integration tests for mysqldump and mysql direct connect driver

agasheesh:


agasheesh:
@nevinheintze - Added you as a reviewer to provide early feedback on the decision to add mysql container in the same .yaml file. 

I am working on adding integration_test.go under mysql/ which will contain end-to-end integration tests for mysql.

nevinheintze (commented):
Looks good -- thanks!
I'm not 100% confident in all of the options flags, but it seems reasonable.

agasheesh:
Hi @nevinheintze - this is ready for another pass. The tests are very basic in nature, though they cover both mysqldump and direct connect. Once this skeleton is in, we can think about what exact tests should we have - not just for mysql, but for postgresql and dynamodb as well.

main.go
```diff
@@ -26,21 +26,17 @@ import (
...
 	_ "github.com/lib/pq"
 
+	"github.com/cloudspannerecosystem/harbourbridge/cmd"
 	"github.com/cloudspannerecosystem/harbourbridge/conversion"
 	"github.com/cloudspannerecosystem/harbourbridge/internal"
 	"github.com/cloudspannerecosystem/harbourbridge/web"
 )
 
 var (
-	badDataFile      = "dropped.txt"
```
nevinheintze, line 35: I think splitting up flags across different files/modules can make it hard to track them down.

Would like to chat about options for keeping them together. Having them in main is a good place, but accessing them from other modules becomes tricky (have to pass them as args, which gets messy). How about having a separate flags module?

The other option which we'd briefly discussed is having separate entry points for different ways to run HB e.g. for commandline vs ui. Let's discuss this a bit more.

agasheesh: I agree that spreading flags across different modules is not a good idea. In fact, go recommendation is to have all command line flags being part of main package itself. For the purpose of this PR, using a separate flags submodule seems reasonable. Though, there are multiple ways of dealing with the issue here:

1. Collect related flags in structs such that passing around the flag values to cmd.CommandLine is less of an issue.

2. Define separate subcommands -- we need to think through all the current and future use cases when we jump on this.

Given these options, I agree that we should spend a little more time thinking about this before moving forward with this PR.

main.go
```diff
@@ -74,10 +69,12 @@ Sample usage:
 func main() {
 	flag.Usage = usage
 	flag.Parse()
+
 	if webapi {
```
nevinheintze, line 73: Not related to this PR, but what about adding a comment:

// Note: the web interface does not use any commandline flags.

main.go
nevinheintze: I like your option 1 i.e. have a struct with all of the stuff we collect or compute from flags and environment variables. And then pass that to CommandLine. I think that's the cleanest thing to do for the moment. 
But let's not do it now, and add to TODO for the moment (see comment below).

main.go
```diff
@@ -53,7 +49,6 @@ func init() {
 	flag.StringVar(&instanceOverride, "instance", "", "instance: Spanner instance to use")
 	flag.StringVar(&filePrefix, "prefix", "", "prefix: file prefix for generated files")
 	flag.StringVar(&driverName, "driver", "pg_dump", "driver name: flag for accessing source DB or dump files (accepted values are \"pg_dump\", \"postgres\", \"mysqldump\", and \"mysql\")")
-	flag.Int64Var(&schemaSampleSize, "schema-sample-size", int64(100000), "schema-sample-size: the number of rows to use for inferring schema (only for DynamoDB)")
```
nevinheintze, line 56: For the moment, suggest we leave schema-sample-size in this file, and just pass schemaSampleSize as a parameter to cmd.CommandLIne. Yeah, it adds an extra parameter, but there's already a lot of parameters in that function and at least its consistent and it also keeps all flags together.

I like your suggestion of collecting all of the config state in a struct. Basically grab all of the components we pass to CommandLine like: driverName, project, instance, dbName, dataOnly, schemaOnly, sessionJSON, filePrefix and schemaSampleSize. In other words all of the stuff we collect or compute from flags and environment variables.

But let's add a TODO for that and defer for a following PR.
Maybe we could even go a step further, and move code for computing these config values in main.go (currently 
GetProject, GetInstance and some related function are in conversion.go, however they aren't used by the webapi). That way we could have a really clean interface from main: main either calls web.WebAPI() or it calls cmd.CommandLine. That would be nice.

cmd/cmd.go
```diff
@@ -0,0 +1,98 @@
...
+func init() {
+	flag.Int64Var(&schemaSampleSize, "schema-sample-size", int64(100000), "schema-sample-size: the number of rows to use for inferring schema (only for DynamoDB)")
+}
+
+// CommandLine provides the core processing for HarbourBridge when run as a command-line tool.
+// It performs the following steps:
+// 1. Run schema conversion (if dataOnly is set to false)
+// 2. Create database (if schemaOnly is set to false)
+// 3. Run data conversion (if schemaOnly is set to false)
+// 4. Generate report (if schemaOnly is set to false)
```
nevinheintze, line 45: Suggest dropping "(if schemaOnly is set to false)" since we generate a report in either case (see the call to report at line 61).

main.go
```diff
@@ -86,8 +83,16 @@ func main() {
 	}
 	defer conversion.Close(lf)
 
+	if schemaOnly && dataOnly {
```
nevinheintze, line 86: Aside (and not suggesting any change in the PR): I find the dataOnly and schemaOnly booleans very confusing in the logic of the rest of this function! Double negatives like "if !schemaOnly" and "if !dataOnly" are hard to keep straight.

I think flag names are correct (and correspond to similar flags in mysqldump and pg_dump).
However, internally, I think we should translate them to internal booleans like "processSchema" and "processData". That would make the code a bit more self-evident.

testing/mysql/integration_test.go
```diff
@@ -0,0 +1,203 @@
...
+	now := time.Now()
+	dbName, _ := conversion.GetDatabaseName(conversion.MYSQLDUMP, now)
+	dbPath := fmt.Sprintf("projects/%s/instances/%s/databases/%s", projectID, instanceID, dbName)
+	dataFilepath := "../../test_data/mysqldump.test.out"
+	filePrefix := filepath.Join(tmpdir, dbName+".")
+	f, err := os.Open(dataFilepath)
+	if err != nil {
+		t.Fatalf("failed to open the test data file: %v", err)
+	}
+	err = cmd.CommandLine(conversion.MYSQLDUMP, projectID, instanceID, dbName /*dataOnly=*/, false /*schemaOnly=*/, false /*sessionJSON=*/, "", &conversion.IOStreams{In: f, Out: os.Stdout}, filePrefix, now)
```
nevinheintze, line 134: I like these comments, but shouldn't they come after the comma i.e ", /*dataOnly=*/ false"

testing/mysql/integration_test.go
```diff
@@ -0,0 +1,203 @@
...
+		t.Skip("Integration tests skipped")
+	}
+	tmpdir, err := ioutil.TempDir(".", "int-test-")
+	if err != nil {
+		log.Fatal(err)
+	}
+	return tmpdir
+}
+
+func TestIntegration_MYSQLDUMP_SimpleUse(t *testing.T) {
```
nevinheintze, line 119: More a question than anything -- what is the suggested function order for test files like this?
Would it make sense for the Test_XXX functions to be earlier in the file, since they are the key focus of the file. Maybe right after TestMain?

nevinheintze (approved):
Looks good -- thanks! Just a few minor comments.
And sorry again for the delay.

testing/mysql/integration_test.go
agasheesh: Seems to be an issue with gofmt that I am using in my local editor. What's more is if I do fix these argument comments e.g., `, dbName, /*dataOnly=*/false, ` etc then our lint integration starts complaining due to `goimports` library also expecting the wrong formatting. Have fixed these manually while I look for how to fix these.

main.go
agasheesh: I also had a hard time reading through the code due to double negatives. I think once we nail down on the submodule structure of commands and refactor towards the same, then many of these issues should go away since I expect separate modules for processSchema and processData then.

testing/mysql/integration_test.go
agasheesh: Seems to be an issue with auto formatting done by gofmt. What's more is that if I do fix these manually, then our lint workflow starts failing with goimports command also expecting this wrong output. Will investigate this more, and fix the lint workflow.

testing/mysql/integration_test.go
agasheesh: I could not find a recommendation specific for test files. However, the general pattern I have seen is following:

- variables, constants and flags
- init()
- initialization/ test setup functions()
- TEST_XXX code
- other helper functions

In this case, the helper functions we have above TEST_XXX functions are setup/teardown related. Other helper functions e.g., checkResults, checkBigint are below.

main.go
agasheesh: Done @https://github.com/cloudspannerecosystem/harbourbridge/pull/121/commits/96fafc8df69686a5aeac12ee5bec2c5656666d1e.

main.go
agasheesh: Done @https://github.com/cloudspannerecosystem/harbourbridge/pull/121/commits/96fafc8df69686a5aeac12ee5bec2c5656666d1e.

main.go
agasheesh: I agree that we need to take a hard look at the content of conversion.go since right now it has become a dump all module which contains many unrelated functions. I have added a TODO for deferring the same for a future PR. Will also look into moving GetProject, GetInstance and related functions in that PR. Done @https://github.com/cloudspannerecosystem/harbourbridge/pull/121/commits/96fafc8df69686a5aeac12ee5bec2c5656666d1e.

cmd/cmd.go
agasheesh: Good catch! Done @ https://github.com/cloudspannerecosystem/harbourbridge/pull/121/commits/96fafc8df69686a5aeac12ee5bec2c5656666d1e.

testing/mysql/integration_test.go
agasheesh: Done @https://github.com/cloudspannerecosystem/harbourbridge/pull/121/commits/96fafc8df69686a5aeac12ee5bec2c5656666d1e

PR #138, opened 2021-05-26
#136: fix: Handle non-boolean values for boolean columns.

### LINKED ISSUE #136: Handle data conversion of non-boolean values for MySQL TINYINT(1)/BOOLEAN
NOTE: TINYINT(1) is a synonym for BOOL and BOOLEAN.

MySQL does not require values for BOOLEAN to actually be boolean values.
Instead, it allows any legal value for tinyint to be stored in a BOOLEAN i.e. -128 to 127.

At the moment, HarbourBridge generates an error and rejects rows where boolean columns have non-boolean values.

Proposed resolution: instead of generating an error, convert integer values to booleans using the standard heuristic: 0 is false and non-zero is true.

nevinheintze:
This PR changes the behaviour during data conversion for boolean columns with non-boolean values: instead of generating an error, we now try to convert an int to a bool.

Old behaviour: HarbourBridge generates an error and rejects rows when boolean columns have non-boolean values.

New behaviour: We try to parse the value as an int64. If that succeeds, then we convert the int to a bool (0 maps to false, everything else maps to true), and we log an "unexpected" event detailing the conversion. If the conversion to int64 fails, the behavior is the same as current behaviour (we log an error and drop the row).

Closes #136

### DIFF
--- a/mysql/data.go
+++ b/mysql/data.go
@@ -129,12 +129,14 @@ func convScalar(conv *internal.Conv, spannerType ddl.Type, srcTypeName string, T
 func convBool(conv *internal.Conv, val string) (bool, error) {
 	b, err := strconv.ParseBool(val)
 	if err != nil {
-		// MySQL allows non-boolean values in boolean columns.
-		// If we can't parse as boolean, try to re-parse as INT64
-		// and treat as true if value is non-zero.
-		// Note: if ParseBool fails, then val is probably a non-zero number.
+		// MySQL uses TINYINT(1) to implement BOOL/BOOLEAN, and does not
+		// enforce/validate boolean values i.e. any value that can be stored
+		// in a TINYINT (-128 to 127) can be stored in BOOL/BOOLEAN.
+		// If ParseBool(val) fails, this is very likely the cause.
+		// To handle this, re-parse as INT64 and treat as true if value is non-zero.
+		// Note: if ParseBool(val) fails, then val is probably a non-zero number.
 		i, err2 := convInt64(val)
-		if err2 == nil {
+		if err2 == nil && i >= -128 && i <= 127 {
 			b = i != 0
 			conv.Unexpected(fmt.Sprintf("Expected boolean value, but found integer value %v; mapping it to %v\n", val, b))
 			return b, err2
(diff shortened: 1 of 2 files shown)


mysql/data.go
```diff
@@ -126,9 +126,19 @@ func convScalar(spannerType ddl.Type, srcTypeName string, TimezoneOffset string,
...
+func convBool(conv *internal.Conv, val string) (bool, error) {
 	b, err := strconv.ParseBool(val)
 	if err != nil {
+		// MySQL allows non-boolean values in boolean columns.
+		// If we can't parse as boolean, try to re-parse as INT64
+		// and treat as true if value is non-zero.
+		// Note: if ParseBool fails, then val is probably a non-zero number.
+		i, err2 := convInt64(val)
+		if err2 == nil {
+			b = i != 0
```
agasheesh, line 138: consider adding check that this value is in range -127 to 127 since we don't want to convert to bool otherwise

mysql/data.go
```diff
@@ -126,9 +126,19 @@ func convScalar(spannerType ddl.Type, srcTypeName string, TimezoneOffset string,
...
 
-func convBool(val string) (bool, error) {
+func convBool(conv *internal.Conv, val string) (bool, error) {
 	b, err := strconv.ParseBool(val)
 	if err != nil {
+		// MySQL allows non-boolean values in boolean columns.
+		// If we can't parse as boolean, try to re-parse as INT64
+		// and treat as true if value is non-zero.
+		// Note: if ParseBool fails, then val is probably a non-zero number.
+		i, err2 := convInt64(val)
```
agasheesh, line 136: we should consider adding a test for convBool since the behavior is slightly non-trivial now

agasheesh (approved):

mysql/data.go
nevinheintze: Done.
Just to clarify: the allowed range for tinyint(1) is -128 to 127.

mysql/data.go
nevinheintze: Done.
Added tests for this change and also for other bool case we didn't test before.
Also added a negative test e.g. that 128 fails with an error.

nevinheintze:
Since there is no urgency, I'm going to hold off submitting till Monday.
Please take another look if you have time.

agasheesh (approved):
Thanks for all the changes. LGTM.

PR #141, opened 2021-05-27
Handle negative expressions e.g., negative int64, float64 column values in mysqldump parser

agasheesh:
Pingcap parser that we use to parse mysqldump returns an AST node with type ValueExpr for positive int64 and float64 values. Whereas, for negative values it returns an UnaryExpression which contains the unary operator and the corresponding operand value. We are adding handling for for UnaryExpressions with Minus operator e.g., "-123" or "-123.456" such that the same can be converted to valid int64 and float64 column types in the intermediate representation. 

Before this change, mysqldump parser would simply drop these rows altogether.

### DIFF
--- a/mysql/mysqldump.go
+++ b/mysql/mysqldump.go
@@ -757,7 +757,7 @@ func getNegativeUnaryVals(valExpr *driver.ValueExpr) (string, error) {
 		if err != nil {
 			return "", fmt.Errorf("unexpected UnaryOperationExpr with value %v", val)
 		}
-		return fmt.Sprintf("-%v", floatVal), nil
+		return fmt.Sprintf("%v", -1*floatVal), nil
 	default:
 		return "", fmt.Errorf("unexpected UnaryOperationExpr value with type %T", val)
 	}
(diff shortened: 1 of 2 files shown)


mysql/mysqldump.go
```diff
@@ -726,13 +728,41 @@ func getVals(row []ast.ExprNode) ([]string, error) {
...
+func getNegativeUnaryVals(valExpr *driver.ValueExpr) (string, error) {
+	switch val := valExpr.GetValue().(type) {
+	case int64:
+		return fmt.Sprintf("%v", -1 * val), nil
+	case *types.MyDecimal:
+		floatVal, err := val.ToFloat64()
+		if err != nil {
+			return "", fmt.Errorf("unexpected UnaryOperationExpr with value %v", val)
+		}
+		return fmt.Sprintf("-%v", floatVal), nil
```
nevinheintze, line 760: For consistency, maybe:

return fmt.Sprintf("%v", -1 * floatVal), nil

nevinheintze (approved):
Looks good -- many thanks!
Just one minor comment.

mysql/mysqldump.go
agasheesh: Thanks for noticing, this was a leftover from an alternative approach with convering types.MyDecimal to string directly. Agreed that "-1*val" is consistent and better. Done @https://github.com/cloudspannerecosystem/harbourbridge/pull/141/commits/d454c1a0dcc2e8bd42c8386b3802dd6a41c9a86c.

PR #143, opened 2021-06-07
Generate legal spanner DDL and dump it along with other downloadables

### LINKED ISSUE #123: Spanner DDL generated by HB from Postgres schema is not executable from the Google Cloud Console.
Issue#1 - Google Cloud's Cloud Spanner DDL editor cannot parse the comments generated by HB
Issue#2 - Spanner keywords are used as names, e.g. “default” as a column name. Such identifiers should be wrapped in backticks by HB.

Deep1998:
Fixes #123 - Spanner DDL generated by HB from Postgres schema is not executable from the Google Cloud Console.

This change dumps the following files through both CLI and UI.

- Session json (mysqldump_2021-06-08_f491-17fc.session.json)
- Readable Schema (mysqldump_2021-06-08_f491-17fc.schema.txt) `(comments: true, protectIds: false)`
- Report Summary (mysqldump_2021-06-08_f491-17fc.report.txt)
- Legal DDL Schema (mysqldump_2021-06-08_f491-17fc.schema.ddl.txt) `(comments: false, protectIds: true)`

In the UI, we first write them during session creation inside the folder `frontend/harbour_bridge_downloads/<dbName>/` .
Subsequently as the user edits the schema and the session file gets updated, the 4 updated downloadables overwrite the earlier dumped files.

Note: The session file used by HB is same as the one written out during create_session i.e. only a single copy of session.json is maintained.

### DIFF
--- a/conversion/conversion.go
+++ b/conversion/conversion.go
@@ -499,6 +499,7 @@ func GetInstance(project string, out *os.File) (string, error) {
 		fmt.Fprintf(out, "Could not find any Spanner instances for project %s\n", project)
 		return "", fmt.Errorf("no Spanner instances for %s", project)
 	}
+
 	// Note: we could ask for user input to select/confirm which Spanner
 	// instance to use, but that interacts poorly with piping pg_dump/mysqldump data
 	// to the tool via stdin.
@@ -538,25 +539,25 @@ func getInstances(project string) ([]string, error) {
 
 // WriteSchemaFile writes DDL statements in a file. It includes CREATE TABLE
 // statements and ALTER TABLE statements to add foreign keys.
+// The parameter name should end with a .txt.
 func WriteSchemaFile(conv *internal.Conv, now time.Time, name string, out *os.File) {
 	f, err := os.Create(name)
 	if err != nil {
 		fmt.Fprintf(out, "Can't create schema file %s: %v\n", name, err)
 		return
 	}
-	// The schema file we write out includes comments, includes foreign keys
+
+	// The schema file we write out below is optimized for reading. It includes comments, foreign keys
 	// and doesn't add backticks around table and column names. This file is
 	// intended for explanatory and documentation purposes, and is not strictly
 	// legal Cloud Spanner DDL (Cloud Spanner doesn't currently support comments).
-	// Change 'Comments' to false and 'ProtectIds' to true to write out a
-	// schema file that is legal Cloud Spanner DDL.
-	ddl := conv.SpSchema.GetDDL(ddl.Config{Comments: true, ProtectIds: false, Tables: true, ForeignKeys: true})
-	if len(ddl) == 0 {
-		ddl = []string{"\n-- Schema is empty -- no tables found\n"}
+	spDDL := conv.SpSchema.GetDDL(ddl.Config{Comments: true, ProtectIds: false, Tables: true, ForeignKeys: true})
+	if len(spDDL) == 0 {
+		spDDL = []string{"\n-- Schema is empty -- no tables found\n"}
 	}
 	l := []string{
 		fmt.Sprintf("-- Schema generated %s\n", now.Format("2006-01-02 15:04:05")),
-		strings.Join(ddl, ";\n\n"),
+		strings.Join(spDDL, ";\n\n"),
 		"\n",
 	}
 	if _, err := f.WriteString(strings.Join(l, "")); err != nil {
@@ -564,6 +565,32 @@ func WriteSchemaFile(conv *internal.Conv, now time.Time, name string, out *os.Fi
 		return
 	}
 	fmt.Fprintf(out, "Wrote schema to file '%s'.\n", name)
+
+	// Convert <file_name>.<ext> to <file_name>.ddl.<ext>.
+	nameSplit := strings.Split(name, ".")
+	nameSplit = append(nameSplit[:len(nameSplit)-1], "ddl", nameSplit[len(nameSplit)-1])
+	name = strings.Join(nameSplit, ".")
+	f, err = os.Create(name)
+	if err != nil {
+		fmt.Fprintf(out, "Can't create legal schema ddl file %s: %v\n", name, err)
+		return
+	}
+
+	// We change 'Comments' to false and 'ProtectIds' to true below to write out a
+	// schema file that is a legal Cloud Spanner DDL.
+	spDDL = conv.SpSchema.GetDDL(ddl.Config{Comments: false, ProtectIds: true, Tables: true, ForeignKeys: true})
+	if len(spDDL) == 0 {
+		spDDL = []string{"\n-- Schema is empty -- no tables found\n"}
+	}
+	l = []string{
+		strings.Join(spDDL, ";\n\n"),
+		"\n",
+	}
+	if _, err = f.WriteString(strings.Join(l, "")); err != nil {
+		fmt.Fprintf(out, "Can't write out legal schema ddl file: %v\n", err)
+		return
+	}
+	fmt.Fprintf(out, "Wrote legal schema ddl to file '%s'.\n", name)
 }
 
 // WriteSessionFile writes conv struct to a file in JSON format.
@@ -585,6 +612,25 @@ func WriteSessionFile(conv *internal.Conv, name string, out *os.File) {
 		return
 	}
 	fmt.Fprintf(out, "Wrote session to file '%s'.\n", name)
+}
+
+// WriteConvGeneratedFiles creates a directory labeled downloads with the current timestamp
+// where it writes the sessionfile, report summary and DDLs then returns the directory where it writes.
+func WriteConvGeneratedFiles(conv *internal.Conv, dbName string, driver string, BytesRead int64, out *os.File) (string, error) {
+	now := time.Now()
+	dirPath := "harbour_bridge_output/" + dbName + "/"
+	err := os.MkdirAll(dirPath, os.ModePerm)
+	if err != nil {
+		fmt.Fprintf(out, "Can't create directory %s: %v\n", dirPath, err)
+		return "", err
+	}
+	schemaFileName := dirPath + dbName + "_schema.txt"
+	WriteSchemaFile(conv, now, schemaFileName, out)
+	reportFileName := dirPath + dbName + "_report.txt"
+	Report(driver, nil, BytesRead, "", conv, reportFileName, out)
+	sessionFileName := dirPath + dbName + ".session.json"
+	WriteSessionFile(conv, sessionFileName, out)
+	return dirPath, nil
 }
 
 // ReadSessionFile reads a session JSON file and
--- a/conversion/conversion.go
+++ b/conversion/conversion.go
@@ -43,6 +43,7 @@ import (
 	sp "cloud.google.com/go/spanner"
 	database "cloud.google.com/go/spanner/admin/database/apiv1"
 	instance "cloud.google.com/go/spanner/admin/instance/apiv1"
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	dydb "github.com/aws/aws-sdk-go/service/dynamodb"
 	_ "github.com/go-sql-driver/mysql"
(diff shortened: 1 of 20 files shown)


conversion/conversion.go
```diff
@@ -550,20 +550,41 @@ func WriteSchemaFile(conv *internal.Conv, now time.Time, name string, out *os.Fi
 	// legal Cloud Spanner DDL (Cloud Spanner doesn't currently support comments).
 	// Change 'Comments' to false and 'ProtectIds' to true to write out a
 	// schema file that is legal Cloud Spanner DDL.
```
agasheesh, line 552: We can maybe merge this comment to down below where we now actually generate a 'legal' Cloud Spanner DDL.

conversion/conversion.go
```diff
@@ -550,20 +550,41 @@ func WriteSchemaFile(conv *internal.Conv, now time.Time, name string, out *os.Fi
...
+		strings.Join(spddl, ";\n\n"),
 		"\n",
 	}
 	if _, err := f.WriteString(strings.Join(l, "")); err != nil {
 		fmt.Fprintf(out, "Can't write out schema file: %v\n", err)
 		return
 	}
 	fmt.Fprintf(out, "Wrote schema to file '%s'.\n", name)
+
+	// Write another file titled {name}.ddl without comments as Spanner DDL editor UI cannot process comments
```
agasheesh, line 568: Go comments are complete sentences and end with a period - https://golang.org/doc/effective_go#commentary. Also consider rewriting this comment while merging the "'Comments' to false and 'ProtectIds' to true" text from above.

conversion/conversion.go
```diff
@@ -550,20 +550,41 @@ func WriteSchemaFile(conv *internal.Conv, now time.Time, name string, out *os.Fi
...
+	name = name + ".ddl"
```
agasheesh, line 569: This will generate a file named '<name>.txt.ddl' - I wonder if that extension could be problematic for someone running HB on windows. How about '<name>.ddl.txt'?

conversion/conversion.go
```diff
@@ -550,20 +550,41 @@ func WriteSchemaFile(conv *internal.Conv, now time.Time, name string, out *os.Fi
...
+	f, err = os.Create(name)
+	if err != nil {
+		fmt.Fprintf(out, "Can't create uncommented schema file %s: %v\n", name, err)
```
agasheesh, line 572: 'uncomment schema file' isn't entirely accurate since we also have `protect_ids:true` - how about 'legal schema ddl file' for simplicity?

web/web.go
```diff
@@ -791,6 +792,14 @@ func updateSessionFile() error {
...
 		return err
 	}
+	dbName := sessionState.dbName
+	if dbName == "" {
+		dbName, err = conversion.GetDatabaseName(sessionState.driver, time.Now())
+		if err != nil {
+			return fmt.Errorf("Can not create database name : %v", err)
+		}
+	}
+	conversion.WriteDownloadableFiles(sessionState.conv, dbName, sessionState.driver, ioHelper.BytesRead, ioHelper.Out)
```
agasheesh, line 802: Is this going to write the session file twice?

frontend/js/services/Action.service.js
```diff
@@ -544,6 +544,16 @@ const Actions = (() => {
                 'UpdateCols': {}
             }
             let newColArrayForDuplicateCheck = [];
+            let columnsNamesArray = [];
```
agasheesh, line 547: Thanks for fixing this - as discussed offline, let's move it to a separate PR.

agasheesh:
There are few non-code related things that we should follow in this and following PRs.

1. In the "Linked Issues", we should link the issue that this PR will fix, if any. That way the issue will also automatically get closed.
2. PR message or the commit message should be descriptive enough such that someone tracking a given change can later get a good sense of the changes contained in a given PR without looking at all the changes in the PR. For example, "Write schema config" isn't descriptive enough in this case. 
3. It's also a good idea to add a paragraph or two describing what's changing. Note that this is different from the text in the linked issue which talks about the issue, and not how it's fixed.

conversion/conversion.go
```diff
@@ -538,32 +538,56 @@ func getInstances(project string) ([]string, error) {
...
 func WriteSchemaFile(conv *internal.Conv, now time.Time, name string, out *os.File) {
 	f, err := os.Create(name)
 	if err != nil {
 		fmt.Fprintf(out, "Can't create schema file %s: %v\n", name, err)
 		return
 	}
-	// The schema file we write out includes comments, includes foreign keys
-	// and doesn't add backticks around table and column names. This file is
-	// intended for explanatory and documentation purposes, and is not strictly
-	// legal Cloud Spanner DDL (Cloud Spanner doesn't currently support comments).
```
agasheesh, line 550: I think we should keep this part of the comment here with slightly modification. Something on the lines of:

// The schema file written below is optimized for reading. It includes comments, foreign keys and doesn't
// add backticks around table and column names. Since this file is intended for explanatory and documentation 
// purposes, it is not strictly legal Cloud Spanner DDL (Cloud Spanner doesn't currently support comments).

conversion/conversion.go
```diff
@@ -538,32 +538,56 @@ func getInstances(project string) ([]string, error) {
...
-	// Change 'Comments' to false and 'ProtectIds' to true to write out a
-	// schema file that is legal Cloud Spanner DDL.
-	ddl := conv.SpSchema.GetDDL(ddl.Config{Comments: true, ProtectIds: false, Tables: true, ForeignKeys: true})
-	if len(ddl) == 0 {
-		ddl = []string{"\n-- Schema is empty -- no tables found\n"}
+	spddl := conv.SpSchema.GetDDL(ddl.Config{Comments: true, ProtectIds: false, Tables: true, ForeignKeys: true})
```
agasheesh, line 548: Variable names in golang are written as mixedCaps e.g., spDDL - https://golang.org/doc/effective_go#mixed-caps. (I know it's a little debatable here but it's good to follow consistency in naming)

conversion/conversion.go
```diff
@@ -538,32 +538,56 @@ func getInstances(project string) ([]string, error) {
...
+
+	/*
+		The schema file written above includes comments, includes foreign keys
+		and doesn't add backticks around table and column names.
+		This file is intended for explanatory and documentation purposes, and
+		is not strictly legal Cloud Spanner DDL (Cloud Spanner doesn't currently support comments).
+
+		We change 'Comments' to false and 'ProtectIds' to true below to write out a
+		schema file that is a legal Cloud Spanner DDL.
+	*/
```
agasheesh, line 571: We like C++-style // line comments in Go code :)

Also comments typically talk about the code below, and thus we should only keep something on the lines of:

// We change 'Comments' to false and 'ProtectIds' to true below to write out a
// schema file that is a legal Cloud Spanner DDL.

conversion/conversion.go
```diff
@@ -538,32 +538,56 @@ func getInstances(project string) ([]string, error) {
...
+	name = strings.Replace(name, ".txt", ".ddl.txt", 1)
```
agasheesh, line 572: maybe nitpicking, but is it possible to only replace the last instance of ".txt" with ".ddl.txt"?

conversion/conversion.go
```diff
@@ -587,6 +611,25 @@ func WriteSessionFile(conv *internal.Conv, name string, out *os.File) {
 	fmt.Fprintf(out, "Wrote session to file '%s'.\n", name)
 }
 
+// WriteDownloadableFiles creates a directory labeled downloads with the current timestamp
+// where it writes the sessionfile, report summary and DDLs then returns the directory where it writes.
+func WriteDownloadableFiles(conv *internal.Conv, dbName string, driver string, BytesRead int64, out *os.File) (string, error) {
+	now := time.Now()
+	dirPath := "frontend/harbour_bridge_downloads/" + dbName + "/"
```
agasheesh, line 618: We can skip the prefix "harbour_bridge" since frontend/ folder will always be inside harbourbridge repo. 

Also I wonder what is the behavior when instead of running harbourbridge from the repo, a pre-built harbourbridge is used? Is it better to write at the current working directory?

conversion/conversion.go
```diff
@@ -587,6 +611,25 @@ func WriteSessionFile(conv *internal.Conv, name string, out *os.File) {
 	fmt.Fprintf(out, "Wrote session to file '%s'.\n", name)
 }
 
+// WriteDownloadableFiles creates a directory labeled downloads with the current timestamp
+// where it writes the sessionfile, report summary and DDLs then returns the directory where it writes.
+func WriteDownloadableFiles(conv *internal.Conv, dbName string, driver string, BytesRead int64, out *os.File) (string, error) {
```
agasheesh, line 616: I can't think of a better name for this variable, but if you think of something that seems better, please feel free to update. DownloadFiles seems too generic.

web/web.go
```diff
@@ -770,26 +770,12 @@ func dropSecondaryIndex(w http.ResponseWriter, r *http.Request) {
...
+// latest sessionState.conv while also dumping schemas and report.
 func updateSessionFile() error {
-	filePath := sessionState.sessionFile
-	if filePath == "" {
-		return fmt.Errorf("Session file path is empty")
-	}
-	f, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE, 0644)
-	if err != nil {
-		return err
-	}
```
agasheesh, line 782: Do we know why was this code required before, and why is it okay to drop it now?

agasheesh (commented):
Looks good mostly, just some minor comments/questions.

web/web.go
Deep1998: Not sure of the pros of opening, clearing then writing compared to simply doing a write. But I did verify the new changes and overwriting does the job.

conversion/conversion.go
```diff
@@ -538,32 +538,56 @@ func getInstances(project string) ([]string, error) {
...
+		strings.Join(spDDL, ";\n\n"),
 		"\n",
 	}
 	if _, err := f.WriteString(strings.Join(l, "")); err != nil {
 		fmt.Fprintf(out, "Can't write out schema file: %v\n", err)
 		return
 	}
 	fmt.Fprintf(out, "Wrote schema to file '%s'.\n", name)
+
+	// converting <file_name>.<ext> to <file_name>.ddl.<ext>.
```
agasheesh, line 567: nit; s/converting/Convert

comments are full sentences and are stated as a fact, rather than as a verb - https://golang.org/doc/effective_go#commentary

conversion/conversion.go
```diff
@@ -538,32 +538,56 @@ func getInstances(project string) ([]string, error) {
...
+	nameSplit := strings.Split(name, ".")
+	nameSplit = append(nameSplit[:len(nameSplit)-1], "ddl", nameSplit[len(nameSplit)-1])
+	name = strings.Join(nameSplit, ".")
+	f, err = os.Create(name)
+	if err != nil {
+		fmt.Fprintf(out, "Can't create legal schema ddl file %s: %v\n", name, err)
+		return
+	}
+	//	We change 'Comments' to false and 'ProtectIds' to true below to write out a
+	//	schema file that is a legal Cloud Spanner DDL.
```
agasheesh, line 577: nit; stray empty spaces

also it's a good idea to leave an empty line before a comment since comments typically refer to some group of code below

conversion/conversion.go
```diff
@@ -587,6 +611,25 @@ func WriteSessionFile(conv *internal.Conv, name string, out *os.File) {
 	fmt.Fprintf(out, "Wrote session to file '%s'.\n", name)
 }
 
+// WriteConvGeneratedFiles creates a directory labeled downloads with the current timestamp
+// where it writes the sessionfile, report summary and DDLs then returns the directory where it writes.
+func WriteConvGeneratedFiles(conv *internal.Conv, dbName string, driver string, BytesRead int64, out *os.File) (string, error) {
+	now := time.Now()
+	dirPath := "harbour_bridge_downloads/" + dbName + "/"
```
agasheesh, line 618: how about we call this directory "output"? since these aren't necessary downloaded files but rather output from running the command

web/web.go
agasheesh: Sounds good. Thanks for verifying.

agasheesh (approved):

conversion/conversion.go
```diff
@@ -538,32 +538,56 @@ func getInstances(project string) ([]string, error) {
...
 	// and doesn't add backticks around table and column names. This file is
 	// intended for explanatory and documentation purposes, and is not strictly
 	// legal Cloud Spanner DDL (Cloud Spanner doesn't currently support comments).
-	// Change 'Comments' to false and 'ProtectIds' to true to write out a
-	// schema file that is legal Cloud Spanner DDL.
-	ddl := conv.SpSchema.GetDDL(ddl.Config{Comments: true, ProtectIds: false, Tables: true, ForeignKeys: true})
-	if len(ddl) == 0 {
-		ddl = []string{"\n-- Schema is empty -- no tables found\n"}
+	spDDL := conv.SpSchema.GetDDL(ddl.Config{Comments: true, ProtectIds: false, Tables: true, ForeignKeys: true})
+	if len(spDDL) == 0 {
```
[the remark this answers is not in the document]
Deep1998: I changed this to spDDL because during the second call below to conv.SpSchema.GetDDL(**ddl**.Config{}) the ddl import name clashes with the variable name.

PR #144, opened 2021-06-09
Renaming a column then reverting to original name raises error in the UI

### LINKED ISSUE #142: Renaming a column then reverting to original name raises error in the UI
## Expected Behavior
If a col is renamed and then reverted to original name, it should do so successfully.

## Actual Behavior
Instead throws an error

## Steps to Reproduce the Problem
This may not happen for all cases. Use the cart.mysqldump in the repo and use user_id column for the steps to reproduce.
Open web UI
1. Change a column name (new_col_name) then save changes
2. Rename back to original name (old_col_name)
3. Clicking save may throw "old_col_name already exists"

Deep1998:
Fixes #142 

The error came up due to the columnNamesList being in a different order from what is seen in the UI. This was due to converting the keys of the name map to array which messed up the order. 

The fix was to generate the columnNamesList by iterating the UI for columnNames ensuring the order is consistent.

Also added unit tests to simulate this.

### DIFF
--- a/frontend/puppeteerTest.js
+++ b/frontend/puppeteerTest.js
@@ -92,6 +92,11 @@ puppeteer.launch(config.launchOptions).then(async(browser) => {
     */
     await resumeSession(page);
 
+    /*
+    14. Rename a column then revert back to original name
+    */
+    await renameAndRevertColumnName(page);
+    
     // Close Browser
     await browser.close();
 });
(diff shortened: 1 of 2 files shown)


frontend/puppeteerTest.js
```diff
@@ -560,6 +565,54 @@ const resumeSession = async(page) => {
     await passed(13);
 };
 
+const renameAndRevertColumnName = async(page) => {
+    await page.waitForSelector(homePage.customerTable);
```
agasheesh, line 569: I had a hard time following what is happening in the test. Can you please add block comments. I see that the rest of the file doesn't have block comments and that's something we missed in previous reviews.

Basically some commentary that guides reader through the test. Something like:

// Comment 1
Some block of code

// Comment 2
Next block of code

and so on

agasheesh (approved):
Looks really good. Thanks for the fix and for adding a test to catch future regressions. Just left a minor comment regarding some comments to the test (and optionally to the Action.service.js as well, since a future reader may not understand why did we chose to generate the column array this way).

PR #146, opened 2021-06-17
Added flag to skip foreign key creation for the CLI

### LINKED ISSUE #145: Improve performance of Foreign Key processing
The current behavior in HarbourBridge is to process schema change statements for creation of foreign keys in a linear order. Each schema update for foreign key creation can take a long time in Cloud Spanner since it needs to do data validation as well as create backing indexes for the foreign key, where applicable. For a source database with a large number of foreign keys, this can become problematic since each foreign key creation takes a long time and the overall time taken can be in hours. This can be solved using the following approaches:

### Alternative 1.1: Batch creation for multiple foreign key in the same schema update. 
There are some caveats to the same though:
#### Caveat 1.2.1: If there are too many schema updates happening at the same time, Cloud Spanner may throttle some of them. See https://cloud.google.com/spanner/docs/schema-updates#frequency for more details.

We will need to do some investigation and come up with a heuristic to combine Alternative 1.1 and Alternative 1.2 for optimal performance.

Foreign keys in HarbourBridge are today created after all the data has already migrated. This is done to avoid the issue where referencing rows are written before the referenced rows when migrating the data. This can be solved by doing the data migration in the same order as the one implied by DAG created by following foreign key dependencies. However, this is not an option when using mysqldump or pgdump since those cannot be read out of order to migrate tables in a specific order.

## Problem Statement 2: Create foreign keys during data migration for direct connect mode
There is some investigation/experimentation needed to understand if this would actually help with the overall data migration time.

For some of the customers, it may make sense to not have HarbourBridge perform the foreign key creations in the first place if they care more about reducing the down time and are willing to apply foreign keys themselves after the migration is over. To support such use cases, we should also provide a command line flag to skip foreign key creation during data migration. This is specially important if the data migration is followed by CDC migration since foreign keys may cause the migration of CDC change events to fail and get retried multiple times and thus increase the overall application downtime during migration.

Deep1998:
Throw an exception when both schema-only and skip-foreign-keys are set.
This flag does not omit the Foreign Key statements in the DDL.

Fixes 145 part 3

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,10 @@ Any data in the source database will be ignored.
 A spanner database will be created based on the schema state provided
 by a session file (`-session`) and data will be migrated.
 
+`-skip-foreign-keys` Specifies that foreign key creation will be skipped after 
+data migration is complete. It cannot be used with schema-only mode. This flag does not 
+omit Foreign Key statements in the generated DDL.
+
 `-session` Specifies a session file that contains all schema and data 
 conversion state endcoded as JSON.
 
--- a/main.go
+++ b/main.go
@@ -41,6 +41,7 @@ var (
 	verbose          bool
 	schemaOnly       bool
 	dataOnly         bool
+	skipForeignKeys  bool
 	sessionJSON      string
 	webapi           bool
 )
@@ -54,6 +55,7 @@ func init() {
 	flag.BoolVar(&verbose, "v", false, "verbose: print additional output")
 	flag.BoolVar(&schemaOnly, "schema-only", false, "schema-only: in this mode we do schema conversion, but skip data conversion")
 	flag.BoolVar(&dataOnly, "data-only", false, "data-only: in this mode we skip schema conversion and just do data conversion (use the session flag to specify the session file for schema and data mapping)")
+	flag.BoolVar(&skipForeignKeys, "skip-foreign-keys", false, "skip-foreign-keys: in this mode we do not update the foreign keys after data migration is complete")
 	flag.StringVar(&sessionJSON, "session", "", "session: specifies the file we restore session state from (used in schema-only to provide schema and data mapping)")
 	flag.BoolVar(&webapi, "web", false, "web: run the web interface (experimental)")
 }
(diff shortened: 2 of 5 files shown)


main.go
```diff
@@ -54,6 +55,7 @@ func init() {
 	flag.BoolVar(&verbose, "v", false, "verbose: print additional output")
 	flag.BoolVar(&schemaOnly, "schema-only", false, "schema-only: in this mode we do schema conversion, but skip data conversion")
 	flag.BoolVar(&dataOnly, "data-only", false, "data-only: in this mode we skip schema conversion and just do data conversion (use the session flag to specify the session file for schema and data mapping)")
+	flag.BoolVar(&skipForeignKeys, "skip-foreign-keys", false, "skip-foreign-keys: in this mode we do not update the foreign keys after data migration is complete")
```
agasheesh, line 58: this does not specify a mode unlike dataOnly and schemaOnly - how about we say something like 

"if true, skip creating foreign keys after data migration is complete (ddl statements for foreign keys can still be found in the downloaded schema.ddl.json file and the same can be applied separately)"

agasheesh (approved):
Looks good. Thanks for all the changes.

README.md
```diff
@@ -284,6 +284,10 @@ Any data in the source database will be ignored.
 A spanner database will be created based on the schema state provided
 by a session file (`-session`) and data will be migrated.
 
+`-skip-foreign-keys` Specifies that foreign key creation will be skipped after 
+data migration is complete. It cannot be used with schema-only mode. This flag does not
```
nevinheintze, line 288: Suggest a slight rewording:

`-skip-foreign-keys` Controls whether we add foreign key constraints after
data migration is complete. This flag cannot be used with schema-only mode,
and does not affect the generation of foreign key statements during schema
processing i.e. foreign key constraints will still appear in the generated Spanner
DDL files.

testing/mysql/integration_test.go
```diff
@@ -118,7 +118,7 @@ func TestIntegration_MYSQLDUMP_SimpleUse(t *testing.T) {
 	if err != nil {
```
nevinheintze, line 118: We should add a simple test for skip-foreign-keys -- maybe a follow-up PR?

nevinheintze (approved):
Really sorry for the slow review.
Just a couple of small comments.

Looks good -- thanks!

README.md
Deep1998: Updated

testing/mysql/integration_test.go
Deep1998: Agreed, I'll put it up in the next PR.

PR #150, opened 2021-07-13
Boilerplate code to run a dynamodb integration test

Deep1998:
Boilerplate code for DynamoDB tests to run a basic flow before adding proper test cases.
Currently the flow spins up a dynamodb container -> adds one row -> migrates to cloud spanner -> reads from the spanner instance.

Introduced the env variable DYNAMODB_ENDPOINT_OVERRIDE (specific to harbourbridge) to provide a custom endpoint for the dynamodb instance. By default, aws uses the AWS_REGION env variable to derive the endpoint (No env variable support from AWS to override the endpoint).

### DIFF
--- a/conversion/conversion.go
+++ b/conversion/conversion.go
@@ -43,6 +43,7 @@ import (
 	sp "cloud.google.com/go/spanner"
 	database "cloud.google.com/go/spanner/admin/database/apiv1"
 	instance "cloud.google.com/go/spanner/admin/instance/apiv1"
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	dydb "github.com/aws/aws-sdk-go/service/dynamodb"
 	_ "github.com/go-sql-driver/mysql"
@@ -216,7 +217,14 @@ func dataFromSQL(driver string, config spanner.BatchWriterConfig, client *sp.Cli
 func schemaFromDynamoDB(sampleSize int64) (*internal.Conv, error) {
 	conv := internal.MakeConv()
 	mySession := session.Must(session.NewSession())
-	client := dydb.New(mySession)
+	cfg := aws.Config{}
+	endpointOverride := os.Getenv("DYNAMODB_ENDPOINT_OVERRIDE")
+	if endpointOverride != "" {
+		cfg = aws.Config{
+			Endpoint: aws.String(endpointOverride),
+		}
+	}
+	client := dydb.New(mySession, &cfg)
 	err := dynamodb.ProcessSchema(conv, client, []string{}, sampleSize)
 	if err != nil {
 		return nil, err
@@ -226,8 +234,14 @@ func schemaFromDynamoDB(sampleSize int64) (*internal.Conv, error) {
 
 func dataFromDynamoDB(config spanner.BatchWriterConfig, client *sp.Client, conv *internal.Conv) (*spanner.BatchWriter, error) {
 	mySession := session.Must(session.NewSession())
-	dyclient := dydb.New(mySession)
-
+	cfg := aws.Config{}
+	endpointOverride := os.Getenv("DYNAMODB_ENDPOINT_OVERRIDE")
+	if endpointOverride != "" {
+		cfg = aws.Config{
+			Endpoint: aws.String(endpointOverride),
+		}
+	}
+	dyclient := dydb.New(mySession, &cfg)
 	dynamodb.SetRowStats(conv, dyclient)
 	totalRows := conv.Rows()
 	p := internal.NewProgress(totalRows, "Writing data to Spanner", internal.Verbose())
--- a/testing/dynamodb/integration_test.go
+++ b/testing/dynamodb/integration_test.go
@@ -1,0 +1,244 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


testing/dynamodb/integration_test.go
```diff
@@ -0,0 +1,244 @@
...
+			t.Fatal(err)
+			break
+		}
+		var year, title, plot string
+		if err := row.Columns(&year, &title, &plot); err != nil {
+			log.Println("Error reading into variables: ", err)
+			t.Fatal(err)
+			break
+		}
+		log.Printf("Found the following:\nYear: %s\nTitle: %s\nPlot: %s\n", year, title, plot)
```
agasheesh, line 236: I am assuming this will get cleaned up in a follow up PR with actual comparison between Dynamodb and Spanner rows.

conversion/conversion.go
```diff
@@ -216,7 +217,14 @@ func dataFromSQL(driver string, config spanner.BatchWriterConfig, client *sp.Cli
...
 	mySession := session.Must(session.NewSession())
-	client := dydb.New(mySession)
+	cfg := aws.Config{}
+	endpointOverride := os.Getenv("DYNAMODB_ENDPOINT_OVERRIDE")
+	if endpointOverride != "" {
+		cfg = aws.Config{
+			Endpoint: aws.String(endpointOverride),
+		}
+	}
+	client := dydb.New(mySession, &cfg)
```
agasheesh, line 227: consider moving this to a common function

agasheesh (approved):
Thanks for contributing this, looks really good.

testing/dynamodb/integration_test.go
Deep1998: Correct let me fix that in this only

conversion/conversion.go
Deep1998: Done

conversion/conversion.go
```diff
@@ -213,10 +214,22 @@ func dataFromSQL(driver string, config spanner.BatchWriterConfig, client *sp.Cli
 	return writer, nil
 }
 
+func getDynamoDBClientConfig() aws.Config {
+	cfg := aws.Config{}
+	endpointOverride := os.Getenv("DYNAMODB_ENDPOINT_OVERRIDE")
+	if endpointOverride != "" {
+		cfg = aws.Config{
+			Endpoint: aws.String(endpointOverride),
+		}
```
agasheesh, line 223: It's more typical to write this as `cfg.Endpoint = aws.String(endpointOverride)` instead of redeclaring cfg.

conversion/conversion.go
```diff
@@ -213,10 +214,22 @@ func dataFromSQL(driver string, config spanner.BatchWriterConfig, client *sp.Cli
...
+	}
+	return cfg
+}
+
 func schemaFromDynamoDB(sampleSize int64) (*internal.Conv, error) {
 	conv := internal.MakeConv()
 	mySession := session.Must(session.NewSession())
-	client := dydb.New(mySession)
+	cfg := getDynamoDBClientConfig()
+	client := dydb.New(mySession, &cfg)
```
agasheesh, line 232: consider inlining getDynamoDBClientConfig() since cfg isn't used again later in the code; similarly in dataFromDynamoDB below

conversion/conversion.go
```diff
@@ -226,8 +239,8 @@ func schemaFromDynamoDB(sampleSize int64) (*internal.Conv, error) {
 
 func dataFromDynamoDB(config spanner.BatchWriterConfig, client *sp.Client, conv *internal.Conv) (*spanner.BatchWriter, error) {
 	mySession := session.Must(session.NewSession())
-	dyclient := dydb.New(mySession)
-
+	cfg := getDynamoDBClientConfig()
+	dyclient := dydb.New(mySession, &cfg)
```
agasheesh, line 243: I know this is existing code, but having dyclient here and client above seems a bit odd; consider just renaming this to client

testing/dynamodb/integration_test.go
```diff
@@ -0,0 +1,247 @@
+// Copyright 2020 Google LLC
+//
```
agasheesh, line 2: 2021 :)

testing/dynamodb/integration_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	"cloud.google.com/go/spanner"
+	database "cloud.google.com/go/spanner/admin/database/apiv1"
+	"google.golang.org/api/iterator"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	dydb "github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute"
+	databasepb "google.golang.org/genproto/googleapis/spanner/admin/database/v1"
```
agasheesh, line 40: thanks for sorting the imports cleanly :)

testing/dynamodb/integration_test.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+		var year, title, plot string
+		if err := row.Columns(&year, &title, &plot); err != nil {
+			log.Println("Error reading into variables: ", err)
+			t.Fatal(err)
+			break
+		}
+		assert.Equal(t, year, "2015")
+		assert.Equal(t, title, "The Big New Movie")
+		assert.Equal(t, plot, "Nothing happens at all.")
```
agasheesh, line 239: this is okay for now, but we should change these functions to take thee expected row as input instead of comparing against literals ~ similarly in corresponding packages in mysql and postgres

testing/dynamodb/integration_test.go
Deep1998: Haha good catch

testing/dynamodb/integration_test.go
Deep1998: Adding a future todo in comment

PR #153, opened 2021-07-19
removed unique from column def

### LINKED ISSUE #105: Refactor: cleanup handling of UNIQUE
Our handling of UNIQUE is confusing and our generation of schema (schema.go) is broken. Fortunately this doesn't cause us to generate broken Spanner schema.

Here are some more details:

a) In schema.go: we represent UNIQUE constraints within Column and also within Index. This means that UNIQUE constraints are redundantly represented which is pretty confusing. We should have one method for representing all UNIQUE constraints. See Discussion section for details of choices we have.

b) UpdateCols is broken: it should not push multi-column UNIQUE constraints down to the column level --- see the two example 'test' tables in the Discussion section.

c) We should avoid doing Spanner-specific transformations when we map into schema (schema.go). Where possible, Spanner-specific transformations should be in toddl.go. At present we have comments in e.g. mysqldump.go indicating Spanner-specific transformations.

**DISCUSSION**
How should we represent UNIQUE constraints in schema.go?
First, note that the UNIQUE constraint at the Column level is not sufficient for representing multi-column UNIQUE constraints. In particular:
   CREATE TABLE test (id bigint, x bigint, UNIQUE(id, x));
   CREATE TABLE test (id bigint UNIQUE, x bigint UNIQUE);
are not the same -- the first says that, when combined, id and x are unique, but the latter says they are individually unique (which is a stronger constraint).

Now, one option would be to add a multi-column constraint to schema.go and representing all UNIQUE constraints as multi-column constraints (dropping Unique from Column and from Index). However, this is not clear-cut since source databases have multiple ways to define UNIQUE constraints: you can either do it declaratively (as a property of a column or table), or you can do it using an index (essentially focusing on how the constraint is implemented). For example in MySQL DDL UNIQUE constraints can be either be:
i) A column level constraint (as part of a CREATE TABLE statement)
ii) A table level constraint (as part of a CREATE TABLE statement)
iii) An index (as part of a CREATE TABLE statement)
iv) Added via an ALTER TABLE constraint (changing column constraints, table constraints or index definitions)
iv) Added via a CREATE UNIQUE INDEX statement (which internally maps to an alter table statement).

In MySQL infoschema tables, UNIQUE is represented both as part of the TABLE_CONSTRAINTS table as well as part of the STATISTICS table (which describes indexes).

Under the hood, all databases implement UNIQUE as an index.
Maybe we should just handle them as indexes after all?

bharadwaj-aditya:
Fixes #105 

Removed the unique attribute from columns and kept it as a constraint everywhere

- [x] Tests pass

### DIFF
--- a/mysql/mysqldump.go
+++ b/mysql/mysqldump.go
@@ -209,7 +209,7 @@ func processCreateTable(conv *internal.Conv, stmt *ast.CreateTableStmt) {
 		return
 	}
 	tableName, err := getTableName(stmt.Table)
-	internal.VerbosePrintf("processing create table elem=%s stmt=%s\n", tableName, stmt)
+	internal.VerbosePrintf("processing create table elem=%s stmt=%v\n", tableName, stmt)
 	if err != nil {
 		logStmtError(conv, stmt, fmt.Errorf("can't get table name: %w", err))
 		return
--- a/mysql/mysqldump.go
+++ b/mysql/mysqldump.go
@@ -209,6 +209,7 @@ func processCreateTable(conv *internal.Conv, stmt *ast.CreateTableStmt) {
 		return
 	}
 	tableName, err := getTableName(stmt.Table)
+	internal.VerbosePrintf("processing create table elem=%s stmt=%s\n", tableName, stmt)
 	if err != nil {
 		logStmtError(conv, stmt, fmt.Errorf("can't get table name: %w", err))
 		return
@@ -328,13 +329,10 @@ func updateCols(conv *internal.Conv, ct ast.ConstraintType, colNames []*ast.Inde
 		colName := column.Column.OrigColName()
 		cd := colDef[colName]
 		switch ct {
-		case ast.ConstraintUniq:
-			cd.Unique = true
 		case ast.ConstraintCheck:
 			cd.Ignored.Check = true
 		case ast.ConstraintPrimaryKey:
 			cd.NotNull = true
-			cd.Unique = true
 		}
 		colDef[colName] = cd
 	}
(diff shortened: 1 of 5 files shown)


bharadwaj-aditya:
@agasheesh @nevinheintze - please review

mysql/mysqldump.go
```diff
@@ -328,13 +329,10 @@ func updateCols(conv *internal.Conv, ct ast.ConstraintType, colNames []*ast.Inde
 		colName := column.Column.OrigColName()
 		cd := colDef[colName]
 		switch ct {
-		case ast.ConstraintUniq:
-			cd.Unique = true
```
agasheesh, line 332: I didn't know that this was done only for mysqldump and not for pgdump :O

agasheesh (approved):
LGTM, please also get a LGTM from @nevinheintze for this PR.

nevinheintze (approved):
Thanks Aditya -- looks good!

At some point it probably makes sense to cleanup our TODO statements about spanner-specific transformations e.g. mysqldump.go. Probably should be a separate PR.

Suggest:
a) adding a comment in schema.go, where we define "type Index" and say something like: This is the only way to represent unique constraints in schema.go. All source database unique constraints will be transformed into this representation, including:
i) A column level constraint (as part of a CREATE TABLE statement)
ii) A table level constraint (as part of a CREATE TABLE statement)
iii) An index (as part of a CREATE TABLE statement)
iv) Added via an ALTER TABLE constraint (changing column constraints, table constraints or index definitions)
iv) Added via a CREATE UNIQUE INDEX statement (which internally maps to an alter table statement).
We use this single representation of unique constraints to simplify their processing and avoid having to handle lots of cases for the same concept. Our choice of an index representation for unique is largely motivated by the fact that databases typically implement UNIQUE via an index.

b) Modifying the comments to say something like: 
Convert unique column constraint in MySQL to a unique index in schema
representation. Note that schema represents all unique constraints as
indexes.
And delete the TODOs.

PR #159, opened 2021-08-20
Added support for Secondary Indexes for dynamoDB

Deep1998:
We convert both global and local secondary indexes now. Earlier these were not converted.

### DIFF
--- a/dynamodb/schema.go
+++ b/dynamodb/schema.go
@@ -128,13 +128,21 @@ func analyzeMetadata(client dynamoClient, s *schema.Table) error {
 		s.PrimaryKeys = append(s.PrimaryKeys, schema.Key{Column: *i.AttributeName})
 	}
 
-	// Secondary indexes
+	// Secondary indexes from GlobalSecondaryIndexes
 	for _, i := range result.Table.GlobalSecondaryIndexes {
 		var keys []schema.Key
 		for _, j := range i.KeySchema {
 			keys = append(keys, schema.Key{Column: *j.AttributeName})
 		}
-		// s.SecIndexes = append(s.SecIndexes, index{Name: *i.IndexName, Keys: keys})
+		s.Indexes = append(s.Indexes, schema.Index{Name: *i.IndexName, Keys: keys})
+	}
+
+	// Secondary indexes from LocalSecondaryIndexes
+	for _, i := range result.Table.LocalSecondaryIndexes {
+		var keys []schema.Key
+		for _, j := range i.KeySchema {
+			keys = append(keys, schema.Key{Column: *j.AttributeName})
+		}
 		s.Indexes = append(s.Indexes, schema.Index{Name: *i.IndexName, Keys: keys})
 	}
 
--- a/dynamodb/toddl.go
+++ b/dynamodb/toddl.go
@@ -28,6 +28,23 @@ import (
 // Spanner. It uses the source schema in conv.SrcSchema, and writes
 // the Spanner schema to conv.SpSchema.
 func schemaToDDL(conv *internal.Conv) error {
+	// Tracks Spanner names that have been used for foreign key constraints
+	// and indexes. We use this to ensure we generate unique names when
+	// we map from MySQL to Spanner since Spanner requires all foreign
+	// key and index names to be distinct (you can't use the same name
+	// for a foreign key constraint and an index).
+	usedNames := make(map[string]bool)
+	// As Spanner uses same namespace for table names, foreign key constraint
+	// names and index names, we need to pre-populate usedNames with Spanner table
+	// names to handle collision with foreign key names and index names.
+	for _, srcTable := range conv.SrcSchema {
+		spTableName, err := internal.GetSpannerTable(conv, srcTable.Name)
+		if err != nil {
+			conv.Unexpected(fmt.Sprintf("Couldn't map source table %s to Spanner: %s", srcTable.Name, err))
+			continue
+		}
+		usedNames[spTableName] = true
+	}
 	for _, srcTable := range conv.SrcSchema {
 		spTableName, err := internal.GetSpannerTable(conv, srcTable.Name)
 		if err != nil {
@@ -64,6 +81,7 @@ func schemaToDDL(conv *internal.Conv) error {
 			ColNames: spColNames,
 			ColDefs:  spColDef,
 			Pks:      cvtPrimaryKeys(conv, srcTable.Name, srcTable.PrimaryKeys),
+			Indexes:  cvtIndexes(conv, spTableName, srcTable.Name, srcTable.Indexes, usedNames),
 			Comment:  comment}
 	}
 	return nil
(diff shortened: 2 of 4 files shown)


dynamodb/toddl.go
```diff
@@ -28,6 +28,23 @@ import (
 // Spanner. It uses the source schema in conv.SrcSchema, and writes
 // the Spanner schema to conv.SpSchema.
 func schemaToDDL(conv *internal.Conv) error {
+	// Tracks Spanner names that have been used for foreign key constraints
+	// and indexes. We use this to ensure we generate unique names when
+	// we map from MySQL to Spanner since Spanner requires all foreign
```
bharadwaj-aditya, line 33: the comment seems to be Mysql specific

dynamodb/toddl.go
```diff
@@ -116,3 +134,32 @@ func cvtPrimaryKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Key)
...
+		var spKeys []ddl.IndexKey
+		for _, k := range srcIndex.Keys {
+			spCol, err := internal.GetSpannerCol(conv, srcTable, k.Column, true)
+			if err != nil {
+				conv.Unexpected(fmt.Sprintf("Can't map index key column name for table %s", srcTable))
+				continue
+			}
+			spKeys = append(spKeys, ddl.IndexKey{Col: spCol, Desc: k.Desc})
+		}
+		if srcIndex.Name == "" {
```
bharadwaj-aditya, line 150: can index names be empty in dynamo ?

bharadwaj-aditya (changes requested):

dynamodb/toddl.go
Deep1998: Not really. The api does not let me skip the name field. But should we still keep it as its non-disruptive?

dynamodb/schema.go
```diff
@@ -128,13 +128,21 @@ func analyzeMetadata(client dynamoClient, s *schema.Table) error {
...
 
-	// Secondary indexes
+	// Secondary indexes from GlobalSecondaryIndexes
 	for _, i := range result.Table.GlobalSecondaryIndexes {
 		var keys []schema.Key
 		for _, j := range i.KeySchema {
 			keys = append(keys, schema.Key{Column: *j.AttributeName})
 		}
-		// s.SecIndexes = append(s.SecIndexes, index{Name: *i.IndexName, Keys: keys})
+		s.Indexes = append(s.Indexes, schema.Index{Name: *i.IndexName, Keys: keys})
```
hengfengli, line 137: Does this mean we treat these two types of indexes in the same way later? Why don't we store GlobalSecondaryIndexes in `SecIndexes`?

dynamodb/schema.go
```diff
@@ -128,13 +128,21 @@ func analyzeMetadata(client dynamoClient, s *schema.Table) error {
...
+	}
+
+	// Secondary indexes from LocalSecondaryIndexes
+	for _, i := range result.Table.LocalSecondaryIndexes {
```
hengfengli, line 141: These two code blocks look quite same. Should we create a function for them?

dynamodb/toddl.go
```diff
@@ -116,3 +134,32 @@ func cvtPrimaryKeys(conv *internal.Conv, srcTable string, srcKeys []schema.Key)
...
+			// Generate a name if index name is empty in MySQL.
```
hengfengli, line 151: The comment is about MySQL. Please update it.

hengfengli (commented):
LGTM. Left a few comments. A question is that in the current approach, do we create the index first and then insert data? Or we insert the data first and then create the index? Which way do we use for creating indexes?

dynamodb/schema.go
Deep1998: SecIndexes might be a deprecated field. Currently, the struct field offers only one field for this i.e. `Indexes`.
```
type Table struct {
	Name        string
	ColNames    []string          
	ColDefs     map[string]Column
	PrimaryKeys []Key
	ForeignKeys []ForeignKey
	Indexes     []Index
}
```

Deep1998:
@hengfengli Currently we only support creating indexes first along with the database and then insert the data.

dynamodb/toddl.go
```diff
@@ -28,6 +28,19 @@ import (
 // Spanner. It uses the source schema in conv.SrcSchema, and writes
 // the Spanner schema to conv.SpSchema.
 func schemaToDDL(conv *internal.Conv) error {
+	// Tracks Spanner names that have been used for indexes. We use this to ensure we generate unique names when
+	// we map from DynamoDB to Spanner since we want Spanner table names and index names to be distinct.
+	usedNames := make(map[string]bool)
```
agasheesh, line 33: Consider adding test coverage for usedNames in case index name collides with table name (assuming this can happen in dynamodb and that's why we added this handling).

dynamodb/toddl.go
agasheesh: I would lean slightly towards not having this if statement in that case since the comment below lets readers believe that index names could be empty in dynamodb.

dynamodb/schema.go
```diff
@@ -128,13 +128,21 @@ func analyzeMetadata(client dynamoClient, s *schema.Table) error {
...
+		var keys []schema.Key
+		for _, j := range i.KeySchema {
+			keys = append(keys, schema.Key{Column: *j.AttributeName})
+		}
 		s.Indexes = append(s.Indexes, schema.Index{Name: *i.IndexName, Keys: keys})
```
agasheesh, line 146: From my cursory reading of DynamoDB documentation, it didn't become apparent that we could map both GlobalSecondaryIndex and LocalSecondaryIndex to secondary indexes in Spanner (which are always global in the case of Spanner). It would help to have a paragraph explaining the same above these for loops.

dynamodb/schema.go
```diff
@@ -128,13 +128,21 @@ func analyzeMetadata(client dynamoClient, s *schema.Table) error {
 		s.PrimaryKeys = append(s.PrimaryKeys, schema.Key{Column: *i.AttributeName})
 	}
 
-	// Secondary indexes
+	// Secondary indexes from GlobalSecondaryIndexes
```
agasheesh, line 131: nit; This comment is not a complete sentence. Similarly below.

dynamodb/toddl.go
```diff
@@ -28,6 +28,19 @@ import (
...
+	// We need to pre-populate usedNames with Spanner table
+	// names to handle collision with table names, foreign key names and index names.
```
agasheesh, line 35: nit; Formatting looks a little weird in the sense that newline started all of a sudden after Spanner table. Can you fix it?

dynamodb/schema.go
```diff
@@ -123,18 +123,31 @@ func analyzeMetadata(client dynamoClient, s *schema.Table) error {
...
 	for _, i := range result.Table.KeySchema {
 		s.PrimaryKeys = append(s.PrimaryKeys, schema.Key{Column: *i.AttributeName})
 	}
 
-	// Secondary indexes
+	// DynamoDB supports 2 types of indexes: Global Secondary Indexes (GSI) and Local Secondary Indexes (LSI).
+	// In GSI, dydb creates another global table for indexes that scales seperately.
+	// As for LSI, every partition in dydb maintains its own local index for that partition.
+	// For spanner, we should convert both these types as how dydb implements them is irrelevant.
+
```
Deep1998, line 135: @bharadwaj-aditya please review if this explanation is sufficient.

bharadwaj-aditya: This looks fine. Also add the link to the doc for details - https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LSI.html

PR #162, opened 2021-08-30
Refactoring to remove duplicate code across dbs

bharadwaj-aditya:
There is a large amount of code that is replicated in the implementation of each database. This PR aim's to remove most of the code duplication in across these and structure the code in a more intuitive manner.
These changes will be verified based on manual smoke tests and passing of existing unit tests.

- [x] Tests pass


agasheesh:
Hi @bharadwaj-aditya - thanks for sending out this PR. I see that all of our integration tests are failing with this PR. Can you please take a look at the same?

bharadwaj-aditya:
Looking into it

agasheesh:
Hi @bharadwaj-aditya - thanks a lot for cleaning this up. I think the approach makes sense. Basically, the common interface encapsulates the function that is different across different source databases such that they can be concretely implemented, where the common functionality is implemented as free standing functions. I have following comments:

1. Can we do the pure renaming refactoring in first PR i.e., move all of the code in sources/. There are few reasons for doing this: 1) This will let us keep the history if you use `git mv` for moving the files, 2) It will let the follow up PRs be focused only on actual changes of interest. 

2. It's not clear to me if all of the interface functions need to be public e.g., GetTablesQuery in BaseInfoSchema. This is a good opportunity to clean up public/private boundaries between modules.

3. I am slightly concerned by all the tests are failing. Can we do the refactoring such that the first pass doesn't change the functionality. Follow up PRs may break the tests but then we will be able to take a closer look at the functionality changes.

@nevinheintze - Please feel free to take a pass on this PR such that we can reconcile our thoughts along with @bharadwaj-aditya and make progress.

agasheesh (commented):
Thanks Aditya for the offline discussion to help explain the reason for failing tests. Noting down the discussion points from our discussion to keep @nevinheintze in loop.

Re 1. We discussed the concerns about being able to roll back the changes, and agreed that we can first safely roll out changes to move source code to sources/ which shouldn't require a rollback or fixing any tests. This will be then followed by actual refactoring PR, which we can rollback standalone if we run into any issues. This will also help with thorough code reviews.

Re 3. We agreed that the current approach of extracting out queries for table, indexes etc is probably overkill and error prone since different databases implement info schema differently. Thus, the path forward here is for each source database implementation to implement its own GetTables() function instead of GetTablesQuery() and so on.

bharadwaj-aditya:
Closing PR - Replacement PR's are 
https://github.com/cloudspannerecosystem/harbourbridge/pull/179
https://github.com/cloudspannerecosystem/harbourbridge/pull/181

PR #163, opened 2021-09-07
Initial set of changes support existing dbs

Deep1998:
Initial set of changes to support existings DBs. Currently commented out identifying dialect.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -39,7 +39,7 @@ var (
 func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnly, schemaOnly, skipForeignKeys bool, schemaSampleSize int64, sessionJSON string, ioHelper *conversion.IOStreams, outputFilePrefix string, now time.Time) error {
 	var conv *internal.Conv
 	var err error
-	dbExists, err := conversion.CheckExistingDb(projectID, instanceID, dbName)
+	dbExists, err := conversion.VerifyDb(projectID, instanceID, dbName)
 	if err != nil {
 		return err
 	}
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -66,26 +66,26 @@ func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnl
 		}
 	}
 
-	db, err := conversion.CreateOrUpdateDatabase(projectID, instanceID, dbName, dbExists, conv, ioHelper.Out)
+	dbURI, err := conversion.CreateOrUpdateDatabase(projectID, instanceID, dbName, dbExists, conv, ioHelper.Out)
 	if err != nil {
 		return fmt.Errorf("can't create/update database: %v", err)
 	}
 
-	client, err := conversion.GetClient(db)
+	client, err := conversion.GetClient(dbURI)
 	if err != nil {
-		return fmt.Errorf("can't create client for db %s: %v", db, err)
+		return fmt.Errorf("can't create client for db %s: %v", dbURI, err)
 	}
 
 	bw, err := conversion.DataConv(driver, ioHelper, client, conv, dataOnly)
 	if err != nil {
-		return fmt.Errorf("can't finish data conversion for db %s: %v", db, err)
+		return fmt.Errorf("can't finish data conversion for db %s: %v", dbURI, err)
 	}
 	if !skipForeignKeys {
 		if err = conversion.UpdateDDLForeignKeys(projectID, instanceID, dbName, conv, ioHelper.Out); err != nil {
-			return fmt.Errorf("can't perform update operation on db %s with foreign keys: %v", db, err)
+			return fmt.Errorf("can't perform update schema on db %s with foreign keys: %v", dbURI, err)
 		}
 	}
-	banner := conversion.GetBanner(now, db)
+	banner := conversion.GetBanner(now, dbURI)
 	conversion.Report(driver, bw.DroppedRowsByTable(), ioHelper.BytesRead, banner, conv, outputFilePrefix+reportFile, ioHelper.Out)
 	conversion.WriteBadData(bw, conv, banner, outputFilePrefix+badDataFile, ioHelper.Out)
 	return nil
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -39,7 +39,7 @@ var (
 func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnly, schemaOnly, skipForeignKeys bool, schemaSampleSize int64, sessionJSON string, ioHelper *conversion.IOStreams, outputFilePrefix string, now time.Time) error {
 	var conv *internal.Conv
 	var err error
-	targetDb, existingDb, err := conversion.VerifyTargetDb(targetDb, projectID, instanceID, dbName)
+	dbExists, err := conversion.CheckExistingDb(projectID, instanceID, dbName)
 	if err != nil {
 		return err
 	}
@@ -66,27 +66,23 @@ func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnl
 		}
 	}
 
-	db, err := conversion.CreateOrUpdateDatabase(targetDb, projectID, instanceID, dbName, existingDb, conv, ioHelper.Out)
+	db, err := conversion.CreateOrUpdateDatabase(projectID, instanceID, dbName, dbExists, conv, ioHelper.Out)
 	if err != nil {
-		fmt.Printf("\nCan't create/update database: %v\n", err)
-		return fmt.Errorf("can't create/update database")
+		return fmt.Errorf("can't create/update database: %v", err)
 	}
 
 	client, err := conversion.GetClient(db)
 	if err != nil {
-		fmt.Printf("\nCan't create client for db %s: %v\n", db, err)
-		return fmt.Errorf("can't create Spanner client")
+		return fmt.Errorf("can't create client for db %s: %v", db, err)
 	}
 
 	bw, err := conversion.DataConv(driver, ioHelper, client, conv, dataOnly)
 	if err != nil {
-		fmt.Printf("\nCan't finish data conversion for db %s: %v\n", db, err)
-		return fmt.Errorf("can't finish data conversion")
+		return fmt.Errorf("can't finish data conversion for db %s: %v", db, err)
 	}
 	if !skipForeignKeys {
 		if err = conversion.UpdateDDLForeignKeys(projectID, instanceID, dbName, conv, ioHelper.Out); err != nil {
-			fmt.Printf("\nCan't perform update operation on db %s with foreign keys: %v\n", db, err)
-			return fmt.Errorf("can't perform update schema with foreign keys")
+			return fmt.Errorf("can't perform update operation on db %s with foreign keys: %v", db, err)
 		}
 	}
 	banner := conversion.GetBanner(now, db)
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -39,6 +39,10 @@ var (
 func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnly, schemaOnly, skipForeignKeys bool, schemaSampleSize int64, sessionJSON string, ioHelper *conversion.IOStreams, outputFilePrefix string, now time.Time) error {
 	var conv *internal.Conv
 	var err error
+	targetDb, isNewDb, err := conversion.VerifyTargetDb(targetDb, projectID, instanceID, dbName)
+	if err != nil {
+		return err
+	}
 	if !dataOnly {
 		conv, err = conversion.SchemaConv(driver, targetDb, ioHelper, schemaSampleSize)
 		if err != nil {
@@ -62,7 +66,7 @@ func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnl
 		}
 	}
 
-	db, err := conversion.CreateDatabase(projectID, instanceID, dbName, conv, ioHelper.Out)
+	db, err := conversion.CreateOrUpdateDatabase(targetDb, projectID, instanceID, dbName, isNewDb, conv, ioHelper.Out)
 	if err != nil {
 		fmt.Printf("\nCan't create database: %v\n", err)
 		return fmt.Errorf("can't create database")
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -39,7 +39,7 @@ var (
 func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnly, schemaOnly, skipForeignKeys bool, schemaSampleSize int64, sessionJSON string, ioHelper *conversion.IOStreams, outputFilePrefix string, now time.Time) error {
 	var conv *internal.Conv
 	var err error
-	targetDb, isNewDb, err := conversion.VerifyTargetDb(targetDb, projectID, instanceID, dbName)
+	targetDb, existingDb, err := conversion.VerifyTargetDb(targetDb, projectID, instanceID, dbName)
 	if err != nil {
 		return err
 	}
@@ -66,10 +66,10 @@ func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnl
 		}
 	}
 
-	db, err := conversion.CreateOrUpdateDatabase(targetDb, projectID, instanceID, dbName, isNewDb, conv, ioHelper.Out)
+	db, err := conversion.CreateOrUpdateDatabase(targetDb, projectID, instanceID, dbName, existingDb, conv, ioHelper.Out)
 	if err != nil {
-		fmt.Printf("\nCan't create database: %v\n", err)
-		return fmt.Errorf("can't create database")
+		fmt.Printf("\nCan't create/update database: %v\n", err)
+		return fmt.Errorf("can't create/update database")
 	}
 
 	client, err := conversion.GetClient(db)
--- a/conversion/conversion.go
+++ b/conversion/conversion.go
@@ -428,27 +428,22 @@ func VerifyTargetDb(targetDb, project, instance, dbName string) (string, bool, e
 	dbPath := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName)
 	_, err = adminClient.GetDatabase(ctx, &adminpb.GetDatabaseRequest{Name: dbPath})
 	if err != nil {
-		// Differentiate between "db does not exist" and other errors? Assuming err means db does not exist for now
-		return targetDb, true, nil
+		if containsAny(strings.ToLower(err.Error()), []string{"database not found"}) {
+			return targetDb, false, nil
+		}
+		return "", false, fmt.Errorf("can't get database info: %s", err)
 	}
-	//if resp.DatabaseDialect == "POSTGRESQL" {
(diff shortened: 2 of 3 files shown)


conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+
+	dbPath := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName)
+	_, err = adminClient.GetDatabase(ctx, &adminpb.GetDatabaseRequest{Name: dbPath})
+	if err != nil {
+		// Differentiate between "db does not exist" and other errors? Assuming err means db does not exist for now
+		return targetDb, true, nil
+	}
+	//if resp.DatabaseDialect == "POSTGRESQL" {
+	//	targetDb = TARGET_EXPERIMENTAL_POSTGRES
+	//} else {
```
Deep1998, line 436: Commented for now since go protos are not public yet

conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+	adminClient, err := database.NewDatabaseAdminClient(ctx)
+	if err != nil {
+		return "", false, fmt.Errorf("can't create admin client: %w", analyzeError(err, project, instance))
+	}
+	defer adminClient.Close()
+
+	dbPath := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName)
+	_, err = adminClient.GetDatabase(ctx, &adminpb.GetDatabaseRequest{Name: dbPath})
+	if err != nil {
+		// Differentiate between "db does not exist" and other errors? Assuming err means db does not exist for now
```
Deep1998, line 431: Should there be checking on what type of error it is?

conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+	//}
+	dbDdl, err := adminClient.GetDatabaseDdl(ctx, &adminpb.GetDatabaseDdlRequest{Database: dbPath})
+	if err != nil {
+		fmt.Printf("\nCan't fetch database ddl: %v\n", err)
+		return "", false, fmt.Errorf("can't fetch database ddl")
+	}
+	if len(dbDdl.Statements) != 0 {
+		fmt.Printf("\nWriting to existing databases with non-empty ddl not supported\n")
+		return "", false, fmt.Errorf("Writing to existing databases with non-empty ddl not supported")
+	}
```
Deep1998, line 447: Note: When running is schema-only mode with a dbname that already exists with some DDL, we will end up throwing this error.

cmd/cmd.go
```diff
@@ -39,6 +39,10 @@ var (
 func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnly, schemaOnly, skipForeignKeys bool, schemaSampleSize int64, sessionJSON string, ioHelper *conversion.IOStreams, outputFilePrefix string, now time.Time) error {
 	var conv *internal.Conv
 	var err error
+	targetDb, isNewDb, err := conversion.VerifyTargetDb(targetDb, projectID, instanceID, dbName)
```
bharadwaj-aditya, line 42: why not move this block inside the createOrGetDatabase  ?

conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
 	return fcopy, n, nil
 }
 
+func VerifyTargetDb(targetDb, project, instance, dbName string) (string, bool, error) {
+	ctx := context.Background()
+	adminClient, err := database.NewDatabaseAdminClient(ctx)
+	if err != nil {
+		return "", false, fmt.Errorf("can't create admin client: %w", analyzeError(err, project, instance))
+	}
+	defer adminClient.Close()
```
agasheesh, line 426: you should do this before "if err != nil" else, the function will return before issuing the defer and thus before closing the adminClient

conversion/conversion.go
agasheesh: yes, it's a good practice in general for checking on specific error code and return a more specific error message to the end user

conversion/conversion.go
agasheesh: we should not check in commented out code and thus I would suggest dropping this (but this looks good otherwise when we are ready to go public)

conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+		return targetDb, true, nil
+	}
+	//if resp.DatabaseDialect == "POSTGRESQL" {
+	//	targetDb = TARGET_EXPERIMENTAL_POSTGRES
+	//} else {
+	//	targetDb = TARGET_SPANNER
+	//}
+	dbDdl, err := adminClient.GetDatabaseDdl(ctx, &adminpb.GetDatabaseDdlRequest{Database: dbPath})
+	if err != nil {
+		fmt.Printf("\nCan't fetch database ddl: %v\n", err)
```
agasheesh, line 441: We don't need to print here since the error return will get propagated and printed anyway. Similarly below.

conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+func CreateOrUpdateDatabase(targetDb, project, instance, dbName string, isNewDb bool, conv *internal.Conv, out *os.File) (string, error) {
+	ctx := context.Background()
+	adminClient, err := database.NewDatabaseAdminClient(ctx)
+	if err != nil {
+		return "", fmt.Errorf("can't create admin client: %w", analyzeError(err, project, instance))
+	}
+	defer adminClient.Close()
+
+	if isNewDb {
+		// Add support for dialect to create spangers dbs?
```
bharadwaj-aditya, line 460: the snippet of checking dialect etc from above needs to come here

conversion/conversion.go
agasheesh: I think it's fair to call out when customers point to an existing Spanner database even in schema-only mode. I am though surprised that we allow users to pass in targetDb name in schema_only mode?

conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+		dbPath, err := CreateDatabase(project, instance, dbName, conv, out)
```
bharadwaj-aditya, line 461: if a database already exists, this will throw an error

conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+	//	targetDb = TARGET_SPANNER
+	//}
+	dbDdl, err := adminClient.GetDatabaseDdl(ctx, &adminpb.GetDatabaseDdlRequest{Database: dbPath})
+	if err != nil {
+		fmt.Printf("\nCan't fetch database ddl: %v\n", err)
+		return "", false, fmt.Errorf("can't fetch database ddl")
+	}
+	if len(dbDdl.Statements) != 0 {
+		fmt.Printf("\nWriting to existing databases with non-empty ddl not supported\n")
+		return "", false, fmt.Errorf("Writing to existing databases with non-empty ddl not supported")
```
agasheesh, line 446: Consider dropping the double-negation and write this as "HarbourBridge only supports writing to existing databases with an empty schema."

conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+	if isNewDb {
+		// Add support for dialect to create spangers dbs?
+		dbPath, err := CreateDatabase(project, instance, dbName, conv, out)
+		if err != nil {
+			fmt.Printf("\nCan't create database: %v\n", err)
+			return "", fmt.Errorf("can't create database")
+		}
+		return dbPath, nil
+	}
+	dbPath := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName)
```
bharadwaj-aditya, line 468: adda a todo here, to handle scenarios where database is not new, but has some schema/data. We need to take a call at this point around how to proceed

conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+	return targetDb, false, nil
+}
+
+func CreateOrUpdateDatabase(targetDb, project, instance, dbName string, isNewDb bool, conv *internal.Conv, out *os.File) (string, error) {
+	ctx := context.Background()
+	adminClient, err := database.NewDatabaseAdminClient(ctx)
+	if err != nil {
+		return "", fmt.Errorf("can't create admin client: %w", analyzeError(err, project, instance))
+	}
+	defer adminClient.Close()
```
agasheesh, line 457: It's a good practice to create context and adminClient only once i.e., in CommandLine function and pass the same around to other functions.

conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+
+	if isNewDb {
```
agasheesh, line 459: instead of isNewDb, consider renaming as `existingDb`

conversion/conversion.go
bharadwaj-aditya: This problem is a little complete because there are a lot of scenarios that need to be handled around various types and stages of migration

bharadwaj-aditya (changes requested):

conversion/conversion.go
```diff
@@ -417,6 +417,70 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+		// Add support for dialect to create spangers dbs?
+		dbPath, err := CreateDatabase(project, instance, dbName, conv, out)
+		if err != nil {
+			fmt.Printf("\nCan't create database: %v\n", err)
+			return "", fmt.Errorf("can't create database")
+		}
+		return dbPath, nil
```
agasheesh, line 466: How about we create a new function "func UpdateDatabase()" similar to CreateDatabase below and we call one of them in this function i.e., CreateOrUpdate - it seems cleaner that way, but open to different ways of imlpementing this.

cmd/cmd.go
Deep1998: I put this check before so that for existing DBs, we do the schema conversion with the correct TargetDB value. Moving it there would mean identifying targetDB after schema conv is completed. Also for cases, when user has a DB with some DDL in it, I feel it is better to fail fast rather than after schema conv is done.

conversion/conversion.go
Deep1998: Yes se do allow passing of the DBname in schema-only currently, I personally pass it with schema-only mode for more readable file names compared to pgdump-<date>-<hash>

conversion/conversion.go
Deep1998: But in case an error occurs, won't the adminClient not get created in the first place so no need to close it?

conversion/conversion.go
```diff
@@ -417,6 +417,58 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+	if err != nil {
+		return "", false, fmt.Errorf("can't fetch database ddl")
+	}
+	if len(dbDdl.Statements) != 0 {
+		return "", false, fmt.Errorf("HarbourBridge supports writing to existing databases only if they have an empty schema.")
+	}
+	return targetDb, true, nil
+}
+
+func CreateOrUpdateDatabase(targetDb, project, instance, dbName string, existingDb bool, conv *internal.Conv, out *os.File) (string, error) {
```
agasheesh, line 446: This function also doesn't need targetDb in this PR.

agasheesh (commented):
Consider creating an adminClient at the top level and pass the same around to reduce the boilerplate code.

cmd/cmd.go
```diff
@@ -62,10 +66,10 @@ func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnl
 		}
 	}
 
-	db, err := conversion.CreateDatabase(projectID, instanceID, dbName, conv, ioHelper.Out)
+	db, err := conversion.CreateOrUpdateDatabase(targetDb, projectID, instanceID, dbName, existingDb, conv, ioHelper.Out)
 	if err != nil {
-		fmt.Printf("\nCan't create database: %v\n", err)
-		return fmt.Errorf("can't create database")
+		fmt.Printf("\nCan't create/update database: %v\n", err)
+		return fmt.Errorf("can't create/update database")
```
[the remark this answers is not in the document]
bharadwaj-aditya: Agreed on this. The typical way to handle this is to add debug/info logs at lower levels in the stack and print all the error/panic logs at the same level throughout the code

conversion/conversion.go
```diff
@@ -417,6 +417,58 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+		return "", false, fmt.Errorf("can't create admin client: %w", analyzeError(err, project, instance))
+	}
+	defer adminClient.Close()
+
+	dbPath := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName)
+	_, err = adminClient.GetDatabase(ctx, &adminpb.GetDatabaseRequest{Name: dbPath})
+	if err != nil {
+		if containsAny(strings.ToLower(err.Error()), []string{"database not found"}) {
+			return targetDb, false, nil
+		}
```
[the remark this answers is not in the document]
agasheesh: Resolved per offline conversation. Plan here is to cleanup error code handling in a follow up PR.

cmd/cmd.go
```diff
@@ -62,27 +66,23 @@ func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnl
 		}
 	}
 
-	db, err := conversion.CreateDatabase(projectID, instanceID, dbName, conv, ioHelper.Out)
+	db, err := conversion.CreateOrUpdateDatabase(projectID, instanceID, dbName, dbExists, conv, ioHelper.Out)
```
agasheesh, line 69: nit; maybe replace db with dbURI for consistency?

cmd/cmd.go
```diff
@@ -62,27 +66,23 @@ func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnl
...
 	if err != nil {
-		fmt.Printf("\nCan't finish data conversion for db %s: %v\n", db, err)
-		return fmt.Errorf("can't finish data conversion")
+		return fmt.Errorf("can't finish data conversion for db %s: %v", db, err)
 	}
 	if !skipForeignKeys {
 		if err = conversion.UpdateDDLForeignKeys(projectID, instanceID, dbName, conv, ioHelper.Out); err != nil {
-			fmt.Printf("\nCan't perform update operation on db %s with foreign keys: %v\n", db, err)
-			return fmt.Errorf("can't perform update schema with foreign keys")
+			return fmt.Errorf("can't perform update operation on db %s with foreign keys: %v", db, err)
```
agasheesh, line 85: nit; I actually liked 'update schema' better, how about "can't perform update schema on db %s with foreign keys: %v"

conversion/conversion.go
```diff
@@ -417,6 +417,76 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
 	return fcopy, n, nil
 }
 
+// VerifyDb checks whether the db exists and if it does, verifies if the schema is what we currently support.
+func VerifyDb(project, instance, dbName string) (dbExists bool, err error) {
+	ctx := context.Background()
```
agasheesh, line 422: Consider cleaning up some of this in a follow up PR as following: 
1. Create dbURI in cmd.CommandLine and pass that around to all these helper functions instead of passing project, instance, dbName parameters (unless a specific function requires these parameters separately); 
2. Pass ctx and adminClient from cmd.CommandLine to these functions

conversion/conversion.go
```diff
@@ -417,6 +417,76 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+
+	if dbExists {
+		dbPath, err := UpdateDatabase(project, instance, dbName, conv, out)
+		if err != nil {
+			fmt.Printf("\nCan't update database schema: %v\n", err)
+			return "", fmt.Errorf("can't update database schema")
+		}
+		return dbPath, nil
+	} else {
+		// Add support for dialect to create spangers dbs?
```
agasheesh, line 480: we are trying to avoid use of spangres keyword publicly in public; consider dropping this comment or replace spangres with PG :)

conversion/conversion.go
```diff
@@ -448,6 +518,31 @@ func CreateDatabase(project, instance, dbName string, conv *internal.Conv, out *
...
+func UpdateDatabase(project, instance, dbName string, conv *internal.Conv, out *os.File) (string, error) {
+	fmt.Fprintf(out, "Updating schema for database %s in instance %s ... ", dbName, instance)
+	ctx := context.Background()
+	adminClient, err := database.NewDatabaseAdminClient(ctx)
+	if err != nil {
+		return "", fmt.Errorf("can't create admin client: %w", analyzeError(err, project, instance))
+	}
+	defer adminClient.Close()
+
+	dbPath := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName)
```
agasheesh, line 531: nit; s/dbPath/dbURI for consistency

conversion/conversion.go
```diff
@@ -448,6 +518,31 @@ func CreateDatabase(project, instance, dbName string, conv *internal.Conv, out *
...
+	op, err := adminClient.UpdateDatabaseDdl(ctx, &adminpb.UpdateDatabaseDdlRequest{
+		Database:   dbPath,
+		Statements: conv.SpSchema.GetDDL(ddl.Config{Comments: false, ProtectIds: true, Tables: true, ForeignKeys: false}),
+	})
+	if err != nil {
+		return "", fmt.Errorf("Can't submit update schema request: %s", err)
```
agasheesh, line 537: nit; s/Can't/can't for consistency

testing/conversion/conversion_test.go
```diff
@@ -94,8 +95,11 @@ func dropDatabase(t *testing.T, dbPath string) {
 	}
 }
 
-func BuildConv(t *testing.T, numCols, numFks int) *internal.Conv {
+func BuildConv(t *testing.T, numCols, numFks int, makeEmpty bool) *internal.Conv {
 	conv := internal.MakeConv()
```
agasheesh, line 99: nit; maybe just create another helper function as following (but feel free to ignore this comment if you think that would add more distraction):

func BuildEmptyConv() *internal.Conv {
       return internal.MakeConv()
}

agasheesh (approved):
Thanks for all the changes. Looks really good. Just a few minor comments.

testing/conversion/conversion_test.go
Deep1998: I put it in the same function instead of a separate one because when calling the BuildConv() from test cases, I dont need to make 2 flows, one for empty conv and one for non-empty conv.

Ex:
`for _, tc := range testCases {
		dbURI, err := conversion.CreateDatabase(projectID, instanceID, tc.dbName, BuildConv(t, 2, 0, tc.emptySchema), os.Stdout)
		`
With 2 separate functions, I need to add an if statement and do some branching here.

bharadwaj-aditya (approved):
LGTM

bharadwaj-aditya:
Left a couple of minor comments. Please have a look. But looks fine overall

PR #168, opened 2021-09-15
Handle name collisions differing only in case

Deep1998:
Store and compare all names in the usedNames map in lower case. 

MySQL allows table names with that differ only in case which is not allowed in spanner. That will be picked up in a separate PR since the table name collision function is used in many places.

### DIFF
--- a/dynamodb/toddl.go
+++ b/dynamodb/toddl.go
@@ -28,15 +28,6 @@ import (
 // Spanner. It uses the source schema in conv.SrcSchema, and writes
 // the Spanner schema to conv.SpSchema.
 func schemaToDDL(conv *internal.Conv) error {
-	// We need to pre-populate conv.UsedNames with Spanner table
-	// names to handle collision with index names.
-	for _, srcTable := range conv.SrcSchema {
-		_, err := internal.GetSpannerTable(conv, srcTable.Name)
-		if err != nil {
-			conv.Unexpected(fmt.Sprintf("Couldn't map source table %s to Spanner: %s", srcTable.Name, err))
-			continue
-		}
-	}
 	for _, srcTable := range conv.SrcSchema {
 		spTableName, err := internal.GetSpannerTable(conv, srcTable.Name)
 		if err != nil {
--- a/dynamodb/toddl.go
+++ b/dynamodb/toddl.go
@@ -28,12 +28,8 @@ import (
 // Spanner. It uses the source schema in conv.SrcSchema, and writes
 // the Spanner schema to conv.SpSchema.
 func schemaToDDL(conv *internal.Conv) error {
-	// conv.UsedNames tracks Spanner names that have been used for table names, foreign key constraints
-	// and indexes. We use this to ensure we generate unique names when
-	// we map from MySQL to Spanner since Spanner requires all these names to be
-	// distinct and should not differ only in case.
 	// We need to pre-populate conv.UsedNames with Spanner table
-	// names to handle collision with foreign key names and index names.
+	// names to handle collision with index names.
 	for _, srcTable := range conv.SrcSchema {
 		_, err := internal.GetSpannerTable(conv, srcTable.Name)
 		if err != nil {
--- a/dynamodb/toddl.go
+++ b/dynamodb/toddl.go
@@ -17,6 +17,7 @@ package dynamodb
 import (
 	"fmt"
 	"strconv"
+	"strings"
 	"unicode"
 
 	"github.com/cloudspannerecosystem/harbourbridge/internal"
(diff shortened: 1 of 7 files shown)


agasheesh:
Thanks for catching this and sending out a fix. Can you please also add test cases that would have failed without this fix? Such that we can catch any future regression in the behavior.

Deep1998:
Modified dump files so integration test covers this case

agasheesh:
The changes in this LGTM. I think it's worth adding changes for handling name collision for table in this PR itself so that we can reason about name collisions across table, index and fk.

agasheesh (commented):
Thanks for the offline discussion about various approaches on how handle name collisions for table names differing only in capitalization. Approach to make usedNames part of the conv struct itself and using the same consistently throughout the codebase sounds good to me.

dynamodb/toddl.go
```diff
@@ -28,18 +28,18 @@ import (
...
-	// we map from DynamoDB to Spanner since we want Spanner table names and index names to be distinct.
-	usedNames := make(map[string]bool)
-	// We need to pre-populate usedNames with Spanner table names to handle collisions
-	// with table names, foreign key names and index names.
+	// conv.UsedNames tracks Spanner names that have been used for table names, foreign key constraints
+	// and indexes. We use this to ensure we generate unique names when
+	// we map from MySQL to Spanner since Spanner requires all these names to be
+	// distinct and should not differ only in case.
+	// We need to pre-populate conv.UsedNames with Spanner table
+	// names to handle collision with foreign key names and index names.
```
agasheesh, line 36: I wonder if this comment makes sense here anymore? How name collisions are handled and usedNames are tracked are transparent to this function, and that seem to be by design. 

I would expect readers to look at mapping.go to understand how userNames are handled. Thus, suggest that we move some of the text there.

test_data/mysqldump.test.out
```diff
@@ -41,6 +43,16 @@ INSERT INTO `cart` VALUES ('901e-a6cfc2b502dc','abc-123',1,'2020-07-20 05:10:26'
 /*!40000 ALTER TABLE `cart` ENABLE KEYS */;
 UNLOCK TABLES;
 
+DROP TABLE IF EXISTS `CART`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `CART` (
```
agasheesh, line 49: I think it's worth adding a comment saying that this Table and IDX index are purposefully added with the same name differing only in cases.

agasheesh (approved):

dynamodb/toddl.go
```diff
@@ -28,18 +28,14 @@ import (
...
-	// Tracks Spanner names that have been used for indexes. We use this to ensure we generate unique names when
-	// we map from DynamoDB to Spanner since we want Spanner table names and index names to be distinct.
-	usedNames := make(map[string]bool)
-	// We need to pre-populate usedNames with Spanner table names to handle collisions
-	// with table names, foreign key names and index names.
+	// We need to pre-populate conv.UsedNames with Spanner table
+	// names to handle collision with index names.
 	for _, srcTable := range conv.SrcSchema {
-		spTableName, err := internal.GetSpannerTable(conv, srcTable.Name)
+		_, err := internal.GetSpannerTable(conv, srcTable.Name)
```
bharadwaj-aditya, line 34: This block is the same as the one below (Line 41), can it be removed ?

internal/mapping.go
```diff
@@ -30,23 +30,29 @@ import (
 // a) the new table name is legal
 // b) the new table name doesn't clash with other Spanner table names
 // c) we consistently return the same name for this table.
+//
+// conv.UsedNames tracks Spanner names that have been used for table names, foreign key constraints
+// and indexes. We use this to ensure we generate unique names when
+// we map from MySQL to Spanner since Spanner requires all these names to be
```
bharadwaj-aditya, line 36: Change from "we map from MySQL ... " to "we map from source databases .... not differ only in case" This behaviour is not enforced in all source databases, an example of such is Mysql"

internal/mapping.go
```diff
@@ -30,23 +30,29 @@ import (
...
 	if srcTable == "" {
 		return "", fmt.Errorf("bad parameter: table string is empty")
 	}
+	// Once computed, return cached result.
 	if sp, found := conv.ToSpanner[srcTable]; found {
 		return sp.Name, nil
 	}
 	spTable, _ := FixName(srcTable)
-	if _, found := conv.ToSource[spTable]; found {
+	if _, found := conv.UsedNames[strings.ToLower(spTable)]; found {
```
bharadwaj-aditya, line 47: can we re-use the getSpannerId method here ?

bharadwaj-aditya (commented):
Largely looks ok, But can use a bit of code cleanup

internal/mapping.go
Deep1998: Not directly, but moved the common code to a new function

bharadwaj-aditya (approved):
LGTM

PR #169, opened 2021-09-16
This should fail

Deep1998:


Deep1998:
Used to test that the change actually fails the integration test. Index names only differ in case.

PR #170, opened 2021-09-16
Refactored context and database client

Deep1998:
Changed parameters to pass context and database client rather than generating a new one each time.

Also modified parameters to use dbURI wherever possible rather than constructing a new one using project, instance and dbname.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -16,9 +16,11 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"time"
 
+	database "cloud.google.com/go/spanner/admin/database/apiv1"
 	"github.com/cloudspannerecosystem/harbourbridge/conversion"
 	"github.com/cloudspannerecosystem/harbourbridge/internal"
 )
@@ -36,10 +38,9 @@ var (
 // 2. Create database (if schemaOnly is set to false)
 // 3. Run data conversion (if schemaOnly is set to false)
 // 4. Generate report
-func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnly, schemaOnly, skipForeignKeys bool, schemaSampleSize int64, sessionJSON string, ioHelper *conversion.IOStreams, outputFilePrefix string, now time.Time) error {
+func CommandLine(ctx context.Context, driver, targetDb, dbURI string, dataOnly, schemaOnly, skipForeignKeys bool, schemaSampleSize int64, sessionJSON string, ioHelper *conversion.IOStreams, outputFilePrefix string, now time.Time) error {
 	var conv *internal.Conv
 	var err error
-
 	if !dataOnly {
 		conv, err = conversion.SchemaConv(driver, targetDb, ioHelper, schemaSampleSize)
 		if err != nil {
@@ -62,13 +63,17 @@ func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnl
 			return err
 		}
 	}
-
-	dbURI, err := conversion.CreateOrUpdateDatabase(projectID, instanceID, dbName, conv, ioHelper.Out)
+	adminClient, err := database.NewDatabaseAdminClient(ctx)
+	if err != nil {
+		return fmt.Errorf("can't create admin client: %w", conversion.AnalyzeError(err, dbURI))
+	}
+	defer adminClient.Close()
+	err = conversion.CreateOrUpdateDatabase(ctx, adminClient, dbURI, conv, ioHelper.Out)
 	if err != nil {
 		return fmt.Errorf("can't create/update database: %v", err)
 	}
 
-	client, err := conversion.GetClient(dbURI)
+	client, err := conversion.GetClient(ctx, dbURI)
 	if err != nil {
 		return fmt.Errorf("can't create client for db %s: %v", dbURI, err)
 	}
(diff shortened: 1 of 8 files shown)


agasheesh:
I will wait for https://github.com/cloudspannerecosystem/harbourbridge/pull/163 to be re-submitted with fixes, before reviewing this PR since the diff is kind of incorrect at right now.

cmd/cmd.go
```diff
@@ -66,12 +76,12 @@ func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnl
 		}
 	}
 
-	dbURI, err := conversion.CreateOrUpdateDatabase(projectID, instanceID, dbName, dbExists, conv, ioHelper.Out)
+	err = conversion.CreateOrUpdateDatabase(ctx, adminClient, projectID, instanceID, dbName, dbExists, conv, ioHelper.Out)
```
agasheesh, line 79: We should also clean up passing of projectID, instanceID and dbName as part of this cleanup by precomputing dbURI at the top level once instance flag is validated and passing the same around.

Deep1998: I cleaned up project and instance everywhere except this function and the ones this calls because this eventually calls analyzeError() which takes the instance and project and input to print a clean error message. I was wondering what we should do with that if we decide to remove that from here?

conversion/conversion.go
```diff
@@ -543,33 +519,38 @@ func UpdateDatabase(project, instance, dbName string, conv *internal.Conv, out *
...
 
-// UpdateDDLForeignKeys updates the Spanner database with foreign key
-// constraints using ALTER TABLE statements.
-func UpdateDDLForeignKeys(project, instance, dbName string, conv *internal.Conv, out *os.File) error {
-	ctx := context.Background()
-	adminClient, err := NewDatabaseAdminClient(ctx)
-	if err != nil {
-		return fmt.Errorf("can't create admin client: %w", analyzeError(err, project, instance))
+func splitDbURI(dbURI string) (project, instance, dbName string, err error) {
+	splitURI := strings.Split(dbURI, "/")
```
Deep1998, line 532: Wanted to use fmt.SScanf instead but keep getting `Unexpected EOF`.

cmd/cmd.go
```diff
@@ -36,10 +38,9 @@ var (
 // 2. Create database (if schemaOnly is set to false)
 // 3. Run data conversion (if schemaOnly is set to false)
 // 4. Generate report
-func CommandLine(driver, targetDb, projectID, instanceID, dbName string, dataOnly, schemaOnly, skipForeignKeys bool, schemaSampleSize int64, sessionJSON string, ioHelper *conversion.IOStreams, outputFilePrefix string, now time.Time) error {
+func CommandLine(ctx context.Context, driver, targetDb, projectID, instanceID, dbName string, dataOnly, schemaOnly, skipForeignKeys bool, schemaSampleSize int64, sessionJSON string, ioHelper *conversion.IOStreams, outputFilePrefix string, now time.Time) error {
```
agasheesh, line 41: I think we can go even further and pass dbURI to CommandLine() itself. The guiding principle is that dbURI is supposed to be a validated URI i.e., we should have already verified that projectID, instanceID are valid by the time CommandLine is called. The way we can make that work is by changing `conversion.AnalyzeError` to also take dbURI as the parameter.

That said, this PR already does a large amount of cleanup and we should get this submitted and move forward. The other cleanup that I have suggested can be easily done as a follow up.

conversion/conversion.go
agasheesh: Using `strings.Split` is the typical way to go about this, and thus looks fine to me.

conversion/conversion.go
```diff
@@ -543,33 +519,38 @@ func UpdateDatabase(project, instance, dbName string, conv *internal.Conv, out *
...
+	if len(splitURI) != 6 {
+		return "", "", "", fmt.Errorf("cannot split bad db URI: %s", dbURI)
+	}
```
agasheesh, line 535: I think a better approach might be to restructure this function as following:

parseDbURI(dbURI string) (projectId, instanceId, dbName string) {
    split := strings.split("/databases/")
    instanceURI, dbName := split[0], split[1]
    projectId, instanceId := parseInstanceURI(instanceURI)
}

parseInstanceURI() (...) {
   ...
}

agasheesh (approved):

cmd/cmd.go
agasheesh: As discussed in other comment and offline, we can use the splitDbURI (or parseDbURI) in analyzeError() to split passed in dbURI into component projectId, instanceId and dbName etc.

conversion/conversion.go
```diff
@@ -543,33 +519,38 @@ func UpdateDatabase(project, instance, dbName string, conv *internal.Conv, out *
...
 }
 
-// UpdateDDLForeignKeys updates the Spanner database with foreign key
-// constraints using ALTER TABLE statements.
-func UpdateDDLForeignKeys(project, instance, dbName string, conv *internal.Conv, out *os.File) error {
-	ctx := context.Background()
-	adminClient, err := NewDatabaseAdminClient(ctx)
-	if err != nil {
-		return fmt.Errorf("can't create admin client: %w", analyzeError(err, project, instance))
+func splitDbURI(dbURI string) (project, instance, dbName string, err error) {
```
agasheesh, line 531: I don't see a lot of value in returning error from this function. We create dbURI internally in HB itself and thus it's unlikely that it will be incorrect except for an explicit bug in HB codebase such that the same should be caught in unit tests and/or integration tests.

Deep1998: I added some error checking here to catch errors when project/instance is "" which has come up quite a few times but the correct error is not returned. But error checking like that should be moved to main.go where we fetch these values in the first place so removing it from here.

PR #172, opened 2021-09-17
Upgrade pg_query_go library to add handling for unexpected parse errors

agasheesh:
HarbourBridge has been so far relying on a long deprecated fork of pg_query_go library. Our customers have reported crashes while parsing pg_dump files in the past and the root cause for most of them (where available) was being unable to handle unexpected Nodes in the AST generated. Thus, this PR upgrades the library to it's recommended version and adds handling for unexpected conditions to be reported at the end of report.txt.

Verified for a pg_dump file shared by a customer that was crashing for 28 different Create Index statements, that the same can be now parsed using HB and all of those 28 Create Index statement parse errors are reported as Unexpected conditions in report.txt.


nevinheintze (commented):
Had a quick look -- looks good!
Thanks for catching this: switching over to github.com/pganalyze and picking up the latest version makes a huge amount of sense.
The old github.com/lfittl/pg_query_go was getting stale. Glad to see the successor library is getting attention/updates now.

Let me know when this is ready for a full review.

agasheesh:
Hi @nevinheintze - this is ready for a full review.

postgres/pgdump_test.go
```diff
@@ -783,7 +783,7 @@ func TestProcessPgDump_WithUnparsableContent(t *testing.T) {
 	if err == nil {
 		t.Fatalf("Expect an error, but got nil")
 	}
-	if !strings.Contains(err.Error(), "Error parsing") {
+	if !strings.Contains(err.Error(), "error parsing") {
```
Deep1998, line 786: Is it more robust to do `!strings.Contains(strings.ToLower(err.Error()), "error parsing")`

postgres/pgdump.go
```diff
@@ -699,45 +707,54 @@ func getCols(conv *internal.Conv, table string, l []nodes.Node) (cols []string,
...
 }
 
-func getString(node nodes.Node) (string, error) {
-	switch n := node.(type) {
-	case nodes.String:
-		return n.Str, nil
+func getString(node *pg_query.Node) (string, error) {
+	switch n := node.GetNode().(type) {
+	case *pg_query.Node_String_:
+		return trimQuote(strings.TrimPrefix(n.String_.String(), "str:")), nil
```
Deep1998, line 748: Is it a good idea to replace this with trimString()?

Deep1998 (commented):
Looks very good! Had a few minor comments.

postgres/pgdump_test.go
agasheesh: Good point. Done @ https://github.com/cloudspannerecosystem/harbourbridge/pull/172/commits/3701d42b42a0409b2981d46d8efa2c4f2ef60ba4.

postgres/pgdump.go
agasheesh: Done @ https://github.com/cloudspannerecosystem/harbourbridge/pull/172/commits/3701d42b42a0409b2981d46d8efa2c4f2ef60ba4.

.gitignore
```diff
@@ -1 +1,3 @@
 .idea/
+main
```
nevinheintze, line 2: Just curious on the background for adding these?

postgres/pgdump.go
```diff
@@ -259,84 +256,84 @@ func processAlterTableStmt(conv *internal.Conv, n nodes.AlterTableStmt) {
...
 				logStmtError(conv, n, err)
 				return
 			}
 			colNames = append(colNames, name)
 			colDef[name] = col
 			constraints = append(constraints, cdConstraints...)
-		case nodes.Constraint:
+		case *pg_query.Node_Constraint:
 			// Note: there should be at most one Constraint node in
 			// n.TableElts.Items. We don't check this. We just keep
```
nevinheintze, line 297: Should this be just "n.TableElts" now?
i.e. drop the .Items?

postgres/pgdump.go
```diff
@@ -389,87 +386,88 @@ func processCopyStmt(conv *internal.Conv, n nodes.CopyStmt) *copyOrInsert {
...
-func processVariableSetStmt(conv *internal.Conv, n nodes.VariableSetStmt) {
-	if n.Name != nil && *n.Name == "timezone" {
-		if len(n.Args.Items) == 1 {
-			switch c := n.Args.Items[0].(type) {
-			case nodes.A_Const:
-				tz, err := getString(c.Val)
+func processVariableSetStmt(conv *internal.Conv, n *pg_query.VariableSetStmt) {
+	if n.Name == "timezone" {
+		if len(n.Args) == 1 {
+			node := n.Args[0]
```
nevinheintze, line 409: Maybe arg instead of node?
node is very generic; arg gives some sense of what this represents.
Maybe even arg0 (but that's probably too long for golang var names!)

postgres/pgdump.go
```diff
@@ -684,57 +689,62 @@ func toForeignKeys(fk constraint) (fkey schema.ForeignKey) {
...
-				switch st := c.Val.(type) {
-				case nodes.String:
-					values = append(values, st.Str)
-				case nodes.Integer:
-					// For uniformity, convert to string and handle everything in
-					// dataConversion(). If performance of insert statements becomes a
-					// high priority (it isn't right now), then consider preserving int64
-					// here to avoid the int64 -> string -> int64 conversions.
-					values = append(values, strconv.FormatInt(st.Ival, 10))
+func getVals(conv *internal.Conv, vll []*pg_query.Node, n *pg_query.InsertStmt) (values []string) {
```
nevinheintze, line 707: I'm pretty sure the original code here is just broken (and the new code preserves the bug).

In short, getVals should actually return [][]string.

Ok, the long details ...

In pgdump, the default way to write out data to the dump file is to use COPY-FROM. However, you can also get pgdump to generate INSERT statements. The very first version of harbourbridge used the insert version (it was kind of easier to start with). Now, when pgdump generates INSERTS, it only generates them with one row at a time i.e.

INSERT INTO public.products (product_id, description, price, date_added) VALUES ('abc-123', 'Blue suede shoes', '141.99', '2020-06-06');

And that was all that I implemented at the time.
However, PostgreSQL supports multi-row inserts i.e.:

INSERT INTO public.products (product_id, description, price, date_added) VALUES
('abc-123', 'Blue suede shoes', '141.99', '2020-06-06'),
('axd-673', 'Antique typewriter', '99.99', '2020-06-07'),
('zxi-631', 'Glass vase', '55.50', '2020-06-10');

These get completely mangled by getVals! You get something like:

vals = ["abc-123" "Blue suede shoes" "141.99" "2020-06-06" "axd-673" "Antique typewriter" "99.99" "2020-06-07" "zxi-631" "Glass vase" "55.50" "2020-06-10"]

i.e. all of the values from all of the rows concatenated together!

To fix this, we should:
a) return [][]string from getVals
b) change copyOrInsert to handle [][]string for vals (maybe we should call it rows?)
c) change ProcessPgDump to handle multi-row inserts i.e. iterate over the [][]string and call ProcessDataRow on each row.
d) add a test for multi-row INSERT statements (in fact, this should probably be the first step!)

I suggest we put a TODO for this in the code and defer for another PR. I'll file an issue for this to capture the state.

nevinheintze (approved):
Thanks Asheesh -- looks good.
Switching over to github.com/pganalyze/pg_query_go/v2 is critical (we need to do it and it resolves our parsing reliability issue), and thanks for all of the cleanup as well.

Added a few very minor comments.
Let's defer the getVals issue to another PR.

postgres/pgdump.go
nevinheintze: I've added issue #176 for this.

.gitignore
agasheesh: I sometimes do a `go build main.go` which results in a `main` binary being added to the git diff. On the other hand, `web/harbour_bridge_output/` is generated when you run the frontend tests locally (we should fix the tests to cleanup though @Deep1998 FYI).

postgres/pgdump.go
agasheesh: Thanks for the detailed history, I was also kind of wondering about the same when I looked at this code but didn't change functionality in this PR. Doing this in a follow up with appropriate testing makes sense to me and I have assigned the github issue to myself. Added TODO. Done @https://github.com/cloudspannerecosystem/harbourbridge/pull/172/commits/3044120ca85a3b4386f12ca5a583fe4436cd0c54.

postgres/pgdump.go
agasheesh: Using `arg` makes sense since we check above for `len(n.Args) == 1`. Done @https://github.com/cloudspannerecosystem/harbourbridge/pull/172/commits/3044120ca85a3b4386f12ca5a583fe4436cd0c54.

postgres/pgdump.go
agasheesh: Done @https://github.com/cloudspannerecosystem/harbourbridge/pull/172/commits/3044120ca85a3b4386f12ca5a583fe4436cd0c54.

Deep1998 (approved):
Thanks for this change, looks great! This should fix a lot of the parsing issues we keep getting from customers. And thanks for pointing out the test output file cleanup, that is something we should take up next (maybe parallely with other PRs since it should be a small change)

agasheesh:
Thanks for quick and thorough reviews!

PR #175, opened 2021-09-21
Add subcommand for schema migration

agasheesh:
This PR adds a new subcommand for doing schema migration which can be run as following:

```
$ go run github.com/cloudspannerecosystem/harbourbridge schema < examples/cart.pg_dump
```

which is equivalent to

```
$ go run github.com/cloudspannerecosystem/harbourbridge -schema-only < examples/cart.pg_dump
```

For the time being, we will support both ways of executing this command. This has a slight downside that the top-level flags cannot be support for subcommands right now. This is since it's not possible to differentiate between global mode (`$ harbourbridge -v -schema-only) ...` and subcommand mode (`$ harbourbridge -v schema` ...) when passing a global flag `v` in this example.

Here's a full list of all the newly introduced subcommand subcommands in addition to schema subcommand.

(listing all the subcommands)
```
$ go run github.com/cloudspannerecosystem/harbourbridge commands
help
commands
schema
```

(help page for all the subcommands)
```
$ go run github.com/cloudspannerecosystem/harbourbridge help
Usage: harbourbridge <flags> <subcommand> <subcommand args>

Subcommands:
	commands         list all command names
	help                     describe subcommands and their syntax
	schema               generate schema for target db from source db schema
```

(detailed help for schema subcommand)
```
$ go run github.com/cloudspannerecosystem/harbourbridge schema -help
harbourbridge schema [file]...

Convert schema for source db specified by driver. Source db dump files can be
specified by either dump-file flag or from stdin. Connection profile for source
databases in direct connect mode can be specified by setting appropriate
environment variables. The schema flags are:
  -driver string
    	Flag for specifying source DB or dump files (accepted values are "pg_dump", "postgres", "mysqldump", "mysql", and "dynamodb") (default "pg_dump")
  -dump-file string
    	Location of dump file to process
  -prefix string
    	File prefix for generated files
  -prompt
    	Prompt before executing
  -schema-sample-size int
    	Number of rows to use for inferring schema (only for DynamoDB) (default 100000)
  -target-db string
    	Specifies the target DB, defaults to spanner
```


main.go
```diff
@@ -75,6 +78,27 @@ Sample usage:
...
+	if err != nil {
+		fmt.Printf("\nCan't set up log file: %v\n", err)
+		panic(fmt.Errorf("can't set up log file"))
+	}
+	defer conversion.Close(lf)
+
+	// TODO: Remove this check and always run HB in subcommands mode once
+	// global command line mode is deprecated. We can also enable support for
+	// top-level flags in subcommand then.
+	if os.Args[1] != "" && !strings.HasPrefix(os.Args[1], "-") {
```
Deep1998, line 92: For `go run main.go schema ` we would need os.Args[3] right? Do we only plan on supporting it for built binaries?

testing/postgres/integration_test.go
```diff
@@ -165,6 +165,29 @@ func TestIntegration_PGDUMP_Command(t *testing.T) {
...
 
+func TestIntegration_PGDUMP_SchemaCommand(t *testing.T) {
+	t.Parallel()
+
+	tmpdir := prepareIntegrationTest(t)
+	defer os.RemoveAll(tmpdir)
+
+	dataFilepath := "../../test_data/pg_dump.test.out"
+	// Be aware that when testing with the command, the time `now` might be
+	// different between file prefixes and the contents in the files. This
```
Deep1998, line 176: Offtopic, but as part of integration tests, should we also check if the 4 files are written during the process. We can specify the dbName param in these commands to avoid the name having a random hash and check if the files exist.

cli/schema.go
```diff
@@ -0,0 +1,101 @@
...
+func (cmd *SchemaCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
+	input := loadInput(cmd.dumpFilePath)
+	ioHelper := &conversion.IOStreams{In: input, Out: os.Stdout}
+	if ioHelper.SeekableIn != nil {
+		defer ioHelper.In.Close()
+	}
+
+	// If filePrefix not explicitly set, use generated dbName.
+	if cmd.filePrefix == "" {
+		dbName, err := conversion.GetDatabaseName(cmd.driverName, time.Now())
```
Deep1998, line 67: Why don't we allow dbName to be overridden by the user?

cli/schema.go
```diff
@@ -0,0 +1,101 @@
...
+	now := time.Now()
+	conversion.WriteSchemaFile(conv, now, cmd.filePrefix+schemaFile, ioHelper.Out)
+	conversion.WriteSessionFile(conv, cmd.filePrefix+sessionFile, ioHelper.Out)
+	conversion.Report(cmd.driverName, nil, ioHelper.BytesRead, "", conv, cmd.filePrefix+reportFile, ioHelper.Out)
+	return subcommands.ExitSuccess
+}
+
+// Load the dump file if parameter has been passed by the user.
+// If no parameter has been passed, then read from standard input
+func loadInput(dumpFile string) *os.File {
```
Deep1998, line 90: This function is also in main.go. Should we move it to a common place and call it from both the places? Or we can take that in a separate PR too since the intern project is dependant on this.

cli/schema.go
```diff
@@ -0,0 +1,101 @@
...
+
+	now := time.Now()
+	conversion.WriteSchemaFile(conv, now, cmd.filePrefix+schemaFile, ioHelper.Out)
+	conversion.WriteSessionFile(conv, cmd.filePrefix+sessionFile, ioHelper.Out)
+	conversion.Report(cmd.driverName, nil, ioHelper.BytesRead, "", conv, cmd.filePrefix+reportFile, ioHelper.Out)
+	return subcommands.ExitSuccess
+}
+
+// Load the dump file if parameter has been passed by the user.
+// If no parameter has been passed, then read from standard input
```
Deep1998, line 89: Nit: missing full stop at the end of sentence.

cli/schema.go
```diff
@@ -0,0 +1,101 @@
...
+func (cmd *SchemaCmd) Synopsis() string {
+	return "generate schema for target db from source db schema"
+}
+
+// Usage returns usage info of the command.
+func (cmd *SchemaCmd) Usage() string {
+	return fmt.Sprintf(`%v schema [file]...
+
+Convert schema for source db specified by driver. Source db dump files can be
+specified by either dump-file flag or from stdin. Connection profile for source
```
Deep1998, line 42: nit: Should we say "source db dump **data**" or some other better phrasing as we end the sentence with "from stdin"? I know we use dumpfiles realistically in stdin but this way we stay consistent with what we actually implement in harbourbridge i.e. reading data from stdin.

cli/schema.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// SetFlags sets the flags.
+func (cmd *SchemaCmd) SetFlags(f *flag.FlagSet) {
+	f.BoolVar(&cmd.prompt, "prompt", false, "Prompt before executing")
+	f.StringVar(&cmd.filePrefix, "prefix", "", "File prefix for generated files")
+	f.StringVar(&cmd.driverName, "driver", "pg_dump", "Flag for specifying source DB or dump files (accepted values are \"pg_dump\", \"postgres\", \"mysqldump\", \"mysql\", and \"dynamodb\")")
+	f.Int64Var(&cmd.schemaSampleSize, "schema-sample-size", int64(100000), "Number of rows to use for inferring schema (only for DynamoDB)")
+	f.StringVar(&cmd.dumpFilePath, "dump-file", "", "Location of dump file to process")
+	f.StringVar(&cmd.targetDb, "target-db", "", "Specifies the target DB, defaults to spanner.")
```
Deep1998, line 55: nit. For consistency, should we make all descriptions not have full stop at the end.

cli/schema.go
```diff
@@ -0,0 +1,101 @@
...
+
+Convert schema for source db specified by driver. Source db dump files can be
+specified by either dump-file flag or from stdin. Connection profile for source
+databases in direct connect mode can be specified by setting appropriate
+environment variables. The schema flags are:
+`, path.Base(os.Args[0]))
+}
+
+// SetFlags sets the flags.
+func (cmd *SchemaCmd) SetFlags(f *flag.FlagSet) {
```
Deep1998, line 49: Why dont we also add the dbName flag here?

cli/schema.go
Deep1998, line 67: With the dbName flag, we should also override dbName here

Deep1998 (commented):
Overall looks very good! Have left a few minor comments/questions. Also, it looks like the prompt flag is not being used yet. Is it a future change?

cli/schema.go
agasheesh: We don't want users to have to specify Spanner database info i.e., instance, dbName, (or project_id as part of environment variables). Users will though still have ability to specify `prefix` to name the output files appropriately.

cli/schema.go
agasheesh: As discussed in the other comment, we don't want users to specify dbName flag as part of schema migration.

testing/postgres/integration_test.go
agasheesh: I agree that we should validate the output produced by schema subcommand as well as actually look at the schema produced to make sure it's what we expect. But to do that we will need to have a golden output for schema generated, to prevent any regression. How about we do this in a follow up PR? (We will though set `prefix` since dbName won't be available for schema subcommand).

cli/schema.go
agasheesh: Yes, flag descriptions in SetFlags in subcommand should not end with period. Thanks for catching, done @ https://github.com/cloudspannerecosystem/harbourbridge/pull/175/commits/be4ddfc12ff3c3c86abfdaa613cfa9f0d1fc2a12.

cli/schema.go
agasheesh: Even from stdin, we read dump file only. Dump file itself can contain either/or/both of schema and data statements. I have though rephrased the statement now as "Source db dump file can be specified by either dump-file flag or piped to stdin". Done @https://github.com/cloudspannerecosystem/harbourbridge/pull/175/commits/be4ddfc12ff3c3c86abfdaa613cfa9f0d1fc2a12. WDYT?

cli/schema.go
agasheesh: Updated the comment in the now refactored function. Done @https://github.com/cloudspannerecosystem/harbourbridge/pull/175/commits/be4ddfc12ff3c3c86abfdaa613cfa9f0d1fc2a12.

cli/schema.go
agasheesh: I was initially planning to do such cleanups in follow up PRs. But since we are already here, done now @https://github.com/cloudspannerecosystem/harbourbridge/pull/175/commits/be4ddfc12ff3c3c86abfdaa613cfa9f0d1fc2a12.

cmd/schema.go
```diff
@@ -0,0 +1,85 @@
...
+specified by either dump-file flag or piped to stdin. Connection profile for source
+databases in direct connect mode can be specified by setting appropriate
+environment variables. The schema flags are:
+`, path.Base(os.Args[0]))
+}
+
+// SetFlags sets the flags.
+func (cmd *SchemaCmd) SetFlags(f *flag.FlagSet) {
+	f.StringVar(&cmd.filePrefix, "prefix", "", "File prefix for generated files")
+	f.StringVar(&cmd.driverName, "driver", "pg_dump", "Flag for specifying source DB or dump files (accepted values are \"pg_dump\", \"postgres\", \"mysqldump\", \"mysql\", and \"dynamodb\")")
```
nevinheintze, line 51: Just wondering out loud here, so feel free to ignore this.

Should we consider renaming driver to source?
We kind of have one chance to do flag renaming as part of the subcommand change, and as we add more capabilities to harbourbridge (including more target databases), I'm wondering if source and target are better/more-symmetric/more-futureproof that driver and targetDb?

Or should we adopt the terminology that DMS uses: source and destination?
(see https://cloud.google.com/database-migration/docs/overview).

cmd/schema.go
```diff
@@ -0,0 +1,85 @@
...
+		dbName, err := conversion.GetDatabaseName(cmd.driverName, time.Now())
+		if err != nil {
+			log.Fatalf("can't generate database name for prefix: %v\n", err)
+		}
+		cmd.filePrefix = dbName + "."
+	}
+
+	var conv *internal.Conv
+	var err error
+	conv, err = conversion.SchemaConv(cmd.driverName, cmd.targetDb, ioHelper, cmd.schemaSampleSize)
```
nevinheintze, line 75: Maybe drop the two var declarations, and make this an assignment?

cmd/schema.go
```diff
@@ -0,0 +1,85 @@
...
+func (cmd *SchemaCmd) SetFlags(f *flag.FlagSet) {
+	f.StringVar(&cmd.filePrefix, "prefix", "", "File prefix for generated files")
+	f.StringVar(&cmd.driverName, "driver", "pg_dump", "Flag for specifying source DB or dump files (accepted values are \"pg_dump\", \"postgres\", \"mysqldump\", \"mysql\", and \"dynamodb\")")
+	f.Int64Var(&cmd.schemaSampleSize, "schema-sample-size", int64(100000), "Number of rows to use for inferring schema (only for DynamoDB)")
+	f.StringVar(&cmd.dumpFilePath, "dump-file", "", "Location of dump file to process")
+	f.StringVar(&cmd.targetDb, "target-db", "", "Specifies the target DB, defaults to spanner")
+}
+
+func (cmd *SchemaCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
+	input := common.OpenDump(cmd.dumpFilePath)
```
nevinheintze, line 58: It's a bit confusing to call OpenDump when we might not have a dump file or maybe we are using a direct-connect driver, so we don't even need the dump file.
(This was also an issue with the previous code with the loadInput function, which was even less meaningful).

I don't have a good resolution for this.

I can't think of a way to improve the naming of OpenDump -- OptionallyOpenDump or ConfigureInput seem worse ...

And I can't think of a good comment to add. Something like:

// OpenDump file it is is specified.

seems pretty worthless.

nevinheintze (approved):
Looks good -- just a few minor comments.
Many thanks!

cmd/schema.go
agasheesh: Yes, in general we want to move to terminology that DMS uses: source and destination as well as reuse the concept of connection profiles for connecting to source and destination databases. Since we only get once change at this, I wanted to be a little thorough in this renaming work and thus was planning to defer this as an immediate follow up work. WDYT?

cmd/schema.go
agasheesh: I thought about it a bit and realized that OpenDump/LoadInput wasn't adding much meaning whereas, what we really wanted was to get reference to a new IOStreams struct depending on whether we are in dump mode etc. So refactored out the code to create the following helper function.

```
// NewIOStreams returns a new IOStreams struct such that input stream is set
// to open file descriptor for dumpFile if driver is PGDUMP or MYSQLDUMP.
// Input stream defaults to stdin. Output stream is always set to stdout.
func NewIOStreams(driver string, dumpFile string) IOStreams {
	io := IOStreams{In: os.Stdin, Out: os.Stdout}
	if (driver == PGDUMP || driver == MYSQLDUMP) && dumpFile != "" {
		fmt.Printf("\nLoading dump file from path: %s\n", dumpFile)
		f, err := os.Open(dumpFile)
		if err != nil {
			fmt.Printf("\nError reading dump file: %v err:%v\n", dumpFile, err)
			log.Fatal(err)
		}
		io.In = f
	}
	return io
}
```

Let me know what do you think?

cmd/schema.go
agasheesh: Done.

cmd/schema.go
nevinheintze: Yup, sounds good - doesn't need to be part of this PR.

cmd/schema.go
nevinheintze: Yes, I think that is much better.
Thanks for thinking about this and coming up with a good solution.

cli/schema.go
Deep1998: This is the best rephrasing according to me!

testing/postgres/integration_test.go
Deep1998: Sounds good!

Deep1998 (approved):
LGTM! Thanks for the changes!

PR #177, opened 2021-09-23
Support migration to existing spanner databases

Deep1998:
Added a fix for the schema-only failure.

### DIFF
--- a/conversion/conversion.go
+++ b/conversion/conversion.go
@@ -428,27 +428,22 @@ func VerifyTargetDb(targetDb, project, instance, dbName string) (string, bool, e
 	dbPath := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName)
 	_, err = adminClient.GetDatabase(ctx, &adminpb.GetDatabaseRequest{Name: dbPath})
 	if err != nil {
-		// Differentiate between "db does not exist" and other errors? Assuming err means db does not exist for now
-		return targetDb, true, nil
+		if containsAny(strings.ToLower(err.Error()), []string{"database not found"}) {
+			return targetDb, false, nil
+		}
+		return "", false, fmt.Errorf("can't get database info: %s", err)
 	}
-	//if resp.DatabaseDialect == "POSTGRESQL" {
-	//	targetDb = TARGET_EXPERIMENTAL_POSTGRES
-	//} else {
-	//	targetDb = TARGET_SPANNER
-	//}
 	dbDdl, err := adminClient.GetDatabaseDdl(ctx, &adminpb.GetDatabaseDdlRequest{Database: dbPath})
 	if err != nil {
-		fmt.Printf("\nCan't fetch database ddl: %v\n", err)
 		return "", false, fmt.Errorf("can't fetch database ddl")
 	}
 	if len(dbDdl.Statements) != 0 {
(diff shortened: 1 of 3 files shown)


conversion/conversion.go
```diff
@@ -417,12 +417,85 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+		return dbExists, err
+	}
+	if dbExists {
+		err = ValidateDDL(ctx, adminClient, dbURI)
+	}
+	return dbExists, err
+}
+
+// CheckExistingDb checks whether the dbURI exists or not.
+func CheckExistingDb(ctx context.Context, adminClient *database.DatabaseAdminClient, dbURI string) (bool, error) {
```
bharadwaj-aditya, line 440: make the methods private where it is not expected to be called from outside

conversion/conversion.go
```diff
@@ -444,17 +517,42 @@ func CreateDatabase(project, instance, dbName string, conv *internal.Conv, out *
...
+	adminClient, err := database.NewDatabaseAdminClient(ctx)
+	if err != nil {
+		return "", fmt.Errorf("can't create admin client: %w", analyzeError(err, project, instance))
+	}
+	defer adminClient.Close()
+
+	dbURI := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName)
+	op, err := adminClient.UpdateDatabaseDdl(ctx, &adminpb.UpdateDatabaseDdlRequest{
+		Database:   dbURI,
+		Statements: conv.SpSchema.GetDDL(ddl.Config{Comments: false, ProtectIds: true, Tables: true, ForeignKeys: false}),
```
bharadwaj-aditya, line 537: why is ForeignKeys false ? Is this handled elsewhere ?

bharadwaj-aditya (changes requested):

conversion/conversion.go
Deep1998: These are being called in the unit tests thats why kept it public.

conversion/conversion.go
Deep1998: Yes foreign keys we create at the very end after data migration. This only creates tables and indexes.

bharadwaj-aditya (approved):
LGTM

conversion/conversion.go
agasheesh: I think it's fair to add comment to the same effect to help future readers understand this.

agasheesh:
The title and description for this PR is pretty thin, which makes it hard for future readers to navigate the changes. In general, there should be enough detail in title for readers to "find" the appropriate PR and in description for readers to not to have to look at the code to understand the changes, unless they want to.

conversion/conversion.go
```diff
@@ -417,12 +417,85 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
...
+	if err != nil {
+		return dbExists, err
+	}
+	if dbExists {
+		err = ValidateDDL(ctx, adminClient, dbURI)
+	}
+	return dbExists, err
+}
+
+// CheckExistingDb checks whether the dbURI exists or not.
```
agasheesh, line 439: `whether the dbURI` -> `whether database with dbURI`

agasheesh (approved):

PR #179, opened 2021-09-23
Moved database code to sources directory

### LINKED ISSUE #178: Source Refactoring
During the process of adding support for various database sources in Harbourbridge, there has been a large amount of duplicated code that has build up for each databases. There are two aspects that need to be handled as a part of the code cleanup

1. Structure of the codebase - to move the code of each source DB into its relevant directory, and group it under a common parent directory. This will ensure that the structure of the codebase is maintained in an easy to understand manner
2. Extracting common code across databases - A large amount of code across the implementation of each database is common, as it handles things like reading from dump files, information schema etc which are relatively similar across most databases. We need to extract out common functionality across the various sources, to maintain a common code flow, reduce bloat, make adding new databases easier and ensure that bug fixes go to all the source databases.
3. Move implementation details of specific databases from the common top level code (probably a factory or registration type structure). This will ensure that when a source database is added, there is no leak of source db logic into the common processing code across harbour bridge

Each of the above can be handled in an independent change request, but the exercise should be taken up before adding support for more databases.

bharadwaj-aditya:
Fixes part 1 of #178

This moves all the databases into the sources directory. Once this is complete, we can take up the further aspects of #178 

- [x] Tests pass


agasheesh (approved):
Thanks for this cleanup. LGTM.

Issue #176, opened 2021-09-23
Multi-row INSERT statements broken for PostgreSQL


nevinheintze:
We only handle single row INSERT statements. Insert statements with multiple rows are broken: we construct a row with all of the values in the multiple rows concatenated together.

Some history ...

In pgdump, the default way to write out data to the dump file is to use COPY-FROM. However, you can also get pgdump to generate INSERT statements. The very first version of harbourbridge used the insert version (it was kind of easier to start with). When pgdump generates INSERTS, it only generates them with one row at a time i.e.

INSERT INTO public.products (product_id, description, price, date_added) VALUES ('abc-123', 'Blue suede shoes', '141.99', '2020-06-06');

And so to get harbourbridge bootstrapped, that's all that was implemented. Then we switched to using COPY-FROM, and (mostly) stopped using INSERT.

However, PostgreSQL does support multi-row inserts i.e.:

INSERT INTO public.products (product_id, description, price, date_added) VALUES
('abc-123', 'Blue suede shoes', '141.99', '2020-06-06'),
('axd-673', 'Antique typewriter', '99.99', '2020-06-07'),
('zxi-631', 'Glass vase', '55.50', '2020-06-10');

These get completely mangled at the moment! The main issue is in getVals, which builds the following row for this multi-row INSERT:

vals = ["abc-123" "Blue suede shoes" "141.99" "2020-06-06" "axd-673" "Antique typewriter" "99.99" "2020-06-07" "zxi-631" "Glass vase" "55.50" "2020-06-10"]

i.e. all of the values from all of the rows concatenated together!

To fix this, we should:
a) return [][]string from getVals
b) change copyOrInsert to handle [][]string for vals (maybe we should call it rows?)
c) change ProcessPgDump to handle multi-row inserts i.e. iterate over the [][]string and call ProcessDataRow on each row.
d) add a test for multi-row INSERT statements (in fact, this should probably be the first step!)

Here's a complete dump file to test this out:

CREATE TABLE public.products (
    product_id character varying(20) NOT NULL,
    description character varying(1000),
    price numeric(10,2),
    date_added date
);

INSERT INTO public.products (product_id, description, price, date_added) VALUES
('abc-123', 'Blue suede shoes', '141.99', '2020-06-06'),
('axd-673', 'Antique typewriter', '99.99', '2020-06-07'),
('zxi-631', 'Glass vase', '55.50', '2020-06-10');

PR #185, opened 2021-09-27
Add support for processing multi-row insert statements in pgdump

agasheesh:
Fixes https://github.com/cloudspannerecosystem/harbourbridge/issues/176.

### DIFF
--- a/sources/postgres/pgdump.go
+++ b/sources/postgres/pgdump.go
@@ -31,7 +31,7 @@ type copyOrInsert struct {
 	stmt  stmtType
 	table string
 	cols  []string
-	vals  []string // Empty for COPY-FROM.
+	rows  [][]string // Empty for COPY-FROM.
 }
 
 type stmtType int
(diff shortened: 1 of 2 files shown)


sources/postgres/pgdump.go
```diff
@@ -733,8 +733,11 @@ func getVals(conv *internal.Conv, vll []*pg_query.Node, n *pg_query.InsertStmt)
 		default:
 			conv.Unexpected(fmt.Sprintf("Processing %v statement: found %s in ValuesList", printNodeType(n), printNodeType(vals)))
 		}
+		if len(values) > 0 {
```
nevinheintze, line 736: I'm not sure if the conditional around the append is needed or helpful.

The relevant case is if ​every value in a row hit an error case i.e. one of the default cases in the switches.
​In that case we'd  skip the row.

Now, if we just had a few bad values, then len(values) would be non-zero, and we'd go and append the row, and we'd hit an error when we call ProcessDataRow since we'd find that the number of values and columns didn't match.

So now we have a situation where if there are some errors, we'll handle it in ProcessDataRow, but if all values in a row are errors, then we'll handle it here. That seems confusing. Also it'll break our statistics for rows (ProcessDataRow keeps stats for bad rows and total rows).

Suggest dropping this and unconditionally appending values to rows, and handling all errors in ProcessDataRows.

nevinheintze (approved):
Thanks Asheesh!
Just one minor comment -- looks good!

sources/postgres/pgdump.go
agasheesh: This is a good observation. Thanks for catching this, fixed now.

PR #188, opened 2021-09-30
Changed progress reporting for foreign keys to fractions

Deep1998:
Changed progress reporting to also support fractional updates. Also changed the maxworkers during foreign key creation to 20.

### DIFF
--- a/conversion/conversion.go
+++ b/conversion/conversion.go
@@ -83,7 +83,7 @@ const (
 
 var (
 	// Set the maximum number of concurrent workers during foreign key creation.
-	MaxWorkers = 10
+	MaxWorkers = 20
 )
 
 func SchemaConv(driver string, targetDb string, ioHelper *IOStreams, schemaSampleSize int64) (*internal.Conv, error) {
@@ -201,7 +201,7 @@ func dataFromSQL(driver string, config spanner.BatchWriterConfig, client *sp.Cli
 		return nil, err
 	}
 	totalRows := conv.Rows()
-	p := internal.NewProgress(totalRows, "Writing data to Spanner", internal.Verbose())
+	p := internal.NewProgress(totalRows, "Writing data to Spanner", internal.Verbose(), false)
 	rows := int64(0)
 	config.Write = func(m []*sp.Mutation) error {
 		_, err := client.Apply(context.Background(), m)
@@ -251,7 +251,7 @@ func dataFromDynamoDB(config spanner.BatchWriterConfig, client *sp.Client, conv 
 	dydbClient := dydb.New(mySession, getDynamoDBClientConfig())
 	dynamodb.SetRowStats(conv, dydbClient)
 	totalRows := conv.Rows()
-	p := internal.NewProgress(totalRows, "Writing data to Spanner", internal.Verbose())
+	p := internal.NewProgress(totalRows, "Writing data to Spanner", internal.Verbose(), false)
 
 	rows := int64(0)
 	config.Write = func(m []*sp.Mutation) error {
@@ -310,7 +310,7 @@ func schemaFromDump(driver string, targetDb string, ioHelper *IOStreams) (*inter
 	ioHelper.BytesRead = n
 	conv := internal.MakeConv()
 	conv.TargetDb = targetDb
-	p := internal.NewProgress(n, "Generating schema", internal.Verbose())
+	p := internal.NewProgress(n, "Generating schema", internal.Verbose(), false)
 	r := internal.NewReader(bufio.NewReader(f), p)
 	conv.SetSchemaMode() // Build schema and ignore data in dump.
 	conv.SetDataSink(nil)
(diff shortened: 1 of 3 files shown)


internal/progress.go
```diff
@@ -83,3 +95,31 @@ func (p *Progress) report(firstCall bool) {
...
+	if p.verbose {
+		fmt.Printf("%s: %d/%d\n", p.message, p.progress, p.total)
+		return
+	}
+	if firstCall {
+		fmt.Printf("%s: %d/%d", p.message, p.progress, p.total)
+	} else {
+		delStr := strings.Repeat("\b", countDigits(p.progress-1)+countDigits(p.total)+1)
+		fmt.Printf(delStr+"%d/%d", p.progress, p.total)
+	}
```
Deep1998, line 109: Assuming that previously reported percentage is `progress-1` since we only call it incrementally during foreign key updates. In the future, we should store previous reported progress and use that to calculate num of digits.

conversion/conversion.go
```diff
@@ -582,7 +582,7 @@ func UpdateDDLForeignKeys(ctx context.Context, adminClient *database.DatabaseAdm
 		return nil
 	}
 	msg := fmt.Sprintf("Updating schema of database %s with foreign key constraints ...", dbURI)
```
Deep1998, line 584: Should we add some sort of a warning message before updating foreign keys that says creation can be slow due to high number of foreign keys?

Deep1998, line 584: Should we add some sort of a warning message before updating foreign keys that says
"Foreign Key Updates can be slow (even if no data is loaded) due to limitations on spanner side" or something similar?

internal/progress.go
```diff
@@ -16,26 +16,35 @@
...
-	pct      int    // Percentage done i.e. progress/total * 100
-	message  string // Name of task being monitored.
-	verbose  bool   // If true, print detailed info about each progress step.
+	total      int64  // How much we have to do.
+	progress   int64  // How much we have done so far.
+	pct        int    // Percentage done i.e. progress/total * 100
+	message    string // Name of task being monitored.
+	verbose    bool   // If true, print detailed info about each progress step.
+	fractional bool   // If true, report progress in fractions instead of percentages.
 }
```
agasheesh, line 35: I think this change makes in sense in general, but for the case of FKs I wonder if we can also output the absolute number of foreign keys processed vs remaining (potentially also in-flight).

Deep1998: Currently, this change displays processed / total for ex: "40/100", should we output 40 processed, 60 remaining instead?

conversion/conversion.go
```diff
@@ -83,7 +83,7 @@ const (
 
 var (
 	// Set the maximum number of concurrent workers during foreign key creation.
-	MaxWorkers = 10
+	MaxWorkers = 20
```
agasheesh, line 86: I remember you mentioned some caveats around changing this to a very high number e.g., 100. It might be worth adding a comment describing how to set this value and how it relates to database admin QPS quota.

conversion/conversion.go
agasheesh: I think that sounds reasonable, especially if number of foreign keys is > 100 or something. In that case, you can also mention about the MaxWorkers variable that customers can tweak.

internal/progress.go
agasheesh: What you have right now is perfect, I was mislead by the variable name initially and thought that it's going to print something like 0.1%, 0.2% and so on ...

internal/progress.go
```diff
@@ -83,3 +96,19 @@ func (p *Progress) report(firstCall bool) {
...
+
+func (p *Progress) reportFraction(firstCall bool) {
+	if p.verbose {
+		fmt.Printf("%s: %d/%d\n", p.message, p.progress, p.total)
+		return
+	}
+	if firstCall {
+		fmt.Printf("%s: %d/%d", p.message, p.progress, p.total)
+	} else {
+		delStr := strings.Repeat("\b", len(strconv.Itoa(int(p.progress-1)))+1+len(strconv.Itoa(int(p.total))))
```
agasheesh, line 108: I am not sure what this is doing but my best guess is that this is a way to overwrite progress bar instead of printing on a new line on the terminal. Either way it's probably worth adding a comment.

conversion/conversion.go
```diff
@@ -582,7 +582,7 @@ func UpdateDDLForeignKeys(ctx context.Context, adminClient *database.DatabaseAdm
 		return nil
 	}
 	msg := fmt.Sprintf("Updating schema of database %s with foreign key constraints ...", dbURI)
-	p := internal.NewProgress(int64(len(fkStmts)), msg, internal.Verbose())
+	p := internal.NewProgress(int64(len(fkStmts)), msg, internal.Verbose(), true)
```
agasheesh, line 586: I wonder whether it's cleaner to initiate Progress struct as following instead

p := &Progress{total: int64(len(fkStmts)), ...}

Just thinking out loud, don't need to change in this PR.

agasheesh (approved):

conversion/conversion.go
bharadwaj-aditya: Ideally, variables that don't need to be set outside of the struct should not be exposed outside the package. That prevents misuse from parts of code which must not, and makes it easy when performing refactoring.
The approach of using NewProgress - is more reminiscent of OO programming, but i guess in these scenarios it has its merit

bharadwaj-aditya (approved):
LGTM

Issue #194, opened 2021-10-07
Support to migrate Json column types


bharadwaj-aditya:
## Actual Behavior
Json is migrated to spanner as a String.

## Expected Behavior
Columns of type Json in source databases (eg-Mysql, Postgres) should be migrated to spanner to columns of Json

Different databases handle Json differently. 
In Postgres, it is implemented as a native column type.
In Mysql 8 and above it is implemented via a check_constraint for Json on top of a longtext 
In Mysql 5.7-8 the representation of the field is different from Mysql 8 (Needs exploration)

manitgupta:
JSON migration is now supported in Harbourbridge.

PR #195, opened 2021-10-07
added json support for postgres databases

bharadwaj-aditya:
Fixes postgres part of #https://github.com/cloudspannerecosystem/harbourbridge/issues/194

- [x] Tests pass

### DIFF
--- a/sources/postgres/pgdump_test.go
+++ b/sources/postgres/pgdump_test.go
@@ -69,6 +69,7 @@ func TestProcessPgDump(t *testing.T) {
 		{"varchar", ddl.Type{Name: ddl.String, Len: ddl.MaxLength}},
 		{"varchar(42)", ddl.Type{Name: ddl.String, Len: int64(42)}},
 		{"json", ddl.Type{Name: ddl.Json}},
+		{"jsonb", ddl.Type{Name: ddl.Json}},
 	}
 	for _, tc := range scalarTests {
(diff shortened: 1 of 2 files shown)


sources/postgres/toddl_test.go
```diff
@@ -106,14 +106,15 @@ func TestToExperimentalSpannerType(t *testing.T) {
 	name := "test"
 	srcSchema := schema.Table{
 		Name:     name,
-		ColNames: []string{"a", "b", "c", "d", "e", "f"},
+		ColNames: []string{"a", "b", "c", "d", "e", "f", "g"},
 		ColDefs: map[string]schema.Column{
```
agasheesh, line 110: This test function is specific to capturing unsupported types for PG Spanner - is TestProcessPgDump a better place to capture new types that HB now supports? You would be able to add test for both schema and data conversion there.

sources/postgres/data.go
```diff
@@ -132,6 +132,8 @@ func convScalar(spannerType ddl.Type, srcTypeName string, location *time.Locatio
 		return val, nil
 	case ddl.Timestamp:
 		return convTimestamp(srcTypeName, location, val)
+	case ddl.Json:
+		return val, nil
```
agasheesh, line 136: Should we wrap this in a more specific type e.g., NullJSON - https://github.com/googleapis/google-cloud-go/blob/4537b45d2611ce480abfb5d186b59e7258ec872c/spanner/value.go#L774 ?

PR #204, opened 2021-10-19
Remove sql dependency from common

### LINKED ISSUE #189: Standardizing DynamoDb implementation to use common structure
## Expected Behavior
The DynamoDb implementation should use the same interfaces as the common db code in sources/common/ as outlined in Issue #178 and PR #181 

## Actual Behavior
Current implementation is completely independent of the structure used in Mysql and Postgres.
Making it follow the same code flow will ensure a more standard code base and provide a template for further non-relational DB integrations.

charvisingla:
Fixes #189 
These changes aim to remove the sql dependency from common package to make is usable for dynamodb (or any other non sql database) implementation. It includes the following changes:
1. Merging GetColumns and ProcessColumns together.
Mysql and Postgres get the columns names first and then fetch their definitions. However, for dynamo, sample data is scanned and column names and definitions are derived from that together. Merging these 2 methods would make the common implementation useful for dynamo as well.
2. Remove the sql data rows specific logic from common and refactor ProcessData to make it useful for Dynamodb implementation.
3. Move sql.db inside InfoSchema structs.
Mysql and Postgres use sql.db client while dynamo uses dydb client library. After the above 2 changes, sql.db usage in common ceases to exist so there is no need to keep it in common just to pass it down as method arguments to mysql/postgres infoschema implementations.

- [x] Tests pass

### DIFF
--- a/sources/common/infoschema.go
+++ b/sources/common/infoschema.go
@@ -15,7 +15,7 @@
 package common
 
 import (
-	"database/sql"
+	sql "database/sql"
 	"fmt"
 
 	"github.com/cloudspannerecosystem/harbourbridge/internal"
@@ -28,14 +28,14 @@ type InfoSchema interface {
 	GetToDdl() ToDdl
 	GetTableName(schema string, tableName string) string
 	GetTables() ([]SchemaAndName, error)
-	GetColumns(table SchemaAndName) (*sql.Rows, error) //TODO - merge this method and ProcessColumns for cleaner interface
-	ProcessColumns(conv *internal.Conv, cols *sql.Rows, constraints map[string][]string) (map[string]schema.Column, []string)
-	GetRowsFromTable(conv *internal.Conv, table SchemaAndName) (*sql.Rows, error)
+	GetColumns(table SchemaAndName) (interface{}, error) //TODO - merge this method and ProcessColumns for cleaner interface
+	ProcessColumns(conv *internal.Conv, cols interface{}, constraints map[string][]string) (map[string]schema.Column, []string)
+	GetRowsFromTable(conv *internal.Conv, srcTable string) (interface{}, error)
 	GetRowCount(table SchemaAndName) (int64, error)
 	GetConstraints(conv *internal.Conv, table SchemaAndName) ([]string, map[string][]string, error)
 	GetForeignKeys(conv *internal.Conv, table SchemaAndName) (foreignKeys []schema.ForeignKey, err error)
 	GetIndexes(conv *internal.Conv, table SchemaAndName) ([]schema.Index, error)
-	ProcessDataRows(conv *internal.Conv, srcTable string, srcCols []string, srcSchema schema.Table, spTable string, spCols []string, spSchema ddl.CreateTable, rows *sql.Rows)
+	ProcessData(conv *internal.Conv, srcTable string, srcSchema schema.Table, spTable string, spCols []string, spSchema ddl.CreateTable)
 }
 
 // SchemaAndName contains the schema and name for a table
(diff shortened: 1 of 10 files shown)


sources/common/infoschema.go
```diff
@@ -26,15 +25,14 @@ import (
...
-	ProcessColumns(conv *internal.Conv, cols *sql.Rows, constraints map[string][]string) (map[string]schema.Column, []string)
-	GetRowsFromTable(conv *internal.Conv, db *sql.DB, table SchemaAndName) (*sql.Rows, error)
-	GetRowCount(db *sql.DB, table SchemaAndName) (int64, error)
-	GetConstraints(conv *internal.Conv, db *sql.DB, table SchemaAndName) ([]string, map[string][]string, error)
-	GetForeignKeys(conv *internal.Conv, db *sql.DB, table SchemaAndName) (foreignKeys []schema.ForeignKey, err error)
-	GetIndexes(conv *internal.Conv, db *sql.DB, table SchemaAndName) ([]schema.Index, error)
-	ProcessDataRows(conv *internal.Conv, srcTable string, srcCols []string, srcSchema schema.Table, spTable string, spCols []string, spSchema ddl.CreateTable, rows *sql.Rows)
+	GetTables() ([]SchemaAndName, error)
+	GetColumns(conv *internal.Conv, table SchemaAndName, constraints map[string][]string, primaryKeys []string) (map[string]schema.Column, []string, error)
+	GetRowsFromTable(conv *internal.Conv, srcTable string) (interface{}, error)
```
bharadwaj-aditya, line 30: why is this parameter changed from schemaAndName to string ?

sources/mysql/infoschema_test.go
```diff
@@ -127,6 +120,28 @@ func TestProcessInfoSchemaMYSQL(t *testing.T) {
...
 			cols:  []string{"INDEX_NAME", "COLUMN_NAME", "SEQ_IN_INDEX", "COLLATION", "NON_UNIQUE"},
+		}, {
+			query: "SELECT (.+) FROM information_schema.COLUMNS (.+)",
+			args:  []driver.Value{"test", "product"},
+			cols:  []string{"column_name", "data_type", "column_type", "is_nullable", "column_default", "character_maximum_length", "numeric_precision", "numeric_scale", "extra"},
+			rows: [][]driver.Value{
+				{"product_id", "text", "text", "NO", nil, nil, nil, nil, nil},
+				{"product_name", "text", "text", "NO", nil, nil, nil, nil, nil}},
+		}, {
+			query: "SELECT (.+) FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS (.+)",
```
bharadwaj-aditya, line 131: Have we increased the number of queries fired during the evaluation ? What is the requirement for these 3 new queries ?

sources/postgres/data_test.go
```diff
@@ -280,12 +280,6 @@ func buildConv(spTable ddl.CreateTable, srcTable schema.Table) *internal.Conv {
 	conv := internal.MakeConv()
 	conv.SpSchema[spTable.Name] = spTable
 	conv.SrcSchema[srcTable.Name] = srcTable
-	conv.ToSource[spTable.Name] = internal.NameAndCols{Name: srcTable.Name, Cols: make(map[string]string)}
```
bharadwaj-aditya, line 283: Is this not required anymore ?

sources/postgres/infoschema_test.go
```diff
@@ -127,6 +119,28 @@ func TestProcessInfoSchema(t *testing.T) {
...
 			cols:  []string{"index_name", "column_name", "column_position", "is_unique", "order"},
+		}, {
+			query: "SELECT (.+) FROM information_schema.COLUMNS (.+)",
+			args:  []driver.Value{"public", "product"},
+			cols:  []string{"column_name", "data_type", "data_type", "is_nullable", "column_default", "character_maximum_length", "numeric_precision", "numeric_scale"},
+			rows: [][]driver.Value{
+				{"product_id", "text", nil, "NO", nil, nil, nil, nil},
+				{"product_name", "text", nil, "NO", nil, nil, nil, nil}},
+		}, {
+			query: "SELECT (.+) FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS (.+)",
```
bharadwaj-aditya, line 130: Same as above. Why are these 3 new queries required ?

sources/postgres/infoschema_test.go
```diff
@@ -445,15 +432,14 @@ func TestConvertSqlRow_MultiCol(t *testing.T) {
 			query: "SELECT (.+) FROM pg_index (.+)",
 			args:  []driver.Value{"public", "test"},
 			cols:  []string{"index_name", "column_name", "column_position", "is_unique", "order"},
-		},
-		// Note: go-sqlmock mocks specify an ordered sequence
```
bharadwaj-aditya, line 449: Did not follow why this is removed ?

bharadwaj-aditya (changes requested):

sources/postgres/infoschema_test.go
charvisingla: This was a redundant call and there was a TODO to remove this. This removal was also needed for dynamodb refactoring. I think I resolved 2-3 TODOs with this PR. Would it be better to create several small PRs for these multiple changes?

sources/common/infoschema.go
charvisingla: GetRowsFromTable is called from ProcessData. schemaAndName object was populated using GetTables call. Due to the change in ProcessData implementation to not make a redundant call to GetTables, we no longer have the schemaAndName object. While we can build this object explicitly to pass to GetRowsFromTable method, it seemed like an overkill. We already have conv access inside GetRowsFromTable to fetch the schema information so I used that.

sources/mysql/infoschema_test.go
bharadwaj-aditya: Got it. Looks like the diff was showing a little differently and the test_ref queries from below have moved up

sources/postgres/infoschema_test.go
bharadwaj-aditya: Different PR would be good considering the size of this one. But i think this one should be fine

bharadwaj-aditya (commented):
Loks fine overall. Just one pending comment on the postgres data_test. If that is fine, then looks fine from my end.

sources/postgres/data_test.go
charvisingla: This setup is not needed as its done through code here https://github.com/cloudspannerecosystem/harbourbridge/blob/3e5affad94d0691b260e9f5e8497e7a23147e4d0/internal/mapping.go#L107.

agasheesh:
There is a lot happening in this PR - I can see at least 3 distinct changes all bundled together. It wasn't clear to me whether it was required for them to be in the same PR. 

3 distinct changes I could spot - 

1. Move db inside InfoSchema structs and thus remove sql dependency from common. It's not clear to me why this change was required in the first place.
2. Resolving the TODO for merging GetColumns and ProcessColumns together. Seems like most of the changes to the tests are related to this.
3. ProcessDataRows refactoring. This part was mostly unclear to me - why that change was required.

Ideally, you should have raised all of these as separate PRs - and even then any cleanup should be its own separate PR. That would have helped reviewers with faster turn around times and overall quicker turnaround as well.

It's probably not worth at this point breaking this up since multiple rounds of reviews have been done. What I would suggest is that we should at least add more details in the PR description such that reviewers won't need to guess what, why and how.

charvisingla:
While it seems that there are 3 distinct changes in this PR, all of them aim to remove sql dependency from common package in order to make it usable for dynamodb implementation. The overlap of these changes with TODOs cleanup was coincidental.

> 1. Move db inside InfoSchema structs and thus remove sql dependency from common. It's not clear to me why this change was required in the first place.

sql.db is a concept/library valid for mysql and postgres while it doesn't have any meaning for dynamo so it doesn't need to be in common. Moreover, after the other changes in this PR, its usage in common would also cease to exist so it didn't make any sense to keep it in common just to pass it down as method arguments. Besides, its initialized just once in conversion.go so should anyways be an object level parameter.
> 2. Resolving the TODO for merging GetColumns and ProcessColumns together. Seems like most of the changes to the tests are related to this.

Merging these 2 were needed to make this common implementation useful for dynamo. Mysql and Postgres get the columns names first and then fetch their definitions. However, for dynamo, sample data is scanned and column names and definitions are derived from that together. So, it made sense to merge these in common instead of making redundant different GetColumns and ProcessColumns for dynamo. 
> 3. ProcessDataRows refactoring. This part was mostly unclear to me - why that change was required.

This refactoring was required to remove the sql data rows specific logic from common as well as to sync the implementation with dynamo by removing redundant GetTables call.

I initially approached these as changes needed for removing sql from common so I missed to see them as distinct 3 changes and creating separate independent PRs for them. I'll take care of this in future. I'll also add the above reasoning in the PR description for now.

bharadwaj-aditya (approved):
LGTM

agasheesh:
Thanks Charvi for updating the PR description and for providing thorough justifications on my previous comment. Makes so much sense now when I read the rationale. It also makes sense why it's okay to have all the changes in the same PR, since they are quite related.

PR #205, opened 2021-10-19
added -verbose flag to meet gcloud criteria

Isabella-Pham:
Current verbose flag `-v` does not meet gCloud guidelines (one letter flags aren't allowed), thus I added support for a `-verbose` flag.

HarbourBridge will support both flags, however gCloud will only support `-verbose`.

- [ x] Tests pass
- [ x] Appropriate changes to README are included in PR

### DIFF
--- a/main.go
+++ b/main.go
@@ -58,6 +58,7 @@ func setupGlobalFlags() {
 	flag.StringVar(&driverName, "driver", "pg_dump", "driver name: flag for accessing source DB or dump files (accepted values are \"pg_dump\", \"postgres\", \"mysqldump\", and \"mysql\")")
 	flag.Int64Var(&schemaSampleSize, "schema-sample-size", int64(100000), "schema-sample-size: the number of rows to use for inferring schema (only for DynamoDB)")
 	flag.BoolVar(&verbose, "v", false, "verbose: print additional output")
+	flag.BoolVar(&verbose, "verbose", false, "verbose: print additional output")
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -67,6 +68,16 @@ func setupGlobalFlags() {
 	flag.StringVar(&targetDb, "target-db", conversion.TARGET_SPANNER, "target-db: Specifies the target DB. Defaults to spanner")
 }
 
+func didSetVerboseAlready() bool{
+	numTimesSet := 0
+    flag.Visit(func(f *flag.Flag) {
```
agasheesh, line 73: can you please fix the formatting here?

main.go
```diff
@@ -67,6 +68,16 @@ func setupGlobalFlags() {
 	flag.StringVar(&targetDb, "target-db", conversion.TARGET_SPANNER, "target-db: Specifies the target DB. Defaults to spanner")
 }
 
+func didSetVerboseAlready() bool{
```
agasheesh, line 71: can you think of a better name for this function - something around setVerboseTwice or similar ?

main.go
```diff
@@ -108,6 +119,10 @@ func main() {
 		return
 	}
 
+	if(didSetVerboseAlready()){
+		panic(fmt.Errorf("can't use both -v and -verbose flags, must use only one"))
```
agasheesh, line 123: s/`can't use`/`cannot set`

and would drop `, must use only one`, since it's not a must to set either in fact

agasheesh (approved):

PR #208, opened 2021-10-27
Add support for source-profile and target-profile in subcommands.

agasheesh:
This PR introduces support for specifying various options for specifying source and target through 2 new flags: `-source-profile` and `-target-profile`. Both of these flags a list of key value pairs as specified in the example usage below. 

For backward compatibility, we provide helper functions to convert SourceProfile and TargetProfile to legacy flags `-driver` and `-target-db`, but eventually we will remove usage of these legacy flags and SourceProfile and TargetProfile structs will be passed throughout the codebase consistently.

This PR also contains updated copies of various READMEs which we will keep updating with new subcommands and flags replace the original READMEs on master once we are ready to switch to new command line interface.

Example usage with new flags structure:

# Specifying source profile

## PostgreSQL dump

```sh
pg_dump mydb | harbourbridge schema -source=postgresql

harbourbridge schema -source=postgres < my_pg_dump_file

# Can specify file format explicitly, which is dump by default and is the only supported format today.

harbourbridge schema -source=postgresql -source-profile="format=dump,file=<full path to dump file>"

harbourbridge schema -source=postgresql -source-profile="format=dump,file=gs://<bucket_name>/<file_path>"
```

## PostgreSQL direct connect

If dump file is not specified on the stdin or through flag in source-profile, assume direct connect.

```sh
harbourbridge schema -source=postgres
```

## MySQL dump

```sh
mysqldump mydb | harbourbridge schema -source=mysql

harbourbridge schema -source=mysql < my_mysqldump_file

harbourbridge schema -prefix=mydb. -source=mysql < my_mysqldump_file
```

## MySQL direct connect

If dump file is not specified on the stdin or through flag in source-profile, assume direct connect.

```sh
export MYSQLHOST=<>
export MYSQLPORT=<>
export MYSQLUSER=<>
export MYSQLDATABASE=<>
harbourbridge schema -source=mysql
```

## DynamoDB
```sh
harbourbridge schema -source=dynamodb

harbourbridge schema -source=dynamodb -source-profile="schema-sample-size=500000"
```

# Specifying target profile

Default target is Spanner, which is also the only supported target database today.
```sh
harbourbridge schema -source=mysql -target=spanner < my_mysqldump_file
```

To specify PG Spanner database, dialect can be specified
```sh
harbourbridge schema -source=mysql -target-profile="dialect=postgresql" < my_mysqldump_file
```

Spanner instance name can be specified

```sh
harbourbridge schema -source=mysql -target-profile="instance=my-spanner-instance" < my_mysqldump_file
```

Spanner database name can be specified

```sh
harbourbridge schema -source=mysql -target-profile="dbname=my-spanner-database-name" < my_mysqldump_file
```


cmd/README.md
```diff
@@ -0,0 +1,521 @@
...
+The tool ingests schema and data from either a pg_dump/mysqldump file or directly 
+from the source database, and supports both schema and data migration. For schema
+migration, HarbourBridge automatically builds a Spanner schema from the schema
+of the source database. This schema can be customized using the HarbourBridge
+schema assistant. For data migration, HarbourBridge creates a new Spanner
+database using the Spanner schema built during schema migration, and populates
+it with data from the source database.
+
+For more details on schema customization and use of the schema assistant, see
+[web/README](web/README.md). The rest of this README describes the command-line
```
Deep1998, line 16: The web/README link seems to be broken.

cmd/README.dynamodb.md
```diff
@@ -0,0 +1,141 @@
...
+## Data Conversion
+
+### A Scan for Entire Table
+
+Data conversion proceeds table by table. For each table, we use the Scan API to
+read data. Each read has a size limit up to 1MB. By using the returned token, we
+make a subsequent call to continue retrieving data from the table.
+
+The row result contains the data type and data itself. According to our
+[inferred schema](#schema-inference), we will parse the row to a format that
```
Deep1998, line 137: Broken link here as well. I guess all links that are relative links (starting with #) would fail

cmd/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+```sh
+# Generate pg_dump file.
+pg_dump mydb > mydb.pg_dump
+
+harbourbridge schema -source=postgresql -source-profile="format=dump" < mydb.pg_dump
+# Or, (By default, format is "dump" when specifying input files)
+harbourbridge schema -source=postgresql < mydb.pg_dump
+# Or, you can specify full path to the dump file
+harbourbridge schema -source=postgresql -source-profile="format=dump,file=<full path to dump file>"
```
Deep1998, line 47: nit: full path to dump file > absolute path to dump file?

cmd/README.md
```diff
@@ -0,0 +1,521 @@
...
+# Target is Spanner by default, but you can also specify it explicitly.
+harbourbridge schema -source=postgresql -source-profile="format=dump" -target=spanner < mydb.pg_dump
+```
+
+Similarly, you can use the tool for other source databases.
+
+```sh
+# Generate Spanner schema from a MySQL dump file
+mysqldump mydb > mydb.mysqldump
+harbourbridge schema -source=mysql < mydb.mysqldump
```
Deep1998, line 60: Should there be a -source-profile="format=dump" here?

cmd/README.mysql.md
```diff
@@ -0,0 +1,268 @@
...
+The tool can also be applied to an existing mysqldump file:
+
+```sh
+harbourbridge schema -source=mysql < my_mysqldump_file
+```
+
+To specify a particular Spanner instance to use, run:
+
+```sh
+harbourbridge schema -source=mysql -target-profile="instance=my-spanner-instance" < my_mysqldump_file
```
Deep1998, line 33: Should we still keep the instance flag in schema mode?

cmd/README.postgres.md
```diff
@@ -0,0 +1,257 @@
...
+The tool can also be applied to an existing pg_dump file:
+
+```sh
+harbourbridge schema -source=postgres < my_pg_dump_file
+```
+
+To specify a particular Spanner instance to use during data migration, run:
+
+```sh
+harbourbridge data -source=postgres -target-profile="instance=my-spanner-instance" < my_pg_dump_file
```
Deep1998, line 36: Same here regarding instance flag with schema mode.

cmd/README.mysql.md
```diff
@@ -0,0 +1,268 @@
...
+mysqldump mydb | harbourbridge schema -source=mysql
+```
+
+The tool can also be applied to an existing mysqldump file:
+
+```sh
+harbourbridge schema -source=mysql < my_mysqldump_file
+```
+
+To specify a particular Spanner instance to use, run:
```
Deep1998, line 30: Nit:  Specify during data mode like the pg readme?

cmd/schema.go
```diff
@@ -46,38 +46,66 @@ environment variables. The schema flags are:
 
 // SetFlags sets the flags.
 func (cmd *SchemaCmd) SetFlags(f *flag.FlagSet) {
+	f.StringVar(&cmd.source, "source", "", "Flag for specifying source DB, (accepted values are \"PostgreSQL\", \"MySQL\", and \"DynamoDB\")")
+	f.StringVar(&cmd.sourceProfile, "source-profile", "", "Flag for specifying connection profile for source database")
```
Deep1998, line 50: Should we add an example of source-profile and target profile in the desc?

cmd/source_profile_test.go
```diff
@@ -0,0 +1,55 @@
...
+			want:         SourceProfileFile{format: "dump"},
+		},
+		{
+			name:         "format and path param, no file piped",
+			params:       map[string]string{"format": "dump", "file": "file1.mysqldump"},
+			pipedToStdin: false,
+			want:         SourceProfileFile{format: "dump", path: "file1.mysqldump"},
+		},
+		{
+			name:         "only path param, no file piped -- defauly dump format",
```
Deep1998, line 41: nit typo: default

cmd/source_profile.go
```diff
@@ -0,0 +1,274 @@
...
+		// variables.
+		conn, err := NewSourceProfileConnection(source, params)
+		return SourceProfile{ty: SourceProfileTypeConnection, conn: conn}, err
+	}
+}
+
+var filePipedToStdin = func() bool {
+	stat, _ := os.Stdin.Stat()
+	// Data is being piped to stdin, if true. Else, stdin is from a terminal.
+	return (stat.Mode() & os.ModeCharDevice) == 0
```
Deep1998, line 273: Question: this is true i.e. piped to stdin for both supported ways to ingest a dump? Using the | and <  operators in the command?

cmd/source_profile_test.go
```diff
@@ -0,0 +1,55 @@
...
+			name:         "format and path param, file piped -- piped file takes precedence",
+			params:       map[string]string{"format": "dump", "file": "file1.mysqldump"},
+			pipedToStdin: true,
+			want:         SourceProfileFile{format: "dump"},
+		},
+		{
+			name:         "format and path param, no file piped",
+			params:       map[string]string{"format": "dump", "file": "file1.mysqldump"},
+			pipedToStdin: false,
+			want:         SourceProfileFile{format: "dump", path: "file1.mysqldump"},
```
Deep1998, line 38: For the case when format dump is specified, no file is specified, piped is False, will we throw an error? I may be wrong but as per my understanding, it will default to direct connect currently.

cmd/source_profile.go
```diff
@@ -0,0 +1,274 @@
...
+		{
+			conn.ty = SourceProfileConnectionTypeMySQL
+			conn.mysql = NewSourceProfileConnectionMySQL(params)
+		}
+	case "postgresql", "postgres", "pg":
+		{
+			conn.ty = SourceProfileConnectionTypePostgreSQL
+			conn.pg = NewSourceProfileConnectionPostgreSQL(params)
+		}
+	case "dynamodb":
```
Deep1998, line 158: Should we add dydb too?

cmd/schema.go
```diff
@@ -46,38 +46,66 @@ environment variables. The schema flags are:
 
 // SetFlags sets the flags.
 func (cmd *SchemaCmd) SetFlags(f *flag.FlagSet) {
+	f.StringVar(&cmd.source, "source", "", "Flag for specifying source DB, (accepted values are \"PostgreSQL\", \"MySQL\", and \"DynamoDB\")")
```
nevinheintze, line 49: Given that we accept other values than just these three, maybe let's be less prescriptive here and instead of "accepted values" just say:
e.g.  \"PostgreSQL\", \"MySQL\", and \"DynamoDB\"
Also, how about using ` strings here to make this string a bit easier to read.

cmd/source_profile.go
```diff
@@ -0,0 +1,274 @@
...
+func NewSourceProfileFile(params map[string]string) SourceProfileFile {
+	profile := SourceProfileFile{}
+	if !filePipedToStdin() {
+		profile.path = params["file"]
+	}
+	if format, ok := params["format"]; ok {
+		profile.format = format
+		// TODO: Add check that format takes values from ["dump", "csv", "avro", ... etc]
+	} else {
+		fmt.Printf("User did not specify `format` in source-profile, defaulting to \"dump\"\n")
```
nevinheintze, line 33: Suggest rephrasing this as:

"source-profile format defaulting to \"dump\"\n"

cmd/source_profile.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func (src SourceProfile) ToLegacyDriver(source string) (string, error) {
+	switch src.ty {
+	case SourceProfileTypeFile:
+		{
+			switch strings.ToLower(source) {
+			case "mysql":
+				return "mysqldump", nil
+			case "postgresql", "postgres", "pg":
```
nevinheintze, line 195: We used to default to pg if driver was not specified.
Suggest either:
1) we preserve the old default, or
2) we explicitly catch the case where source is the empty string and print a message such as: "Please specify source e.g. PostgreSQL, MySQL, DynamoDB.

Note that at the moment, if you run:
  go run github.com/cloudspannerecosystem/harbourbridge schema < examples/cart.pg_dump 

You just get:
  exit status 1

We should print an error message.

cmd/source_profile.go
```diff
@@ -0,0 +1,274 @@
...
+				return "pg_dump", nil
+			default:
+				return "", fmt.Errorf("invalid source database %v for source-profile type file", strings.ToLower(source))
+			}
+		}
+	case SourceProfileTypeConnection:
+		{
+			switch strings.ToLower(source) {
```
nevinheintze, line 203: Same comment here -- we don't handle case of empty string for source.

cmd/source_profile.go
```diff
@@ -0,0 +1,274 @@
...
+	switch src.ty {
+	case SourceProfileTypeFile:
+		{
+			switch strings.ToLower(source) {
+			case "mysql":
+				return "mysqldump", nil
+			case "postgresql", "postgres", "pg":
+				return "pg_dump", nil
+			default:
+				return "", fmt.Errorf("invalid source database %v for source-profile type file", strings.ToLower(source))
```
nevinheintze, line 198: Suggest rephrasing as:

invalid source=%s specification in source profile

I think that makes it easier for users to track down the issue (and "source database" is kind of ambiguous here, since it could refer to the source database type or the source database instance).

Also, I'm not sure users will know what "source-profile type file" means?
I realize you want to say what mode we are running in, since e.g. DynamoDB isn't supported with a file input. I'm wondering whether it would be clearer to separate out the DynamoDB case and generate a targeted message in this case like:
"Dump files are not supported with DynamoDB."

cmd/schema.go
```diff
@@ -46,38 +46,66 @@ environment variables. The schema flags are:
...
-	f.Int64Var(&cmd.schemaSampleSize, "schema-sample-size", int64(100000), "Number of rows to use for inferring schema (only for DynamoDB)")
-	f.StringVar(&cmd.dumpFilePath, "dump-file", "", "Location of dump file to process")
-	f.StringVar(&cmd.targetDb, "target-db", "", "Specifies the target DB, defaults to spanner")
 }
 
 func (cmd *SchemaCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
-	ioHelper := conversion.NewIOStreams(cmd.driverName, cmd.dumpFilePath)
+	sourceProfile, err := NewSourceProfile(cmd.sourceProfile, cmd.source)
+	if err != nil {
+		log.Fatal(err)
```
nevinheintze, line 59: Due to a series of unfortunate adventures, log.Fatal doesn't actually do anything: log messages are dropped on the floor (see background on SetupLogFile in conversion/conversion.go).
That means that all information about why we failed here is lost, which is pretty confusing for users. They just see an "exit status 1" message.

Suggest switching to panic(err) here like we do in main.go.
Longer term, we can resurrect log, but we'll have to deal with the spam issue noted in SetupLogFile that was the reason we initially decided to silence log messages.

cmd/schema.go
```diff
@@ -46,38 +46,66 @@ environment variables. The schema flags are:
...
+	}
+	driverName, err := sourceProfile.ToLegacyDriver(cmd.source)
+	if err != nil {
+		log.Fatal(err)
```
nevinheintze, line 63: Same comment here: suggest using panic(err).
And again at line 69.

nevinheintze (approved):
Thanks Asheesh for all the work on this -- looks good and is a much needed overhaul of all of our flags an options.

As discussed, I've only done a high level review, not a thorough line-by-line review.
Please get LG from Deep and Aditya before merging.

cmd/source_profile.go
```diff
@@ -0,0 +1,274 @@
...
+//
+// Example: -source-profile="file=/tmp/abc, format=dump"
+// Example: -source-profile="file=gcs://bucket_name/cart.txt, format=dump"
+//
+// Format 2. Specify source connection parameters. If none specified, then read
+// from envrironment variables.
+//
+// Format 3. Specify a config file that specifies source connection profile.
+//
+func NewSourceProfile(str string, source string) (SourceProfile, error) {
```
nevinheintze, line 238: Suggest changing "str" arg name to either just "s" or something more descriptive e.g. "profile".
Just "str" isn't very useful.

cmd/target_profile.go
```diff
@@ -0,0 +1,117 @@
...
+// correspond to regular Cloud Spanner database and PG Cloud Spanner database
+// respectively.
+//
+// If dbname is not specified, then HarbourBridge will autogenerate the same
+// and create a database with the same name.
+//
+// Example: -target-profile="instance=my-instance1,dbname=my-new-db1"
+// Example: -target-profile="instance=my-instance1,dbname=my-new-db1,dialect=PostgreSQL"
+//
+func NewTargetProfile(str string) (TargetProfile, error) {
```
nevinheintze, line 82: Suggest changing "str" arg name to either just "s" or something more descriptive e.g. "profile".
Just "str" isn't very useful.

cmd/README.md
agasheesh: These copies in cmd/ are only supposed to be used as a staging area until we are ready to copy them over to their original locations. Thus, it's okay if the links are temporary broken. We will get an opportunity to review the same again in a separate PR focused on updating READMEs. I will add a temporary banner to top of these READMEs to reflect the same.

cmd/README.dynamodb.md
agasheesh: These copies in cmd/ are only supposed to be used as a staging area until we are ready to copy them over to their original locations. Thus, it's okay if the links are temporary broken. We will get an opportunity to review the same again in a separate PR focused on updating READMEs. I will add a temporary banner to top of these READMEs to reflect the same.

cmd/README.md
agasheesh: I didn't want to say absolute path since wasn't sure if that works for a file stored on GCS etc. Full path seemed more generic.

cmd/README.mysql.md
agasheesh: Good catch, updated these sample commands to use data subcommand instead.

cmd/README.mysql.md
agasheesh: Done.

cmd/source_profile_test.go
agasheesh: Done.

cmd/source_profile.go
agasheesh: Yes.

cmd/source_profile_test.go
agasheesh: This is an interesting corner case, and I suspect there may be many more such where users pass in invalid combination of flags and source-profile key-value pairs. I agree that we should catch as many of these invalid conditions as possible and return appropriate errors, but curious to know @bharadwaj-aditya's opinion as well before I make this change.

cmd/source_profile.go
agasheesh: Not sure if dydb is common enough to warrant adding support similar to postgres or pg. Does it make sense to defer this and add the same when users demand for the same?

cmd/schema.go
agasheesh: Done.

cmd/source_profile.go
agasheesh: Done.

cmd/source_profile.go
agasheesh: I intended to do implemented 2. but as pointed in the other comments, seems like log.Fatalf doesn't work as intended. I have switched to using panic() to return the appropriate message.

cmd/schema.go
agasheesh: Done.

cmd/schema.go
agasheesh: Thanks for this background, I have switched to panic for now everywhere I used log.Fatal previously.

cmd/source_profile.go
agasheesh: `source=` isn't part of the source profile but is rather is a separate flag (similarly for `target =`). I agree with catching DynamoDB case separately and have implemented the same. I have also improved the overall error messaging.

cmd/README.md
Deep1998: Makes sense

cmd/source_profile.go
agasheesh: I think just "s" works here, since the scope for this argument is pretty limited. Done.

cmd/target_profile.go
agasheesh: I think just "s" works here, since the scope for this argument is pretty limited. Done.

cmd/source_profile.go
agasheesh: Now, we return an error "please specify a valid source database using -source flag, received source = %v" from here. Also, added a top-level error just for checking empty source flag.

cmd/source_profile.go
agasheesh: I have added top-level error checking for source database which checks for empty source - "cannot leave -source flag empty, please specify source databases e.g., -source=postgres etc"; more specific checks are done separately where context is available.

cmd/source_profile.go
agasheesh: I will leave it as it is for now, but should be a quick follow up change if required.

cmd/source_profile_test.go
agasheesh: I now consider presence of either/both of "file" or "format" as the user's intention to use a file input. Have added corresponding error handling.

cmd/schema.go
agasheesh: I have added examples for both now. Done.

cmd/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+`-session` Specifies a session file that contains all schema and data 
+conversion state endcoded as JSON.
+
+### Source profile (`-source-profile`)
+
+HarbourBridge accepts the following options for --source-profile,
+specified as "key1=value1,key2=value,..." pairs:
+
+`-schema-sample-size` Specifies the number of rows to use for inferring schema
```
bharadwaj-aditya, line 317: suggest we move this to the dynamodb specific readme. We should keep only generic stuff here. I think it makes sense to move any DB specific stuff into the source profile, so this flag would go there

cmd/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+#### 1.1 pg_dump
+
+If you are using pg_dump (-driver=pg_dump), check that pg_dump is
+correctly configured to connect to your PostgreSQL
+database. Note that pg_dump uses the same options as psql to connect to your
+database. See the [psql](https://www.postgresql.org/docs/9.3/app-psql.html) and
+[pg_dump](https://www.postgresql.org/docs/9.3/app-pgdump.html) documentation.
+
+Access to a PostgreSQL database is typically configured using the
```
bharadwaj-aditya, line 398: We should probably remove this from the readme, and keep the env variables on the deprecation track. Or if it is vital to have, keep it as a secondary option in the readme

cmd/README.md
```diff
@@ -0,0 +1,521 @@
...
+_PGHOST_, _PGPORT_, _PGUSER_, _PGDATABASE_ environment variables,
+which are standard across PostgreSQL utilities.
+
+It is also possible to configure access via pg_dump's command-line options
```
bharadwaj-aditya, line 402: These flags will move into the source profile

cmd/README.md
```diff
@@ -0,0 +1,521 @@
...
+If you are using mysqldump (-driver=mysqldump), check that mysqldump is
+correctly configured to connect to your MySQL
+database via the command-line options `--host`, `--port`, and `--user`.
+Note that mysqldump uses the same options as mysql to connect to your
+database. See the [mysql](https://dev.mysql.com/doc/refman/8.0/en/mysql-commands.html)
+and [mysqldump](https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html) documentation.
+
+#### 1.4 Direct access to MySQL
+
+In this case, HarbourBridge connects directly to the MySQL
```
bharadwaj-aditya, line 424: same as postgres. Move these flags into the source profile

cmd/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+The following steps can help diagnose common issues encountered while running
+HarbourBridge.
+
+### 1. Verify driver configuration
+
+First, check that the driver is correctly configured to connect to your
+database. Driver configuration varies depending on the driver used.
+
+#### 1.1 pg_dump
```
bharadwaj-aditya, line 390: considering we have a DB specific readme, does it make sense to add it here ? We should just try to link it to the DB specific readme

cmd/schema.go
```diff
@@ -35,49 +34,75 @@ func (cmd *SchemaCmd) Synopsis() string {
...
-		dbName, err := conversion.GetDatabaseName(cmd.driverName, time.Now())
+		dbName, err := conversion.GetDatabaseName(driverName, time.Now())
 		if err != nil {
-			log.Fatalf("can't generate database name for prefix: %v\n", err)
+			panic(fmt.Errorf("can't generate database name for prefix: %v", err))
 		}
 		cmd.filePrefix = dbName + "."
 	}
 
+	schemaSampleSize := int64(100000)
```
bharadwaj-aditya, line 89: wondering if it makes sense to push the profile parsing logic into db specific functions or modules. There will be some duplication of code, but we can emit better error messages and separate out the code more cleanly

cmd/source_profile.go
```diff
@@ -0,0 +1,283 @@
...
+	pwd  string // Same as MYSQLPWD environment variable
+}
+
+func NewSourceProfileConnectionMySQL(params map[string]string) SourceProfileConnectionMySQL {
+	// TODO: Move parsing of environment variables in this function.
+	mysql := SourceProfileConnectionMySQL{}
+	if host, ok := params["host"]; ok {
+		mysql.host = host
+	}
+	if port, ok := params["port"]; ok {
```
bharadwaj-aditya, line 62: should set the default for stuff like port: 3306. Which rarely change.
Another option is to even default host to localhost

cmd/source_profile.go
```diff
@@ -0,0 +1,283 @@
...
+	db   string // Same as PGDATABASE environment variable
+	pwd  string // Same as PGPASSWORD environment variable
+}
+
+func NewSourceProfileConnectionPostgreSQL(params map[string]string) SourceProfileConnectionPostgreSQL {
+	pg := SourceProfileConnectionPostgreSQL{}
+	if host, ok := params["host"]; ok {
+		pg.host = host
+	}
+	if port, ok := params["port"]; ok {
```
bharadwaj-aditya, line 90: fallback to default port: 5432

cmd/source_profile.go
```diff
@@ -0,0 +1,283 @@
...
+type SourceProfileConfig struct {
+	path string
+}
+
+func NewSourceProfileConfig(path string) SourceProfileConfig {
+	return SourceProfileConfig{path: path}
+}
+
+type SourceProfile struct {
+	ty     SourceProfileType
```
bharadwaj-aditya, line 182: nit: suggest we change this variable to typ/stype/prof_type instead ? ty is not very clear

cmd/source_profile.go
```diff
@@ -0,0 +1,283 @@
...
+	file   SourceProfileFile
+	conn   SourceProfileConnection
+	config SourceProfileConfig
+}
+
+func (src SourceProfile) ToLegacyDriver(source string) (string, error) {
```
bharadwaj-aditya, line 188: add a comment on this function. The name is not very clear

cmd/source_profile.go
```diff
@@ -0,0 +1,283 @@
...
+			}
+		}
+	case SourceProfileTypeConfig:
+		return "", fmt.Errorf("specifying source-profile using config not implemented")
+	default:
+		return "", fmt.Errorf("invalid source-profile, could not infer type")
+	}
+}
+
+// Source profile is passed as a list of key value pairs on the command line.
```
bharadwaj-aditya, line 223: move this comment to the top of toLegacyDriver

cmd/source_profile.go
```diff
@@ -0,0 +1,283 @@
...
+// Format 3. Specify a config file that specifies source connection profile.
+//
+func NewSourceProfile(s string, source string) (SourceProfile, error) {
+	if source == "" {
+		return SourceProfile{}, fmt.Errorf("cannot leave -source flag empty, please specify source databases e.g., -source=postgres etc")
+	}
+
+	params := make(map[string]string)
+
+	if len(s) > 0 {
```
bharadwaj-aditya, line 247: we should use some existing library to parse csv instead, to handle scenarios where the parameters have commas within the quotes

cmd/source_profile_test.go
bharadwaj-aditya: this handling sounds reasonable to me

cmd/target_profile.go
```diff
@@ -0,0 +1,117 @@
...
+//
+// If dbname is not specified, then HarbourBridge will autogenerate the same
+// and create a database with the same name.
+//
+// Example: -target-profile="instance=my-instance1,dbname=my-new-db1"
+// Example: -target-profile="instance=my-instance1,dbname=my-new-db1,dialect=PostgreSQL"
+//
+func NewTargetProfile(s string) (TargetProfile, error) {
+	params := make(map[string]string)
+	if len(s) > 0 {
```
bharadwaj-aditya, line 84: move this parsing logic into a common utility between source and destination, and just work on the output which should be a map

bharadwaj-aditya (changes requested):

cmd/target_profile.go
agasheesh: Done.

cmd/source_profile.go
agasheesh: Done.

cmd/README.md
agasheesh: Done.

cmd/README.md
agasheesh: I agree that we should clean this up, but would prefer if we do that work as a follow up since it's independent from this PR.

cmd/README.md
agasheesh: We need to clean up our handling of environemnt variables vs connection paramters specified in source-profile. I think it makes sense to clean this up along with that change. There is a TODO in source_profile.go to make this change.

cmd/README.md
agasheesh: I wasn't sure what this means, but wonder if this is also something that should be cleaned up separately.

cmd/README.md
agasheesh: I plan to clean up the READMEs for environment variables as part of resolving the TODO in NewSourceProfileConnectionMySQL in source_profile.go.

cmd/source_profile.go
agasheesh: Umm, I thought we should add a separate comment to ToLegacyDriver based on the code review comment, which I have done now. ToLegacyDriver will get deprecated soon, and thus I was wondering if we should just keep this comment here.

cmd/source_profile.go
agasheesh: Done.

cmd/schema.go
agasheesh: That's the intention, but we would be able to do that in follow up refactoring when we pass around SourceProfile itself instead of passing around multiple arguments to conversion.SchemaConv and conversion.DataConv etc. Until then, this check needs to live here.

cmd/source_profile.go
agasheesh: Done.

cmd/source_profile.go
agasheesh: Done.

cmd/source_profile.go
agasheesh: Using ty to denote type is quite common in Go (I could easily find many examples in google3). Further the scope of ty is quite limited to switch statements only. Was it not clear from the usage in switch statements what does ty mean?

cmd/schema.go
```diff
@@ -16,11 +16,11 @@ import (
...
-	filePrefix       string
-	driverName       string
-	schemaSampleSize int64
-	dumpFilePath     string
-	targetDb         string
+	source        string
+	sourceProfile string
+	target        string
+	targetProfile string
+	filePrefix    string // TODO: move filePrefix to global flags
```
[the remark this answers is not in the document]
agasheesh: Done.

bharadwaj-aditya (approved):
LGTM

PR #209, opened 2021-10-28
Add support for Spanner database with PostgreSQL as a target database

agasheesh:

- [x] Tests pass


conversion/conversion.go
```diff
@@ -506,19 +506,33 @@ func CreateDatabase(ctx context.Context, adminClient *database.DatabaseAdminClie
...
-		ExtraStatements: schema,
-	})
+	}
+	if conv.TargetDb == TARGET_EXPERIMENTAL_POSTGRES {
+		// PG doesn't allow backticks around the database name.
+		req.CreateStatement = "CREATE DATABASE " + dbName
+		req.DatabaseDialect = adminpb.DatabaseDialect_POSTGRESQL
+	} else {
+		req.CreateStatement = "CREATE DATABASE `" + dbName + "`"
+		// PG doesn't allow extra statements during database creation.
```
nevinheintze, line 518: Cut-and-paste duplicate comment?
The one at line 513 seems appropriate, but this one seems out of place?

conversion/conversion.go
```diff
@@ -506,19 +506,33 @@ func CreateDatabase(ctx context.Context, adminClient *database.DatabaseAdminClie
...
+		schema := conv.SpSchema.GetDDL(ddl.Config{Comments: false, ProtectIds: true, Tables: true, ForeignKeys: false, TargetDb: conv.TargetDb})
+		req.ExtraStatements = schema
```
nevinheintze, line 520: Suggest:
req.ExtraStatements = conv.SpSchema.GetDDL(ddl.Config{Comments: false, ProtectIds: true, Tables: true, ForeignKeys: false, TargetDb: conv.TargetDb})

conversion/conversion.go
```diff
@@ -506,19 +506,33 @@ func CreateDatabase(ctx context.Context, adminClient *database.DatabaseAdminClie
...
-	schema := conv.SpSchema.GetDDL(ddl.Config{Comments: false, ProtectIds: true, Tables: true, ForeignKeys: false})
-	op, err := adminClient.CreateDatabase(ctx, &adminpb.CreateDatabaseRequest{
+	req := &adminpb.CreateDatabaseRequest{
 		Parent:          fmt.Sprintf("projects/%s/instances/%s", project, instance),
-		CreateStatement: "CREATE DATABASE `" + dbName + "`",
-		ExtraStatements: schema,
-	})
+	}
+	if conv.TargetDb == TARGET_EXPERIMENTAL_POSTGRES {
+		// PG doesn't allow backticks around the database name.
```
nevinheintze, line 513: Suggest rephrasing comment as:
// TARGET_EXPERIMENTAL_POSTGRES doesn't support:
// a) backticks around the database name, or
// b) DDL statements as part of a CreateDatabase operation (so schema
//      must be set using a separate UpdateDatabase operation).

And then deleting the PG comment in the else branch (lin 518).
I think splitting the comments is kind of confusing because both comments are really for this branch.

conversion/conversion.go
```diff
@@ -506,19 +506,33 @@ func CreateDatabase(ctx context.Context, adminClient *database.DatabaseAdminClie
...
+		req.CreateStatement = "CREATE DATABASE " + dbName
+		req.DatabaseDialect = adminpb.DatabaseDialect_POSTGRESQL
+	} else {
+		req.CreateStatement = "CREATE DATABASE `" + dbName + "`"
+		// PG doesn't allow extra statements during database creation.
+		schema := conv.SpSchema.GetDDL(ddl.Config{Comments: false, ProtectIds: true, Tables: true, ForeignKeys: false, TargetDb: conv.TargetDb})
```
nevinheintze, line 519: Not sure if the temporary variable is worth it?

sources/postgres/data.go
```diff
@@ -158,11 +158,14 @@ func convBytes(val string) ([]byte, error) {
 	return b, err
 }
 
-func convDate(val string) (civil.Date, error) {
+func convDate(conv *internal.Conv, val string) (interface{}, error) {
 	d, err := civil.ParseDate(val)
 	if err != nil {
 		return d, fmt.Errorf("can't convert to date: %w", err)
 	}
+	if conv.TargetDb == "experimental_postgres" {
```
nevinheintze, line 166: Suggest a comment along the lines of:

// In experimental_postgres mode, the civil.ParseDate call is only used as a validation step, and we
// throw its result away and return 'val'.

nevinheintze, line 166: Use TARGET_EXPERIMENTAL_POSTGRES constant?

sources/postgres/data.go
```diff
@@ -186,13 +189,16 @@ func convInt64(val string) (int64, error) {
 // into a string representing a valid Spanner numeric.
 // Ideally we would just return a *big.Rat, but spanner.Mutation
 // doesn't currently support use of *big.Rat.
-// TODO: return *big.Rat when client library supports it.
-func convNumeric(val string) (string, error) {
-	r := new(big.Rat)
-	if _, ok := r.SetString(val); !ok {
-		return "", fmt.Errorf("can't convert %q to big.Rat", val)
+func convNumeric(conv *internal.Conv, val string) (interface{}, error) {
+	if conv.TargetDb == "experimental_postgres" {
```
nevinheintze, line 193: Use TARGET_EXPERIMENTAL_POSTGRES constant?

sources/postgres/data.go
```diff
@@ -186,13 +189,16 @@ func convInt64(val string) (int64, error) {
 // into a string representing a valid Spanner numeric.
 // Ideally we would just return a *big.Rat, but spanner.Mutation
 // doesn't currently support use of *big.Rat.
-// TODO: return *big.Rat when client library supports it.
```
nevinheintze, line 189: The new code returns big.Rat for regularly spanner NUMERIC types --- does that mean spanner.Mutation now supports big.Rat?

If so, thanks for resolving the TODO.
Can you also clean up the comment -- we don't need the "Ideally we would just return a *big.Rat, but spanner.Mutation doesn't currently support use of *big.Rat.".
Also, please update the convNumeric comment - it no longer returns a string.

spanner/ddl/ast.go
```diff
@@ -112,7 +174,12 @@ func (c Config) quote(s string) string {
 // comment. These are returned as separate strings to support formatting
 // needs of PrintCreateTable.
 func (cd ColumnDef) PrintColumnDef(c Config) (string, string) {
-	s := fmt.Sprintf("%s %s", c.quote(cd.Name), cd.T.PrintColumnDefType())
+	var s string
+	if c.TargetDb == "experimental_postgres" {
```
nevinheintze, line 178: Use TARGET_EXPERIMENTAL_POSTGRES constant?

spanner/ddl/ast.go
```diff
@@ -181,16 +248,25 @@ func (ct CreateTable) PrintCreateTable(config Config) string {
...
 	var colComment []string
 	var keys []string
-	for i, cn := range ct.ColNames {
-		s, c := ct.ColDefs[cn].PrintColumnDef(config)
-		s = "\n    " + s
-		if i < len(ct.ColNames)-1 {
-			s += ","
-		} else {
-			s += " "
+	if config.TargetDb == "experimental_postgres" {
```
nevinheintze, line 251: Use TARGET_EXPERIMENTAL_POSTGRES constant?

spanner/ddl/ast.go
```diff
@@ -181,16 +248,25 @@ func (ct CreateTable) PrintCreateTable(config Config) string {
...
+		for i, cn := range ct.ColNames {
+			s, c := ct.ColDefs[cn].PrintColumnDef(config)
+			s = "\n    " + s
+			if i < len(ct.ColNames)-1 {
+				s += ","
+			} else {
+				s += ","
```
nevinheintze, line 258: Isn't this the same as the if branch i.e. we're appending "," in both cases?
Can't we just drop the if statement completely?

spanner/ddl/ast.go
```diff
@@ -181,16 +248,25 @@ func (ct CreateTable) PrintCreateTable(config Config) string {
...
+			}
+			col = append(col, s)
+			colComment = append(colComment, c)
+		}
+	} else {
+		for _, cn := range ct.ColNames {
```
nevinheintze, line 264: I'm confused: shouldn't this case be the same as the old code?
e.g. where did the "if i < len(ct.ColNames)-1" go to?

spanner/ddl/ast.go
```diff
@@ -211,7 +287,10 @@ func (ct CreateTable) PrintCreateTable(config Config) string {
 	if ct.Parent != "" {
 		interleave = ",\nINTERLEAVE IN PARENT " + config.quote(ct.Parent)
 	}
-	return fmt.Sprintf("%sCREATE TABLE %s (%s\n) PRIMARY KEY (%s)%s", tableComment, config.quote(ct.Name), cols, strings.Join(keys, ", "), interleave)
+	if config.TargetDb == "experimental_postgres" {
```
nevinheintze, line 290: Use TARGET_EXPERIMENTAL_POSTGRES constant?

spanner/ddl/ast_test.go
```diff
@@ -102,15 +101,43 @@ func TestPrintCreateTable(t *testing.T) {
...
+  col1 INT64 NOT NULL,
+  col2 STRING(MAX),
+  col3 BYTES(42),
+) PRIMARY KEY (col1 DESC)`,
+		},
+		{
+			"quote",
+			true,
+			t1,
+"CREATE TABLE `mytable` (\n"+
```
nevinheintze, line 121: For consistency, suggest using ` string here as well?

nevinheintze (commented):
I've done a fairly complete pass.
I haven't approved because I haven't done a really thorough review.
I'm happy to do so, or also happy for Aditya and Deep to to so and approve.

sources/postgres/data.go
agasheesh: We couldn't use the constant before because of circular dependency. Have now moved all constants to a top-level common package and now have updated to use constant as well.

sources/postgres/data.go
agasheesh: We couldn't use the constant before because of circular dependency. Have now moved all constants to a top-level common package and now have updated to use constant as well.

spanner/ddl/ast.go
agasheesh: We couldn't use the constant before because of circular dependency. Have now moved all constants to a top-level common package and now have updated to use constant as well.

spanner/ddl/ast.go
agasheesh: We couldn't use the constant before because of circular dependency. Have now moved all constants to a top-level common package and now have updated to use constant as well.

spanner/ddl/ast.go
agasheesh: We couldn't use the constant before because of circular dependency. Have now moved all constants to a top-level common package and now have updated to use constant as well.

sources/postgres/data.go
agasheesh: Yes, client library does support big.Rat now, but I ran into trouble updating the test cases to expect big.Rat instead of string. I have thus reverted back cleaning up this TODO for now.

conversion/conversion.go
agasheesh: Done.

conversion/conversion.go
agasheesh: This was leftover from some previous refactoring. Cleaned up.

sources/postgres/data.go
agasheesh: Thanks for flagging this, this is non-trivial code indeed and deserved it's own explanation.

spanner/ddl/ast_test.go
agasheesh: gofmt formatted ` ... ` strings as following:

			`CREATE TABLE mytable (
  col1 INT64 NOT NULL,
  col2 STRING(MAX),
  col3 BYTES(42),
) PRIMARY KEY (col1 DESC)`,

For consistency, have modified all strings to use " ... " concatanated with newlines.

spanner/ddl/ast.go
agasheesh: I have now cleaned up the handling here. It turned out that `if i < len(ct.ColNames)-1` wasn't needed for either dialect and that both the dialects could produce the same output for column definitions. Now, the only difference is how PRIMARY KEY and INTERLEAVE and printed and those are handled below separately.

spanner/ddl/ast.go
agasheesh: I have now cleaned up the handling here. It turned out that `if i < len(ct.ColNames)-1` wasn't needed for either dialect and that both the dialects could produce the same output for column definitions. Now, the only difference is how PRIMARY KEY and INTERLEAVE and printed and those are handled below separately.

spanner/ddl/ast.go
```diff
@@ -200,18 +267,28 @@ func (ct CreateTable) PrintCreateTable(config Config) string {
...
 		tableComment = "--\n-- " + ct.Comment + "\n--\n"
 	}
+
 	var interleave string
 	if ct.Parent != "" {
-		interleave = ",\nINTERLEAVE IN PARENT " + config.quote(ct.Parent)
+		if config.TargetDb == constants.TARGET_EXPERIMENTAL_POSTGRES {
+			interleave = " INTERLEAVE IN PARENT " + config.quote(ct.Parent)
+		} else {
+			interleave = ",\nINTERLEAVE IN PARENT " + config.quote(ct.Parent)
```
Deep1998, line 284: Do we need separate handling for pg and normal spanner? In PrintCreateTable just above, it seems the ',' was added at the end of stmt for both normal and pg spanner.

spanner/ddl/ast_test.go
```diff
@@ -37,7 +37,27 @@ func TestPrintScalarType(t *testing.T) {
...
+
+func TestPrintScalarTypePG(t *testing.T) {
+	tests := []struct {
+		in       Type
+		expected string
+	}{
+		{Type{Name: Bool}, "BOOL"},
+		{Type{Name: Int64}, "INT8"},
+		{Type{Name: Float64}, "FLOAT8"},
+		{Type{Name: String, Len: MaxLength}, "VARCHAR(2621440)"},
```
Deep1998, line 52: Nit: consider adding a test with 2621440 < len < MaxLength

Deep1998 (approved):
LGTM! Have a few minor comments and nits.

sources/postgres/toddl_test.go
```diff
@@ -155,9 +155,9 @@ func TestToExperimentalSpannerType(t *testing.T) {
 			"b": ddl.ColumnDef{Name: "b", T: ddl.Type{Name: ddl.Float64}},
 			"c": ddl.ColumnDef{Name: "c", T: ddl.Type{Name: ddl.Bool}},
 			"d": ddl.ColumnDef{Name: "d", T: ddl.Type{Name: ddl.String, Len: int64(6)}},
-			"e": ddl.ColumnDef{Name: "e", T: ddl.Type{Name: ddl.String, Len: ddl.MaxLength}},
-			"f": ddl.ColumnDef{Name: "f", T: ddl.Type{Name: ddl.String, Len: ddl.MaxLength}},
-			"g": ddl.ColumnDef{Name: "g", T: ddl.Type{Name: ddl.Json}},
+			"e": ddl.ColumnDef{Name: "e", T: ddl.Type{Name: ddl.Numeric}},
+			"f": ddl.ColumnDef{Name: "f", T: ddl.Type{Name: ddl.Date}},
+			"g": ddl.ColumnDef{Name: "g", T: ddl.Type{Name: ddl.String, Len: ddl.MaxLength}},
```
bharadwaj-aditya, line 160: JSON is being checked as a String. Is this intentional ?

spanner/ddl/ast.go
```diff
@@ -181,17 +252,13 @@ func (ct CreateTable) PrintCreateTable(config Config) string {
...
-	for i, cn := range ct.ColNames {
+	for _, cn := range ct.ColNames {
 		s, c := ct.ColDefs[cn].PrintColumnDef(config)
-		s = "\n    " + s
-		if i < len(ct.ColNames)-1 {
-			s += ","
-		} else {
-			s += " "
-		}
+		s = "\t" + s + ",\n"
```
bharadwaj-aditya, line 257: is this because a trailing , is supported by spanner ?

bharadwaj-aditya (commented):
Mostly clarifications

bharadwaj-aditya (approved):
LGTM

spanner/ddl/ast.go
```diff
@@ -83,6 +105,49 @@ func (ty Type) PrintColumnDefType() string {
...
+	// with IsArray set to true. In the unlikely event, set to string type.
+	if ty.IsArray {
+		str = PGVarchar
+		ty.Len = PGMaxLength
+	}
+	// PG doesn't support variable length Bytea and thus doesn't support
+	// setting length (or max length) for the Bytes.
+	if ty.Name == String || ty.Name == Date || ty.IsArray {
+		str += "("
+		if ty.Len == MaxLength || ty.Len == PGMaxLength {
```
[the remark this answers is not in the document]
agasheesh: We want to treat only these two specific cases as the cases with MaxLength. If the length parameter is somehow something else, it's better to catch that error and potentially fix the bug.

spanner/ddl/ast.go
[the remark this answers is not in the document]
agasheesh: Yes, "PG spanner only supports PRIMARY KEY() inside the CREATE TABLE() while Spanner DDL supports it on the outside". Added a comment explaining the difference.

spanner/ddl/ast_test.go
agasheesh: That's not a valid state i.e., it will never be the case that we generate a String with a max length other than MaxLength or PGMaxLength.

conversion/conversion.go
```diff
@@ -506,19 +506,33 @@ func CreateDatabase(ctx context.Context, adminClient *database.DatabaseAdminClie
...
+	op, err := adminClient.CreateDatabase(ctx, req)
 	if err != nil {
 		return fmt.Errorf("can't build CreateDatabaseRequest: %w", AnalyzeError(err, dbURI))
 	}
 	if _, err := op.Wait(ctx); err != nil {
 		return fmt.Errorf("createDatabase call failed: %w", AnalyzeError(err, dbURI))
 	}
 	fmt.Fprintf(out, "Created database successfully.\n")
+
+	if conv.TargetDb == TARGET_EXPERIMENTAL_POSTGRES {
```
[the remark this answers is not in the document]
agasheesh: Yes, this is a legacy flag and today we convert subcommand flags to legacy flags both for source-profilee and target-profile. We can get rid of this flag when we eventually move to passing around SourceProfile and TargetProfile across the codebase. We can also potentially store SourceProfile and TargetProfile in conv structure to validate when running data migration after a schema migration.

go.mod
```diff
@@ -29,3 +29,12 @@ require (
 // We need keep the replacement since google.golang.org/grpc/naming isn't
 // available in higher versions.
 replace google.golang.org/grpc => google.golang.org/grpc v1.29.1
+
+// DO NOT MERGE to mainline -- temporary workaround until PG support is
```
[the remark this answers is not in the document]
agasheesh: Done.

sources/mysql/mysqldump_test.go
```diff
@@ -728,7 +729,7 @@ CREATE TABLE test (a text PRIMARY KEY, b text);`,
 	INSERT INTO test (id, a, b, c) VALUES (1,'2019-10-29',4.444,5.44444);
 	`,
 			expectedData: []spannerData{
-				spannerData{table: "test", cols: []string{"id", "a", "b", "c"}, vals: []interface{}{int64(1), getDate("2019-10-29"), float64(4.444), "5.444440000"}}},
+				spannerData{table: "test", cols: []string{"id", "a", "b", "c"}, vals: []interface{}{int64(1), getDate("2019-10-29"), float64(4.444), big.NewRat(136111, 25000)}}},
```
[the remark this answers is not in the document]
agasheesh: I have resolved the TODO at https://github.com/cloudspannerecosystem/harbourbridge/blob/c44ff08a43ce2b36c7d6c511b80ca0c8a0ed4697/sources/postgres/data.go#L189 by returning big.Rat instead of String for Decimal/Numeric types to Go client library. Thus, the test case is updated appropriately.

sources/postgres/toddl_test.go
agasheesh: Yes, JSON is not supported in PG Spanner. Thus, we convert JSON to String for PostgreSQL as the target dialect. 

As a side note based on our offline discussion, we don't think we should take up the work for validating the Date in this PR for the cases when we send it as String to Spanner. I have reverted the changes for special handling of Date and created a separate Github issue which tracks Date validation in HarbourBridge. Our handling is not complete anyway today and we only accept ISO 8601 format.

spanner/ddl/ast.go
```diff
@@ -26,29 +26,51 @@ import (
...
+
 	// Numeric represent NUMERIC type.
 	Numeric string = "NUMERIC"
 	// Json represent JSON type.
 	Json string = "JSON"
+
 	// MaxLength is a sentinel for Type's Len field, representing the MAX value.
 	MaxLength = math.MaxInt64
+	// PGMaxLength represents sentinel for Type's Len field in PG.
+	PGMaxLength = 2621440
```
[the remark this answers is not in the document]
agasheesh: Sounds good, have moved all PG data types together.

spanner/ddl/ast.go
agasheesh: Yes, trailing , is supported by spanner and this change makes the code uniform across both dialects.

PR #211, opened 2021-11-02
Updated spanner emulator version

Deep1998:
We skip JSON checks for interleaved tables (becuase taht uses a different dump) and for direct connect tests (because we use mariadb:10.3 which does not support JSON type, it instead converts it to longtext).

### DIFF
--- a/testing/mysql/integration_test.go
+++ b/testing/mysql/integration_test.go
@@ -22,12 +22,14 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/cloudspannerecosystem/harbourbridge/common/constants"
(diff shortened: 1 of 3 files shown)


Deep1998:
Latest emulator image does not support json yet.

Deep1998:
ETA: Dec 21

testing/mysql/integration_test.go
```diff
@@ -341,6 +342,9 @@ func checkResults(t *testing.T, dbURI string) {
 	defer client.Close()
 
 	checkBigInt(ctx, t, client)
+	if json {
```
Deep1998, line 345: skip json checks for interleaved table test since that uses a different dump.

bharadwaj-aditya (changes requested):

bharadwaj-aditya (approved):
LGTM

PR #212, opened 2021-11-03
Add data and eval subcommands to harbourbridge command line interface

agasheesh:


testing/mysql/integration_test.go
```diff
@@ -218,6 +206,85 @@ func TestIntegration_MySQLDUMP_DataOnly(t *testing.T) {
...
+	defer dropDatabase(t, dbURI)
+	checkResults(t, dbURI)
+}
+
+func TestIntegration_MySQLDUMP_EvalSubcommand(t *testing.T) {
+	onlyRunForEmulatorTest(t)
+	tmpdir := prepareIntegrationTest(t)
+	defer os.RemoveAll(tmpdir)
+
+	dbName := "test-data-only-mode"
```
Deep1998, line 278: nit: test-eval-mode

cmd/common.go
```diff
@@ -41,3 +46,34 @@ func parseProfile(s string) (map[string]string, error) {
 	}
 	return params, nil
 }
+
+func getSpannerIds(ctx context.Context, targetProfile TargetProfile, now time.Time, driverName string, out *os.File) (string, string, string, error) {
```
Deep1998, line 50: nit: This function refers to spannerIDs differently from this [spannerID](https://github.com/cloudspannerecosystem/harbourbridge/blob/master/internal/mapping.go#L162). Should be fine functionally but wondering if we can think of a better name? Nothing comes to my mind right now.

cmd/data.go
```diff
@@ -0,0 +1,147 @@
...
+	flag.BoolVar(&cmd.skipForeignKeys, "skip-foreign-keys", false, "Skip creating foreign keys after data migration is complete (ddl statements for foreign keys can still be found in the downloaded schema.ddl.txt file and the same can be applied separately)")
+	f.StringVar(&cmd.filePrefix, "prefix", "", "File prefix for generated files")
+}
+
+func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
+	var err error
+	defer func() {
+		if err != nil {
+			fmt.Printf("FATAL error: %v\n", err)
+		}
```
Deep1998, line 64: I think we can later refactor these parts into schema specific changes and data specific, and call them from Execute() of schema, data and eval modes since there is much repetition across these modes.

Deep1998 (approved):
Overall looks very good, had a couple of minor comments. Thanks for the quick changes.

testing/mysql/integration_test.go
agasheesh: Done.

cmd/common.go
agasheesh: That's a good find. I have renamed it to getResourceIds since each of project, instance and database are various resources.

PR #213, opened 2021-11-03
Use source profile connection params for direct connect in subcommand…

Deep1998:
Refactor the **driverConfig** variable to **sqlConnectionStr** which is a more suitable name and avoids clashes.

For schema, data and eval in subcommand mode, we can pass the sqlConnectionStr throughout as a parameter. 

If `sqlConnectionStr == ""`,  it means user is running HB in legacy mode. We follow the original flow then where the params are fetched from env variables.

In subcommand mode i.e. `sqlConnectionStr != ""`, the expectation is user will provide all connection params in the source-profile. If host/user/dbname are missing, we throw an error asking to specify those. If the user does not provide any flag (to support backward compatibility), we read the params from env variables. This is only temporary and will be removed once we fully switch to subcommand mode.

For DynamoDB, the code path is kept same as before because the SDK itself only supports reading from the env variables. We cannot specify credentials in the go client.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -41,6 +41,8 @@ func CommandLine(ctx context.Context, driver, targetDb, dbURI string, dataOnly, 
 	var conv *internal.Conv
 	var err error
 	if !dataOnly {
+		// We pass an empty string to the sqlConnectionStr parameter as this is the legacy codepath,
+		// which reads the environment variables and constructs the string later on.
 		conv, err = conversion.SchemaConv(driver, "", targetDb, ioHelper, schemaSampleSize)
 		if err != nil {
 			return err
@@ -77,6 +79,8 @@ func CommandLine(ctx context.Context, driver, targetDb, dbURI string, dataOnly, 
 		return fmt.Errorf("can't create client for db %s: %v", dbURI, err)
 	}
 
+	// We pass an empty string to the sqlConnectionStr parameter as this is the legacy codepath,
+	// which reads the environment variables and constructs the string later on.
 	bw, err := conversion.DataConv(driver, "", ioHelper, client, conv, dataOnly)
 	if err != nil {
 		return fmt.Errorf("can't finish data conversion for db %s: %v", dbURI, err)
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -77,7 +77,7 @@ func CommandLine(ctx context.Context, driver, targetDb, dbURI string, dataOnly, 
 		return fmt.Errorf("can't create client for db %s: %v", dbURI, err)
 	}
 
-	bw, err := conversion.DataConv(driver, ioHelper, client, conv, dataOnly)
+	bw, err := conversion.DataConv(driver, "", ioHelper, client, conv, dataOnly)
 	if err != nil {
 		return fmt.Errorf("can't finish data conversion for db %s: %v", dbURI, err)
 	}
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -41,7 +41,7 @@ func CommandLine(ctx context.Context, driver, targetDb, dbURI string, dataOnly, 
 	var conv *internal.Conv
 	var err error
 	if !dataOnly {
-		conv, err = conversion.SchemaConv(driver, targetDb, ioHelper, schemaSampleSize)
+		conv, err = conversion.SchemaConv(driver, "", targetDb, ioHelper, schemaSampleSize)
 		if err != nil {
 			return err
 		}
--- a/cmd/data.go
+++ b/cmd/data.go
@@ -128,21 +128,8 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 		err = fmt.Errorf("can't create/update database: %v", err)
 		return subcommands.ExitFailure
 	}
-	sqlConnectionStr := ""
-	if sourceProfile.ty == SourceProfileTypeConnection {
-		// For DynamoDB, the SDK only reads credentials from env variables.
-		// There is no way to pass them through the code hence, don't need to override
-		// sqlConnectionStr for DynamoDB.
-		if sourceProfile.conn.ty == SourceProfileConnectionTypeMySQL {
-			connParams := sourceProfile.conn.mysql
-			sqlConnectionStr = conversion.GetMYSQLConnectionStr(connParams.host, connParams.port, connParams.user, connParams.pwd, connParams.db)
-		} else if sourceProfile.conn.ty == SourceProfileConnectionTypePostgreSQL {
-			connParams := sourceProfile.conn.pg
-			sqlConnectionStr = conversion.GetPGSQLConnectionStr(connParams.host, connParams.port, connParams.user, connParams.pwd, connParams.db)
-		}
-	}
 
-	bw, err := conversion.DataConv(driverName, sqlConnectionStr, &ioHelper, client, conv, true)
+	bw, err := conversion.DataConv(driverName, getSQLConnectionStr(sourceProfile), &ioHelper, client, conv, true)
 	if err != nil {
 		err = fmt.Errorf("can't finish data conversion for db %s: %v", dbURI, err)
 		return subcommands.ExitFailure
--- a/cmd/data.go
+++ b/cmd/data.go
@@ -128,8 +128,21 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 		err = fmt.Errorf("can't create/update database: %v", err)
 		return subcommands.ExitFailure
 	}
+	sqlConnectionStr := ""
+	if sourceProfile.ty == SourceProfileTypeConnection {
+		// For DynamoDB, the SDK only reads credentials from env variables.
+		// There is no way to pass them through the code hence, don't need to override
+		// sqlConnectionStr for DynamoDB.
+		if sourceProfile.conn.ty == SourceProfileConnectionTypeMySQL {
+			connParams := sourceProfile.conn.mysql
+			sqlConnectionStr = conversion.GetMYSQLConnectionStr(connParams.host, connParams.port, connParams.user, connParams.pwd, connParams.db)
+		} else if sourceProfile.conn.ty == SourceProfileConnectionTypePostgreSQL {
+			connParams := sourceProfile.conn.pg
+			sqlConnectionStr = conversion.GetPGSQLConnectionStr(connParams.host, connParams.port, connParams.user, connParams.pwd, connParams.db)
+		}
+	}
 
-	bw, err := conversion.DataConv(driverName, &ioHelper, client, conv, true)
+	bw, err := conversion.DataConv(driverName, sqlConnectionStr, &ioHelper, client, conv, true)
 	if err != nil {
 		err = fmt.Errorf("can't finish data conversion for db %s: %v", dbURI, err)
 		return subcommands.ExitFailure
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -97,27 +97,10 @@ func (cmd *EvalCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 		cmd.filePrefix = dbName + "."
 	}
 
-	schemaSampleSize := int64(100000)
-	sqlConnectionStr := ""
-	if sourceProfile.ty == SourceProfileTypeConnection {
-		if sourceProfile.conn.ty == SourceProfileConnectionTypeDynamoDB {
-			if sourceProfile.conn.dydb.schemaSampleSize != 0 {
-				schemaSampleSize = sourceProfile.conn.dydb.schemaSampleSize
-			}
-			// For DynamoDB, the SDK only reads credentials from env variables.
-			// There is no way to pass them through the code hence, don't need to override
-			// sqlConnectionStr for DynamoDB.
-		} else if sourceProfile.conn.ty == SourceProfileConnectionTypeMySQL {
-			connParams := sourceProfile.conn.mysql
-			sqlConnectionStr = conversion.GetMYSQLConnectionStr(connParams.host, connParams.port, connParams.user, connParams.pwd, connParams.db)
-		} else {
-			connParams := sourceProfile.conn.pg
-			sqlConnectionStr = conversion.GetPGSQLConnectionStr(connParams.host, connParams.port, connParams.user, connParams.pwd, connParams.db)
-		}
-	}
+	sqlConnectionStr := getSQLConnectionStr(sourceProfile)
 
 	var conv *internal.Conv
-	conv, err = conversion.SchemaConv(driverName, sqlConnectionStr, targetDb, &ioHelper, schemaSampleSize)
+	conv, err = conversion.SchemaConv(driverName, sqlConnectionStr, targetDb, &ioHelper, getSchemaSampleSize(sourceProfile))
 	if err != nil {
 		panic(err)
 	}
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -98,15 +98,26 @@ func (cmd *EvalCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 	}
 
 	schemaSampleSize := int64(100000)
+	sqlConnectionStr := ""
 	if sourceProfile.ty == SourceProfileTypeConnection {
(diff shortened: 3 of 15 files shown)


cmd/data.go
```diff
@@ -128,8 +128,21 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 		err = fmt.Errorf("can't create/update database: %v", err)
 		return subcommands.ExitFailure
 	}
+	sqlConnectionStr := ""
```
bharadwaj-aditya, line 131: extract this out into a method. The current method is already quite big

cmd/eval.go
```diff
@@ -98,15 +98,26 @@ func (cmd *EvalCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 	}
 
 	schemaSampleSize := int64(100000)
+	sqlConnectionStr := ""
 	if sourceProfile.ty == SourceProfileTypeConnection {
 		if sourceProfile.conn.ty == SourceProfileConnectionTypeDynamoDB {
 			if sourceProfile.conn.dydb.schemaSampleSize != 0 {
 				schemaSampleSize = sourceProfile.conn.dydb.schemaSampleSize
 			}
+			// For DynamoDB, the SDK only reads credentials from env variables.
```
bharadwaj-aditya, line 107: use the same method as data.go for this, to keep the logic common

cmd/schema.go
```diff
@@ -95,15 +95,26 @@ func (cmd *SchemaCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interfa
 	}
 
 	schemaSampleSize := int64(100000)
+	sqlConnectionStr := ""
 	if sourceProfile.ty == SourceProfileTypeConnection {
 		if sourceProfile.conn.ty == SourceProfileConnectionTypeDynamoDB {
 			if sourceProfile.conn.dydb.schemaSampleSize != 0 {
 				schemaSampleSize = sourceProfile.conn.dydb.schemaSampleSize
 			}
+			// For DynamoDB, the SDK only reads credentials from env variables.
```
bharadwaj-aditya, line 104: same as schema and eval.go. Pull this into a common method

conversion/conversion.go
```diff
@@ -125,55 +125,67 @@ func DataConv(driver string, ioHelper *IOStreams, client *sp.Client, conv *inter
 	}
 }
 
-func driverConfig(driver string) (string, error) {
+func generateSQLConnectionStr(driver string) (string, error) {
 	switch driver {
 	case POSTGRES:
-		return pgDriverConfig()
+		return generatePGSQLConnectionStr()
```
bharadwaj-aditya, line 131: will this print the error to specify env variables if the source connection profile structure is used instead ?

main.go
```diff
@@ -110,6 +110,9 @@ func main() {
 		flag.Parse()
 		os.Exit(int(subcommands.Execute(ctx)))
 	}
+	fmt.Printf("\nWarning: Found usage of deprecated flags. Support for these " +
+		"flags will be discontinued soon. It is recommended to use Harbourbridge " +
+		"in subcommand mode.\n\n")
```
bharadwaj-aditya, line 115: rephrase this to use HarbourBridge with connection profiles.

bharadwaj-aditya (changes requested):
Some refactoring required. Also, it is not clear where the driverConfig methods will be invoked from. Make sure you test the old flows, subcommand flows and web flows for log output also

conversion/conversion.go
Deep1998: This part of code is only hit if ran in the legacy mode. So if source profile is specified, this part of code will never be hit. 
That said, it is a bit confusing right now because both subcommand and legacy mode code paths intersect at SchemaConv() and DataConv()  calls and supporting both modes has caused the function calls to be a little haphazard. Once we deprecate the old mode, this half of the code should be unnecessary and can be scrapped.

cmd/source_profile.go
```diff
@@ -84,26 +103,42 @@ type SourceProfileConnectionPostgreSQL struct {
...
+	} else if !hostOk && !userOk && !dbOk && !portOk && !pwdOk {
+		// No connection params provided through source-profile. Fetching from env variables.
+		fmt.Printf("Connection parameters not specified in source-profile. Reading from environment variables...\n")
+		pg.host = os.Getenv("PGHOST")
+		pg.user = os.Getenv("PGUSER")
+		pg.db = os.Getenv("PGDATABASE")
+		pg.port = os.Getenv("PGPORT")
+		pg.pwd = os.Getenv("PGPASSWORD")
+	} else {
+		return pg, fmt.Errorf("please specify host, port, user and db_name in the source profile")
```
agasheesh, line 125: s\source profile\source-profile (here and elsewhere in case I missed this in my PRs)

conversion/conversion.go
```diff
@@ -103,7 +103,7 @@ func SchemaConv(driver string, targetDb string, ioHelper *IOStreams, schemaSampl
 	}
 }
 
-func DataConv(driver string, ioHelper *IOStreams, client *sp.Client, conv *internal.Conv, dataOnly bool) (*spanner.BatchWriter, error) {
+func DataConv(driver, sqlConnectionStr string, ioHelper *IOStreams, client *sp.Client, conv *internal.Conv, dataOnly bool) (*spanner.BatchWriter, error) {
```
agasheesh, line 106: This is okay for now, but should be cleaned up to be replaced with *cmd.SourceProfile such that the same can be nil when running in legacy mode. But the issue here is that cmd package depends on conversion package, and this would cause a circular dependency. Thus, happy to defer this cleanup for later.

agasheesh (commented):
The main change that I would ask for would be to also update NewSourceProfileConnectionDynamoDB in this same PR.

conversion/conversion.go
Deep1998: Yes will file an issue to track.

main.go
```diff
@@ -110,6 +110,9 @@ func main() {
 		flag.Parse()
 		os.Exit(int(subcommands.Execute(ctx)))
 	}
+	fmt.Printf("\nWarning: Found usage of deprecated flags. Support for these " +
+		"flags will be discontinued soon. It is recommended to use Harbourbridge " +
+		"in subcommand mode with connection profiles.\n\n")
```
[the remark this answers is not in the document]
Deep1998: Added link to the relevant documentation.

bharadwaj-aditya (approved):
LGTM

cmd/cmd.go
```diff
@@ -41,7 +41,7 @@ func CommandLine(ctx context.Context, driver, targetDb, dbURI string, dataOnly,
 	var conv *internal.Conv
 	var err error
 	if !dataOnly {
-		conv, err = conversion.SchemaConv(driver, targetDb, ioHelper, schemaSampleSize)
+		conv, err = conversion.SchemaConv(driver, "", targetDb, ioHelper, schemaSampleSize)
```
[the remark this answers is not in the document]
agasheesh: Maybe worth adding a comment though why an empty value of sqlConnectionStr is passed. Both here and below.

cmd/source_profile.go
```diff
@@ -55,27 +56,45 @@ type SourceProfileConnectionMySQL struct {
...
-	if port, ok := params["port"]; ok {
-		mysql.port = port
-	} else { // Set default port for mysql, which rarely changes.
-		mysql.port = "3306"
+	host, hostOk := params["host"]
+	user, userOk := params["user"]
+	db, dbOk := params["db_name"]
+	port, portOk := params["port"]
+	pwd, pwdOk := params["password"]
+	if !(hostOk || userOk || dbOk || portOk || pwdOk) {
```
agasheesh, line 66: Maybe add a comment saying on the lines of that we don't users to mix and match params from source-profile and environment variables. We try to get all the required params from source-profile first, and if none are set then we try to read the same from environment variables.

cmd/source_profile.go
```diff
@@ -55,27 +56,45 @@ type SourceProfileConnectionMySQL struct {
...
+	} else {
+		// Partial params provided through source-profile. Ask user to provide all.
+		return mysql, fmt.Errorf("please specify host, port, user and db_name in the source-profile")
 	}
-	if user, ok := params["user"]; ok {
-		mysql.user = user
+
+	// Throw same error if the input entered is empty.
+	if mysql.host == "" || mysql.user == "" || mysql.db == "" {
+		return mysql, fmt.Errorf("found empty string for host/user/db. please specify host, port, user and db_name in the source-profile")
```
agasheesh, line 86: This is incorrect since we could reach here if let's say a user has set only some of the environment variables but not all the required ones. I suggest that we rewrite the conditional such that we make the distinction whether we are reading from source-profile vs environment variables and handle those appropriately. Something as following:

```
if (hostOk || userOk || dbOk || portOk || pwdOk) {
    // Read connection params from source-profile.
    if (! (hostOk && userOk && dbOk ) ) {
       // Missing required connection params, return error.
    }
    // return error if mysql.host == "" etc
} else {
    // User did not specify connection params in source-profile at all, fall back to reading from environment variables.
   mysql.host = os.Getenv("MYSQLHOST")
   ....
   // return error if mysql.host == "" etc, error should mention environment variables and not params from source-profile
}
```

agasheesh: Similarly for NewSourceProfileConnectionPostgreSQL.

cmd/source_profile.go
```diff
@@ -86,58 +105,86 @@ type SourceProfileConnectionPostgreSQL struct {
...
+	return pg, nil
 }
 
 type SourceProfileConnectionDynamoDB struct {
-	awsAccessKeyID     string // Same as AWS_ACCESS_KEY_ID environment variable
-	awsSecretAccessKey string // Same as AWS_SECRET_ACCESS_KEY environment variable
-	awsRegion          string // Same as AWS_REGION environment variable
-	dydbEndpoint       string // Same as DYNAMODB_ENDPOINT_OVERRIDE environment variable
-	schemaSampleSize   int64  // Number of rows to use for inferring schema (default 100,000)
+	// We don't need any connection parameters here as SDK reads them from the env variables.
```
agasheesh, line 148: I don't see any harm in keeping the parsed flags for later reference either. It's similar to what we are doing for SourceProfileConnectionPostgreSQL and SourceProfileConnectionMySQL - there too we could just store `sqlConnectionStr` instead of storing all of host, port, user, db etc. In fact it takes more effort for a reader to understand the comment instead.

A case in point would be when we eventually move to passing around SourceProfile to SchemaConv and DataConv, then source database specific code such as setting environment variables for DynamoDB would move inside sources/dynamodb/.

cmd/source_profile.go
```diff
@@ -86,58 +105,86 @@ type SourceProfileConnectionPostgreSQL struct {
...
+		// Override env variables for these params as source-profile takes precedence.
+		os.Setenv("AWS_ACCESS_KEY_ID", awsAccessKeyID)
+		os.Setenv("AWS_SECRET_ACCESS_KEY", awsSecretAccessKey)
+		os.Setenv("AWS_REGION", awsRegion)
+		// Endpoint is optional. If not provided, the SDK infers endpoint from AWS_REGION.
+		// We should explicitly set it to "" if not provided to handle the case
+		// when the user already has an env variable DYNAMODB_ENDPOINT_OVERRIDE.
+		if !endpointOk {
+			dydbEndpoint = ""
+		}
```
agasheesh, line 182: Is this the right thing to do? i.e., should we try to clear the endpoint without telling the user - I think that will get rather confusing. Updating the environment variables is probably okay but not clearing them IMO. If user did wanted to clear it, they would explicitly set it to empty string in source-profile.

Also, the case of DynamoDB is different from PostgreSQL and MySQL I think. There it was clear that we want to deprecate setting env variables. But for the case of DynamoDB it's perfectly alright to set through environment variables (and is the preferred way), and I wonder if it's better to rewrite this as following instead?

if dydb.awsAccessKeyID, ok := params["awsAccessKeyID"]; ok {
   os.Setenv("AWS_ACCESS_KEY_ID", awsAccessKeyID)
}
...

and so on.

cmd/source_profile.go
```diff
@@ -154,12 +201,20 @@ func NewSourceProfileConnection(source string, params map[string]string) (Source
 	case "mysql":
 		{
 			conn.ty = SourceProfileConnectionTypeMySQL
-			conn.mysql = NewSourceProfileConnectionMySQL(params)
+			mysqlconn, err := NewSourceProfileConnectionMySQL(params)
+			conn.mysql = mysqlconn
```
agasheesh, line 205: do we need mysqlconn temporary variable? (similarly below)

cmd/source_profile.go
```diff
@@ -230,8 +285,8 @@ func (src SourceProfile) ToLegacyDriver(source string) (string, error) {
 	}
 }
 
-// Source profile is passed as a list of key value pairs on the command line.
-// Following 3 formats are supported as valid source profiles.
+// Source-profile is passed as a list of key value pairs on the command line.
+// Following 3 formats are supported as valid source-profiles.
```
agasheesh, line 289: (nit) This looks a little odd - how about "Flag source-profile is passed ....... as a valid source-profile"

cmd/source_profile_test.go
```diff
@@ -53,3 +53,100 @@ func TestNewSourceProfileFile(t *testing.T) {
...
+		{
+			name:          "no mandatory params but optional provided",
+			params:        map[string]string{"dydbEndpoint": "b"},
+			errorExpected: true,
+		},
+		{
+			name:          "partial mandatory params",
+			params:        map[string]string{"awsRegion": "a"},
+			errorExpected: true,
+		},
```
agasheesh, line 140: This is the error case that I am not too sure about for DynamoDB (as explained in the comment in source_profile.go also). Since setting through environment variables will stay around for DynamoDB, I wonder if it makes sense to treat all params as optional and best-effort only.

conversion/conversion.go
```diff
@@ -73,10 +73,10 @@ var (
 	MaxWorkers = 20
 )
 
-func SchemaConv(driver string, targetDb string, ioHelper *IOStreams, schemaSampleSize int64) (*internal.Conv, error) {
+func SchemaConv(driver, sqlConnectionStr, targetDb string, ioHelper *IOStreams, schemaSampleSize int64) (*internal.Conv, error) {
```
agasheesh, line 76: These public functions now deserve a comment - even if it just explains the behavior for sqlConnectionStr (which is that it's expected to be empty for dump files and dynamodb -- and for direct connect to mysql and postgres, empty value reflects that the command is being executed from legacy mode with global flags). Similarly for DataConv.

testing/mysql/integration_test.go
```diff
@@ -126,6 +126,29 @@ func TestIntegration_MYSQLDUMP_Command(t *testing.T) {
...
+	t.Parallel()
+
+	tmpdir := prepareIntegrationTest(t)
+	defer os.RemoveAll(tmpdir)
+
+	dbName := "mysql-dc-eval-src-params"
+	dbURI := fmt.Sprintf("projects/%s/instances/%s/databases/%s", projectID, instanceID, dbName)
+	filePrefix := filepath.Join(tmpdir, dbName+".")
+
+	host, user, db_name, password := os.Getenv("MYSQLHOST"), os.Getenv("MYSQLUSER"), os.Getenv("MYSQLDATABASE"), os.Getenv("MYSQLPWD")
```
agasheesh, line 140: how about we extend this integration test slightly and have a helper functions that lets you clear the environment variables and set them back?

cmd/source_profile.go
```diff
@@ -117,27 +172,30 @@ type SourceProfileConnectionDynamoDB struct {
...
+	}
+	if dydb.awsSecretAccessKey, ok = params["awsSecretAccessKey"]; ok {
+		os.Setenv("AWS_SECRET_ACCESS_KEY", dydb.awsAccessKeyID)
+	}
+	if dydb.awsRegion, ok = params["awsRegion"]; ok {
+		os.Setenv("AWS_REGION", dydb.awsAccessKeyID)
+	}
+	if dydb.dydbEndpoint, ok = params["dydbEndpoint"]; ok {
+		os.Setenv("DYNAMODB_ENDPOINT_OVERRIDE", dydb.awsAccessKeyID)
+	}
```
Deep1998, line 198: dydb param override

bharadwaj-aditya: This looks fine

PR #214, opened 2021-11-03
using unique keys instead of synth keys if unique key exist

### LINKED ISSUE #35: Use UNIQUE + NOT NULL when there is no primary key
Currently, if a table does not have a primary key, we create a synthetic primary key.

However, some tables have columns that are UNIQUE and NOT NULL but are not primary keys. Such columns would be good candidates to use as the primary key (and better than adding synthetic keys).

shreyakhajanchi:
Fixes #35 - Use UNIQUE key if PRIMARY key is absent and UNIQUE key is present.

This change checks if primary key exists for a table, if not ,it further checks if a UNIQUE key exists, if yes then it uses the UNIQUE key as PRIMARY key instead of creating a synthetic primary key.


sources/common/infoschema.go
```diff
@@ -162,8 +162,19 @@ func processTable(conv *internal.Conv, db *sql.DB, table SchemaAndName, infoSche
 	}
 	name := infoSchema.GetTableName(table.Schema, table.Name)
 	var schemaPKeys []schema.Key
-	for _, k := range primaryKeys {
-		schemaPKeys = append(schemaPKeys, schema.Key{Column: k})
+	if len(primaryKeys) == 0 {
```
Deep1998, line 165: Suggest adding a small comment regarding the behaviour.

Deep1998 (approved):
Thanks for the quick fix. Looks good! One minor comment.

Deep1998:
Similar changes should be replicated to dump file flow as well before merging.

sources/common/infoschema.go
```diff
@@ -162,8 +162,19 @@ func processTable(conv *internal.Conv, db *sql.DB, table SchemaAndName, infoSche
...
+		for col, constraintList := range constraints {
+			for _, constraint := range constraintList {
+				if constraint == "UNIQUE" {
+					schemaPKeys = append(schemaPKeys, schema.Key{Column: col})
+					break
+				}
```
Deep1998, line 171: This piece currently takes all cols with UNIQUE as part of primary key. Should we just take one rather than all?

Deep1998 (commented):
Another minor comment.

internal/convert.go
```diff
@@ -256,12 +258,36 @@ func (conv *Conv) SampleBadRows(n int) []string {
...
+			if len(ct.Indexes) != 0 {
+				for i, index := range ct.Indexes {
+					if index.Unique {
+						for _, indexKey := range index.Keys {
+							ct.Pks = append(ct.Pks, ddl.IndexKey{Col: indexKey.Col, Desc: indexKey.Desc})
+							conv.UniquePKey[t] = append(conv.UniquePKey[t], indexKey.Col)
+						}
+						primaryKeyPopulated = true
+						if len(ct.Indexes) == 1 {
+							ct.Indexes = nil
```
Deep1998, line 273: Should this be nil or []? It seems like in other places we handle it as [].

internal/convert.go
```diff
@@ -256,12 +258,36 @@ func (conv *Conv) SampleBadRows(n int) []string {
...
+						} else {
+							tempIndex := ct.Indexes[:i]
+							tempIndex = append(tempIndex, ct.Indexes[i+1:]...)
+							ct.Indexes = tempIndex
```
Deep1998, line 277: nit: This piece works correctly but I think we can refactor this whole if-else block with `ct.Indexes = append(ct.Indexes[:i], ct.Indexes[i+1:]...)` for better readability. Even when ct.Indexes is singleton, this would yield an empty list which is consistent with other places.

internal/convert_test.go
```diff
@@ -84,25 +84,161 @@ func TestGetBadRows(t *testing.T) {
...
+						"a": {Name: "a", T: ddl.Type{Name: ddl.Int64}},
+						"b": {Name: "b", T: ddl.Type{Name: ddl.Float64}},
+					},
+					Pks: []ddl.IndexKey{{Col: "a"}, {Col: "b"}}},
+			},
+			uniqueKey: []string{"a", "b"},
+		},
+		{
+			name: "in case two unique keys are present first one is taken as primary key",
+			inputSchema: map[string]ddl.CreateTable{
```
Deep1998, line 148: Very thorough tests. Consider making it exhaustive by adding a case when there is already a primary key and expected schema is same as input schema.

internal/report.go
```diff
@@ -161,6 +163,13 @@ func buildTableReportBody(conv *Conv, srcTable string, issues map[string][]Schem
 				l = append(l, fmt.Sprintf("Column '%s' was added because this table didn't have a primary key. Spanner requires a primary key for every table", *syntheticPK))
 			}
 		}
+		if uniquePK != nil {
+			// Warning about using a column with unique constraint as primary key
+			// in case primary key is absent.
+			if p.severity == warning {
+				l = append(l, fmt.Sprintf("Column '%s' was used as primary key because it had UNIQUE constraint and this table didn't have a primary key. Spanner requires a primary key for every table", strings.Join(uniquePK, ", ")))
```
Deep1998, line 170: We don't mention that we dropped the UNIQUE constraint. How about we rephrase this to
"UNIQUE constraint on column(s) '%s' replaced with primary key since this table didn't have one. Spanner requires a primary key for every table"

Deep1998 (commented):
Thanks for the fix! Looks very good. Have a few nits.

internal/convert.go
bharadwaj-aditya: This block is nested to 8 levels. Makes it pretty difficult to read.
Suggest using the approach suggested here, or moving the block out into a new method

bharadwaj-aditya (changes requested):

internal/convert_test.go
shreyakhajanchi: Added test case

internal/convert.go
shreyakhajanchi: thanks for pointing out, modified.

internal/report.go
shreyakhajanchi: rephrased as suggested

internal/convert.go
shreyakhajanchi: Modified the code block to use suggested approach.

bharadwaj-aditya (approved):
LGTM

Deep1998 (approved):

PR #216, opened 2021-11-03
Cleanup integration tests to replace cmd.CommandLine with bash execution

agasheesh:
We had two exact same copies of some of the integration tests before, where one of them would use cmd.CommandLine as the point of entry whereas the other would use "bash -c harbourbridge -flag1 ..." as the entry point. For the end to end integration tests it doesn't add a lot of value to create duplicate integration tests just calling into cmd.CommandLine which we plan to deprecate anyway.

### DIFF
--- a/testing/dynamodb/integration_test.go
+++ b/testing/dynamodb/integration_test.go
@@ -26,8 +26,8 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


nevinheintze (approved):
Looks good -- thanks for the cleanup.

PR #218, opened 2021-11-09
Update all the READMEs to reflect new subcommands and flags structure

agasheesh:
- [x] Tests pass


README.md
```diff
@@ -31,34 +31,50 @@ critical things like tuning performance and getting the most out of
...
+at specific sections for PostgreSQL, MySQL etc for detailed examples.
+
+For example, to generate Spanner schema from a PostgreSQL dump file, run
 
 ```sh
-# By default, the driver is "pg_dump".
-pg_dump mydb | harbourbridge
-# Or,
-pg_dump mydb | harbourbridge -driver=pg_dump
+# Generate pg_dump file.
```
nevinheintze, line 40: This generates a fairly intimidating set of instructions! And it's not clear as a first time user what I should do since many of the instructions are alternatives: https://screenshot.googleplex.com/m3vBNwSeKJirNSL

I would keep the first set of instructions super-simple. I think the simpler the better.
How about just:

# Generate pg_dump file.
pg_dump mydb > mydb.pg_dump
# Run HarbourBridge
harbourbridge schema -source=postgresql < mydb.pg_dump

And then have a separate block of instructions for a GCS bucket. Maybe something like:

You can also specify path to a file stored in GCS (Google Cloud Storage) via:

harbourbridge schema -source=postgresql -source-profile="format=dump,file=gs://<bucket_name>/<file_path>"

And just leave it at that. 
I would drop the other options -- this is the intro part of the README, and I don't think we need to be complete here.

sources/dynamodb/README.md
```diff
@@ -32,16 +32,22 @@ To use the tool on a DynamoDB database (it will migrate all tables),
...
 ```
 
-Due the schemaless nature of DynamoDB, the tool infers the schema based on a
-certain amount of sampled data, by default, 100000 rows. If a table has more
-rows than the default value, we only use 100000 rows for estimating the schema.
-You can change this value via the flag `schema-sample-size`. For example,
+### Source profile (`-source-profile`)
+
+HarbourBridge accepts the following options for --source-profile for DynamoDB,
+specified as "key1=value1,key2=value,..." pairs:
```
nevinheintze, line 41: Maybe drop "specified as "key1=value1,key2=value,..." pairs"
since there is only one option to specify?

sources/mysql/README.md
```diff
@@ -24,41 +24,41 @@ on a MySQL database (via go's database/sql package).
 To use HarbourBridge on a MySQL database called mydb using mysqldump, run:
```
nevinheintze, line 24: This first command just does schema conversion, and then a few examples later there is an example of data conversion in the context of the instance parameter. This is a bit confusing (and isn't the data conversion example incomplete -- don't you need a session file?). I think you need to explicitly introduce schema and data mode. How about something like:

You can directly pipe the output of mysqldump to HarbourBridge. For example:

```sh
mysqldump mydb | harbourbridge schema -source=mysql
```

will build a Spanner schema from the MySQL database 'mydb'. You can also run the tool on an existing mysqldump file:

```sh
harbourbridge schema -source=mysql < my_mysqldump_file
```

Both of these are examples of HarbourBridge schema subcommand. There are two additional subcommands: data and eval. The data subcommand does data conversion, and the eval subcommand performs an all-in-one combinated schema and data conversion.

then give examples of the two commands

and then describe the target-profile options.

Hmm. I guess one issue here is the repetition of this for both postgres and dynamo.

sources/postgres/README.md
```diff
@@ -24,41 +24,44 @@ on a PostgreSQL database (via go's database/sql package).
 To use HarbourBridge on a PostgreSQL database called mydb using pg_dump output,run:
 
 ```sh
```
nevinheintze, line 26: Same comment as for mysql. I think we need to introduce the subcommands at some point.

sources/dynamodb/README.md
```diff
@@ -32,16 +32,22 @@ To use the tool on a DynamoDB database (it will migrate all tables),
 run
 
 ```sh
-harbourbridge -driver=dynamodb
+harbourbridge schema -source=dynamodb
```
nevinheintze, line 35: This will just do schema conversion?
However the description in line 31 indicates ("migrate all tables") indicates full migration and not just schema migration?

Does it make sense to use the eval subcommand here?

BTW: while thinking about the README, I'm having second thoughts about 'eval' as a subcommand name. It kind of sounds like some kind of experimental mode for HarbourBridge. I'm wondering if the long-winded schema-and-data might be better/self-documenting? What do you think?

README.md
```diff
@@ -31,34 +31,50 @@ critical things like tuning performance and getting the most out of
...
+pg_dump mydb > mydb.pg_dump
```
nevinheintze, line 41: Am wondering why not keep the original command structure and just pipe the output of pg_dump to HarbourBridge? i.e. something like:

pg_dump mydb | harbourbridge schema -source=postgresql

Or do you think that the dump file is the more common usage and we should focus on that?

nevinheintze (commented):
First round of comments.
I definitely agree with your suggestion earlier today that we combine everything into one README -- this is kind of getting unwieldy. But let's not do it in this PR.

README.md
agasheesh: I have now dropped the examples from the Intro section altogether and we simply ask users to look at the quick start guide. This takes care of the duplication as well as moves documentation about GCS to where we describe `file` param in `-source-profile` flag.

sources/dynamodb/README.md
agasheesh: Done.

sources/dynamodb/README.md
```diff
@@ -32,16 +32,22 @@ To use the tool on a DynamoDB database (it will migrate all tables),
...
+### Source profile (`-source-profile`)
+
+HarbourBridge accepts the following options for --source-profile for DynamoDB,
+specified as "key1=value1,key2=value,..." pairs:
+
+`-schema-sample-size` Due to the schemaless nature of DynamoDB, the tool infers
+the schema based on a certain amount of sampled data, by default, 100000 rows.
+If a table has more rows than the default value, we only use 100000 rows for 
+estimating the schema. This flag lets you specify the number of rows to use for
+inferring schema. The default value is 100,000.
```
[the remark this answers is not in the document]
agasheesh: Done.

README.md
agasheesh: From a performance point of view, it's not ideal to pipe rather large dump files to HarbourBridge since stdin provided in this manner is not seekable and we end up creating a copy. I have thus tried to move away from providing examples where we suggest to just pipe the output for pg_dump to HB. I have though deleted these examples altogether as explained in an earlier reply.

sources/mysql/README.md
agasheesh: I have now added examples of schema, data and oneshot to all the source database specific READMEs. I didn't see the repetition across different READMEs as an issue since each of those READMEs along with the main README should be complete on it's own. It's likely a give user will focus on only one of the source databases at a time. Examples with data subcommand now also talk about requiring session file.

sources/dynamodb/README.md
agasheesh: I have fixed the text to say schema migration and have later provided an example of running with oneshot subcommand. As I am reading the documentation now, oneshot seems to make sense. Let me know your thoughts.

README.md
```diff
@@ -351,14 +392,29 @@ data conversion can be found here:
...
+See [DynamoDB example usage](sources/dynamodb/README.md#example-dynamodb-usage)
+for troubleshooting direct access to DynamoDB.
 
-#### 1.1 pg_dump
+### 2. Generating dump files
 
-If you are using pg_dump (-driver=pg_dump), check that pg_dump is
+#### 2.1 Generating pg_dump file
+
+If you are using pg_dump , check that pg_dump is
```
bharadwaj-aditya, line 417: It would make sense to move this into the source specific readme

cmd/oneshot.go
```diff
@@ -13,8 +13,8 @@ import (
 	"github.com/google/subcommands"
 )
 
-// EvalCmd struct with flags.
-type EvalCmd struct {
+// OneshotCmd struct with flags.
```
bharadwaj-aditya, line 16: Oneshot might be misleading if/when we add support to stop and start migrations.
I think being descriptive and saying schema-and-data in this case might make more sense.

cmd/source_profile.go
```diff
@@ -119,22 +119,22 @@ type SourceProfileConnectionDynamoDB struct {
 func NewSourceProfileConnectionDynamoDB(params map[string]string) (SourceProfileConnectionDynamoDB, error) {
 	// TODO: Populate environment variables for DynamoDB if passed as flags.
 	dydb := SourceProfileConnectionDynamoDB{}
-	if awsAccessKeyID, ok := params["awsAccessKeyID"]; ok {
+	if awsAccessKeyID, ok := params["aws-access-key-id"]; ok {
```
bharadwaj-aditya, line 122: do we expect all the keys to be hiphenated ? or is this just the case for AWS ?

README.md
```diff
@@ -17,48 +17,27 @@ For more details on schema customization and use of the schema assistant, see
...
-HarbourBridge accepts pg_dump/mysqldump's standard plain-text format, but 
-not archive or custom formats. More details on usage can be found in 
-[Example usage](#example-usage) section.
-
-HarbourBridge automatically determines the cloud project and Spanner instance to
-use, and generates a new Spanner database name (prefixed with `{driver}_` and
-today's date). Command-line flags can be used to explicitly set the Spanner
-instance or database name. See [Options](#options).
+HarbourBridge automatically determines the cloud project to use, and generates
+a new Spanner database name (prefixed with `{source}_` and today's date).
```
nevinheintze, line 38: I'm not sure {source} will make too much sense for first-time readers.
Suggest just deleting "prefixed with `{source}_` and today's date)".

README.md
```diff
@@ -143,34 +123,43 @@ This workflow also allows you to modify or customize the HarbourBridge codebase.
 To use the tool on a PostgreSQL database called mydb, run
 
 ```sh
-pg_dump mydb | harbourbridge -driver=pg_dump
+pg_dump mydb > mydb.pg_dump
+harbourbridge schema -source=postgresql < mydb.pg_dump
```
nevinheintze, line 127: This command is for schema mode, but the text below (the paragraph starting "This command ...") describes what happens for schema-and-data mode.

We should be consistent here. A couple of options:

Option 1:  Just focus on schema migration. Suggest changing the text above the command to something like:
```
 To use the tool for schema-migration on a PostgreSQL database called mydb, run
```
and then in the description below lay out what the tool does.

Option 2: Focus on schema and data migration. In this case change the command and keep the existing descriptions.

I've a slight preference for Option 2 because it shows the full range of HarbourBridge functionality, and is a good starting point for first users. But I don't care strongly.

cmd/oneshot.go
nevinheintze: Yes, I've also been thinking about this a bunch too and have downloaded PR218 and have been playing with the command-line interface.

I apologize Asheesh -- when we discussed this previously, I said oneshot was ok, but I'm less sure now.

The problem with oneshot is that it isn't clear what it does. Moreover, oneshot has some misleading connotations e.g. the standard definition of one-shot is "something that happens only once" (e.g. https://dictionary.cambridge.org/dictionary/english/one-shot). We're using oneshot to mean "all-in-one" but that's a less common usage I think.

I know I'm backtracking a bit here, but I agree with Aditya. I think schema-and-data is better (even if its too long).
Some alternative options:
 schema-and-data
 schema+data
 schema-data
My quick take on these:
 schema-and-data: a solid option, but kind of verbose
  schema+data: a good option, but is it ok to have "+" in the command name?
  schema-data: an ok option, but potentially confusing: do you mean the schema's data?

README.md
```diff
@@ -117,21 +96,22 @@ Spanner instance.
...
 
 You can make a copy of the HarbourBridge codebase from the github repository
-and use "go run":
+and use "go run".
 
 ```sh
 git clone https://github.com/cloudspannerecosystem/harbourbridge
 cd harbourbridge
-go run github.com/cloudspannerecosystem/harbourbridge -driver=pg_dump < my_pg_dump_file
+go run github.com/cloudspannerecosystem/harbourbridge schema -source=postgres < my_pg_dump_file
```
nevinheintze, line 110: How about making this something that a first-time user can cut-and-paste and it'll just work?
In which case, maybe change the go run command to be:

go run github.com/cloudspannerecosystem/harbourbridge help

I know this is not very useful, but I think having something users can simply cut-and-paste for installation instructions is a big plus.

README.md
```diff
@@ -143,34 +123,43 @@ This workflow also allows you to modify or customize the HarbourBridge codebase.
...
-set the `--instance` flag. The new Cloud Spanner database will have a name of 
-the form `{DRIVER}_{DATE}_{RANDOM}`, where`{DRIVER}` is the drivername 
-specified,`{DATE}`is today's date, and`{RANDOM}` is a random suffix for 
-uniqueness.
+database with the data from the source database.
+If the project has multiple instances, then list of available instances
+will be shown and you will have to pick one of the available instances and
+set the `--instance` flag in target-profile. The new Cloud Spanner database
+will have a name of the form `{SOURCE}_{DATE}_{RANDOM}`, where`{SOURCE}` is
+the source specified,`{DATE}`is today's date, and`{RANDOM}` is a random
```
nevinheintze, line 161: Maybe "is the value of the source flag" instead of "the source specified"?

README.md
```diff
@@ -273,75 +261,128 @@ By default, these files are prefixed by the name of the Spanner database (with a
...
 
-`-session` Specifies a session file that contains all schema and data 
+`-session` Specifies a session file that contains all schema and data
 conversion state endcoded as JSON.
 
-`-dump-file` Specifies the path where the dump file is stored. 
-If the location is Google Cloud Storage (GCS), you can use the following format:
-`-dump-file=gs://{bucket_name}/{path/to/dump_file}`. Please ensure you have
-read pemissions to the GCS bucket you would like to use. 
+#### `-source-profile`
```
nevinheintze, line 334: The formatting of this looks kind of odd: there is a list of command line flags, and then it continues as a header.
Suggest the following change: extend the Command line flags section to be a full list of flags i.e add:

`-source-profile` Specifies detailed parameters for the source database such as connection parameters. See Source Profile (linked to section) for details.

`-target-profile` Specifies detailed parameters for the target database. See Target Profile (linked to section) for details.

And then rename the sections "Source Profile" and "Target Profile" respectively.

sources/dynamodb/README.md
```diff
@@ -1,54 +1,100 @@
...
-using data from an existing PostgreSQL or MySQL or DynamoDB database. This
+HarbourBridge is a stand-alone open source tool for Cloud Spanner evaluation and migration,
+using data from an existing database. This
 README provides details of the tool's DynamoDB capabilities. For general
-HarbourBridge information see this [README](https://github.com/cloudspannerecosystem/harbourbridge#harbourbridge-turnkey-spanner-evaluation).
+HarbourBridge information see this [README](https://github.com/cloudspannerecosystem/harbourbridge#harbourbridge-spanner-evaluation-and-migration).
 
 ## Example DynamoDB Usage
 
-The following examples assume `harbourbridge` alias has been setup as
```
nevinheintze, line 10: I think it is worth noting the alias and referring to the section in the main README. How about:

The following examples assume a `harbourbridge` alias has been setup as described in the Installing HarbourBridge section of the main README (with link).

nevinheintze: And same comment for the mysql/README.md and postgres/README.md.

sources/mysql/README.md
```diff
@@ -1,83 +1,88 @@
...
+harbourbridge schema -source=mysql -source-profile="host=<>,port=<>,user=<>,db_name=<>"
 ```
 
-It is assumed that _MYSQLHOST_, _MYSQLPORT_, _MYSQLUSER_, _MYSQLDATABASE_ environment
-variables are set. Password can be specified either in the _MYSQLPWD_ environment
-variable or provided at the password prompt.
+Parameters `port` and `password` are optional. Port (`port`) defaults to `3306`
+for MySQL source. Password can be provided at the password prompt.
+
+(⚠ Deprecated ⚠) Set environment variables `MYSQLHOST`, `MYSQLPORT`,
```
nevinheintze, line 80: Not suggest a change here, but just noting our plan that we should deprecate the MySQL environment variables because they are non-standard.
But we should keep the Postgres ones, because they are standard (and commonly used).

sources/mysql/README.md
```diff
@@ -259,12 +264,12 @@ type, which supports UTF-8.
 As noted earlier when discussing [schema conversion of
 Spatial datatype](#spatial-datatype), Spanner does not support spatial datatypes and are
 mapped to `STRING(MAX)` Spanner type. Data conversion for spatial datatypes is different
-for both `mysql` and `mysqldump` driver.
+for both `mysql` direct connect and `mysqldump` file.
```
nevinheintze, line 267: Suggest rephrasing:

There are differences in data conversion for spatial datatypes depending on whether direct connect or a `mysqldump` file is used:

sources/postgres/README.md
```diff
@@ -1,83 +1,92 @@
...
-To use HarbourBridge on a PostgreSQL database called mydb using pg_dump output,run:
+The tool can used to migrate schema from an existing pg_dump file:
 
 ```sh
-pg_dump mydb | harbourbridge -driver=pg_dump
+harbourbridge schema -source=postgresql < my_pg_dump_file
 ```
 
-The tool can also be applied to an existing pg_dump file:
+You can use either of `postgresql`, `postgres`, or `pg` to `-source` to specify
```
nevinheintze, line 21: Suggest rephrasing:

You can use any of `postgresql`, `postgres`, or `pg` as the argument to the `-source` flag. They all specify
PostgreSQL as the source database.

sources/postgres/README.md
```diff
@@ -1,83 +1,92 @@
...
+harbourbridge schema -source=postgres -source-profile="host=<>,port=<>,user=<>,db_name=<>"
 ```
 
-It is assumed that _PGHOST_, _PGPORT_, _PGUSER_, _PGDATABASE_ environment
-variables are set. Password can be specified either in the _PGPASSWORD_ environment
-variable or provided at the password prompt.
+Parameters `port` and `password` are optional. Port (`port`) defaults to `5432`
+for PostgreSQL source. Password can be provided at the password prompt.
+
+(⚠ Deprecated ⚠) Set environment variables `PGHOST`, `PGPORT`, `PGUSER`,
```
nevinheintze, line 84: Suggest not deprecating the environment variables for postgres -- these are standard environment variables commonly used by postgres users.

In contrast, the mysql ones are non-standard harbourbridge-specific, so should be deprecated I think.

nevinheintze (commented):
Thanks for all of the work here Asheesh!
I certainly didn't realize the amount of work it would be for this -- and also thanks for fixing a bunch of incidental things along the way.
 This is a huge improvement.

sources/postgres/README.md
agasheesh: SGTM, have modified the language to say "Alternatively" instead of "Deprecated".

cmd/oneshot.go
agasheesh: I didn't have a preference between oneshot and schema-and-data, but agree that schema-and-data is clearer. Have updated the subcommand and it's references in the READMEs to say schema-and-data now.

README.md
agasheesh: I agree, but would ask to take this clean up in a separate PR, such that this PR can be wrapped up.

cmd/source_profile.go
agasheesh: I was looking at gcloud CLI and noticed that all the flags were hyphenated thus was trying to cleanup and make all the flags and params (keys) hyphenated as well for HB CLI.

README.md
agasheesh: Agree that it's better to just say database name since how we generate the same can change over time, and then the README would become stale. Done.

README.md
agasheesh: I agree that for someone just starting with HB, it's better to describe the full functionality since we describe the various subcommands separately as well. Have updated to use `schema-and-data`.

README.md
agasheesh: Done, it's actually preferable to have `help` subcommand since that's more descriptive than the output of `commands` subcommand. Makes sense to do this, since right after this we do have more concrete examples of using `schema-and-data` subcommand.

README.md
agasheesh: Done.

README.md
agasheesh: Hmm, I did spend quite some thinking about how to structure this, but couldn't make it clean. Makes sense to link to sections, thanks for the advice. Done.

sources/dynamodb/README.md
agasheesh: I went back and forth on this, thanks for the input. Have now added back "noting the alias" text.

sources/mysql/README.md
agasheesh: Acknowledged.

sources/mysql/README.md
agasheesh: Done, much cleaner. Thanks.

sources/postgres/README.md
agasheesh: Done.

agasheesh:
Thanks @nevinheintze and @bharadwaj-aditya for thorough reviews. There were many open ended questions around how to structure the READMEs and the same is reflected in the number of code review iterations that was required for this PR. But thanks to the same, I feel much more comfortable with the current state. PTAL.

README.md
```diff
@@ -273,75 +261,132 @@ By default, these files are prefixed by the name of the Spanner database (with a
...
+```sh
+harbourbridge <subcommand> flags
+```
 
-`-instance` Specifies the Spanner instance to use. The new database will be
-created in this instance. If not specified, the tool automatically determines an
-appropriate instance using gcloud.
+### Getting Help
+
+The command hb help displays the available subcommands and the important global flags. The hb flags command displays the list of all available global flags, including flags from other libraries that HarbourBridge uses.
```
bharadwaj-aditya, line 275: This should be harbourbridge help and harbourbridgeflags

bharadwaj-aditya (approved):
Looks fine to me overall. Couple of minor comments, but those can be addressed independently

cmd/source_profile.go
```diff
@@ -173,27 +173,27 @@ type SourceProfileConnectionDynamoDB struct {
 
 func NewSourceProfileConnectionDynamoDB(params map[string]string) (SourceProfileConnectionDynamoDB, error) {
 	dydb := SourceProfileConnectionDynamoDB{}
-	if schemaSampleSize, ok := params["schemaSampleSize"]; ok {
+	if schemaSampleSize, ok := params["schema-sample-size"]; ok {
```
Deep1998, line 176: Why did we move away from camel case just for dynamodb

Deep1998 (approved):
LGTM

PR #222, opened 2021-11-10
Replace instructions to run HarbourBridge with git clone and go run

agasheesh:
`go install` or `go get` commands require that go.mod doesn't have replace directives and thus the current instructions don't work anymore. Thus, we are now switching to recommend that customers only use git clone mechanism to setup and install HarbourBridge.

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,34 +123,39 @@ variable if it is not already configured using
 
 ### Installing HarbourBridge
 
-Download the tool to your machine and install it.
+You can make a copy of the HarbourBridge codebase from the github repository
+and use "go run":
 
 ```sh
-GO111MODULE=on go get github.com/cloudspannerecosystem/harbourbridge
+git clone https://github.com/cloudspannerecosystem/harbourbridge
+cd harbourbridge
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -123,34 +123,39 @@ variable if it is not already configured using
...
 
-Download the tool to your machine and install it.
+You can make a copy of the HarbourBridge codebase from the github repository
+and use "go run":
 
 ```sh
-GO111MODULE=on go get github.com/cloudspannerecosystem/harbourbridge
+git clone https://github.com/cloudspannerecosystem/harbourbridge
+cd harbourbridge
+pg_dump mydb | go run github.com/cloudspannerecosystem/harbourbridge -driver=pg_dump
```
bharadwaj-aditya, line 132: we should probably skip pg_dump here and just let them run go run with the flags. The subsequent db specific sections will cover this piping use cases.

bharadwaj-aditya (approved):
LGTM - Just a minor correction on the first usage to make it clearer

README.md
agasheesh: Done.

PR #223, opened 2021-11-11
Add double quotes around the database name for PG Spanner

agasheesh:
Spanner with GoogleSQL dialect requires database names with hypens in CreateDatabase statement to be quoted with backtick characters e.g., \`my-db-name\`, whereas Spanner with PostgreSQL dialect (PGSpanner) doesn't allows recognize backtick characters. Thus, to create database with a name that contains hyphens, the dbname needs to be quoted with double quotes e.g., \"my-pg-db-name\".


Deep1998 (approved):
LGTM

PR #224, opened 2021-11-12
Dynamodb Support for pg spanner

Deep1998:
Added changes to support dynamodb to pg spanner migration.

Manually tested for now.

### DIFF
--- a/sources/dynamodb/data.go
+++ b/sources/dynamodb/data.go
@@ -49,7 +49,14 @@ func cvtRow(attrsMap map[string]*dynamodb.AttributeValue, srcSchema schema.Table
 			srcStrVal = "null"
 		} else {
 			// Convert data to the target type.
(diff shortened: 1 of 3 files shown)


sources/dynamodb/data.go
```diff
@@ -149,22 +197,6 @@ func cvtColValue(attrVal *dynamodb.AttributeValue, srcType string, spType string
 				return nil, fmt.Errorf("failed to convert %v to a json string", attrVal.GoString())
 			}
 			return string(b), nil
-		case typeString:
-			return *attrVal.S, nil
-		case typeStringSet:
```
bharadwaj-aditya, line 154: Do we retain the old functionality for non-pg spanner post these changes ?

Deep1998: Yes we do. I made the changes accordingly to support both and integration tests for dydb passing means older version works fine.

bharadwaj-aditya (approved):
LGTM

bharadwaj-aditya:
You will have some conflicts based on the other DynamoDb PR merged by Charvi. If there are significant changes, I will take another pass at this

PR #225, opened 2021-11-18
Security patch for dependabot alerts

Deep1998:
There are 2 dependabot alerts for the backend libs:

- **github.com/gorilla/websocket :**  a version bump should fix.
- **github.com/dgrijalva/jwt-go :** This library is unmaintained hence no version bump available. This is imported through the **pingcap/tidb** library. There are 2 ways to go about this:
   1. **(Did not work)** Bump TiDB to latest version (that has the fix for this issue): This cannot be done because pingcap/tidb does not version as per golang recommendations i.e. they do not append /v2, /v3 , /v4 ... at the end of their module paths. Hence we cannot bump the version up.
   2. **(Approach taken)** TiDB replaced this lib with a fork named form3tech-oss that has the fix in their latest version. We put the same replace directive in our go.mod without bumping the TiDB version. 
  
**Caveat to the approach taken:**
The form3tech-oss fork has the fix but is unmaintained now. The ideal solution would be to switch to a community maintained fork named golang-jwt but TiDB still uses the former fork hence keeping it same as that.


bharadwaj-aditya (approved):
LGTM

PR #226, opened 2021-11-22
ordering tables so that interleaved tables are populated after their …

shreyakhajanchi:
Currently, while data population interleaving table order is not respected. With this change the interleave tables will be populated after their parent table has been populated.


spanner/ddl/ast.go
```diff
@@ -328,45 +328,55 @@ func NewSchema() Schema {
...
-				// but we will always make progress because interleaved tables can't
-				// have cycles. In principle this could be O(n^2), but in practice chains
-				// of interleaved tables are small.
-				tableQueue = append(tableQueue, tableName)
+		for _, tableName := range sortedTableNames {
+			ddl = append(ddl, s[tableName].PrintCreateTable(c))
+			for _, index := range s[tableName].Indexes {
+				ddl = append(ddl, index.PrintCreateIndex(c))
 			}
+
```
Deep1998, line 379: nit: stray newline

spanner/ddl/ast.go
```diff
@@ -328,45 +328,55 @@ func NewSchema() Schema {
 	return make(map[string]CreateTable)
 }
 
+// Tables are ordered in alphabetical order with one exception: interleaved
+// tables appear after the definition of their parent table.
+func OrderTable(s Schema) []string {
```
Deep1998, line 333: nit: OrderTables?

spanner/ddl/ast.go
```diff
@@ -328,45 +328,55 @@ func NewSchema() Schema {
...
+	var tableNames, sortedTableNames []string
+	for t := range s {
+		tableNames = append(tableNames, t)
+	}
+	sort.Strings(tableNames)
+	tableQueue := tableNames
+	printed := make(map[string]bool)
```
Deep1998, line 340: nit: Since we are not actually printing them here, should we rename this? maybe picked or marked? We should also edit the comments below that say print.

sources/mysql/data_test.go
```diff
@@ -320,6 +320,8 @@ func buildConv(spTable ddl.CreateTable, srcTable schema.Table) *internal.Conv {
 	conv := internal.MakeConv()
 	conv.SpSchema[spTable.Name] = spTable
 	conv.SrcSchema[srcTable.Name] = srcTable
+	conv.ToSource[spTable.Name] = internal.NameAndCols{Name: srcTable.Name, Cols: make(map[string]string)}
+	conv.ToSpanner[srcTable.Name] = internal.NameAndCols{Name: spTable.Name, Cols: make(map[string]string)}
```
Deep1998, line 324: Why did we have to modify this?

bharadwaj-aditya (changes requested):
Please add a unit test for this. Aside from a few formatting changes, looks fine overall

charvisingla (changes requested):

sources/common/infoschema.go
```diff
@@ -74,7 +74,13 @@ func ProcessSchema(conv *internal.Conv, infoSchema InfoSchema) error {
 // If we can't get/process data for a table, we skip that table and process
 // the remaining tables.
 func ProcessData(conv *internal.Conv, infoSchema InfoSchema) {
-	for srcTable, srcSchema := range conv.SrcSchema {
+	// Tables are ordered in alphabetical order with one exception: interleaved
+	// tables appear after the population of their parent table.
+	orderTableNames := ddl.OrderTable(conv.SpSchema)
+
+	for _, spannerTable := range orderTableNames {
+		srcTable := conv.ToSource[spannerTable].Name
```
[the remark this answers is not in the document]
shreyakhajanchi: GetSpannerTable creates a new spanner table name and then assigns the mapping which cannot be done in case of source table name

sources/common/infoschema.go
```diff
@@ -74,7 +74,13 @@ func ProcessSchema(conv *internal.Conv, infoSchema InfoSchema) error {
 // If we can't get/process data for a table, we skip that table and process
 // the remaining tables.
 func ProcessData(conv *internal.Conv, infoSchema InfoSchema) {
-	for srcTable, srcSchema := range conv.SrcSchema {
+	// Tables are ordered in alphabetical order with one exception: interleaved
+	// tables appear after the population of their parent table.
+	orderTableNames := ddl.OrderTable(conv.SpSchema)
+
+	for _, spannerTable := range orderTableNames {
```
Deep1998, line 81: In a very rare scenario, it could be possible that the next table in the list is a child of the current, and the first row of the next table references the last row of the current table. It could be possible that these are batched in a single mutation, and could potentially throw an error due to missing parent row. It is very rare to actually witness this, but I think it makes sense to make this code threadsafe and ensure we drain out the row queue and finish the writes before moving onto the next table in the for loop.

Deep1998 (commented):
Rare but potential bug.

sources/common/infoschema.go
Deep1998: You should be able to use bw.Flush() at the end to drain out the rows. If not, we might have to bubble the wait group counts uptil here.

sources/common/infoschema.go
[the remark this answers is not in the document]
charvisingla: oh ok. Missed that. We should anyways create this method there to fetch the source table. That file should be the only one to directly deal with the mapping (ToSource/ToSpanner). It can also throw explicit errors for input validation and for cases when the mapping/conv is not populated.

sources/common/infoschema.go
bharadwaj-aditya: Making a clean cut and flushing on table change is a good way to do this. There is very little gain we will get optimizing on this, and the possible issues if we don't are numerous.

sources/mysql/data_test.go
shreyakhajanchi: This change was already there but was removed as part of https://github.com/cloudspannerecosystem/harbourbridge/pull/204/files .

spanner/ddl/ast.go
shreyakhajanchi: Thanks for pointing out , made the change

spanner/ddl/ast.go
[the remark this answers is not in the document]
shreyakhajanchi: 1. Placing this method in mapping.go causes circular dependency of mapping.go and ast.go because Schema is defined in ast.go.
2. spSchema is a map therefore unable to sort and place there

sources/common/infoschema.go
shreyakhajanchi: done

sources/common/infoschema.go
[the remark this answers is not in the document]
shreyakhajanchi: done

sources/common/infoschema.go
```diff
@@ -86,6 +93,7 @@ func ProcessData(conv *internal.Conv, infoSchema InfoSchema) {
 		}
 		err := infoSchema.ProcessData(conv, srcTable, srcSchema, spTable, spCols, spSchema)
 		if err != nil {
+			writer.Flush()
```
Deep1998, line 96: I think this writer.Flush() should come right after this if err != nil block.

Deep1998: Doing that will also require a change in how we do Flush right now. We should probably add a similar option to pass a flush() function to the batchwriter, similar to how we pass a data sink. That, way the unit tests also won't require any major change.

test_data/session_test.json
```diff
@@ -0,0 +1,306 @@
+{
```
Deep1998, line 1: Suggest changing the session file name should be changed to something specific describing whats special about it rather than something generic.

Deep1998 (commented):
One minor change regarding the flush().

charvisingla:
> 1. Placing this method in mapping.go causes circular dependency of mapping.go and ast.go because Schema is defined in ast.go.
> 2. spSchema is a map therefore unable to sort and place there

1. Schema is not needed in ordering functionality. Only the tableNames are needed.
2. In order to store the tables in sorted order, we would either need to change spSchema into an appropriate data structure that will preserve the sorted order or keep a sorted list of tablenames. Its fine to not do this immediately since there are only 2 instances that trigger sorting. But It can probably degrade performance for databases with large number of tables so we should add a note to optimize this in future.

sources/common/infoschema.go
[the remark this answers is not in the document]
charvisingla: nvm, saw Deep's comment that this should be after the error block.

charvisingla: By passing it as a function to the batchwriter, I'm assuming you meant adding flush as a function param in conv to avoid passing writer as a param to infoschema. While I wouldn't want to add more to conv but exposing writer to infoschema is also not something that is desired.

Unrelated, 2 possible performance improvements unless they are already captured somewhere
1. I saw that we are flushing the data after entire data is processed. It should be better for performance if we flush data as soon as we have enough rows processed for write thresholds. This would make processing and writing somewhat parallel.
2. Each row and each table is processed sequentially currently, independent tables and rows should also be processed in parallel.

internal/convert.go
```diff
@@ -33,6 +33,7 @@ type Conv struct {
 	ToSource       map[string]NameAndCols              // Maps from Spanner table name to source-DB table name and column mapping.
 	UsedNames      map[string]bool                     // Map storing the names that are already assigned to tables, indices or foreign key contraints.
 	dataSink       func(table string, cols []string, values []interface{})
+	DataFlush      func()         `json:"-"`
```
Deep1998, line 36: What does this struct tag do? Maybe add a comment next to it explaining the field?

Deep1998 (approved):
LGTM

bharadwaj-aditya (approved):
LGTM

sources/common/infoschema.go
shreyakhajanchi: Data is written in parallel currently via go routines and when write threshold reaches the system waits before processing further records.

charvisingla: Writes do happen in parallel but looking at the code it seemed like the first write starts after entire processing. I might be missing something so we can discuss further. These were anyways not related to the PR. Just put it out here as a suggestion for future optimizations.

charvisingla (approved):

PR #237, opened 2021-12-13
Support for CSV: First Cut

Deep1998:
First cut of changes to support CSV in harbourbridge. The functionality is limited in the first cut to just unblock UHG.

Subcommand changes:
Added CSV as a source for data mode. The source profile accepts the path to the manifest file. No changes to target profile.
Sample command:
`data -source=csv -source-profile="manifest=manifest.json" -target-profile="project=my-project,instance=spanner-instance,dbname=spanner-db" `

The manifest structure is largely similar to dataflow's manifest with slight changes:
```
[
    {
      "table_name": "Albums",
      "file_patterns": [
        "/Users/deepchowdhury/Desktop/harbourbridge/Albums.csv"
      ],
      "columns": [
        {"column_name": "a", "type_name": "BOOL"},
        {"column_name": "b", "type_name": "BYTES(124)"},   // Length is optional
        {"column_name": "c", "type_name": "DATE"},
        {"column_name": "d", "type_name": "FLOAT64"},
        {"column_name": "e", "type_name": "INT64"},
        {"column_name": "f", "type_name": "NUMERIC"},
        {"column_name": "g", "type_name": "STRING"},
        {"column_name": "h", "type_name": "TIMESTAMP"},
        {"column_name": "i", "type_name": "JSON"}
      ]
    },
    {
      "table_name": "Singers",
      "file_patterns": [
        "/Users/deepchowdhury/Desktop/harbourbridge/Singers.csv"
      ],
      "columns": [
        {"column_name": "SingerId", "type_name": "INT64"},
        {"column_name": "FirstName", "type_name": "STRING(100)"},  // Length is optional
        {"column_name": "LastName", "type_name": "STRING"}
      ]
    }
]
```
**Note regarding the manifest:**
- The table_name field should be identical to the table name in Spanner schema.
- File patterns do not accept regex expressions. Provide the path inside double quotes.
- The type_name should be identical to the types in spanner schema. Only for `STRING` and `BYTES`, the length can be optionally omitted.

**CSV file format:**
```
bool_col,byte_col,date_col,float_col,int_col,numeric_col,string_col,timestamp_col,json_col // column headers
true,bytevalue,2020-12-09,15.13,100,39.94,Helloworld,2019-10-29 05:30:00,"{""key1"": ""value1"", ""key2"": ""value2""}" // data
```
**Note regarding the CSV file:**
- Each column in a row should be separated by commas.
- The first row needs to have the column name headers. The column names should be identical to the column names in Spanner schema.
- Remove trailing spaces, tabs in the column name headers.

**CSV Data types:**
- We only support scalar data types right now. Sample data format for each provided in the snippet above.
- The only supported data format right now is **RFC3339 full-date format**
- The only supported timestamp format right now is **ISO 8601**
- The format to escape the quotes in json is adding an additional `"` in front of the double quote. `\` does not work. Also enclose the whole data inside "". Some modification might be required since most databases do not export CSVs with escaping quotes like mentioned.

Pending tasks:
1) Interleaved table support
2) PG Spanner support

Stretch Goal:
Implement as a full fledged feature where manifest file is not required.

### DIFF
--- a/cmd/data.go
+++ b/cmd/data.go
@@ -112,6 +112,10 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 
 	// If using CSV mode, follow different code path from here.
 	if driverName == constants.CSV {
+		if targetProfile.conn.sp.dbname == "" {
+			err = fmt.Errorf("dbname is mandatory in target-profile for csv source")
+			return subcommands.ExitFailure
+		}
 		// TODO: refactor this to go through DataConv(). Fix it when refactoring passing of source-profile throughout the code
 		// to avoid excess parameters in the DataConv and SchemaConv functions.
 		bw, err := conversion.DataFromCSV(conv, sourceProfile.csv.manifest, client, targetDb)
--- a/cmd/data.go
+++ b/cmd/data.go
@@ -57,6 +57,48 @@ func (cmd *DataCmd) SetFlags(f *flag.FlagSet) {
 	f.StringVar(&cmd.filePrefix, "prefix", "", "File prefix for generated files")
 }
 
+// TODO: refactor this to go through DataConv(). Fix it when refactoring passing of source-profile throughout the code
+// to avoid excess parameters in the DataConv and SchemaConv functions.
+func executeCSVLoader(ctx context.Context, conv *internal.Conv, sourceProfile SourceProfile, targetProfile TargetProfile, filePrefix string) subcommands.ExitStatus {
+	if targetProfile.conn.sp.dbname == "" {
+		fmt.Printf("dbname is mandatory in target-profile for csv source")
+		return subcommands.ExitFailure
+	}
+	targetDb := targetProfile.ToLegacyTargetDb()
+	ioHelper := conversion.NewIOStreams(constants.CSV, "")
+	if ioHelper.SeekableIn != nil {
+		defer ioHelper.In.Close()
+	}
+	now := time.Now()
+	project, instance, dbName, err := getResourceIds(ctx, targetProfile, now, constants.CSV, ioHelper.Out)
+	if err != nil {
+		return subcommands.ExitUsageError
+	}
+	dbURI := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName)
+	client, err := conversion.GetClient(ctx, dbURI)
+	if err != nil {
+		fmt.Printf("can't create client for db %s: %v", dbURI, err)
+		return subcommands.ExitFailure
+	}
+	defer client.Close()
+
+	delimiter := sourceProfile.csv.delimiter
+	if len(delimiter) != 1 {
+		fmt.Printf("delimiter should only be a single character long, found '%s'", delimiter)
+		return subcommands.ExitFailure
+	}
+	bw, err := conversion.DataFromCSV(conv, sourceProfile.csv.manifest, client, targetDb, sourceProfile.csv.nullStr, rune(delimiter[0]))
+	if err != nil {
+		fmt.Printf("can't finish data conversion for db %s: %v", dbURI, err)
+		return subcommands.ExitFailure
+	}
+	banner := conversion.GetBanner(now, dbURI)
+	conversion.Report(constants.CSV, bw.DroppedRowsByTable(), ioHelper.BytesRead, banner, conv, filePrefix+reportFile, ioHelper.Out)
+	conversion.WriteBadData(bw, conv, banner, filePrefix+badDataFile, ioHelper.Out)
+
+	return subcommands.ExitSuccess
+}
+
 func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
 	var err error
 	defer func() {
@@ -64,6 +106,7 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 			fmt.Printf("FATAL error: %v\n", err)
 		}
 	}()
+	conv := internal.MakeConv()
 
 	sourceProfile, err := NewSourceProfile(cmd.sourceProfile, cmd.source)
 	if err != nil {
@@ -78,8 +121,12 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 	if err != nil {
 		return subcommands.ExitUsageError
 	}
-	targetDb := targetProfile.ToLegacyTargetDb()
+	// If using CSV mode, follow different code path from here.
+	if driverName == constants.CSV {
+		return executeCSVLoader(ctx, conv, sourceProfile, targetProfile, cmd.filePrefix)
+	}
 
+	targetDb := targetProfile.ToLegacyTargetDb()
 	dumpFilePath := ""
 	if sourceProfile.ty == SourceProfileTypeFile && (sourceProfile.file.format == "" || sourceProfile.file.format == "dump") {
 		dumpFilePath = sourceProfile.file.path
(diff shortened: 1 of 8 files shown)


cmd/data.go
```diff
@@ -101,6 +102,29 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
...
+	client, err := conversion.GetClient(ctx, dbURI)
+	if err != nil {
+		err = fmt.Errorf("can't create client for db %s: %v", dbURI, err)
+		return subcommands.ExitFailure
+	}
+	defer client.Close()
+
+	// If using CSV mode, follow different code path from here.
+	if driverName == constants.CSV {
+		// TODO: refactor this to go through DataConv(). Fix it when refactoring passing of source-profile throughout the code
```
Deep1998, line 115: check target profile is correct.

sources/csv/data.go
```diff
@@ -0,0 +1,284 @@
...
+		conv.WriteRow(tableName, tableName, srcCols, cvtVals)
+	}
+}
+
+// convertData currently only supports scalar data types.
+func convertData(conv *internal.Conv, tableName string, srcCols []string, values []string) ([]interface{}, error) {
+	var v []interface{}
+	colDefs := conv.SpSchema[tableName].ColDefs
+	for i, val := range values {
+		colName := srcCols[i]
```
shreyakhajanchi, line 195: Do we want to add any special handling for null values here ?

sources/csv/data.go
```diff
@@ -0,0 +1,284 @@
...
+		x, err := convScalar(colDefs[colName].T, val)
+		if err != nil {
+			return nil, err
+		}
+		v = append(v, x)
+	}
+	return v, nil
+}
+
+func convScalar(spannerType ddl.Type, val string) (interface{}, error) {
```
shreyakhajanchi, line 205: Is it possible to have array as datatype too ?

cmd/data.go
```diff
@@ -101,6 +102,33 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
...
 	conv := internal.MakeConv()
+
+	client, err := conversion.GetClient(ctx, dbURI)
+	if err != nil {
+		err = fmt.Errorf("can't create client for db %s: %v", dbURI, err)
+		return subcommands.ExitFailure
+	}
+	defer client.Close()
+
+	// If using CSV mode, follow different code path from here.
```
bharadwaj-aditya, line 113: Move this into another method. This method is already pretty big

cmd/data.go
```diff
@@ -101,6 +102,33 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
...
+	if driverName == constants.CSV {
+		if targetProfile.conn.sp.dbname == "" {
+			err = fmt.Errorf("dbname is mandatory in target-profile for csv source")
+			return subcommands.ExitFailure
+		}
+		// TODO: refactor this to go through DataConv(). Fix it when refactoring passing of source-profile throughout the code
+		// to avoid excess parameters in the DataConv and SchemaConv functions.
+		bw, err := conversion.DataFromCSV(conv, sourceProfile.csv.manifest, client, targetDb)
```
bharadwaj-aditya, line 121: Do we check if the schema matches the manifest anywhere in this flow ?

sources/csv/data.go
```diff
@@ -0,0 +1,284 @@
...
+
+type Column struct {
+	Column_name string `json:"column_name"`
+	Type_name   string `json:"type_name"`
+}
+
+// Harbourbridge accepts a manifest file in the form of a json which unmarshalls into the Table struct.
+type Table struct {
+	Table_name    string   `json:"table_name"`
+	File_patterns []string `json:"file_patterns"`
```
bharadwaj-aditya, line 42: What are the scenarios where we would have multiple file patterns per table ?
I'm guessing this is to be compatible with the existing structure, but good to know if there is a valid use case

sources/csv/data.go
```diff
@@ -0,0 +1,284 @@
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return 0, fmt.Errorf("can't read row from file: %s", filePath)
+		}
+		count++
+	}
+	// Exclude the first row, that is, column headers.
+	return count - 1, nil
```
bharadwaj-aditya, line 125: We should check for empty files here, else we might run into some edge cases

sources/csv/data.go
```diff
@@ -0,0 +1,284 @@
...
+			colDefs[col.Column_name] = ddl.ColumnDef{Name: col.Column_name, T: ty}
+		}
+		conv.SpSchema[table.Table_name] = ddl.CreateTable{Name: table.Table_name, ColDefs: colDefs}
+
+		for _, filePath := range table.File_patterns {
+			csvFile, err := os.Open(filePath)
+			if err != nil {
+				return fmt.Errorf(fmt.Sprintf("can't read csv file: %s due to: %v\n", filePath, err))
+			}
+			r := csvReader.NewReader(csvFile)
```
bharadwaj-aditya, line 152: What delimiters does this reader expect ? Is it only commas , or separated by double/single quotes. Can the delimiter be provided ?

sources/csv/data.go
```diff
@@ -0,0 +1,284 @@
...
+func convNumeric(val string) (interface{}, error) {
+	r := new(big.Rat)
+	if _, ok := r.SetString(val); !ok {
+		return "", fmt.Errorf("can't convert %q to big.Rat", val)
+	}
+	return r, nil
+}
+
+func convTimestamp(val string) (t time.Time, err error) {
+	t, err = time.Parse("2006-01-02 15:04:05", val)
```
bharadwaj-aditya, line 279: What is this timestamp ?

sources/csv/data.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+		if len(table.File_patterns) == 0 {
+			return fmt.Errorf("no file path provided for table %s", name)
+		}
+		cols := table.Columns
+		if len(cols) == 0 {
+			return fmt.Errorf("`columns` field for table %s is empty", name)
+		}
+		for j, col := range cols {
+			if col.Column_name == "" || col.Type_name == "" {
```
bharadwaj-aditya, line 84: We should check for valid column types supported here before we start the migration.

sources/csv/data.go
```diff
@@ -0,0 +1,284 @@
...
+// across multiple CSV files hence, the manifest accepts a list of file paths in the input.
+func ProcessCSV(conv *internal.Conv, tables []Table) error {
+	for _, table := range tables {
+		// Populating just the table names in the conv for SrcSchema and SpSchema
+		// so the report for row stats is generated.
+		conv.SrcSchema[table.Table_name] = schema.Table{Name: table.Table_name}
+
+		// The map colDefs stores the mapping from column names to their final types.
+		colDefs := make(map[string]ddl.ColumnDef)
+		for _, col := range table.Columns {
```
bharadwaj-aditya, line 138: Might make sense to move this check to the verify phase.

bharadwaj-aditya (changes requested):

sources/csv/data.go
Deep1998: Yes I was planning to take that during the next cut.

sources/csv/data.go
Deep1998: Good point. Modifying it to read the NULL value string from the source-profile which the user inputs (default "").

cmd/data.go
Deep1998: Nopes, the assumption is that the spanner schema will match the manifest and has to be taken care of the user. Once we add the querying spanner schema functionality, then we wont need a manifest file at all.

sources/csv/data.go
Deep1998: This I did mostly for consistency reasons. I think most poc dbs will just have 1 csv file per table. But if some user has a table split over multiple files they can use this functionality. Thought I'm not sure how frequently that occurs.

sources/csv/data.go
Deep1998: Currently only comma is there. But adding custom ones should be straight forward. I'll add it

sources/csv/data.go
Deep1998: Added checks, but for empty fiels, just reporting it in the report, not erroring out because thats what we do in other places currently.

sources/csv/data.go
Deep1998: This is the layout aka a template so the function knows the format. Thats how the function accepts it.

PR #238, opened 2021-12-14
Fix for failing tests

Deep1998:
Current tests are suddenly failing during setup when running
`sudo apt-get install -y mysql-client`
with
```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 mysql-client : Depends: mysql-client-8.0 but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

I tried specifying the older version used in older successful runs of the tests which was 8.0.26-0ubuntu0.20.04.3 via
`sudo apt-get install -y mysql-client=8.0.26-0` but that also fails with version not found.

Currently, removing that line as the mysql CLI still seems to work and the tests pass without that.

### DIFF
--- a/.github/workflows/integration-tests-against-emulator.yaml
+++ b/.github/workflows/integration-tests-against-emulator.yaml
@@ -74,10 +74,6 @@ jobs:
       - run: psql -f test_data/pg_dump.test.out
 
       # init a MySQL database from the test_data
-      - name: Install MySQL client required for loading .sql files
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y mysql-client
       - run: mysql --version
       - run: mysql -v -P 3306 --protocol=tcp -u root -proot test < test_data/mysqldump.test.out
 


.github/workflows/integration-tests-against-emulator.yaml
```diff
@@ -77,7 +77,6 @@ jobs:
       - name: Install MySQL client required for loading .sql files
         run: |
           sudo apt-get update
```
bharadwaj-aditya, line 79: We can remove this line also if we are not installing anything here

bharadwaj-aditya:
How do the mysql tests proceed without installing this ?Is there some default version of mysql installed ?

Deep1998:
I think the client comes installed when installed mariadb image, because when we install mysql locally, we just install the server and the CLI works out of the box. I personally never had to install mysql-client for the CLI to work. 
This line was probably just to double-check because in the prev runs of successful tests, the output of this command used to be `mysql-client is already the newest version (8.0.26-0ubuntu0.20.04.3)`.

bharadwaj-aditya:
Makes sense. Approving

bharadwaj-aditya (approved):
LGTM

PR #239, opened 2021-12-16
CSV README

Deep1998:


README.md
```diff
@@ -310,7 +311,7 @@ This section describes the flags common across all the subcommands. For flags
 specific to a give subcommand run `harbourbridge help <subcommand>`.
 
 `-source` Required flag. Specifies the source database. Supported source
-databases are _'postgres'_, _'mysql'_ and _'dynamodb'_.
+databases are _'postgres'_, _'mysql'_, _'dynamodb'_ and _'csv'_(only in data mode).
```
bharadwaj-aditya, line 314: change this to Specifies the data source as CSV is not a database.

sources/csv/README.md
```diff
@@ -0,0 +1,114 @@
+# HarbourBridge: CSV-to-Spanner Migration
+
+HarbourBridge is a stand-alone open source tool for Cloud Spanner evaluation 
+and migration. We now support loading data from CSVs. This assumes a Spanner
+database with Schema already exists and HarbourBridge loads the data for you.
+
+
+To run HarbourBridge to migrate CSVs, it should be used only with the data
```
bharadwaj-aditya, line 8: Might sound better if we say

"To migrate CSVs via HarbourBridge, provide a manifest file and run it with the data subcommand"
And then describe the manifest file

sources/csv/README.md
```diff
@@ -0,0 +1,114 @@
...
+]
+```
+**CAVEATS:**
+- File patterns do not accept regular expressions. Provide the path inside 
+double quotes.
+- The type_name should be identical to the types in spanner schema. Only 
+for `STRING` and `BYTES`, the length can be optionally omitted.
+
+### CSV File Format
+- Each column in a row should be separated by a `,`(comma).
```
bharadwaj-aditya, line 66: Specify how to handle strings here.

sources/csv/README.md
```diff
@@ -0,0 +1,114 @@
...
+to write to.
+
+## Foreign Keys and Interleaved Tables
+
+HarbourBridge currently does not guarantee data conversion if the Spanner
+schema has Foreign Keys or Interleaved Tables in the final schema.
+It is recommended to create Foreign Keys only after data conversion
+has been completed.
+
+For interleaved tables, it may not be possible to change the schema later.
```
bharadwaj-aditya, line 111: Does the manifest provide an option to give the order in which tables should be imported ? If it does, we should use that here, else we might see a lot of migrations failing.

bharadwaj-aditya (changes requested):

README.md
Deep1998: Good catch

sources/csv/README.md
Deep1998: The order in which the manifest lists is the same order in which it will be imported. Updated the documentation to make it clearer.

PR #240, opened 2021-12-16
Feature :  Sql server support

sagarsearce:
#### This PR adds SQL server data and schema migration functionality
Related to https://github.com/cloudspannerecosystem/harbourbridge/issues/251

#### Brief of the changes 
Added new value for the source flag [ -source = sqlserver OR mssql] and implemented ToDdl and common.InfoSchema interface for the SQL server.

- [x] schema migration
- [x] data migration
- [x] Unit test added for each file


cmd/source_profile.go
```diff
@@ -234,6 +275,14 @@ func NewSourceProfileConnection(source string, params map[string]string) (Source
 				return conn, err
 			}
 		}
+	case "sqlserver":
```
Deep1998, line 278: this should also include mssql

sagarsearce: ya, thank you. will add that.

conversion/conversion.go
```diff
@@ -182,13 +186,36 @@ func GetMYSQLConnectionStr(server, port, user, password, dbname string) string {
 	return fmt.Sprintf("%s:%s@tcp(%s:%s)/%s", user, password, server, port, dbname)
 }
 
+func generateSQLSERVERConnectionStr() (string, error) {
+	server := os.Getenv("SSHOST")
+	port := os.Getenv("SSPORT")
+	user := os.Getenv("SSUSER")
+	dbname := os.Getenv("SSDATABASE")
```
Deep1998, line 193: These and some other bits retain the older funcitonality of env variables. @bharadwaj-aditya when do we stop adding supporting for these?

sources/sqlserver/data.go
```diff
@@ -0,0 +1,263 @@
...
+func ConvertData(conv *internal.Conv, srcTable string, srcCols []string, srcSchema schema.Table, spTable string, spCols []string, spSchema ddl.CreateTable, vals []string) (string, []string, []interface{}, error) {
+	var c []string
+	var v []interface{}
+	if len(spCols) != len(srcCols) || len(spCols) != len(vals) {
+		return "", []string{}, []interface{}{}, fmt.Errorf("ConvertData: spCols, srcCols and vals don't all have the same lengths: len(spCols)=%d, len(srcCols)=%d, len(vals)=%d", len(spCols), len(srcCols), len(vals))
+	}
+	for i, spCol := range spCols {
+		srcCol := srcCols[i]
+		// Skip columns with 'NULL' values. When processing data rows from mysqldump, these values
+		// are represented as nil (by pingcap/tidb/types/parser_driver's ValueExpr), which is
```
Deep1998, line 61: MYSQL specific comment

sagarsearce: please ignore the data files now it is still in progress.

sources/sqlserver/data.go
```diff
@@ -0,0 +1,263 @@
...
+		// converted to the string '<nil>'. When processing data rows obtained from the MySQL driver,
+		// 'NULL' values are represented as "NULL" (because we retrieve the values as strings).
+		if vals[i] == "<nil>" || vals[i] == "NULL" {
```
Deep1998, line 64: I know this is how we do it in MySQL as well, but I think this might be a bug and we will end up preserving it for mssql. If we do a continue here, we end up sending lesser values than the number of columns columns which I think might error out. I think we should start passing nil values in these cases for it to work correctly.

But its best if you verify the hypothesis by having a nullable column with null values, my guess is migrating that row would fail.

sources/sqlserver/infoschema.go
```diff
@@ -0,0 +1,383 @@
...
+	if err != nil {
+		return 0, err
+	}
+	defer rows.Close()
+	var count int64
+	if rows.Next() {
+		err := rows.Scan(&count)
+		return count, err
+	}
+	return 0, nil //Check if 0 is ok to return
```
Deep1998, line 121: Maybe remove this comment, I think it should be ok.

sources/sqlserver/infoschema.go
```diff
@@ -0,0 +1,383 @@
...
+			AND TAB.name=@p1
+			AND TAB.schema_id = SCHEMA_ID(@p2)
+			ORDER BY IX.name ;
+	`
+	rows, err := isi.Db.Query(q2, table.Name, table.Schema)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	// TODO : remove sequence
```
Deep1998, line 329: What is this todo for?

sources/sqlserver/infoschema.go
```diff
@@ -0,0 +1,383 @@
...
+	for rows.Next() {
+		// get RawBytes from data.
+		err := rows.Scan(scanArgs...)
+		if err != nil {
+			conv.Unexpected(fmt.Sprintf("Couldn't process sql data row: %s", err))
+			// Scan failed, so we don't have any data to add to bad rows.
+			conv.StatsAddBadRow(srcTable, conv.DataMode())
+			continue
+		}
+		values := valsToStrings(v)
```
Deep1998, line 80: Wondering if there are any drawbacks to converting all types to strings then back to the relevant final types types. In Postgres, we receive an interface array as well, but there we do the modifications on a case by case basis inside the cvtSQLScalar() function. That said, converting all to strings and back seems cleaner, as we can reuse that for dump mode in the future, and avoids all the case by case handling we have to do.

Deep1998: Are there any data type transformations that could get messed up when converting to string and then to the final dll type?

sources/sqlserver/README.md
```diff
@@ -0,0 +1,68 @@
...
+| ROWVERSION             | TIMESTAMP    |
+| BINARY                 | BYTES        |
+| VARBINARY              | BYTES        |
+| IMAGE                  | BYTES        |
+| XML                    | STRING       |
+| UNIQUEIDENTIFIER       | STRING       |
+| SQL_VARIANT            | STRING       |
+| HIERARCHYID            | STRING       |
+| Spatial Geography Type | STRING       |
+| Spatial Geometry Types | STRING       |
```
Deep1998, line 68: Spatial Geometry Types seems repeated

sources/sqlserver/toddl.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+// Override the types to map to experimental postgres types.
+func overrideExperimentalType(columnType schema.Type, originalType ddl.Type) ddl.Type {
+	if len(columnType.ArrayBounds) > 0 {
+		return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}
+	} else if columnType.Name == "date" {
+		return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}
+	} else if columnType.Name == "json" || columnType.Name == "jsonb" {
```
Deep1998, line 99: I think sql server does noth ave a json tpe so this can be removed.

sources/sqlserver/infoschema.go
```diff
@@ -0,0 +1,383 @@
+// Copyright 2020 Google LLC
```
Deep1998, line 1: 2021 :) probably in other files too

sources/sqlserver/toddl.go
```diff
@@ -0,0 +1,103 @@
...
+// mods) into a Spanner type. This is the core source-to-Spanner type
+// mapping.  toSpannerType returns the Spanner type and a list of type
+// conversion issues encountered.
+func (tdi ToDdlImpl) ToSpannerType(conv *internal.Conv, columnType schema.Type) (ddl.Type, []internal.SchemaIssue) {
+	ty, issues := toSpannerTypeInternal(conv, columnType.Name, columnType.Mods)
+	if conv.TargetDb == constants.TargetExperimentalPostgres {
+		ty = overrideExperimentalType(columnType, ty)
+	} else {
+		if len(columnType.ArrayBounds) > 1 {
+			ty = ddl.Type{Name: ddl.String, Len: ddl.MaxLength}
```
Deep1998, line 44: Looks like sql server does not support arrays. Suggest adding a comment regarding the same and remove the array handling here as well as inside overrideExperimentalType().

Deep1998 (changes requested):
Had a first look, overall looks good! Have a few comments here and there

sources/sqlserver/infoschema.go
sagarsearce: okay, sure.

sources/sqlserver/infoschema.go
sagarsearce: It was regarding maintaining the order of the column when fetching the index from the database, it is already completed, we forgot to remove the todo comment, will remove that in the next commit.

sources/sqlserver/README.md
sagarsearce: one is Geometry and one is Geography is different types in SQL Server, that's we mentioned both.

sources/sqlserver/toddl.go
sagarsearce: agreed, will be changed in the next commit.

sources/sqlserver/infoschema.go
sagarsearce: noted. :)

sources/sqlserver/toddl.go
sagarsearce: agreed, will remove it in the next commit.

frontend/js/pages/SchemaConversionScreen/SchemaConversionScreen.page.js
```diff
@@ -16,81 +16,81 @@ import "../../services/Fetch.service.js";
...
+    clearInterval(this.stateObserver);
+  }
 
-    disconnectedCallback() {
-        clearInterval(this.stateObserver);
+  sendDatatoReportTab(tableNameArray, currentTabContent) {
+    for (let i = 0; i < tableNameArray.length; i++) {
+      let filterdata = {
+        SpSchema: currentTabContent.SpSchema[tableNameArray[i]],
+        SrcSchema: currentTabContent.SrcSchema[currentTabContent.ToSource[tableNameArray[i]].Name],
```
bharadwaj-aditya, line 37: Looks like the type of TableNameArray has changed from a string to a more complex object. Was this for something SQL server specific ?

sources/sqlserver/README.md
```diff
@@ -0,0 +1,68 @@
+# HarbourBridge: SQLServer-to-Spanner Evaluation and Migration
+
+HarbourBridge is a stand-alone open source tool for Cloud Spanner evaluation,
+using data from an existing PostgreSQL,MySQL,SQL server database. This README provides
+details of the tool's SQLserver capabilities. For general HarbourBridge information
```
bharadwaj-aditya, line 5: Please also mention the alias in the first para

sources/sqlserver/README.md
```diff
@@ -0,0 +1,68 @@
...
+```sh
+harbourbridge schema -source=sqlserver -source-profile="host=<>,port=<>,user=<>,db_name=<>"
+```
+
+Parameters `port` and `password` are optional. Port (`port`) defaults to `1433`
+for SQLserver source. Password can be provided at the password prompt.
+
+## Schema Conversion
+
+| MSSQL_Type             | Spanner_Type |
```
bharadwaj-aditya, line 33: Lets use the same name everywhere. Please change this toe SQL Server type

sources/sqlserver/data.go
```diff
@@ -0,0 +1,263 @@
...
+		// If timezone is not specified in mysqldump, we consider UTC time.
+		if TimezoneOffset == "" {
+			TimezoneOffset = "+00:00"
+		}
+		// convert timestamp from format "2006-01-02 15:04:05" to
+		// "2006-01-02T15:04:05+00:00".
+		timeNew := strings.Split(val, " ")
+		timeJoined := strings.Join(timeNew, "T")
+		timeJoined = timeJoined + TimezoneOffset
+		t, err = time.Parse(time.RFC3339, timeJoined)
```
bharadwaj-aditya, line 196: RFC3339 is not the same as ISO8601. What is the default in SQL server ?
https://stackoverflow.com/questions/522251/whats-the-difference-between-iso-8601-and-rfc-3339-date-formats

sources/sqlserver/data.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// processQuote returns the unquoted version of s.
+// Note: The element values of PostgreSQL arrays may have double
+// quotes around them.  The array output routine will put double
+// quotes around element values if they are empty strings, contain
+// curly braces, delimiter characters, double quotes, backslashes, or
+// white space, or match the word NULL. Double quotes and backslashes
+// embedded in element values will be backslash-escaped.  See section
+// 8.14.6.of www.postgresql.org/docs/9.1/arrays.html.
```
bharadwaj-aditya, line 257: This is postgres specific, should not be here ?

sources/sqlserver/infoschema.go
```diff
@@ -0,0 +1,383 @@
...
+}
+
+// GetToDdl function below implement the common.InfoSchema interface.
+func (isi InfoSchemaImpl) GetToDdl() common.ToDdl {
+	return ToDdlImpl{}
+}
+
+// GetTableName returns table name.
+func (isi InfoSchemaImpl) GetTableName(schema string, tableName string) string {
+	if schema == "public" { // Drop 'public' prefix.
```
bharadwaj-aditya, line 41: Is this relevant in SQL server ?

sources/sqlserver/infoschema.go
```diff
@@ -0,0 +1,383 @@
...
+		}
+		values := valsToStrings(v)
+		ProcessDataRow(conv, srcTable, srcCols, srcSchema, spTable, spCols, spSchema, values)
+	}
+	return nil
+}
+
+// GetRowsFromTable returns a sql Rows object for a table.
+func (isi InfoSchemaImpl) GetRowsFromTable(conv *internal.Conv, srcTable string) (interface{}, error) {
+	q := fmt.Sprintf(`SELECT * FROM %s;`, srcTable)
```
bharadwaj-aditya, line 88: In SQL server can we  fire queries across databases ? If yes then this query needs to include database name

sources/sqlserver/infoschema.go
```diff
@@ -0,0 +1,383 @@
...
+// GetTables return list of tables in the selected database.
+func (isi InfoSchemaImpl) GetTables() ([]common.SchemaAndName, error) {
+	q := `
+	SELECT 
+		SCH.name AS table_schema, 
+		TBL.name AS table_name
+	FROM sys.tables AS TBL
+	INNER JOIN sys.schemas AS SCH 
+	ON SCH.schema_id = TBL.schema_id
+	WHERE TBL.type = 'U' AND TBL.is_tracked_by_cdc = 0 AND TBL.is_ms_shipped = 0 AND TBL.name <> 'sysdiagrams'
```
bharadwaj-aditya, line 133: We might need to filter on schema name also here. Check the case where multiple schema's have the same tables

sources/sqlserver/infoschema.go
```diff
@@ -0,0 +1,383 @@
...
+			m[col] = append(m[col], constraint)
+		}
+	}
+	return primaryKeys, m, nil
+}
+
+// GetForeignKeys returns a list of all the foreign key constraints.
+func (isi InfoSchemaImpl) GetForeignKeys(conv *internal.Conv, table common.SchemaAndName) (foreignKeys []schema.ForeignKey, err error) {
+	q := `
+	SELECT
```
bharadwaj-aditya, line 252: check if we need to add schema name as a part of this query

web/sqlserver_typemap.go
```diff
@@ -0,0 +1,117 @@
...
+// Spanner type name is specified and is a potential mapping for this source type,
+// then it will be used to build the returned ddl.Type. If not, the default
+// Spanner type for this source type will be used.
+// Note that toSpannerTypeSQLserver is extensively tested via tests in web_test.go.
+//
+// TODO: Move the type remapping function to toddl.go (once we've merged
+// dynamodb/toddl.go, mysql/toddl.go, sqlserver/toddl.go and postgres/toddl.go).
+// Consider some refactoring to reduce code duplication (although note
+// that this type remapping has to preserve all previous changes done via the UI!)
+func toSpannerTypeSQLserver(srcType string, spType string, mods []int64) (ddl.Type, []internal.SchemaIssue) {
```
bharadwaj-aditya, line 36: General question: Is there a way to get this from toDdl to avoid duplication of code for all the databases ?

bharadwaj-aditya (changes requested):

sources/sqlserver/infoschema.go
```diff
@@ -0,0 +1,383 @@
...
+	SELECT 
+		OBJECT_SCHEMA_NAME (FK.referenced_object_id) AS [schema_name],
+		OBJECT_NAME (FK.referenced_object_id) AS [referenced_table],
+		COL_NAME(FKC.parent_object_id, FKC.parent_column_id) AS [column],  
+		COL_NAME(FKC.referenced_object_id, FKC.referenced_column_id) AS [referenced_column],  
+		FK.name AS [foreign_key_name]
+	FROM sys.foreign_keys AS FK  
+	INNER JOIN sys.foreign_key_columns AS FKC   
+    ON FK.object_id = FKC.constraint_object_id  
+	WHERE FK.parent_object_id = OBJECT_ID(@p1);
```
Deep1998, line 261: Im unsure what OBJECT_ID is, but I see table.Schema goes as an input to it. If we are not using table.Name here since I dont see a @p2 in the query, we can prolly remove it from line 264. Similarly in other queries (if any)

anupam-searce: -OBJECT_ID returns the id of the database object, in this case, the table that is passed in @p1
-Actually, only one argument is passed in the next statement. It is formatted like "<schma name>.<table name>". The query will fetch the detail for the required schema and table only.

sources/sqlserver/infoschema.go
anupam-searce: We are passing the schema and table through @p1. It is formatted like "[schma name].[table name]" in the next statement.

sources/sqlserver/infoschema.go
anupam-searce: Agreed, this query will return all the tables in all schema. 
However, we are using the schema and table both (Either in this format "[schema name].[table name]" or separately as parameters) in other queries where we fetch detail about a specific table

sources/sqlserver/infoschema.go
anupam-searce: Yes, queries across databases are possible in SQL Server. But here, as no database is passed, the query will be executed on the database which is in the connection object/context and which is supplied through the command line.

Any other use case that can cause any issue?

cmd/source_profile.go
```diff
@@ -200,10 +240,11 @@ func NewSourceProfileConnectionDynamoDB(params map[string]string) (SourceProfile
 }
 
 type SourceProfileConnection struct {
-	ty    SourceProfileConnectionType
-	mysql SourceProfileConnectionMySQL
-	pg    SourceProfileConnectionPostgreSQL
-	dydb  SourceProfileConnectionDynamoDB
+	ty        SourceProfileConnectionType
```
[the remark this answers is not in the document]
anupam-searce: Agreed, will fix that. 
Not sure how that happened. As gofmt (configured in vscode) should always format the code while saving

sources/sqlserver/data.go
anupam-searce: Sure, we have already removed those will be fixed in the next PR update.

sources/sqlserver/README.md
anupam-searce: Sure, changes will be available in the next PR update

cmd/source_profile.go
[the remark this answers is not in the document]
anupam-searce: We have checked that, and the indentation looks fine in the final block

cmd/source_profile.go
```diff
@@ -160,6 +161,45 @@ func NewSourceProfileConnectionPostgreSQL(params map[string]string) (SourceProfi
...
+
+func NewSourceProfileConnectionSqlServer(params map[string]string) (SourceProfileConnectionSqlServer, error) {
+	ss := SourceProfileConnectionSqlServer{}
+	host, hostOk := params["host"]
+	user, userOk := params["user"]
+	db, dbOk := params["db_name"]
+	port, _ := params["port"]
+	pwd, _ := params["password"]
+
+	if hostOk && userOk && dbOk {
```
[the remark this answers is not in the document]
anupam-searce: Agreed, we will update the code to get the values from env variables if those are not provided in the CLI.

sources/sqlserver/README.md
anupam-searce: Updated.

sources/sqlserver/infoschema.go
anupam-searce: You are right, fixed and ignored "dbo" prefix. It and will be available in the next PR

sources/sqlserver/infoschema.go
Deep1998: If possible, its best we keep it consistent across all queries. Since we do specify the db name in other places, maybe we should do that here as well.

sources/sqlserver/data.go
```diff
@@ -57,10 +57,7 @@ func ConvertData(conv *internal.Conv, srcTable string, srcCols []string, srcSche
 	}
 	for i, spCol := range spCols {
 		srcCol := srcCols[i]
-		// Skip columns with 'NULL' values. When processing data rows from mysqldump, these values
-		// are represented as nil (by pingcap/tidb/types/parser_driver's ValueExpr), which is
-		// converted to the string '<nil>'. When processing data rows obtained from the MySQL driver,
-		// 'NULL' values are represented as "NULL" (because we retrieve the values as strings).
+		// Skip columns with 'NULL' values.
 		if vals[i] == "<nil>" || vals[i] == "NULL" {
```
Deep1998, line 61: We should remove the vals[i] == "<nil>" condition.

sources/sqlserver/infoschema.go
anupam-searce: Agreed, will update the queries.

sources/sqlserver/data.go
```diff
@@ -180,84 +171,28 @@ func convNumeric(conv *internal.Conv, val string) (interface{}, error) {
...
 		timeJoined = timeJoined + TimezoneOffset
 		t, err = time.Parse(time.RFC3339, timeJoined)
 	} else {
 		// datetime: data should just consist of date and time.
-		// timestamp conversion should ignore timezone. We mimic this using Parse
-		// i.e. treat it as UTC, so it will be stored 'as-is' in Spanner.
+		// timestamp conversion should ignore timezone.
+		// val will be in this format "2021-12-15 07:39:52.943 +0000 UTC"
+		// the part after time can be ignored
+		if idx := strings.Index(val, "+"); idx != -1 {
```
Deep1998, line 189: nit: Since this part is common in to both if and else, move this to before the if-else block.
Also, comments convention is that sentences should end with a full stop and start with a capital letter (unless specifying variable/function names that start with a small letter like val in this case).

sources/sqlserver/data_test.go
```diff
@@ -0,0 +1,299 @@
...
+			cols:  []string{"b", "c", "a"},
+			vals:  []string{"6.6", "1", "6"},
+			ecols: []string{"b", "c", "a"},
+			evals: []interface{}{float64(6.6), true, int64(6)},
+		},
+		{
+			name:  "Null column",
+			cols:  []string{"a", "b", "c"},
+			vals:  []string{"6", "<nil>", "1"},
+			ecols: []string{"a", "c"},
```
Deep1998, line 136: This should also change since for nil values, we plan on going with `"NULL"` instead of `"<nil>"` as per the valsToStrings function.

sources/sqlserver/toddl.go
```diff
@@ -76,27 +67,25 @@ func toSpannerTypeInternal(conv *internal.Conv, id string, mods []int64) (ddl.Ty
...
 	case "ntext", "text", "xml":
 		return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, nil
-	case "smalldatetime", "datetimeoffset", "datetime2", "datetime", "timestamp":
+	case "smalldatetime", "datetimeoffset", "datetime2", "datetime":
 		return ddl.Type{Name: ddl.Timestamp}, []internal.SchemaIssue{internal.Timestamp}
 	case "time":
 		return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.Time}
 	case "varchar", "char", "nvarchar", "nchar":
-		if len(mods) > 0 && mods[0] > 0 && mods[0] <= StringLimit {
+		if len(mods) > 0 && mods[0] > 0 && mods[0] <= stringLimit {
```
Deep1998, line 75: Suggest adding a comment explaining this custom string and byte limit.

Deep1998 (commented):
Had another run through, minor comments.

frontend/js/pages/SchemaConversionScreen/SchemaConversionScreen.page.js
anupam-searce: This change is to fix an existing issue in the UI where it is unable to extract the table information when the table schema is different than the default one.

sources/sqlserver/data.go
anupam-searce: You are right, removed.

sources/sqlserver/data.go
anupam-searce: The method that had that block has been updated. Now the query returns the dates in the desired format. So few of the codes were removed.

sources/sqlserver/data_test.go
anupam-searce: Correct, updated with "NULL"

sources/sqlserver/toddl.go
anupam-searce: Added a comment there

conversion/conversion.go
anupam-searce: These unused env variables are now removed.

cmd/source_profile.go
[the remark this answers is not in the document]
Deep1998: I was wondering if we should do that here, since we plan on deprecating the env variables anyways. Its best if we dont fetch that here either.

conversion/conversion.go
Deep1998: I'm looking at the latest commit at it looks like we still retain the support for env variables.

sources/sqlserver/README.md
```diff
@@ -0,0 +1,70 @@
+# HarbourBridge: SQLServer-to-Spanner Evaluation and Migration
+
+HarbourBridge is a stand-alone open source tool for Cloud Spanner evaluation,
+using data from an existing PostgreSQL,MySQL,SQL server database. This README provides
```
Deep1998, line 4: nit: maybe remove the PGSQL and MYSQL here since this is a sql server readme.

conversion/conversion.go
```diff
@@ -129,12 +130,15 @@ func connectionConfig(driver string, sqlConnectionStr string) (interface{}, erro
...
 	case constants.MYSQL:
-		// If empty, this is called as part of the legacy mode witih global CLI flags.
-		// When using source-profile mode is used, the sqlConnectionStr is already populated.
 		if sqlConnectionStr == "" {
 			return generateMYSQLConnectionStr()
 		}
 		return sqlConnectionStr, nil
+	case constants.SQLSERVER:
+		if sqlConnectionStr == "" {
+			return generateSQLSERVERConnectionStr()
```
Deep1998, line 139: Since we don't support the env variables anymore, sqlConnectionStr should never be empty. Consider refactoring this and adding a comment on why its left empty.

sources/sqlserver/toddl.go
Deep1998: I dont see byteLimit used anywhere.

sources/sqlserver/toddl.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"github.com/cloudspannerecosystem/harbourbridge/common/constants"
+	"github.com/cloudspannerecosystem/harbourbridge/internal"
+	"github.com/cloudspannerecosystem/harbourbridge/schema"
+	"github.com/cloudspannerecosystem/harbourbridge/spanner/ddl"
+)
+
+const (
+	stringLimit int64 = 2621440
```
Deep1998, line 26: Instead of redefining the limit here, consider defining a var StringMaxLength along with the spanner vars and using that value here. We already do it for [PGMaxLength](https://github.com/cloudspannerecosystem/harbourbridge/blob/master/spanner/ddl/ast.go#L69)

Deep1998: Note: We use StringMaxLength only to differentiate between an integer length and using `MAX`. We still assign `ddl.MaxLength` if the source length is > StringMaxLength (2621440)

sources/sqlserver/infoschema.go
Deep1998: Ah right my bad, missed the fmt.Sprintf

Deep1998 (commented):
Took another parse, have minor comments mainly regarding the deprecation of env variables. Looks pretty good overall, close to merge.

cmd/source_profile.go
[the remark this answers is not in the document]
bharadwaj-aditya: we won't be completely deprecating them. They idea is to keep the way of accessing HarbourBridge close to the way the client is generally used. So we support the environment variables also, though our suggest way will be using the source profile.

conversion/conversion.go
anupam-searce: Yes, I may have added that comment early. These should be gone by now.

sources/sqlserver/data.go
anupam-searce: This method has been updated and now the SQL query returns the DateTime in ISO8601

sources/sqlserver/toddl.go
anupam-searce: Right, that unused constant has been removed recently.

conversion/conversion.go
anupam-searce: Agreed, updated the code accordingly. Changes should be available now.

sources/sqlserver/README.md
anupam-searce: PG and MySQL ref removed.

sources/sqlserver/toddl.go
anupam-searce: Right, removed the constant from the current package and moved it to the ddl package.

web/sqlserver_typemap.go
anupam-searce: Right, we can do that and we added an action item. As planned we will do this in a later phase.

Deep1998:
LGTM!
Just one thing to go over, with the current changes we kind of still support sql server through the legacy cmd mode (not sure if it works though). Do we want to add some checks on the allowed drivers for legacy mode or do we go with the stance that new engines are not tested so it is not recommended to use legacy mode for them but we dont error out if they put sql server in the driver?

frontend/js/components/DataTable/DataTable.component.js
```diff
@@ -184,9 +184,9 @@ class DataTable extends HTMLElement {
         let maxPossiblePageNumber =  Math.ceil(tableColumnsArrayLength / columnPerPage);
         let tableColumnsArrayCurrent = [];
         let z = 0;
-        for(let i= pageNumber*columnPerPage ;i < Math.min(tableColumnsArrayLength, pageNumber*columnPerPage+15) ; i++ )
-        {
-            tableColumnsArrayCurrent[z]=tableColumnsArray[i];
+        for (let i = pageNumber * columnPerPage; i < Math.min(tableColumnsArrayLength, pageNumber * columnPerPage + 15); i++) {
+            if (tableColumnsArray[i] != 'synth_id')
```
bharadwaj-aditya, line 188: why is this ignored ? Do we not render the synth id ?

bharadwaj-aditya:
> LGTM! Just one thing to go over, with the current changes we kind of still support sql server through the legacy cmd mode (not sure if it works though). Do we want to add some checks on the allowed drivers for legacy mode or do we go with the stance that new engines are not tested so it is not recommended to use legacy mode for them but we dont error out if they put sql server in the driver?

Let us throw up an error if the driver is manually set which is not one of the previously supported databases - i.e. mysql, postgres or Dynamo

bharadwaj-aditya (approved):
LGTM - just one minor comment to check why the change was made

frontend/js/components/DataTable/DataTable.component.js
anupam-searce: As per the current structure, we are fetching the source/spanner column information using the source-spanner map. Here there is no column in the source and hence it can cause issues. Moreover, there is no source column to display in the grid for the synthetic key.
Do we need to display the synthetic key even though we do not have any column at the source?

Deep1998 (approved):

Issue #245, opened 2021-12-21
Handle NULL values during data conversion


Deep1998:
Currently, there is no handling for NULL data types during data conversion for pg_dump.

Issue #244, opened 2021-12-21
Handle Float and Null values (possibly others) in pgdump parser


Deep1998:
pgdump.go, getRows() currentl only has Node_String_ and Node_Integer in the case handling. This errors out when dump file has values Null or 15.124 since these categories come under neither. We should add handling for each data type over here.

PR #246, opened 2021-12-23
Pg spanner misc fixes

### LINKED ISSUE #241: TargetDB handling in getDDL() calls
func (s Schema) GetDDL(c Config) takes a config that accepts targetDB in the struct which is left out in a few places like the WriteSchemaFile() causing the generated schema files to be of GOOGLESQL dialect when specified dialect is PG.

ALso, FK statements still use backticks `` instead of "" for PG spanner. Check if any other handling is required.

Deep1998:
Fixes #241, #244 and #245 

- Added handling for float and null values
- For pg dump, null values were not getting skipped earlier.
- Quoting was incorrect for pg dialect, causing foreign key statements to fail
- The schema files were of spanner dialect since targetdb was not passed as a param.

### DIFF
--- a/sources/postgres/data.go
+++ b/sources/postgres/data.go
@@ -76,7 +76,10 @@ func ConvertData(conv *internal.Conv, srcTable string, srcCols []string, vals []
 	}
 	for i, spCol := range spCols {
 		srcCol := srcCols[i]
-		if vals[i] == "\\N" { // PostgreSQL representation of empty column in COPY-FROM blocks.
+		// "\\N" is for PostgreSQL representation of empty column in COPY-FROM blocks.
+		// TODO: Consider using NullString to differentiate between an actual column having "NULL" as a string
+		// and NULL values.
+		if vals[i] == "\\N" || vals[i] == "NULL" {
 			continue
 		}
 		spColDef, ok1 := spSchema.ColDefs[spCol]
--- a/sources/postgres/pgdump.go
+++ b/sources/postgres/pgdump.go
@@ -732,6 +732,12 @@ func getRows(conv *internal.Conv, vll []*pg_query.Node, n *pg_query.InsertStmt) 
 						// high priority (it isn't right now), then consider preserving int64
 						// here to avoid the int64 -> string -> int64 conversions.
 						values = append(values, strconv.FormatInt(int64(c.Integer.Ival), 10))
+					case *pg_query.Node_Float:
+						values = append(values, c.Float.Str)
+					case *pg_query.Node_Null:
+						values = append(values, "NULL")
+					// TODO: There are other Node types like Node_IntList, Node_List, Node_BitString etc that
+					// need to be explored.
 					default:
 						conv.Unexpected(fmt.Sprintf("Processing %v statement: found %s node for A_Const Val", printNodeType(n), printNodeType(c)))
 					}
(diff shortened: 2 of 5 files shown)


sources/postgres/pgdump.go
```diff
@@ -732,6 +732,12 @@ func getRows(conv *internal.Conv, vll []*pg_query.Node, n *pg_query.InsertStmt)
 						// high priority (it isn't right now), then consider preserving int64
 						// here to avoid the int64 -> string -> int64 conversions.
 						values = append(values, strconv.FormatInt(int64(c.Integer.Ival), 10))
+					case *pg_query.Node_Float:
+						values = append(values, c.Float.Str)
+					case *pg_query.Node_Null:
```
Deep1998, line 737: For rows of the form (NULL, 14.67), the pg_query node type is Node_Null and Node_Float respectively. 

Data dumped with quotes ('NULL', '14.67') would have worked as it would get classified as Node_String_ but without quotes, the pg_query parser assigns a different type to the data.

spanner/ddl/ast.go
```diff
@@ -158,7 +158,11 @@ type Config struct {
 
 func (c Config) quote(s string) string {
 	if c.ProtectIds {
-		return "`" + s + "`"
+		if c.TargetDb == constants.TargetExperimentalPostgres {
+			return "\"" + s + "\""
```
Deep1998, line 162: PG dialect uses " instead of `

sources/postgres/data.go
```diff
@@ -76,7 +76,10 @@ func ConvertData(conv *internal.Conv, srcTable string, srcCols []string, vals []
 	}
 	for i, spCol := range spCols {
 		srcCol := srcCols[i]
-		if vals[i] == "\\N" { // PostgreSQL representation of empty column in COPY-FROM blocks.
+		// "\\N" is for PostgreSQL representation of empty column in COPY-FROM blocks.
+		// TODO: Consider using NullString to differentiate between an actual column having "NULL" as a string
```
Deep1998, line 80: Kept this as a todo because we currently skip if vals[i] == "NULL" across all engines.

sources/postgres/pgdump.go
bharadwaj-aditya: what is the different type that gets assigned ? Add this description in comments within the code

spanner/ddl/ast_test.go
```diff
@@ -88,7 +88,7 @@ func TestPrintColumnDefPG(t *testing.T) {
 		{in: ColumnDef{Name: "col1", T: Type{Name: Int64, IsArray: true}}, expected: "col1 VARCHAR(2621440)"},
 		{in: ColumnDef{Name: "col1", T: Type{Name: Int64}, NotNull: true}, expected: "col1 INT8 NOT NULL"},
 		{in: ColumnDef{Name: "col1", T: Type{Name: Int64, IsArray: true}, NotNull: true}, expected: "col1 VARCHAR(2621440) NOT NULL"},
-		{in: ColumnDef{Name: "col1", T: Type{Name: Int64}}, protectIds: true, expected: "`col1` INT8"},
+		{in: ColumnDef{Name: "col1", T: Type{Name: Int64}}, protectIds: true, expected: "\"col1\" INT8"},
```
bharadwaj-aditya, line 91: how were the backticks working earlier ?

bharadwaj-aditya (changes requested):

spanner/ddl/ast_test.go
Deep1998: Because the quote() function did not have handling for PG. It always defaulted to backticks. This Pr added the changes in quote() above that if dialect is pg, use double quotes.

sources/postgres/pgdump.go
Deep1998: Done

bharadwaj-aditya (approved):
LGTM

PR #248, opened 2021-12-28
Refactoring 1: Cleanup profiles and move conversion helpers methods to utils

Deep1998:
This PR refactors 2 things:

- Move helper functions from `conversion/conversion.go` to `common/utils/utils.go`:
Currently, a lot of helper functions are stashed inside conversion/conversion.go which are also used by other packages. Often, other packages import conversion just to use these helper functions.
This is also required so that source profile does not need to import conversion.go and cyclic imports issue can be resolved. This will come in handy when we refactor the code to pass source and target profiles throughout.

- Breakout source and target profiles code from cmd folder: 
The profile related files are moved into a new package called `profiles`. The folder `cmd/` only has the subcommand files. The cmd directory imports the conversion package and profiles package. Conversion imports just the profiles package.
This refactoring is necessary to solve the cyclic dependency issue as well.


cmd/common.go
```diff
@@ -1,111 +0,0 @@
-package cmd
```
bharadwaj-aditya, line 1: use git move for this. This seems like a new file created in profiles

bharadwaj-aditya (changes requested):
Looks fine overall, but just check the moved files are all done via "git mv" where possible, so that we maintain history

cmd/common.go
Deep1998: I did use git mv for this , but once i add the changes required to this file, it ends up marking this as a new file. It only happens for this files and works for the rest.

Deep1998:
We have this kind of repetition in multiple places like cmd/cmd.go, common/constants/constants.go, conversion/conversion.go. But if we were to resolve this, we can go 2 ways about it. 
To avoid confusion it is a good practice to keep folder name and package name same. 

- With just common/utils.go, we would change the package name to `package common` which seems very generic for util functions. 
- Another alternative is to get rid of the common folder, and just have `constants/constants.go` and `utils/utils.go` in the root dir.
I'm leaning towards keeping the repetition since its happening in multiple places, but if were to change it, the latter seems better.

bharadwaj-aditya:
lets go with the structure defined here and figure out if we want to change at a later stage then. Approving

bharadwaj-aditya (approved):
LGTM

PR #249, opened 2022-01-04
Refactoring 2: Pass profiles across conversion methods instead of flags.

Deep1998:
This change ensures we pass Source Profile and Target Profile across the SchemaConv and Data conv functions.
We also store driver and targetdb in source and target profiles respectively.

### DIFF
--- a/profiles/common.go
+++ b/profiles/common.go
@@ -78,7 +78,7 @@ func GetResourceIds(ctx context.Context, targetProfile TargetProfile, now time.T
 	return project, instance, dbName, err
 }
 
-func GetSQLConnectionStr(sourceProfile *SourceProfile) string {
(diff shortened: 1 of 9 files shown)


profiles/common.go
```diff
@@ -78,16 +78,16 @@ func GetResourceIds(ctx context.Context, targetProfile TargetProfile, now time.T
 	return project, instance, dbName, err
 }
 
-func GetSQLConnectionStr(sourceProfile SourceProfile) string {
+func GetSQLConnectionStr(sourceProfile *SourceProfile) string {
```
bharadwaj-aditya, line 81: might make sense to pass the value rather than the reference everywhere, to safeguard the attributes from change.

bharadwaj-aditya (approved):
Minor comments.
Approving in any case.

PR #250, opened 2022-01-09
Support for CSV: First Cut

Deep1998:
First cut of changes to support CSV in harbourbridge. The functionality is limited in the first cut to just unblock customers.

Subcommand changes:
Added CSV as a source for data mode. The source profile accepts the path to the manifest file. No changes to target profile.
**Sample command:**
```
data -source=csv -source-profile="manifest=manifest.json" -target-profile="project=my-project,instance=spanner-instance,dbname=spanner-db"
```
The manifest structure is largely similar to dataflow's manifest with slight changes:
```
[
    {
      "table_name": "Albums",
      "file_patterns": [
        "/Users/deepchowdhury/Desktop/harbourbridge/Albums.csv"
      ],
      "columns": [
        {"column_name": "a", "type_name": "BOOL"},
        {"column_name": "b", "type_name": "BYTES(124)"},   // Length is optional
        {"column_name": "c", "type_name": "DATE"},
        {"column_name": "d", "type_name": "FLOAT64"},
        {"column_name": "e", "type_name": "INT64"},
        {"column_name": "f", "type_name": "NUMERIC"},
        {"column_name": "g", "type_name": "STRING"},
        {"column_name": "h", "type_name": "TIMESTAMP"},
        {"column_name": "i", "type_name": "JSON"}
      ]
    },
    {
      "table_name": "Singers",
      "file_patterns": [
        "/Users/deepchowdhury/Desktop/harbourbridge/Singers.csv"
      ],
      "columns": [
        {"column_name": "SingerId", "type_name": "INT64"},
        {"column_name": "FirstName", "type_name": "STRING(100)"},  // Length is optional
        {"column_name": "LastName", "type_name": "STRING"}
      ]
    }
]
```
**Note regarding the manifest:**

- The `table_name` field should be identical to the table name in Spanner schema.
- File patterns do not accept regex expressions. Provide the path inside double quotes.
- The type_name should be identical to the types in spanner schema. Only for `STRING` and `BYTES`, the length can be optionally omitted.

**CSV file format:**
```
bool_col,byte_col,date_col,float_col,int_col,numeric_col,string_col,timestamp_col,json_col // column headers
true,bytevalue,2020-12-09,15.13,100,39.94,Helloworld,2019-10-29 05:30:00,"{""key1"": ""value1"", ""key2"": ""value2""}" // data
```
Note regarding the CSV file:

- Each column in a row should be separated by commas.
- The first row needs to have the column name headers. The column names should be identical to the column names in Spanner schema.
- Remove trailing spaces, tabs in the column name headers.

**CSV Data types:**

- We only support scalar data types right now. Sample data format for each provided in the snippet above.
- The only supported data format right now is RFC3339 full-date format
- The only supported timestamp format right now is ISO 8601
- The format to escape the quotes in json is adding an additional " in front of the double quote. \ does not work. Also enclose the whole data inside "". Some modification might be required since most databases do not export CSVs with escaping quotes like mentioned.

**Pending tasks:**

Interleaved table support
PG Spanner support
Stretch Goal:
Implement as a full fledged feature where manifest file is not required.

### DIFF
--- a/cmd/data.go
+++ b/cmd/data.go
@@ -59,47 +59,6 @@ func (cmd *DataCmd) SetFlags(f *flag.FlagSet) {
 	f.StringVar(&cmd.filePrefix, "prefix", "", "File prefix for generated files")
 }
 
-// TODO: refactor this to go through DataConv(). Fix it when refactoring passing of source-profile throughout the code
-// to avoid excess parameters in the DataConv and SchemaConv functions.
-func executeCSVLoader(ctx context.Context, conv *internal.Conv, sourceProfile profiles.SourceProfile, targetProfile profiles.TargetProfile, filePrefix string) subcommands.ExitStatus {
-	if targetProfile.Conn.Sp.Dbname == "" {
-		fmt.Printf("dbname is mandatory in target-profile for csv source")
-		return subcommands.ExitFailure
-	}
-	ioHelper := utils.NewIOStreams(constants.CSV, "")
-	if ioHelper.SeekableIn != nil {
-		defer ioHelper.In.Close()
-	}
-	now := time.Now()
-	project, instance, dbName, err := profiles.GetResourceIds(ctx, targetProfile, now, constants.CSV, ioHelper.Out)
-	if err != nil {
-		return subcommands.ExitUsageError
-	}
-	dbURI := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName)
-	client, err := utils.GetClient(ctx, dbURI)
-	if err != nil {
-		fmt.Printf("can't create client for db %s: %v", dbURI, err)
-		return subcommands.ExitFailure
-	}
-	defer client.Close()
-
-	delimiter := sourceProfile.Csv.Delimiter
-	if len(delimiter) != 1 {
-		fmt.Printf("delimiter should only be a single character long, found '%s'", delimiter)
-		return subcommands.ExitFailure
-	}
-	bw, err := conversion.DataFromCSV(conv, sourceProfile.Csv.Manifest, client, targetProfile.TargetDb, sourceProfile.Csv.NullStr, rune(delimiter[0]))
-	if err != nil {
-		fmt.Printf("can't finish data conversion for db %s: %v", dbURI, err)
-		return subcommands.ExitFailure
-	}
-	banner := utils.GetBanner(now, dbURI)
-	conversion.Report(constants.CSV, bw.DroppedRowsByTable(), ioHelper.BytesRead, banner, conv, filePrefix+reportFile, ioHelper.Out)
-	conversion.WriteBadData(bw, conv, banner, filePrefix+badDataFile, ioHelper.Out)
-
-	return subcommands.ExitSuccess
-}
-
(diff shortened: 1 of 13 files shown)


cmd/data.go
```diff
@@ -102,14 +104,23 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
...
+		err = fmt.Errorf("can't create client for db %s: %v", dbURI, err)
+		return subcommands.ExitFailure
 	}
-	if targetProfile.TargetDb != "" && conv.TargetDb != targetProfile.TargetDb {
-		err = fmt.Errorf("running data migration for Spanner dialect: %v, whereas schema mapping was done for dialect: %v", targetProfile.TargetDb, conv.TargetDb)
-		return subcommands.ExitUsageError
+	defer client.Close()
+
+	// For csv mode, skip loading a session file.
+	if sourceProfile.Driver != constants.CSV {
```
bharadwaj-aditya, line 115: we should avoid mentioning of csv or any specific profile here.
Similar to UseTargetSchema, we can add a method - supportSessionFile or something on the sourceProfile

profiles/source_profile.go
```diff
@@ -290,6 +318,14 @@ func (src SourceProfile) ToLegacyDriver(source string) (string, error) {
 		}
 	case SourceProfileTypeConfig:
 		return "", fmt.Errorf("specifying source-profile using config not implemented")
+	case SourceProfileTypeCsv:
+		{
+			if strings.ToLower(source) == constants.CSV {
+				return constants.CSV, nil
+			} else {
+				return "", fmt.Errorf("found manifest in the source profile but invalid -source flag received = %v, (did you mean csv?)", source)
```
bharadwaj-aditya, line 326: when will this case be encountered ? The sourceProfileType csv will only be encountered with the source is set to CSV, right ?

sources/csv/data.go
```diff
@@ -0,0 +1,292 @@
...
+			}
+			r := csvReader.NewReader(csvFile)
+			r.Comma = delimiter
+			count, err := getCSVRowCount(r)
+			if err != nil {
+				conv.Unexpected(fmt.Sprintf("Couldn't get number of rows for table %s", table.Table_name))
+				continue
+			}
+			if count == 0 {
+				conv.Unexpected(fmt.Sprintf("File %s is empty.", filePath))
```
bharadwaj-aditya, line 120: mention the table name also in this comment

bharadwaj-aditya (commented):
Just a few minor comments

cmd/data.go
Deep1998: Yes we should, my bad I missed this

profiles/source_profile.go
Deep1998: Good catch, I missed updating this. This was required earlier as per the code changes but not needed now.

cmd/data.go
Deep1998: Using UseTargetSchema as well for this differentiation since they pretty much are the same functionally.

bharadwaj-aditya (approved):
LGTM

PR #254, opened 2022-01-12
Feature : Oracle db schema conversion

sagarsearce:
## Oracle Db schema conversion Feature 
- oracle driver package added
- source profile added for oracle db
- info schema interface implemented for schema conversion
- Type map added for oracle 
- unit test added


conversion/conversion.go
```diff
@@ -156,6 +159,11 @@ func getDbNameFromSQLConnectionStr(driver, sqlConnectionStr string) string {
 	case constants.SQLSERVER:
 		splts := strings.Split(sqlConnectionStr, "?database=")
 		return splts[len(splts)-1]
+	case constants.ORACLE:
+		// connection string formate : "oracle://user:password@104.108.154.85:1521/XE"
+		substr := sqlConnectionStr[9:]
+		dbName := strings.Split(substr, ":")[0]
```
Deep1998, line 165: This seems to return user instead of db name, I take it the dbname in the sample string is XE?

sources/oracle/infoschema.go
```diff
@@ -0,0 +1,291 @@
...
+				Columns:      fKeys[k].Cols,
+				ReferTable:   fKeys[k].Table,
+				ReferColumns: fKeys[k].Refcols})
+	}
+	return foreignKeys, nil
+}
+
+// GetIndexes return a list of all indexes for the specified table.
+// Oracle db support several types of index: 1. Normal indexes. (By default, Oracle Database creates B-tree indexes.)
+// 2.Bitmap indexes 3.Partitioned indexes 4. Function-based indexes 5.Domain indexes,
```
Deep1998, line 216: nit:maybe put newlines after each point

sources/oracle/toddl.go
```diff
@@ -0,0 +1,100 @@
...
+	case "NUMBER":
+		// If no scale is avalible then map it to int64, and numeric elsewhere.
+		if len(mods) == 1 && mods[0] >= 1 && mods[0] < 19 {
+			return ddl.Type{Name: ddl.Int64}, nil
+		} else {
+			return ddl.Type{Name: ddl.Numeric}, nil
+		}
+	case "BFILE", "BLOB":
+		return ddl.Type{Name: ddl.Bytes, Len: ddl.MaxLength}, nil
+	case "CHAR", "CHARATER":
```
Deep1998, line 52: CHARACTER?

sources/oracle/toddl.go
```diff
@@ -0,0 +1,100 @@
...
+		return ddl.Type{Name: ddl.String, Len: mods[0]}, nil
+	case "CLOB":
+		return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, nil
+	case "DATE", "DATETIME", "TIMESTAMPWITHTIMEZONE", "TIMESTAMP":
```
Deep1998, line 56: Are none of them a good match for ddl.Date?

sources/oracle/toddl.go
```diff
@@ -0,0 +1,100 @@
...
+		return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, nil
+	default:
+		return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.NoGoodType}
+	}
+}
+
+// Override the types to map to experimental postgres types.
+func overrideExperimentalType(columnType schema.Type, originalType ddl.Type) ddl.Type {
+	if columnType.Name == "DATE" {
+		return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}
```
Deep1998, line 97: This is incorrect since we currently map DATE to timestamp which is supported by PG spanner.
We should either map Date to ddl.Date by default or remove this.

conversion/conversion.go
anupam-searce: Actually, XE is the SID. Here user needs to be considered. In oracle User / Owner is essentially the same as the Schema. In the queries we need to pass the schema.

sources/oracle/infoschema.go
anupam-searce: Done

sources/oracle/toddl.go
anupam-searce: CHARACTER is not a standard type and therefore it has been removed

sources/oracle/toddl.go
anupam-searce: DATE will be mapped to ddl.Date. We have removed other types from the case statement and handled timestamps differently

sources/oracle/toddl.go
anupam-searce: As DATE is now mapped to ddl.Date, this check becomes valid.

bharadwaj-aditya (approved):
LGTM

PR #256, opened 2022-01-18
CSV 2

Deep1998:
This adds the following features on top of csv:

- No longer provide data types in manifest file. HB reads spanner infoschema to fetch the data types.
- Support for PG spanner
- Interleaved table support
- GCS file support
- Array support added

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -18,9 +18,11 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
+	"github.com/cloudspannerecosystem/harbourbridge/common/constants"
 	"github.com/cloudspannerecosystem/harbourbridge/common/utils"
 	"github.com/cloudspannerecosystem/harbourbridge/conversion"
 	"github.com/cloudspannerecosystem/harbourbridge/internal"
@@ -97,7 +99,7 @@ func CommandLine(ctx context.Context, driver, targetDb, dbURI string, dataOnly, 
 
 	// We pass an empty string to the sqlConnectionStr parameter as this is the legacy codepath,
 	// which reads the environment variables and constructs the string later on.
-	bw, err := conversion.DataConv(sourceProfile, targetProfile, ioHelper, client, conv, dataOnly)
+	bw, err := conversion.DataConv(ctx, sourceProfile, targetProfile, ioHelper, client, conv, dataOnly)
 	if err != nil {
 		return fmt.Errorf("can't finish data conversion for db %s: %v", dbURI, err)
 	}
@@ -109,5 +111,7 @@ func CommandLine(ctx context.Context, driver, targetDb, dbURI string, dataOnly, 
 	banner := utils.GetBanner(now, dbURI)
 	conversion.Report(driver, bw.DroppedRowsByTable(), ioHelper.BytesRead, banner, conv, outputFilePrefix+reportFile, ioHelper.Out)
 	conversion.WriteBadData(bw, conv, banner, outputFilePrefix+badDataFile, ioHelper.Out)
+	// Cleanup hb tmp data directory.
+	os.RemoveAll(os.TempDir() + constants.HB_TMP_DIR)
 	return nil
 }
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -43,7 +43,8 @@ var (
 // 3. Run data conversion (if schemaOnly is set to false)
 // 4. Generate report
 func CommandLine(ctx context.Context, driver, targetDb, dbURI string, dataOnly, schemaOnly, skipForeignKeys bool, schemaSampleSize int64, sessionJSON string, ioHelper *utils.IOStreams, outputFilePrefix string, now time.Time) error {
-
+	// Cleanup hb tmp data directory in case residuals remain from prev runs.
+	os.RemoveAll(os.TempDir() + constants.HB_TMP_DIR)
 	// Legacy mode is only supported for MySQL, PostgreSQL and DynamoDB
 	if driver != "" && utils.IsValidDriver(driver) && !utils.IsLegacyModeSupportedDriver(driver) {
 		return fmt.Errorf("legacy mode is not supported for drivers other than %s", strings.Join(utils.GetLegacyModeSupportedDrivers(), ", "))
--- a/common/utils/utils.go
+++ b/common/utils/utils.go
@@ -39,6 +39,12 @@ type IOStreams struct {
 	BytesRead           int64
 }
 
+// Harbourbridge accepts a manifest file in the form of a json which unmarshalls into the ManifestTables struct.
+type ManifestTable struct {
+	Table_name    string   `json:"table_name"`
+	File_patterns []string `json:"file_patterns"`
+}
+
 // NewIOStreams returns a new IOStreams struct such that input stream is set
 // to open file descriptor for dumpFile if driver is PGDUMP or MYSQLDUMP.
 // Input stream defaults to stdin. Output stream is always set to stdout.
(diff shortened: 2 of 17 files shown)


common/utils/utils.go
```diff
@@ -289,6 +294,12 @@ func ContainsAny(s string, l []string) bool {
 	return false
 }
 
+func ValueSetEqual(a, b []string) bool {
```
bharadwaj-aditya, line 297: the name is unclear.
You can rename it to something like checkEqualSets or something

sources/csv/data.go
```diff
@@ -40,12 +44,33 @@ type Column struct {
...
 	File_patterns []string `json:"file_patterns"`
-	Columns       []Column `json:"columns"`
 }
 
-// LoadManifest reads the manifest file and unmarshalls it into a list of Table struct.
+func GetCSVFiles(conv *internal.Conv, sourceProfile profiles.SourceProfile) (tables []Table, err error) {
+	// If manifest file not provided, we assume the csvs exist in the same directory
+	// in table_name.csv format.
+	if sourceProfile.Csv.Manifest == "" {
+		for t := range conv.SpSchema {
```
bharadwaj-aditya, line 53: add a logger at all decision points like this, regarding which approach was chosen

sources/csv/data.go
```diff
@@ -68,37 +93,59 @@ func VerifyManifest(conv *internal.Conv, tables []Table) error {
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			missing = append(missing, name)
+		}
+	}
+	if len(missing) > 0 {
+		fmt.Printf("WARNING: did not find manifest entries for tables [ %s ], proceeding anyways...", strings.Join(missing, ", "))
```
bharadwaj-aditya, line 110: Ignoring and proceeding might be better than proceeding anyways

spanner/batchwriter/batchwriter.go
```diff
@@ -15,7 +15,7 @@
 // Package spanner provides high-level abstractions for working with
 // Cloud Spanner that are not available from the core Cloud Spanner
 // libraries.
-package spanner
+package batchwriter
```
bharadwaj-aditya, line 18: If we need this package, we should probably name it "writer" so we can put other writers we might build later in the same package

bharadwaj-aditya (changes requested):

common/utils/utils.go
```diff
@@ -415,3 +430,26 @@ func IsLegacyModeSupportedDriver(driver string) bool {
...
+	err := common.ProcessSchema(conv, infoSchema)
+	if err != nil {
+		return fmt.Errorf("error trying to read and convert spanner schema: %v", err)
+	}
+	parentTables, err := infoSchema.GetInterleaveTables()
+	if err != nil {
+		// We should ideally throw an error here as it could potentially cause a lot of failed writes.
+		// We raise an unexpected error for now to make it compatible with the integration tests.
+		// In the emulator, the interleave_type column in not supported hence the query fails.
+		conv.Unexpected(fmt.Sprintf("error trying to fetch interleave table info from schema: %v", err))
```
Deep1998, line 446: Check this out

common/utils/utils.go
```diff
@@ -85,22 +90,23 @@ func downloadFromGCS(bucketName string, filePath string) (*os.File, error) {
...
-	tmpfile, err := ioutil.TempFile("", "harbourbridge.gcs.data")
+	tmpDir := os.TempDir() + constants.HB_TMP_DIR
+	os.MkdirAll(tmpDir, os.ModePerm)
+	tmpfile, err := os.Create(tmpDir + "/" + tmpFile)
 	if err != nil {
 		fmt.Printf("saveFile: unable to open temporary file to save dump file from GCS bucket %v", err)
 		log.Fatal(err)
 		return nil, err
 	}
-	syscall.Unlink(tmpfile.Name()) // File will be deleted when this process exits.
```
Deep1998, line 94: This deletes the file as soon as file is closed. Works for dump files but not for csvs which are opened and closed later. Hence moving all downloads to a harbourbridge_tmp_data dir in tmp folder and deleting it at the end.

common/utils/utils.go
```diff
@@ -64,8 +69,8 @@ func NewIOStreams(driver string, dumpFile string) IOStreams {
 	return io
 }
 
-// downloadFromGCS returns the dump file that is downloaded from GCS
-func downloadFromGCS(bucketName string, filePath string) (*os.File, error) {
+// DownloadFromGCS returns the dump file that is downloaded from GCS.
+func DownloadFromGCS(bucketName, filePath, tmpFile string) (*os.File, error) {
```
bharadwaj-aditya, line 73: Is this function used from outside this package ?

cmd/cmd.go
```diff
@@ -109,5 +111,7 @@ func CommandLine(ctx context.Context, driver, targetDb, dbURI string, dataOnly,
 	banner := utils.GetBanner(now, dbURI)
 	conversion.Report(driver, bw.DroppedRowsByTable(), ioHelper.BytesRead, banner, conv, outputFilePrefix+reportFile, ioHelper.Out)
 	conversion.WriteBadData(bw, conv, banner, outputFilePrefix+badDataFile, ioHelper.Out)
+	// Cleanup hb tmp data directory.
+	os.RemoveAll(os.TempDir() + constants.HB_TMP_DIR)
```
bharadwaj-aditya, line 115: you will need to do this at startup too, to handle any residual files due to unclean shutdowns of previous runs

common/utils/utils.go
bharadwaj-aditya: if this is an integration test specific issue, we should change the data that the integration test runs on rather than changing the code here

sources/csv/data.go
```diff
@@ -68,37 +98,61 @@ func VerifyManifest(conv *internal.Conv, tables []Table) error {
...
+func preloadGCSFiles(tables []Table) ([]Table, error) {
+	for i, table := range tables {
+		for j, filePath := range table.File_patterns {
+			u, err := url.Parse(filePath)
 			if err != nil {
-				return fmt.Errorf("can't map to spanner type: %v. Please use the data types as in your spanner database", err)
+				return nil, fmt.Errorf("unable parse file path %s for table %s", filePath, table.Table_name)
+			}
+			if u.Scheme == "gs" {
+				bucketName := u.Host
```
bharadwaj-aditya, line 141: move all of this to utils ? into a method like DownloadFromGcsUrl or something ?

sources/csv/data.go
```diff
@@ -120,14 +174,25 @@ func SetRowStats(conv *internal.Conv, tables []Table, delimiter rune) {
...
 	count := int64(0)
+	srcCols, err := r.Read()
+	if err == io.EOF {
+		return count, nil
+	}
+	if err != nil {
+		return count, fmt.Errorf("can't read csv headers for col names due to: %v", err)
+	}
+	// If the row read was not a header, increase count.
+	if !utils.CheckEqualSets(srcCols, colNames) {
```
bharadwaj-aditya, line 193: this is going to be very slow. We should check this only for the first line and proceed without this check for other lines.
Lets check if there's a more efficient way to do this based on number of line breaks or so.

sources/csv/data.go
```diff
@@ -153,26 +228,38 @@ func ProcessCSV(conv *internal.Conv, tables []Table, nullStr string, delimiter r
...
 			if err == io.EOF {
 				conv.Unexpected(fmt.Sprintf("error processing table %s: file %s is empty.", table.Table_name, filePath))
 				continue
 			}
 			if err != nil {
-				return fmt.Errorf("can't read csv headers for col names due to: %v", err)
+				return fmt.Errorf("can't read row for %s due to: %v", filePath, err)
 			}
+			// If first row is some permutation of Spanner schema columns, we assume the first row is headers.
+			if utils.CheckEqualSets(srcCols, colNames) {
```
bharadwaj-aditya, line 242: don't perform this check for every line. It will be very expensive.

bharadwaj-aditya (changes requested):

common/utils/utils.go
Deep1998: Its an integration test specific issue because the emulator has a different infoschema than spanner i.e. the interleave_type column does not exist in the emulator.

sources/csv/data.go
```diff
@@ -68,37 +98,61 @@ func VerifyManifest(conv *internal.Conv, tables []Table) error {
...
-			ty, err := ToSpannerType(col.Type_name)
+func preloadGCSFiles(tables []Table) ([]Table, error) {
+	for i, table := range tables {
+		for j, filePath := range table.File_patterns {
+			u, err := url.Parse(filePath)
 			if err != nil {
-				return fmt.Errorf("can't map to spanner type: %v. Please use the data types as in your spanner database", err)
+				return nil, fmt.Errorf("unable parse file path %s for table %s", filePath, table.Table_name)
+			}
+			if u.Scheme == "gs" {
```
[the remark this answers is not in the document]
Deep1998: We maintain this []table struct that we create from the manifest that has table name and list of file paths. Some of these file paths can be gcs paths while some can be local.
```
[
{ table_name:"abc", file_patterns: ["/local/path", "gs://bucket/path"]},
{ table_name:"xyz", file_patterns: ["gs://bucket2/path2"]},
]
```
So we iterate these paths and if we find a gcs path, we download it to tmp and replace the gs path with the local path to the download file. So if file paths are not from gcs, we just skip them since they are already local paths.

sources/csv/data.go
Deep1998: Currently, thats what we are doing, only checking for the first line. We read just the first row, check if its a permutation of the table columns. 
If yes, we treat that as a header and move on to the remaining lines where we don't do this matching.
If no, we write off that row because that was a data row.

sources/csv/data.go
Deep1998: Explained in prev comment.

sources/csv/data.go
Deep1998: Makes sense, later on with avro, we will require this function across both.

bharadwaj-aditya (approved):
LGTM

PR #258, opened 2022-01-24
Csv readme

Deep1998:


sources/csv/README.md
```diff
@@ -0,0 +1,122 @@
...
+]
+```
+**CAVEATS:**
+- File patterns do not accept regular expressions. Provide the path inside 
+double quotes.
+
+### CSV File Format
+- Harbourbridge checks the first row and matches it with the spanner columns
+to check if the first row is a permutation of the Spanner table's column names.
+If a match is not found, then Harbourbridge assumes the order of data in csv
```
bharadwaj-aditya, line 70: do we have sanity checks to see the number of columns match ?

bharadwaj-aditya (commented):
Just one comment. I might missed it in the CSV code PR. Please check that bit

sources/csv/README.md
Deep1998: Good catch, I missed this check. Added

bharadwaj-aditya (approved):
LGTM

PR #260, opened 2022-01-25
Oracle database support in UI

sagarsearce:
This Pr adds oracle database support in the user interface.

1. updated web API
2. updated oracle/toddl.go to use central typemap for both web and cli.
3. frontend component updated.


sources/oracle/toddl.go
```diff
@@ -38,51 +38,91 @@ type ToDdlImpl struct {
...
+	// passing empty spType to execute default case.will get other spType from web pkg
+	ty, issues := ToSpannerTypeInternal(conv, "", columnType.Name, columnType.Mods)
 	if conv.TargetDb == constants.TargetExperimentalPostgres {
 		ty = overrideExperimentalType(columnType, ty)
 	}
 	return ty, issues
 }
 
-func toSpannerTypeInternal(conv *internal.Conv, id string, mods []int64) (ddl.Type, []internal.SchemaIssue) {
+func ToSpannerTypeInternal(conv *internal.Conv, spType string, srcType string, mods []int64) (ddl.Type, []internal.SchemaIssue) {
```
bharadwaj-aditya, line 49: this method is named internal, we should not expose it publicly

bharadwaj-aditya (changes requested):
Please also verify if the edit global type function is working fine.

sources/oracle/toddl.go
sagarsearce: We haven’t defined a different type map for oracle in the web package, we are using a mapping from toddl only as we discussed.
we need an extra argument(spanner datatype) to use exiting mapping but we can not update ToSpannerType function because it is part of the ToDdlImpl interface. So as a temporary solution we made this internal function public. Eventually will merge type maps for all the databases and update the interface function.
Should we change the function name or any better approach?

bharadwaj-aditya (changes requested):

bharadwaj-aditya (approved):
LGTM

PR #261, opened 2022-02-03
Csv handling for dialect mismatch

Deep1998:
This PR also handles some refactorings to implement this cleanly:

- Move GetResourceIds from common.go to a struct method of target profile. We do this to cache the project, instance and dbname values into target profile and reuse them.
- struct method in target profile called FetchDatabaseDialect() which fetches the dialect which we then compare with the specified dialect to see if there is a mismatch.
-  Passing that admin client from data.go subcommand to the csv is an inefficient way to handle it hencewe create a new admin client inside FetchDatabaseDialect() for now. We should refactor the code such that all clients are stored in target profile, or some new struct that can be passed throughout the code. Marked as todo.

### DIFF
--- a/cmd/data.go
+++ b/cmd/data.go
@@ -95,10 +95,14 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 	}
 
 	now := time.Now()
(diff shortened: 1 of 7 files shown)


bharadwaj-aditya (approved):
LGTM

PR #262, opened 2022-02-08
added spatial type support in mysql direct connect(web)

### LINKED ISSUE #192: UI: Global Data type mapping box not populated in UI when using direct connect
Screenshot 2021-10-07 at 12 36 50 PM
This box is not populated when mysql in direct connect has unsupported data types such as geometry.
However, using the corresponding dump file with unsupported types works as expected.

Sample schema to replicate this:
```

CREATE TABLE `productshapes` (
  `productId` int NOT NULL AUTO_INCREMENT,
  `productLine` varchar(50) NOT NULL ,
  `productShape` geometry,
  `attributes` JSON,
  PRIMARY KEY (`productId`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```

sagarsearce:
Fixes #192 

added spatial type in web data type map so will not get null in the frontend, hence global data type map will populate properly.

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -1157,7 +1157,7 @@ func addTypeToList(convertedType string, spType string, issues []internal.Schema
 }
 func init() {
 	// Initialize mysqlTypeMap.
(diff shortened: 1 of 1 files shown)


bharadwaj-aditya (approved):
LGTM

PR #265, opened 2022-03-04
chore: update temporary entries for cloud.google.com/go/spanner in go.mod to public versions supporting PG feature

rahul2393:

- [x] Tests pass

### What is the motivation behind this?

We need this because of a 'dependency hell' situation:

1. `grpc-go` makes breaking changes between minor releases of their module.
2. tidb v1.1.0-beta [depends on a `grpc/naming` API which was removed in v1.30](https://github.com/etcd-io/etcd/issues/12124). We depend on etcd v0.5.0-alpha indirectly via tidb v1.1.0-beta.

This situation would be solved by bumping tidb to version which depends on etcd v3.6.0+, which does not make use of the removed grpc-go API.

This module is used to solve this dependency hell, by adding the `google.golang.org/grpc/naming` package into version v1.44 of grpc-go.

It is the smallest patch that can be applied to solve this particular 'dependency hell' issue with grpc-go. It will not fix other issues when packages depend on other, newer packages.

## When can it be removed?

This must be removed when tidb does not need grpc/naming.


bharadwaj-aditya (approved):
LGTM

naming/dns_resolver.go
```diff
@@ -0,0 +1,293 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package naming
```
bharadwaj-aditya, line 19: Please add a comment linking where this snippet has been taken from

naming/naming.go
```diff
@@ -0,0 +1,68 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Package naming defines the naming API and related data structures for gRPC.
+//
+// This package is deprecated: please use package resolver instead.
```
bharadwaj-aditya, line 21: please add a comment as to where this snippet has been taken from

naming/dns_resolver.go
rahul2393: I added in naming/README https://github.com/cloudspannerecosystem/harbourbridge/pull/265/files#diff-e48c26b9bc60a5770dd43c73b76b599d9672681ffef4f90be8b20137c3593514R4

rahul2393: Do you think we need to add here too?

Deep1998:
LGTM, thanks for this fix. Please update the PR description as well though. You can simply copy paste the readme contents.

naming/dns_resolver.go
bharadwaj-aditya: Makes sense to add a comment here too, just to make it clear that this is going to be a copy of the source and not to add any custom code.

rahul2393: Done!! Github is not showing `Outdated` for some reasons.

naming/naming.go
rahul2393: Done!! Github is not showing `Outdated` for some reasons.

PR #270, opened 2022-03-24
removing third party license and sakila files as they are not being used

shreyakhajanchi:
Based on the suggestions in : https://launch.corp.google.com/launch/4138859/approver/1093
For getting OSS compliance moving third party license to a new folder


LICENSE
```diff
@@ -186,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2021 Google Inc.
+   Copyright [yyyy] [name of copyright owner]
```
bharadwaj-aditya, line 190: Is removing the year intentional ?

shreyakhajanchi: yes, the license needs to be kept exactly as is with Copyright [yyyy] [name of copyright owner]

bharadwaj-aditya (approved):
LGTM

PR #276, opened 2022-04-03
verifying spanner endpoint before execution to avoid the client api calls getting stuck

shreyakhajanchi:
In case of incorrect spanner api endpoint, the client APIs get stuck as in the background they have a timeout of 60 min. In order to inform the user about the probable incorrect endpoint added a warning to be printed at the start of the process.

### DIFF
--- a/main.go
+++ b/main.go
@@ -100,6 +100,14 @@ func main() {
 	}
 	defer utils.Close(lf)
 
+	endpoint := os.Getenv("SPANNER_API_ENDPOINT")
+	if endpoint != "" && endpoint != "staging-wrenchworks.sandbox.googleapis.com:443" {
+		fmt.Printf("\nWarning: Endpoint specified may be incorrect: \n" +
+			"For connecting to prod set SPANNER_API_ENDPOINT as blank \n" +
+			"For connecting to emulator set SPANNER_EMULATOR_HOST=localhost:9010 \n" +
+			"For connecting to staging set SPANNER_API_ENDPOINT=staging-wrenchworks.sandbox.googleapis.com:443\n\n")
+	}
+
 	// TODO: Remove this check and always run HB in subcommands mode once
 	// global command line mode is deprecated. We can also enable support for
 	// top-level flags in subcommand then.
--- a/main.go
+++ b/main.go
@@ -101,7 +101,8 @@ func main() {
 	defer utils.Close(lf)
 
 	endpoint := os.Getenv("SPANNER_API_ENDPOINT")
-	if endpoint != "" && endpoint != "staging-wrenchworks.sandbox.googleapis.com:443" {
+	emulator_host := os.Getenv("SPANNER_EMULATOR_HOST")
+	if !((endpoint == "" && (emulator_host == "" || emulator_host == "localhost:9010")) || endpoint == "staging-wrenchworks.sandbox.googleapis.com:443") {
 		fmt.Printf("\nWarning: Endpoint specified may be incorrect: \n" +
 			"For connecting to prod set SPANNER_API_ENDPOINT as blank \n" +
 			"For connecting to emulator set SPANNER_EMULATOR_HOST=localhost:9010 \n" +
--- a/main.go
+++ b/main.go
@@ -101,12 +101,12 @@ func main() {
 	defer utils.Close(lf)
 
 	endpoint := os.Getenv("SPANNER_API_ENDPOINT")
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -100,6 +100,15 @@ func main() {
 	}
 	defer utils.Close(lf)
 
+	endpoint := os.Getenv("SPANNER_API_ENDPOINT")
+	emulator_host := os.Getenv("SPANNER_EMULATOR_HOST")
+	if !((endpoint == "" && (emulator_host == "" || emulator_host == "localhost:9010")) || (endpoint == "staging-wrenchworks.sandbox.googleapis.com:443" && emulator_host == "")) {
```
bharadwaj-aditya, line 105: Please correct this naming error

main.go
```diff
@@ -100,6 +100,15 @@ func main() {
...
+		fmt.Printf("\nWarning: Endpoint specified may be incorrect: \n" +
+			"For connecting to prod set SPANNER_API_ENDPOINT as blank \n" +
+			"For connecting to emulator set SPANNER_EMULATOR_HOST=localhost:9010 \n" +
+			"For connecting to staging set SPANNER_API_ENDPOINT=staging-wrenchworks.sandbox.googleapis.com:443\n\n")
```
bharadwaj-aditya, line 109: It is a good idea to use constants for strings like this, so that if we change it in one place, it reflects everywhere.

bharadwaj-aditya (changes requested):

main.go
shreyakhajanchi: corrected.

main.go
shreyakhajanchi: added constants

main.go
```diff
@@ -100,6 +100,15 @@ func main() {
 	}
 	defer utils.Close(lf)
 
+	endpoint := os.Getenv("SPANNER_API_ENDPOINT")
+	emulatorHost := os.Getenv("SPANNER_EMULATOR_HOST")
+	if !((endpoint == "" && (emulatorHost == "" || emulatorHost == constants.SPANNER_EMULATOR_HOST)) || (endpoint == constants.STAGING_SPANNER_API_ENDPOINT && emulatorHost == "")) {
```
Deep1998, line 105: This if condition is difficult to read. Suggest moving this to a 
func checkSpannerEndpoints() bool {
   endpoint := os.Getenv("SPANNER_API_ENDPOINT")
   emulatorHost := os.Getenv("SPANNER_EMULATOR_HOST")
  // list down the conditions separately.
}

main.go
```diff
@@ -100,6 +100,15 @@ func main() {
...
+		fmt.Printf("\nWarning: Endpoint specified may be incorrect: \n" +
+			"For connecting to prod set SPANNER_API_ENDPOINT as blank \n" +
+			"For connecting to emulator set SPANNER_EMULATOR_HOST=" + constants.SPANNER_EMULATOR_HOST + "\n" +
+			"For connecting to staging set SPANNER_API_ENDPOINT=" + constants.STAGING_SPANNER_API_ENDPOINT + "\n\n")
```
Deep1998, line 109: Do we want to print this staging url to all the users?

main.go
```diff
@@ -100,6 +100,15 @@ func main() {
 	}
 	defer utils.Close(lf)
 
+	endpoint := os.Getenv("SPANNER_API_ENDPOINT")
+	emulatorHost := os.Getenv("SPANNER_EMULATOR_HOST")
+	if !((endpoint == "" && (emulatorHost == "" || emulatorHost == constants.SPANNER_EMULATOR_HOST)) || (endpoint == constants.STAGING_SPANNER_API_ENDPOINT && emulatorHost == "")) {
+		fmt.Printf("\nWarning: Endpoint specified may be incorrect: \n" +
+			"For connecting to prod set SPANNER_API_ENDPOINT as blank \n" +
```
Deep1998, line 107: Would it be clear to the user that we are talking about the env variable SPANNER_API_ENDPOINT?
Also suggest fixing the punctuation a bit.
```
fmt.Printf("\nWarning: Endpoint specified may be incorrect\n" +
			"For connecting to prod, run: export SPANNER_API_ENDPOINT=\"\" \n" +
			"For connecting to emulator, run: export SPANNER_EMULATOR_HOST=" + constants.SPANNER_EMULATOR_HOST + "\n" +
			"For connecting to staging, run: export SPANNER_API_ENDPOINT=" + constants.STAGING_SPANNER_API_ENDPOINT + "\n\n")
```

PR #279, opened 2022-04-14
Adding regex check for handling failure in parsing dump files with specific syntax

shreyakhajanchi:
This regex check filters out the system generated code of the format :
/*!50717 SELECT COUNT(*) INTO @rocksdb_has_p_s_session_variables FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'performance_schema' AND TABLE_NAME = 'session_variables' */;

Such code format fails in parsing, we have added this regex as a workaround and will remove once pingcap provides a permanent fix for this : https://github.com/pingcap/parser/issues/1370

### DIFF
--- a/sources/mysql/mysqldump.go
+++ b/sources/mysql/mysqldump.go
@@ -97,6 +97,7 @@ func readAndParseChunk(conv *internal.Conv, r *internal.Reader) ([]byte, []ast.S
 
 	// Regex for ignoring strings of the form /*!50717 SELECT COUNT(*) INTO @rocksdb_has_p_s_session_variables FROM INFORMATION_SCHEMA.TABLES */;
 	// These system generated SQL statements are currently not supported by parser and return error.
+	// Pingcap Issue : https://github.com/pingcap/parser/issues/1370
 	regexExp := regexp.MustCompile(`^(\/\*[!0-9\s]*SELECT[^\n]*INTO[\s]+@[^\n]*\*\/;\n)$`)
 	for {
 		b := r.ReadLine()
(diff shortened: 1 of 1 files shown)


sources/mysql/mysqldump.go
```diff
@@ -110,6 +114,10 @@ func readAndParseChunk(conv *internal.Conv, r *internal.Reader) ([]byte, []ast.S
 				n += copy(s[n:], l[i])
 			}
 			chunk := string(s)
+			matchStatus := regexExp.Match([]byte(chunk))
+			if matchStatus {
```
bharadwaj-aditya, line 118: log that this is being skipped

sources/mysql/mysqldump.go
```diff
@@ -94,6 +94,10 @@ func processMySQLDump(conv *internal.Conv, r *internal.Reader) error {
 // files containing tens or hundreds of GB of data.
 func readAndParseChunk(conv *internal.Conv, r *internal.Reader) ([]byte, []ast.StmtNode, error) {
 	var l [][]byte
+
+	// Regex for ignoring strings of the form /*!50717 SELECT COUNT(*) INTO @rocksdb_has_p_s_session_variables FROM INFORMATION_SCHEMA.TABLES */;
+	// These system generated SQL statements are currently not supported by parser and return error.
```
bharadwaj-aditya, line 99: add reference to the bug in the comment

shreyakhajanchi: added bug reference

sources/mysql/mysqldump.go
shreyakhajanchi: added log

bharadwaj-aditya (approved):
LGTM

sources/mysql/mysqldump.go
bharadwaj-aditya: this check will slow down the parsing, but let's make optimizing the parsing an independent activity

PR #281, opened 2022-04-18
Replaced to get with go install

Deep1998:
Lints failing since go get has been deprecated for [installing dependencies](https://go.dev/doc/go-get-install-deprecation), it is only recommended now to use it to update go.mog files. Replaced it with go install.

Also honnef.co/go/tools was incompatible with latest go, hence bumped it to the latest version.


bharadwaj-aditya (approved):
LGTM

PR #288, opened 2022-05-04
Upgrade Tidb dependency

manitgupta:
Upgrades the `tidb` dependency to remove dependency on the older `grpc/naming` module, and upgrade `etcd` and `protobuf` dependencies along the way.

* Removes the copied grpc/naming module
* Upgrades etcd and protobuf to latest version
* Fixes imports to use pingcap parser from within the tidb module

Fixes dependabot alerts:
1. https://github.com/cloudspannerecosystem/harbourbridge/security/dependabot/10
2. https://github.com/cloudspannerecosystem/harbourbridge/security/dependabot/12

- [x] Tests pass - `lint` and `integration test against emulator` is passing. 


.github/workflows/lint.yaml
```diff
@@ -30,4 +30,4 @@ jobs:
       - run: go vet .
       - run: go install golang.org/x/tools/cmd/goimports && test -z $(goimports -l .)
       - run: go install golang.org/x/lint/golint && golint .
-      - run: go install honnef.co/go/tools/cmd/staticcheck && staticcheck .
+      - run: go install honnef.co/go/tools/cmd/staticcheck@latest && staticcheck .
```
Deep1998, line 33: Looks like the latest version is 0.3. Do we know why this was needed since we already used 0.3 earlier?

manitgupta: While I am not a 100% sure, the most likely reason is conflicting library versions present in our tool graph. On running it without `@latest`, the action spews errors like so:

```
Run go install honnef.co/go/tools/cmd/staticcheck && staticcheck .
go: downloading honnef.co/go/tools v0.0.1-2020.1.4
panic: interface conversion: types.Type is nil, not *types.Signature

goroutine 8019 [running]:
honnef.co/go/tools/ir.(*builder).expr0(0xc005ed9910, 0xc0052b05a0, {0x94d4e8?, 0xc002039f20}, {0x0, {0x0, 0x0}, {0x0, 0x0}})
	/home/runner/go/pkg/mod/honnef.co/go/tools@v0.0.1-2020.1.4/ir/builder.go:[6](https://github.com/manitgupta/harbourbridge/runs/6285004877?check_suite_focus=true#step:8:6)40 +0x22db
honnef.co/go/tools/ir.(*builder).expr(0x[8](https://github.com/manitgupta/harbourbridge/runs/6285004877?check_suite_focus=true#step:8:8)26b40?, 0xc0052b05a0, {0x[9](https://github.com/manitgupta/harbourbridge/runs/6285004877?check_suite_focus=true#step:8:9)4d4e8?, 0xc002039f20?})
```
[Link](https://github.com/manitgupta/harbourbridge/runs/6285004877?check_suite_focus=true)

`panic: interface conversion: types.Type is nil, not *types.Signature`  looks to be a dependency mismatch error. 

On digging further, I looked at the repo documentation [here](https://staticcheck.io/docs/getting-started/#installation) and installed via the above method.

manitgupta: Actually, while responding to you @Deep1998, I also looked at the new version documentation [here](https://staticcheck.io/changes/2022.1/).

```
Not tied directly to this release, but worth mentioning regardless: Staticcheck has an [official GitHub Action]

(https://github.com/dominikh/staticcheck-action) now, which may simplify your CI pipeline.
```

This is the GitHub Action: https://github.com/dominikh/staticcheck-action

Maybe we can also look into a setting this action up directly in a separate PR?

Deep1998: Makes sense, we can take that as a follow up

Deep1998 (approved):
LGTM

PR #297, opened 2022-05-18
Fixed dbname casing across the codebase

Deep1998:
This fix also makes the parameter casing uniform for dbName. Earlier source profile referred to it as db_name while target referred to it as dbname. It is now referred to as dbName in both places.

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,9 +374,9 @@ such as `csv`, `avro` etc.
 HarbourBridge accepts the following options for --target-profile,
 specified as "key1=value1,key2=value,..." pairs:
 
-`dbname` Specifies the name of the Spanner database to create. This must be a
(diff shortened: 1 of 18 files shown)


manitgupta:
This looks like a backward-incompatible change (please correct me if I am wrong), and customers would need to change their commands to work after this PR. 

Since HB is a migration tool (which by nature shouldn't be used regularly by a single client) I don't think this should be an issue, but I think this change is a good place for us to think about [semantic versioning](https://semver.org/) in HB.

For example, this change would most likely result in a major version update (v1.X.X -> v2.0.0) since the API (CLI command) would become backward incompatible.

bharadwaj-aditya (approved):
LGTM

Issue #294, opened 2022-05-13
Improve information on conversion report


shreyakhajanchi:
Currently, conversion results for two table results with same number of warnings may be classified differently. For instance one maybe classified as POOR and the other as OK.
Cause: This happens because the number of columns in the tables differ.
Solution: Need to improve the information in conversion report to give user better idea on why did we mark the conversion as POOR/GOOD/OK.
This needs to be done for both schema and data conversion.

PR #310, opened 2022-05-27
improved conversion report

its7ARC:
Fixes [#294](https://github.com/cloudspannerecosystem/harbourbridge/issues/294)

- Added percentage of columns mapped cleanly in migration from the source database to Spanner in the conversion report.
- Added percentage of rows mapped in migration from the source database to Spanner in the conversion report.


sources/mysql/report_test.go
```diff
@@ -74,7 +74,7 @@ func TestReport(t *testing.T) {
 		`----------------------------
 Summary of Conversion
 ----------------------------
-Schema conversion: GOOD (most columns mapped cleanly, but some missing primary keys).
+Schema conversion: GOOD (99.99412% of 17006 columns mapped cleanly) + some missing primary keys.
```
Deep1998, line 77: You might want to limit the precision to 2 decimal places

its7ARC: will do.

its7ARC: done

Deep1998 (approved):
Looks like the initial precision setting was part of the logic, we should revert it to that. Approving anyways.

its7ARC:
> Looks like the initial precision setting was part of the logic, we should revert it to that. Approving anyways.

This will be reverted in the next pull request.

.github/workflows/test-coverage.yaml
```diff
@@ -0,0 +1,49 @@
+# Copyright 2022 Google LLC
```
Deep1998, line 1: what are these files?

its7ARC: These .github files haven't been added by me. These aren't actual changes, the current state of the Harbourbridge repository already has these files with the same content in them.

its7ARC:
Closing pull request. Will make new pr's, each adding a specific functionality.

PR #314, opened 2022-06-08
Improved Conversion report.

its7ARC:
Fixes [#294](https://github.com/cloudspannerecosystem/harbourbridge/issues/294)

- Added the following to the conversion report - 
  - Percentage of columns mapped cleanly in migration from the source database to Spanner.
  - Percentage of data rows mapped in migration from the source database to Spanner.

- Above additions have been done for
  - Overall database report
  - Table wise report

- This addition of percentages to the report will remove the ambiguity to infer why a conversion is being classified as _Excellent_, _Good_ or _Bad_.

- The functions for classification of conversions as _good_ and _ok_ have been modified to incorporate floating point precision in order to be more accurate.
  - eg. For the _good_ function in _report.go_: Let _badcount_ = 1 and _total_ = 21.
    - Current function would floor 21/20 as 1 and the output for 1 < 1 would be false => Classification as not good.
    - Function in this commit would compare 1 < 21/20 or 1 < 1.05 = True => Classification as good.
          
- The expected variables in the test files of the report have also been modified to be in sync with the changes in the report file.

### DIFF
--- a/internal/report.go
+++ b/internal/report.go
@@ -325,6 +325,7 @@ func AnalyzeCols(conv *Conv, srcTable, spTable string) (map[string][]SchemaIssue
 // summary rating.
 func rateSchema(cols, warnings int64, missingPKey, summary bool) string {
 	pkMsg := "missing primary key"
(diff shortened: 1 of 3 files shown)


Deep1998 (approved):
LGTM

Issue #309, opened 2022-05-26
Report of the modifications made by HabourBridge to the schema


bharadwaj-aditya:
Feature Request: 
Update the Schema report with the changes that HarbourBridge performed during schema migration.

This can be broken down into two problems

1. Explicit changes by User
2. Implicit changes made by HarbourBridge (When the names have invalid characters, non-unique names, syth_id's etc)

Deep1998:
Report when Harbourbridge implicity fix names/ids. Fixing is fine but would be good if it will log those in the report. I noticed it implicitly fixes names/ids if they:
contains invalid characters
starts with bad character
not unique within a schema (Harbourbridge append numbers at the end)

PR #316, opened 2022-06-08
Addition of implicit name changes to the report

its7ARC:
Fixes [#309 ](https://github.com/cloudspannerecosystem/harbourbridge/issues/309)

This PR ensures  
- The conversion report notifies the following implicit changes that occur in migration from the source database to Spanner.
  - Table name changes
  - Column name changes
  - Foreign key name changes
  - Index name changes
  - Selection of a column with unique constraint as a primary key in case the table doesn't has one in the source database.
  - Addition of a synthetic primary key in case a table does not has a primary key or a column with unique constraint in the source database.
 
- Tabular report of all the name changes that occur in migration from the source database to Spanner.
  - Format : | Source Table | Change | Old Name | New Name |

- Central storage of the name mappings between the source database and Spanner for -
  - Table names
  - Column names
  - Foreign key names
  - Index names
 
- The _expected_ variables in the test files have been modified to be in sync with the changes to the report.


internal/report.go
```diff
@@ -192,6 +209,9 @@ func buildTableReportBody(conv *Conv, srcTable string, issues map[string][]Schem
 				}
 				srcType := srcSchema.ColDefs[srcCol].Type.Print()
 				spType := spSchema.ColDefs[spCol].T.PrintColumnDefType()
+				srcName := fmt.Sprintf(srcSchema.ColDefs[srcCol].Name)
```
Deep1998, line 212: sprintf is used for [formatting strings](https://www.geeksforgeeks.org/fmt-sprintf-function-in-golang-with-examples/). For assigning a string variable like this, we dont need an Sprintf.

internal/report.go
```diff
@@ -434,6 +458,38 @@ func IgnoredStatements(conv *Conv) (l []string) {
 	return l
 }
 
+func reportNameChanges(conv *Conv, w *bufio.Writer) {
+
+	w.WriteString("-----------------------------------------------------------------------------------------------------\n")
+	w.WriteString("Name Changes in Migration\n")
+	w.WriteString("-----------------------------------------------------------------------------------------------------\n")
+	fmt.Fprintf(w, "%25s %15s %25s %25s\n", "Source Table", "Change", "Old Name", "New Name")
```
Deep1998, line 466: nit/Suggestion:you can also do 
`w.Writestring(fmt.Sprintf("%25s %15s %25s %25s\n", "Source Table", "Change", "Old Name", "New Name"))`
as it is consistent with the rest.

internal/report.go
```diff
@@ -172,6 +175,20 @@ func buildTableReportBody(conv *Conv, srcTable string, issues map[string][]Schem
 				l = append(l, fmt.Sprintf("UNIQUE constraint on column(s) '%s' replaced with primary key since this table didn't have one. Spanner requires a primary key for every table", strings.Join(uniquePK, ", ")))
 			}
 		}
+
+		if p.severity == note {
+			for srcKeyName, spKeyName := range conv.ToSpanner[srcTable].ForeignKey {
+				if srcKeyName != spKeyName {
+					l = append(l, fmt.Sprintf("%s, Foreign Key '%s' is mapped to '%s'", IssueDB[NameConvention].Brief, srcKeyName, spKeyName))
```
Deep1998, line 182: @bharadwaj-aditya wondering if we should report it at the beginning as well as in the table wise summary? It might make the report very noisy.

internal/convert.go
```diff
@@ -85,13 +85,16 @@ const (
 	Widened
 	Time
 	StringOverflow
+	NameConvention
```
Deep1998, line 88: Nit: should we try for a more informative name for this issue? How about IllegalName?

its7ARC: will do.

internal/report.go
its7ARC: Are there any benefits of using w.WriteString?

internal/report.go
its7ARC: got it.

internal/report.go
its7ARC: will do this.

internal/convert.go
```diff
@@ -85,13 +85,16 @@ const (
 	Widened
 	Time
 	StringOverflow
+	IllegalName
 )
 
 // NameAndCols contains the name of a table and its columns.
 // Used to map between source DB and Spanner table and column names.
 type NameAndCols struct {
```
Deep1998, line 93: This works but I am leaning towards avoiding channging the strucutre of the NameAndCols struct mainly because it will change the session file that will get exported and hence can cause backwards compatibility issues. An easy workaround is to add another field in conv similar to names and conv but for fks and indexes.

its7ARC: ohkay, will do.

internal/report.go
Deep1998: Lets keep it in both places

internal/convert.go
```diff
@@ -94,6 +97,11 @@ type NameAndCols struct {
 	Cols map[string]string
 }
 
+type FkeyAndIdxs struct {
```
Deep1998, line 100: Brief comment describing the struct, like how we do in nameandcols.

internal/convert.go
```diff
@@ -31,7 +31,9 @@ type Conv struct {
 	SrcSchema                map[string]schema.Table             // Maps source-DB table name to schema information.
 	Issues                   map[string]map[string][]SchemaIssue // Maps source-DB table/col to list of schema conversion issues.
 	ToSpanner                map[string]NameAndCols              // Maps from source-DB table name to Spanner name and column mapping.
+	ToSpannerFkIdx           map[string]FkeyAndIdxs              // Maps from source-DB table name to Spanner table name, foreign keys and index mappings.
```
Deep1998, line 34: consider adding `json:"-"` struct tags so that we dont export these to the session file.

its7ARC: _ToSpanner_ and _ToSource_ are being used in the code base functionally, putting them into a different _"audit_" object will lead to a number of cascading changes throughout the codebase. Adding only _ToSpannerFkIdx_, _ToSourceFkIdx_ to _audit_ will seperate the central storage of column name mapping from the foreign key and index mappings. This will make the mappings slightly more difficult to track.

To avoid peppering them through the conv, it is perhaps better to put them along with the other variables that don't affect the functionality in the _conv_ struct only. eg -along with  _SchemaConversionDuration_, _DataConversionDuration_.

internal/convert.go
its7ARC: Done.

internal/report.go
```diff
@@ -434,6 +458,38 @@ func IgnoredStatements(conv *Conv) (l []string) {
...
+	for srcTableName, spTable := range conv.ToSpanner {
+		if srcTableName != spTable.Name {
+			fmt.Fprintf(w, "%25s %15s %25s %25s\n", srcTableName, "Table Name", srcTableName, spTable.Name)
+		}
+		for srcColName, spColName := range spTable.Cols {
+			if srcColName != spColName {
+				fmt.Fprintf(w, "%25s %15s %25s %25s\n", srcTableName, "Column Name", srcColName, spColName)
+			}
+		}
+		for srcFkName, spFkName := range conv.ToSpannerFkIdx[srcTableName].ForeignKey {
```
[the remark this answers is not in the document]
Deep1998: So as part of the bug, we only plan to cover the implicit changes in this. HarbourBridge does not do these ordering changes via the CLI so we dont need to cover that yet. We can take it up during part 2 of this bug when we make the changes for UI logging.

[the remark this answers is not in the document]
its7ARC: Screenshot 2022-06-13 12 36 55 PM
Screenshot 2022-06-13 12 35 21 PM

In case the foreign key and indexes cannot be mapped to Spanner from the source DB, their report is already present in the codebase.

Deep1998 (approved):
LGTM, thanks for the changes!

PR #320, opened 2022-06-23
Static file embedding for web

shellymathew98:
- Static files are embedded for web.
- Go embed is supported only by go versions 1.16 and above, thus edited the version in go.mod here.

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,6 +19,7 @@ package main
 
 import (
 	"context"
+	"embed"
 	"flag"
 	"fmt"
 	"os"
@@ -38,6 +39,9 @@ import (
 	"github.com/cloudspannerecosystem/harbourbridge/webv2"
 	"github.com/google/subcommands"
 )
+
+//go:embed frontend/*
+var forntendDir embed.FS
 
 var (
 	dbNameOverride   string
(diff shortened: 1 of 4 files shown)


web/routes.go
```diff
@@ -15,14 +15,19 @@
 package web
```
manitgupta, line 15: Can we also fix the `webv2` imports as part of this PR?

manitgupta (commented):
Can you also add details around how you validated the change? 

An ideal test would be like so:
1. Build the binary
2. Copy it over to some unknown location
3. run the binary
4. the webUI should work as expected.

main.go
```diff
@@ -39,6 +40,9 @@ import (
 	"github.com/google/subcommands"
 )
 
+//go:embed frontend/*
+var forntendDir embed.FS
```
manitgupta, line 44: nit: spelling of variable `frontendDir`

shellymathew98:
> Can you also add details around how you validated the change?
> 
> An ideal test would be like so:
> 
> 1. Build the binary
> 2. Copy it over to some unknown location
> 3. run the binary
> 4. the webUI should work as expected.

- I tried placing the binary on a different location and webUI works fine from there.

web/routes.go
manitgupta: Sure.

manitgupta (approved):
LGTM

PR #321, opened 2022-06-23
Add THIRD_PARTY_NOTICES and vendor source

manitgupta:
Adds the third party notices of the libraries used in HarbourBridge.
Adds the source of the `go-sql-driver` library to `vendor/` directory since its distributed under the MPL 2.0 license.


manitgupta:
Third Party Notices and the licensed source should not be included in the source repo.

PR #322, opened 2022-06-23
Changes to HarbourBridge for gCloud Integration

manitgupta:
Adds the following:

1. Updates the password prompt workflow so that user is asked to provide password as part of CLI input or the environment variables when HB is being used from the gCloud SDK.
2. Integrates `zap` logging library.
3. Adds two new flags: `--dry-run` and `--log-level`

### Log level flag
`--log-level` flag is added for `schema`, `data` and `schema-and-data` commands. 
Currently, this flag only duplicates the logging capability of `--verbose` flag that exists in the global command line mode, which was missing the subcommand mode earlier. This is done by using the `zap` structured logging library.

The library logs to both the console and a `harbourbridge.log` file. This provides user a reproducible log of the harbourbridge run.

### Dry-run flag
This PR only adds, but does not implement the `--dry-run` flag. This will be implemented in the subsequent PRs. The idea behind adding this flag as part of this PR is to enable full CLI compatibility with the `gcloud spanner migration` surface.


conversion/conversion.go
```diff
@@ -390,6 +391,7 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
 		return f, n, err
 	}
 	internal.VerbosePrintln("Creating a tmp file with a copy of stdin because stdin is not seekable.")
```
Deep1998, line 393: Should we delete this line?

manitgupta: Yeah, we need to keep these for compatibility with the `--verbose` mode that is present in the global command line mode. Once global mode is deprecated, we can remove these.

logger/logger.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+const LOG_FILE_NAME = "harbourbridge.log"
+
+var Log *zap.Logger
+
+func InitializeLogger(inputLogLevel string) error {
```
Deep1998, line 14: Should we add some case sensitivity handling over here?

logger/logger.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return err
+	}
+	logLevel := zap.NewAtomicLevelAt(*zapLogLevel)
+	core := zapcore.NewTee(
+		zapcore.NewCore(fileEncoder, writer, logLevel),
+		zapcore.NewCore(consoleEncoder, zapcore.AddSync(os.Stdout), logLevel),
+	)
+	Log = zap.New(core, zap.AddCaller(), zap.AddStacktrace(zapcore.ErrorLevel))
+	return nil
```
Deep1998, line 32: This function seems to be using a bunch of zap library functions which do not feel very readable naturally to me. Maybe consider adding some comments that explain what they are doing?

Deep1998 (approved):
LGTM overall, I am still unclear why we added 
```
func init() {
	logger.Log = zap.NewNop()
}
```
in all the test files.

logger/logger.go
manitgupta: The `zap` library takes care of case sensitivity implicitly. For example, `INFO` and `info` are both translated to the `INFO` log level.

manitgupta:
Since HarbourBridge commands run in a script-like fashion, the logger is initialised during execution flow of the CLI commands. This means that the tests need explicit initialisation of the logger in order to run (and not fail trying to find the logger object). I agree that it's not the cleanest solution, but its best option I could think of.

logger/logger.go
manitgupta: Makes sense, adding some comments for readability.

PR #323, opened 2022-06-23
Adding interface support for starting streaming migrations

nareshz:
Things added/changed
1. Common functions for all databases in infoschema interface for starting streaming migration
   * StartChangeDataCapture -> This can be implemented to start capturing live changes on the database before bulk migration.
   * StartStreamingMigration -> This can be implemented to start the processing of captured changes after finishing bulk migration
2. Changes to InfoSchemaImpl object for storing information required for starting streaming migrations.
3. Moved code for automatic triggering of Datastream and Dataflow jobs[#286] to the functions above described for MySQL, Oracle.

- [x] Tests pass


streaming/streaming.go
```diff
@@ -219,12 +220,11 @@ func LaunchStream(ctx context.Context, sourceProfile profiles.SourceProfile, pro
 	if err != nil {
 		return fmt.Errorf("could not create update request: %v", err)
 	}
-
 	_, err = upOp.Wait(ctx)
 	if err != nil {
 		return fmt.Errorf("update stream operation failed: %v", err)
 	}
-	fmt.Println("Stream state set to RUNNING...")
+	fmt.Println("Done")
```
shreyakhajanchi, line 227: Printing only Done might not be indicative of what was done. Print a more verbose msg something like - Stream state set to running

sources/mysql/infoschema.go
```diff
@@ -15,23 +15,29 @@
...
 	"strings"
 
+	sp "cloud.google.com/go/spanner"
 	"github.com/cloudspannerecosystem/harbourbridge/internal"
+	"github.com/cloudspannerecosystem/harbourbridge/profiles"
 	"github.com/cloudspannerecosystem/harbourbridge/schema"
 	"github.com/cloudspannerecosystem/harbourbridge/sources/common"
 	"github.com/cloudspannerecosystem/harbourbridge/spanner/ddl"
+	"github.com/cloudspannerecosystem/harbourbridge/streaming"
 	_ "github.com/go-sql-driver/mysql" // The driver should be used via the database/sql package.
```
shreyakhajanchi, line 31: As a good practice better to split these imports into 3 groups - inbuilt, external and local. Same for other files.

sources/dynamodb/schema.go
```diff
@@ -188,6 +191,14 @@ func (isi InfoSchemaImpl) ProcessData(conv *internal.Conv, srcTable string, srcS
 	return nil
 }
 
+func (isi InfoSchemaImpl) StartChangeDataCapture(ctx context.Context, conv *internal.Conv) (map[string]interface{}, error) {
```
shreyakhajanchi, line 194: We will need to modify the source profile conn for dynamodb in case of streaming will that be a part of different PR ?

sources/oracle/infoschema.go
```diff
@@ -394,6 +400,31 @@ func (isi InfoSchemaImpl) GetIndexes(conv *internal.Conv, table common.SchemaAnd
...
+	return mp, err
+}
+
+// StartStreamingMigration is used for automatic triggering of Dataflow job when
+// performing a streaming migration.
+func (isi InfoSchemaImpl) StartStreamingMigration(ctx context.Context, client *sp.Client, conv *internal.Conv, streamingInfo map[string]interface{}) error {
+	streamingCfg, _ := streamingInfo["streamingCfg"].(streaming.StreamingCfg)
+	err := streaming.StartDataflow(ctx, isi.SourceProfile, isi.TargetProfile, streamingCfg)
+	if err != nil {
+		err = fmt.Errorf("error starting dataflow: %v", err)
```
shreyakhajanchi, line 422: These wrapper errors are almost same as msg already in err : error launching dataflow. We can either modify the messaging or directly return err

shreyakhajanchi (commented):
Added few minor comments. Please ensure that you test MySQL and Oracle flows in the current branch to avoid any issue due to code movement

streaming/streaming.go
```diff
@@ -287,3 +287,24 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
 		" will have to be manually cleaned up via he UI. HarbourBridge will not delete them post completion of the migration.")
 	return nil
 }
+
+func StartDatastream(ctx context.Context, sourceProfile profiles.SourceProfile, targetProfile profiles.TargetProfile) (StreamingCfg, error) {
+	streamingCfg, err := ReadStreamingConfig(sourceProfile.Conn.Mysql.StreamingConfig, targetProfile.Conn.Sp.Dbname)
```
Deep1998, line 292: Consider adding a funciton to multiplex the streaming config, shouldn't be too much effort, but can potentially finish both flows for oracle and mysql? We can return an error for the other type of sources

Deep1998 (commented):
LGTM

streaming/streaming.go
nareshz: The code before had this flow
> <functionToSetStateToRunning() called here>Stream state set to RUNNING

Deep suggested that we should change it because changing stream status take some time
> Setting stream state to RUNNING...<functionToSetStateToRunning() called here>Done

There is one more line added before calling functionToSetStateToRunning() "Setting stream state to RUNNING...". Done is printed as a confirmation to the state set to Runnning.

sources/dynamodb/schema.go
nareshz: As we are planning to introduce a `streaming` flag for starting streaming migration. We can parse this flag in the main cmd package and change the existing SourceProfile.Conn.Streaming flag to start streaming migration for Dynamodb

streaming/streaming.go
nareshz: We don't need to return an error for other types of sources as this function is called only for oracle and mysql. Implementation for other databases is empty. 
I will multiplex the streaming config here based on the source profile connection. If connection is mysql then StreamingConfig for the same will be used and vice versa for oracle.

sources/oracle/infoschema.go
nareshz: yes you are right "error starting dataflow" is repeated two times. We can remove this. I will make the changes.

nareshz:
> Added few minor comments. Please ensure that you test MySQL and Oracle flows in the current branch to avoid any issue due to code movement

shreyakhajanchi (approved):
LGTM

PR #325, opened 2022-06-24
Changes to handle incorrect Spanner API endpoints

### LINKED ISSUE #324: Migration gets stuck in case of incorrect spanner endpoint
In case of an incorrect or unsupported spanner API endpoint the migration seems to be stuck as there is no output in the CLI. The client libraries actually have a default timeout of 60 min therefore the request is retried until the time limit is reached. The user is unaware of any update in the background. We need to add some timeout to these requests to avoid this scenario.

Steps to reproduce:
$gcloud auth application-default login
$export GCLOUD_PROJECT=span-cloud-testing
$export SPANNER_API_ENDPOINT=https://staging-wrenchworks.sandbox.googleapis.com/
$cd harbourbridge
$harbourbridge schema-and-data -source=pg -target-profile="instance=test-instance,dbname=test-db" < ~/singers.pg_dump

source-profile format defaulting to `dump`
Generating schema: 100%
Wrote schema to file 'pg_dump_2022-02-24_f127-4304.schema.txt'.
Wrote legal schema ddl to file 'pg_dump_2022-02-24_f127-4304.schema.ddl.txt'.
Wrote session to file 'pg_dump_2022-02-24_f127-4304.session.json'.
Processed 3203 bytes of pg_dump data (17 statements, 21 rows of data, 0 errors, 0 unexpected conditions).
Schema conversion: EXCELLENT (all columns mapped cleanly).
See file 'pg_dump_2022-02-24_f127-4304.report.txt' for details of the schema and data conversions.
Using Google Cloud project: span-cloud-testing
Using Cloud Spanner instance: test-instance

WARNING: Please check that permissions for this Spanner instance are
appropriate. Spanner manages access control at the database level, and the
database created by HarbourBridge will inherit default permissions from this
instance. All data written to Spanner will be visible to anyone who can
access the created database. Note that pg_dump table-level and row-level
ACLs are dropped during conversion since they are not supported by Spanner.

>>>> HANG <<<<

If the endpoint used is: staging-wrenchworks.sandbox.googleapis.com:443
The above issue doesn't appear.

nareshz:
Fixes #324 

**Issue**: The default timeout limit for client libraries API calls is 60 minutes. Client libraries retries the GetDatabase API call If **Unavailable**/**DeadlineExceeded** error is faced ([More Details](https://github.com/cloudspannerecosystem/harbourbridge/issues/324#issuecomment-1165639181)). When an incorrect Spanner API endpoint is used **Unavailable** error code is returned and this is the reason it is retried for 60 minutes.
>rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp: lookup tcp///staging-wrenchworks.sandbox.googleapis.com/: Servname not supported for ai_socktype" 

**Things changed**: If API call doesn't respond to the request then at an interval of 5 minutes the user is informed through the command line about the problem.

- [x] Tests pass

### DIFF
--- a/conversion/conversion.go
+++ b/conversion/conversion.go
@@ -392,8 +392,13 @@ func getSeekable(f *os.File) (*os.File, int64, error) {
 }
 
 // VerifyDb checks whether the db exists and if it does, verifies if the schema is what we currently support.
+// Timeout limit for request is set to 2 minutes to handle unsupported spanner API endpoints.
 func VerifyDb(ctx context.Context, adminClient *database.DatabaseAdminClient, dbURI string) (dbExists bool, err error) {
-	dbExists, err = CheckExistingDb(ctx, adminClient, dbURI)
+	requestTimeout := 2 * time.Minute
+	reqCtx, cancel := context.WithTimeout(ctx, requestTimeout)
+	defer cancel()
+
+	dbExists, err = CheckExistingDb(reqCtx, adminClient, dbURI)
 	if err != nil {
 		return dbExists, err
 	}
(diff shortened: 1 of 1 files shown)


conversion/conversion.go
```diff
@@ -403,10 +408,14 @@ func VerifyDb(ctx context.Context, adminClient *database.DatabaseAdminClient, db
 	return dbExists, err
 }
 
-// CheckExistingDb checks whether the database with dbURI exists or not.
+// CheckExistingDb checks whether the database with dbURI exists or not. If request is not completed within
+// context deadline then it will return an context error.
 func CheckExistingDb(ctx context.Context, adminClient *database.DatabaseAdminClient, dbURI string) (bool, error) {
 	_, err := adminClient.GetDatabase(ctx, &adminpb.GetDatabaseRequest{Name: dbURI})
 	if err != nil {
+		if err == ctx.Err() {
```
shreyakhajanchi, line 416: better to check for some specific part of error msg rather than comparing the entire err

nareshz: The error which is returned is ctx.Err() -> context deadline exceeded. This comparison won't take much time. 
If we don't compare the full error then doesn't it compromise the correctness?

shreyakhajanchi: in that case we should compare ctx.err with context deadline exceeded not err

nareshz: The value of ctx.err is 'context deadline exceeded' if context expires before API call is completed. So we should compare the error returned with ctx.err, right?

PR #326, opened 2022-06-29
Output files of HarbourBridge in the 'harbour_bridge_output' directory.

its7ARC:
In the CLI mode, the prefix flag is used to explicitly set the paths for the output files generated by HarbourBridge.  eg - `--prefix= ~/abc/name.` will put the output files in the `~/abc` directory with a name `name.report.txt`, `name.schema.txt`, etc.

But if the prefix flag is not set, HarbourBridge puts the output files at the base level of the HarbourBridge directory. ie - where the main.go file is stored. A number of scattered output files therefore cause clutter.

This PR ensures - 
- The output files are put in a seperate `harbour_bridge_output` directory if the prefix flag is not explicitly set.
- The path of these files have been kept in the format of the Web UI output files - 
  - `harbour_bridge_output/dbName/dbName.report.txt`
  - `harbour_bridge_output/dbName/dbName.schema.txt`
  - `harbour_bridge_output/dbName/dbName.schema.ddl.txt`
  - `harbour_bridge_output/dbName/dbName.session.json`
- Putting these files in a seperate directory will also be useful in mounting this directory to the host machine for HarbourBridge in Docker mode.


Deep1998 (approved):

cmd/data.go
```diff
@@ -125,7 +125,12 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 
 	// If filePrefix not explicitly set, use dbName as prefix.
 	if cmd.filePrefix == "" {
-		cmd.filePrefix = dbName + "."
+		dirPath := "harbour_bridge_output/" + dbName + "/"
```
manitgupta, line 128: Please convert this to a formatted string. Applicable at other places as well.

cmd/data.go
```diff
@@ -125,7 +125,12 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
...
+		err = os.MkdirAll(dirPath, os.ModePerm)
```
manitgupta, line 129: Should we check if the directory with the same name already exists before making it?

cmd/data.go
```diff
@@ -125,7 +125,12 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
...
+		if err != nil {
+			fmt.Fprintf(ioHelper.Out, "Can't create directory %s: %v\n", dirPath, err)
+		}
+		cmd.filePrefix = dirPath + dbName + "."
```
manitgupta, line 133: Same as above, please use formatted strings for readability.

cmd/data.go
Deep1998: AFAIK, this function is idempotent. But best to verify.

manitgupta:
Can you also add how does the user specify the `--prefix` flag now? Seems like the `harbour_bridge_output` is only created when `--prefix` is not specified, and when it is, the generated files would still go the root directory?

> Putting these files in a seperate directory will also be useful in mounting this directory to the host machine for HarbourBridge in Docker mode.

For this objective, shouldn't the news files be created inside a directory irrespective of the `--prefix` flag?

cmd/data.go
its7ARC: Done.

cmd/data.go
its7ARC: I verified this, it is idempotent.

cmd/data.go
its7ARC: Done.

cmd/data.go
its7ARC: Done.

cmd/data.go
its7ARC: Done.

cmd/data.go
its7ARC: > Can you also add how does the user specify the `--prefix` flag now? 

Have added this in the PR description. 

> Seems like the `harbour_bridge_output` is only created when `--prefix` is not specified, and when it is, the generated files would still go the root directory?

The `prefix` flag is used to set file paths explicitly. If it is used then we don't have to put the files in the `harbour_bridge_output` directory.

>> Putting these files in a seperate directory will also be useful in mounting this directory to the host machine for HarbourBridge in Docker mode.
>
> For this objective, shouldn't the news files be created inside a directory irrespective of the `--prefix` flag?

Yes they should be. But we don't have a way to differentiate if the migration is happening in the normal mode or the Docker mode. We can document in the usage guide that the prefix flag is not relevant in the Docker Mode.

its7ARC:
> Can you also add how does the user specify the `--prefix` flag now? 

Have added this in the PR description. 

> Seems like the `harbour_bridge_output` is only created when `--prefix` is not specified, and when it is, the generated files would still go the root directory?

The `prefix` flag is used to set the file paths explicitly. If it is used then we don't have to put the files in the `harbour_bridge_output` directory.

>> Putting these files in a seperate directory will also be useful in mounting this directory to the host machine for HarbourBridge in Docker mode.
>
> For this objective, shouldn't the news files be created inside a directory irrespective of the `--prefix` flag?

Yes they should be. But we don't have a way to differentiate if the migration is happening in the normal mode or the Docker mode. We can document in the usage guide that the prefix flag is not relevant in the Docker Mode.

manitgupta:
- `prefix` flag is irrelevant for the user when running harbourbridge as a docker since the directory created inside the docker is irrelevant to the user. The mounted directory is more important.

its7ARC:
I am closing this PR. Will make this change on a separate branch along with the other Docker PR's after the final decision on the functionality of the `prefix` flag in the normal mode and the GCloud CLI mode is decided.

PR #329, opened 2022-07-03
Modifying modes of migration

shreyakhajanchi:
1. Added support for dry run flag
2. Modified modes to support schema only and data only migration
3. Modified respective unit/integration tests

Design doc : [here](https://docs.google.com/document/d/1CK7zD9vBKdpRwti0buZt2d7zdiOiZUfJt9E9lIEfNFg/edit?resourcekey=0-EjcRV0EwCb7e-4JzHa80Kg#heading=h.8w7xkmdacnem)

### DIFF
--- a/cmd/data.go
+++ b/cmd/data.go
@@ -29,7 +29,6 @@ import (
 	"github.com/cloudspannerecosystem/harbourbridge/conversion"
 	"github.com/cloudspannerecosystem/harbourbridge/internal"
 	"github.com/cloudspannerecosystem/harbourbridge/logger"
-	"github.com/cloudspannerecosystem/harbourbridge/profiles"
 	"github.com/cloudspannerecosystem/harbourbridge/proto/migration"
 	"github.com/cloudspannerecosystem/harbourbridge/spanner/writer"
 	"github.com/google/subcommands"
@@ -43,15 +42,12 @@ type DataCmd struct {
 	sourceProfile   string
 	target          string
 	targetProfile   string
-	skipForeignKeys bool
 	sessionJSON     string
 	filePrefix      string // TODO: move filePrefix to global flags
 	writeLimit      int64
 	dryRun          bool
-<<<<<<< HEAD
 	logLevel        string
-=======
->>>>>>> 2778961 (modification wrt report)
+	skipForeignKeys bool
 }
 
 // Name returns the name of operation.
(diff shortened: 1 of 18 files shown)


common/utils/utils.go
```diff
@@ -500,6 +500,63 @@ func ReadSpannerSchema(ctx context.Context, conv *internal.Conv, client *sp.Clie
 	return nil
 }
 
+// CompareSchema compares the spanner schema of two conv objects and returns specific error if they don't match
+func CompareSchema(conv1, conv2 *internal.Conv) error {
```
Deep1998, line 504: looks like this function does not handle cases when the lists/map are differently ordered.

Deep1998: Also what about cases when the session file contents are a subset of the spanner schema contents? We might want to handle that case, where the user added some extra tables but used HB to only load a partial set. It shouldn't be too much of a change on our side but will improve usability.

cmd/data.go
```diff
@@ -112,77 +114,110 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
...
+				fmt.Printf("can't verify target database: %v\n", err)
+				return subcommands.ExitFailure
+			}
+			if !dbExists {
+				fmt.Println("target database doesn't exist")
+				return subcommands.ExitFailure
+			}
+			err = conversion.ValidateTables(ctx, client, conv.TargetDb)
+			if err != nil {
+				fmt.Printf("can't validate the tables: %v\n", err)
```
Deep1998, line 181: nit: If we return an error saying "table xyz has non-zero rows", the whole message would read: 
`can't validate the tables: table xyz has non-zero rows`.
Consider rephrasing this error to: `error validating tables`

cmd/data.go
```diff
@@ -112,77 +114,110 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
...
+		}
 
-	if !cmd.skipForeignKeys {
 		if err = conversion.UpdateDDLForeignKeys(ctx, adminClient, dbURI, conv, ioHelper.Out); err != nil {
-			err = fmt.Errorf("can't perform update schema on db %s with foreign keys: %v", dbURI, err)
+			fmt.Printf("can't perform update schema on db %s with foreign keys: %v\n", dbURI, err)
+			return subcommands.ExitFailure
+		}
+		banner = utils.GetBanner(now, dbURI)
+	} else {
```
Deep1998, line 208: Refactoring: instead of formatting it like
 ```
if dryRun {
  do abc
} else {
  do xyz
}
```
it will be much more readable with something like

```
if dryRun
   funcToDoabc()
else 
   funcToDoxyz()
```
And I think we should be able to reuse funcToDoabc90 and xyz() in all 3 cmd files

**Review by manitgupta — CHANGES_REQUESTED** [review]

File: cmd/data.go [hunk 4]
```diff
@@ -72,9 +74,9 @@ func (cmd *DataCmd) SetFlags(f *flag.FlagSet) {
 	f.StringVar(&cmd.sessionJSON, "session", "", "Specifies the file we restore session state from")
 	f.StringVar(&cmd.target, "target", "Spanner", "Specifies the target DB, defaults to Spanner (accepted values: `Spanner`)")
 	f.StringVar(&cmd.targetProfile, "target-profile", "", "Flag for specifying connection profile for target database e.g., \"dialect=postgresql\"")
-	f.BoolVar(&cmd.skipForeignKeys, "skip-foreign-keys", false, "Skip creating foreign keys after data migration is complete (ddl statements for foreign keys can still be found in the downloaded schema.ddl.txt file and the same can be applied separately)")
```
manitgupta, line 75: `--skip-foreign-keys` in the data command was I guess used in Spanner db creation. Now since creation responsibility is shifted to schema command, please port this flag over to the `schema` command.

shreyakhajanchi: since we are adding the foreign key as part of data.go so retained this flag as it might make sense in this case

cmd/data.go
shreyakhajanchi: Used functions in case of repeated code and modified the PR

common/utils/utils.go
shreyakhajanchi: For tables and columns we are reading from map so ordering won't affect and for indexes we store them in the order of definition.

shreyakhajanchi: Added handling for case where spanner database may have extra table. But in case of same tables we should check for similarity in definition to avoid data write failures.

Deep1998 (approved):
LGTM

cmd/utils.go
```diff
@@ -0,0 +1,84 @@
+/* Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
shreyakhajanchi: Tried placing in common/utils but getting import cycle issue therefore created within same package

PR #331, opened 2022-07-06
Adding Dockerfile of HarbourBridge's Docker image

its7ARC:
This PR consists of -  
-  The `Dockerfile` of HarbourBridge's Docker image. 

This file is necessary to enable Google Cloud Build to build HarbourBridge's Docker images and put them in the Google Container Registry. 

The built images 
- Will support all the migration modes of HarbourBridge - 
  - schema
  - data
  - schema-and-data

- Will support both forms of source databases that HarbourBridge does - 
  - Dump files in the host's file system.
  - Remotely hosted databases.

The `Dockerfile` consists of -
- Build Stage : This stage uses the HarbourBridge source code,  its go.mod dependencies and Golang to compile the software into a binary file `harbourbridge`. This binary is copied into the release stage.
- Release Stage : The image built in this stage will be available to the users. It packages the HarbourBridge binary file `harbourbridge` and GCloud SDK together. 
  - The binary will get executed to perform migrations from the source databases to Spanner.
  - GCloud SDK will authenticate the host's GCloud account and put the migrated database in the relevant Spanner instance.
  - GCloud SDK needs python and an OS to function, thus `Python3` and `Ubuntu OS` are also part of the image.


Dockerfile
```diff
@@ -0,0 +1,66 @@
+# Copyright 2020 Google LLC
```
Deep1998, line 1: 2022 :)

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+# syntax=docker/dockerfile:1
+
+######################### BUILD STAGE 1 ############################
+# This stage builds the HarbourBridge binary using its source code.
+
+# Add the official Docker golang image as the base image for the build stage
+FROM golang:latest AS build
+
+# Set the default work directory as “/hbBin” in the container.
+WORKDIR /hbBin
```
Deep1998, line 25: nit: should we rename this to `harbourbridge` ? keeps it consistent with what it says when we clone it.

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+# Copy HarbourBrdige’s source code into the working directory.
+COPY . ./
```
Deep1998, line 28: Are we sure this is what the command is supposed to do?  this looks like it is copying current directory to current directory. 
Also typo in in comment `HarbourBrdige’s`

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+# Download HarbourBridge’s dependencies mentioned in the go.mod file.
+RUN go mod tidy 
+# Compile the source code and store the binary in the work directory /hbBin.
+RUN make build-static  
+
+######################### Release Stage ###########################
+# This stage consists of the GCloud SDK and the HabrbourBridge binary.
+
+# Add the Ubuntu base image from the Google Container Registry.
+FROM marketplace.gcr.io/google/ubuntu1804:latest
```
Deep1998, line 38: Aviod using latest here too

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+# Set the default working directory.
+# This directory binds with the host directory that is to store the output files generated by HarbourBridge.
+WORKDIR /harbourBridge/harbour_bridge_output
```
Deep1998, line 41: nit: consider using harbourbridge everywhere instead of harbourBridge for consistency

Deep1998: Also, to clarify, we don't need this step right? harbourbridge should have been able to create this automatically to right?

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+# Create directory to store the HarbourBridge binary. 
+RUN mkdir /harbourBridge/bin
+# Create directory to store the dump files. This directory binds with the host directory that contains the dump files of the source database.
+RUN mkdir /harbourBridge/sourceDump
+# Copy the HarbourBridge binary from the build stage's hbBin directory to the release stage's habourBridge/bin directory
+COPY --from=build /hbBin/harbourbridge /harbourBridge/bin
+
+# Add Python3 to support GCloud SDK.
+RUN apt update
+RUN apt install -y python3
```
Deep1998, line 51: I think its best to mention a version everywhere possible

Deep1998 (changes requested):
Left a few comments

Dockerfile
its7ARC: The binary file name is also `harbourbridge`. The path to the binary will become `/harbourbridge/harbourbridge` if we make this change. I think the current name is more descriptive, it tells that this is the binary directory for  `harbourbridge` and the path to the binary is `/hbBin/harbourbridge`.

Dockerfile
its7ARC: No, it is needed. In the present code, HarbourBridge puts the output files in the location from where the migration commands are called. So we've set the point to call the migration command from - the default working directory as `/harbourBridge/harbour_bridge_output`. This is where the output files will be stored after migration. This directory will be bind mounted with the host's file system to put the output files in the latter.

its7ARC: If we change `harbourBridge` to `harbourbridge`, the file paths for the binary file `harbourbridge` will become slightly less clear. eg - `harbourbridge/bin/harbourbridge` in the release stage, `harbourbridge/harbourbridge` in the build stage. Because of this reason we've used `harbourBridge` for the directory names and `harbourbridge` for the binary file name.

Dockerfile
its7ARC: Okay.

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+# limitations under the License.
+
+
+# syntax=docker/dockerfile:1
+
+######################### BUILD STAGE 1 ############################
+# This stage builds the HarbourBridge binary using its source code.
+
+# Add the official Docker golang image as the base image for the build stage
+FROM golang:latest AS build
```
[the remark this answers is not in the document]
its7ARC: Okay

Dockerfile
its7ARC: Done.

Dockerfile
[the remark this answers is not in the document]
its7ARC: Done.

Dockerfile
its7ARC: Done.

Dockerfile
its7ARC: Done.

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# syntax=docker/dockerfile:1
+
+######################### BUILD STAGE 1 ############################
+# This stage builds the HarbourBridge binary using its source code.
+
+# Add the official Docker golang image as the base image for the build stage
```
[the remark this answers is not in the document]
its7ARC: done.

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+RUN mkdir -p /usr/local/gcloud \
+  && tar -C /usr/local/gcloud -xf /tmp/google-cloud-sdk.tar.gz \
+  && /usr/local/gcloud/google-cloud-sdk/install.sh
+
+# Add gcloud to $PATH
+ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
+
+# Default command to execute when the container is run.
+# NOTE : This command will be overridden by the HarborBridge migration commands.
+CMD [ "/harbourBridge/bin/harbourbridge" ]
```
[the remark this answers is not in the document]
its7ARC: > Is the CMD command necessary? since we don't plan on overriding it anyways?

The CMD command is compulsary in Dockerfiles.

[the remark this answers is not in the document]
its7ARC: No, it is not. Will remove it.

Dockerfile
Deep1998: harbourbridge/bin/harbourbridge should be fine, if you see, even mysql uses mysq/bin/mysql.
hbBin we can keep as is since it is temporary. My concern is if the user logs in to the container, they will finally see the conventional productname/bin/productname rather than different casing at each level.

Dockerfile
manitgupta: maybe `harbourbridge_bin` or something which follows `my_dir_type`? I haven't seen directories being named using `camelCase` in linux..

manitgupta:
Please keep naming consistent, see `harbourBridge`, `harbour_bridge`, `harbourbridge` and acronyms like `hbBin` in the PR.

Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+# Create directory to store the HarbourBridge binary. 
+RUN mkdir /harbourBridge/bin
+# Create directory to store the dump files. This directory binds with the host directory that contains the dump files of the source database.
+RUN mkdir /harbourBridge/sourceDump
+# Copy the HarbourBridge binary from the build stage's hbBin directory to the release stage's harbourBridge/bin directory.
+COPY --from=build /hbBin/harbourbridge /harbourBridge/bin
+
+# Add Python3 to support GCloud SDK.
+RUN apt-get update
+RUN apt-get install -y python3.10
```
manitgupta, line 51: Why is this needed? What versions of `python` does `gCloud SDK` support?

its7ARC: GCloud SDK needs Python2.7.9 or higher. Refer - [Python Requirement](https://cloud.google.com/sdk/docs/install#:~:text=The%20Google%20Cloud%20CLI%20requires%20Python%203%20(3.5%20to%203.8%2C%203.7%20recommended)%20and%20Python%202%20(2.7.9%20or%20higher).).

Dockerfile
its7ARC: Done.

Dockerfile
its7ARC: Done. `hbBin` > `harbourbridge_bin`.

Deep1998 (approved):
LGTM

PR #334, opened 2022-07-07
Adding Cloud Build's configuration file to setup GIT-GCR pipeline.

its7ARC:
This PR consists of 
- A `cloudbuild.yaml` file.

The file will be used by Google Cloud Build to establish a pipeline between the HarbourBridge's git repository and the Google Container Registry.
The pipeline will enable Cloud Build to build docker images of HarbourBridge using its latest source code in its git repository's master branch and push the images to GCR.

Trigger for this operation - build and push, will be to click on the `run` button of the `HarbourBridge-Image-Release` trigger in the `Cloud Build > Triggers` console in Google Cloud. 

The tags for the released image in GCR will be 
- Version Number. eg - `1.0.1`. This will be specified by the developer in the release trigger.
- `latest`

The logs of the image build will be stored in the `${PROJECT_ID}_cloudbuild` storage bucket.

### DIFF
--- a/.docker/cloudbuild.yaml
+++ b/.docker/cloudbuild.yaml
@@ -1,0 +1,28 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Build the Docker Image of HarbourBridge using its latest source code on Github and push it to GCR.
+# Add _VERSION_NUMBER and 'latest' as the tags of the image. 
+steps:
+- name : 'gcr.io/cloud-builders/docker'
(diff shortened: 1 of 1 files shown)


.docker/cloudbuild.yaml
```diff
@@ -0,0 +1,28 @@
...
+steps:
+- name : 'gcr.io/cloud-builders/docker'
+  args : ['build', '-t', 'gcr.io/${PROJECT_ID}/harbour-bridge:${_VERSION_NUMBER}', '-t', 'gcr.io/${PROJECT_ID}/harbour-bridge:latest', '.']
+  id : 'build-harbourbridge-image'
+  waitFor: ['-']
+substitutions:
+  _VERSION_NUMBER: 0.0.0 # default value.
+images:
+- 'gcr.io/$PROJECT_ID/harbour-bridge:latest'
+- 'gcr.io/$PROJECT_ID/harbour-bridge:${_VERSION_NUMBER}'
```
manitgupta, line 26: Does this mean two images will be pushed? AFAIK tag and image have 1:1 mapping. 
Also, just to confirm, can there be a case where the `latest` tag is updated with the new image but an image corresponding to the `VERSION_NUMBER` fails to publish?

its7ARC: The 1:1 mapping is when image names are different. In our case, the `latest` and `_VERSION_NUMBER` tags will be added to a single published image.

> Does this mean two images will be pushed? AFAIK tag and image have 1:1 mapping.

No.

> Also, just to confirm, can there be a case where the latest tag is updated with the new image but an image corresponding to the VERSION_NUMBER fails to publish?

Since a single image is published, this case is not possible.
There may be a case when all the tags are not published for the image. But we use the Google's Cloud Build tool to publish, so it is unlikely.

PR #336, opened 2022-07-10
Support for writing data from DynamoDB Streams to Cloud Spanner

nareshz:
Things added
- Converting data read from DynamoDB Streams to Cloud Spanner data based on source and spanner schemas.
- Creating mutations from converted data and processing them to Cloud Spanner.
- Collecting of bad and dropped records if errors are faced.
- Retry logic if insertion fails because of missing parent data.
- Updating customers about the progress of processing every minute. 
- Unit tests.

Things not handled
- Changes to report, badData files and Integration tests.
- Appropriate changes to README for using streaming migration.
- These will be added in subsequent PRs.

### DIFF
--- a/sources/dynamodb/streaming.go
+++ b/sources/dynamodb/streaming.go
@@ -14,21 +14,31 @@
 package dynamodb
 
 import (
+	"context"
+	"encoding/base64"
 	"fmt"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
 	"time"
 
+	sp "cloud.google.com/go/spanner"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbiface"
 	"github.com/aws/aws-sdk-go/service/dynamodbstreams"
 	"github.com/aws/aws-sdk-go/service/dynamodbstreams/dynamodbstreamsiface"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/proto"
 
+	"github.com/cloudspannerecosystem/harbourbridge/common/constants"
+	"github.com/cloudspannerecosystem/harbourbridge/common/metrics"
 	"github.com/cloudspannerecosystem/harbourbridge/internal"
+	"github.com/cloudspannerecosystem/harbourbridge/schema"
+	"github.com/cloudspannerecosystem/harbourbridge/spanner/ddl"
 )
 
 // NewDynamoDBStream initializes a new DynamoDB Stream for a table with NEW_AND_OLD_IMAGES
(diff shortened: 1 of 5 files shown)


sources/dynamodb/streaming.go
```diff
@@ -253,9 +315,133 @@ func getRecords(streamClient dynamodbstreamsiface.DynamoDBStreamsAPI, shardItera
...
+	srcSchema := conv.SrcSchema[srcTable]
+	spTable, err1 := internal.GetSpannerTable(conv, srcTable)
+	spCols, err2 := internal.GetSpannerCols(conv, srcTable, srcSchema.ColNames)
+	spSchema, ok := conv.SpSchema[spTable]
+	var err error
+	if err1 != nil || err2 != nil || !ok {
+		err = fmt.Errorf(fmt.Sprintf("err1=%s, err2=%s, ok=%t", err1, err2, ok))
+	}
+	return srcSchema, spTable, spCols, spSchema, err
+}
```
shreyakhajanchi, line 329: This seems to be a generic method not specific to dynamodb so its better to place it in utility

sources/dynamodb/streaming.go
```diff
@@ -253,9 +315,133 @@ func getRecords(streamClient dynamodbstreamsiface.DynamoDBStreamsAPI, shardItera
...
+	if len(badCols) == 0 {
+		writeRecord(streamInfo, srcTable, spTable, eventName, spCols, spVals, srcSchema)
+	} else {
+		streamInfo.StatsAddBadRecord(srcTable, eventName)
+		streamInfo.CollectBadRecord(eventName, srcTable, srcSchema.ColNames, srcStrVals)
+	}
 	streamInfo.StatsAddRecordProcessed()
 }
+
+// writeRecord checks if the writer to write data to Cloud Spanner is configured or not.
```
shreyakhajanchi, line 361: Please rephrase this to make it clearer

sources/dynamodb/streaming.go
```diff
@@ -253,9 +315,133 @@ func getRecords(streamClient dynamodbstreamsiface.DynamoDBStreamsAPI, shardItera
...
+		time.Sleep(4 * time.Second)
+		retryLimit--
+	}
+	return err
+}
+
+// setWriter initializes the write function used to write mutations to Cloud Spanner.
+func setWriter(streamInfo *StreamingInfo, client *sp.Client, conv *internal.Conv) {
+	streamInfo.write = func(m *sp.Mutation) error {
+		const migrationMetadataKey = "cloud-spanner-migration-metadata"
```
shreyakhajanchi, line 440: Since this const is used at multiple places it would be better to put it in constants file and use it from there

shreyakhajanchi (commented):
Added few comments

sources/dynamodb/streaming.go
```diff
@@ -80,6 +90,58 @@ func catchCtrlC(wg *sync.WaitGroup, streamInfo *StreamingInfo) {
...
+
+		lastFiveMin += arr[counter]
+
+		if timer < 5 {
+			firstFiveMin += arr[counter]
+		}
+
+		lastMin := arr[counter]
+		optimumCondition := ((lastFiveMin*100 <= 5*firstFiveMin) || (lastMin == 0))
+		updateProgress(optimumCondition, false, tillLastMin)
```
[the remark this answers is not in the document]
nareshz: streamInfo.recordsProcessed may increase during the time we are checking optimumCondition and calling updateProgress.
We are using tillLastMin to show that the last time we checked optimumCondition this was the number of records processed.

sources/dynamodb/streaming.go
nareshz: Okay! will do that.

sources/dynamodb/streaming.go
```diff
@@ -253,9 +315,133 @@ func getRecords(streamClient dynamodbstreamsiface.DynamoDBStreamsAPI, shardItera
...
+		m = sp.Delete(spTable, sp.Key{reqSpVals[0]})
+	} else {
+		m = sp.Delete(spTable, sp.Key{reqSpVals[0], reqSpVals[1]})
+	}
+	return m
+}
+
+// parentDataMissingError checks if the error is a parent data missing error.
+func parentDataMissingError(err error) bool {
+	return strings.Contains(err.Error(), "NotFound") && strings.Contains(err.Error(), "Parent row") && strings.Contains(err.Error(), "is missing")
```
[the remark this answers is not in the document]
nareshz: I checked after our discussion and found that Contains is better in terms of performance.

sources/dynamodb/streaming.go
```diff
@@ -407,9 +407,9 @@ func parentDataMissingError(err error) bool {
 }
 
 // writeMutation handles writing of a mutation to Cloud Spanner. If insertion fails
-// because of parent data missing error then it retries for a limit of 1000.
+// because of parent data missing error then it retries for a limit of 100.
 func writeMutation(m *sp.Mutation, streamInfo *StreamingInfo) error {
-	retryLimit := 1000
+	retryLimit := 100
```
manitgupta, line 412: Don't we have a constants or config file in harbourbridge where we can declare these constants?
If not, lets declare `retryLimit` as a constant instead of a variable.

conversion/conversion.go
```diff
@@ -329,7 +327,7 @@ func populateDataConv(conv *internal.Conv, config writer.BatchWriterConfig, clie
 		migrationData := metrics.GetMigrationData(conv, "", "", constants.DataConv)
 		serializedMigrationData, _ := proto.Marshal(migrationData)
 		migrationMetadataValue := base64.StdEncoding.EncodeToString(serializedMigrationData)
-		_, err := client.Apply(metadata.AppendToOutgoingContext(context.Background(), migrationMetadataKey, migrationMetadataValue), m)
+		_, err := client.Apply(metadata.AppendToOutgoingContext(context.Background(), constants.MigrationMetadataKey, migrationMetadataValue), m)
```
manitgupta, line 330: Is this change backward compatible? Can it cause a split with the existing migration usage data that we have/and or break the migration dashboard we have created?

sources/dynamodb/streaming.go
nareshz: we have a constant file but this retryLimit is very specific to DynamoDB. So it wouldn't make sense to put in constants file.

Okay! will change retryLimit to a constant.

sources/dynamodb/streaming.go
```diff
@@ -253,9 +316,119 @@ func getRecords(streamClient dynamodbstreamsiface.DynamoDBStreamsAPI, shardItera
...
+		m = sp.Delete(spTable, sp.Key{reqSpVals[0]})
+	} else {
+		m = sp.Delete(spTable, sp.Key{reqSpVals[0], reqSpVals[1]})
+	}
+	return m
+}
+
+// parentDataMissingError checks if the error is a parent data missing error.
+func parentDataMissingError(err error) bool {
+	return strings.Contains(err.Error(), "NotFound") && strings.Contains(err.Error(), "Parent row") && strings.Contains(err.Error(), "is missing")
```
manitgupta, line 406: This is a pretty brittle piece of code. Depending on a the text of an error is risky. Unlike err types which are not usually change without notice to the client, the error text may change without notice.
If there is no way around this at this time, let's add a comment where on why we did this and create an issue in the repo to check we can come back and fix it later.

nareshz: As this error is very specific in this solution, adding a comment on why we are doing this and a note that this may be subject to change.

PR #340, opened 2022-07-14
DynamoDB Streams: Integration test, Report, Readme

nareshz:
**Things added**

- Integration tests for end to end testing of streaming migration in DynamoDB.
- Changes to report and bad data files for showing progress of streaming migration. Includes
  - Tabular report for all the Insert, Modify, Remove records. (Successfully Written Records / Total Read Records)

**Table Format**
| Table  | Insert | Modify | Remove |
|--------|--------|--------|--------|
| table1 | 1/2    | 2/3    | 3/4    |
| table2 | 2/3    | 1/2    | 1/1    |

- Changes to README for using the streaming migration in DynamoDB.


internal/convert.go
```diff
@@ -150,6 +150,17 @@ type Audit struct {
 	MigrationRequestId       string                                 `json:"-"` // Unique request id generated per migration
 	MigrationType            *migration.MigrationData_MigrationType `json:"-"` // Type of migration: Schema migration, data migration or schema and data migration
 	DryRun                   bool                                   `json:"-"` // Flag to identify if the migration is a dry run
+	StreamingStats           streamingStats                         `json:"-"` // Stores information related to streaming migration process.
+}
+
+// Stores information related to the streaming migration process.
+type streamingStats struct {
+	Streaming        bool                        // Flag for confirmation of streaming migration.
+	Records          map[string]map[string]int64 // Tablewise count of records received for processing, broken down by record type i.e. INSERT, MODIFY & REMOVE.
```
shreyakhajanchi, line 159: Better to rename as TotalRecords

sources/dynamodb/README.md
```diff
@@ -93,6 +93,28 @@ Sample usage:
 harbourbridge schema -source=dynamodb -source-profile="schema-sample-size=500000,aws-access-key-id=<>,..."
 ```
 
+## DynamoDB Streaming Migration Usage
+
+### Note
```
shreyakhajanchi, line 98: We can skip this heading and directly list the details

internal/report.go
```diff
@@ -91,9 +92,134 @@ func GenerateReport(driverName string, conv *Conv, w *bufio.Writer, badWrites ma
...
 		writeUnexpectedConditions(driverName, conv, w)
 	}
+
+	if conv.Audit.StreamingStats.Streaming {
+		writeStreamingMigrationReport(driverName, conv, w)
+	}
 	return summary
 }
 
+func sumNestedMapValues(mp map[string]map[string]int64) int64 {
```
shreyakhajanchi, line 102: can be moved to utility

internal/report.go
```diff
@@ -91,9 +92,134 @@ func GenerateReport(driverName string, conv *Conv, w *bufio.Writer, badWrites ma
...
+func getStreamName(driver string) (string, error) {
+	switch driver {
+	case constants.DYNAMODB:
+		return "DynamoDB Streams", nil
+	default:
+		return "", fmt.Errorf("streaming migration report only printed for DynamoDB currently")
+	}
+}
+
+func getChangeTypes(driver string) []string {
```
shreyakhajanchi, line 121: getRecordTypes seems more suitable here

testing/dynamodb/streaming/integration_test.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+
+	populateDynamoDBStreams(t)
+
+	// Wait for a maximum of 5 minutes and check row count
+	timeLimit = 300
+	for timeLimit > 0 {
+		time.Sleep(2 * time.Second)
+		// Check count of row to be 2.
+		rowCount, err := rowCountTestTable(client)
```
shreyakhajanchi, line 278: Instead of checking the table counts did we discuss to check for the output msg in logs ?

shreyakhajanchi (commented):
Overall looks good. Added few minor comments

testing/dynamodb/streaming/integration_test.go
nareshz: yes, we did discuss that. But the problem was with updating the output every minute. Instead we can just check every 2 minutes whether row count is 2 or not from spanner queries. This will be better in terms of time consumption.

internal/report.go
nareshz: If we move it to utility then there are import cycle issues. That's why this was implemented here.

shreyakhajanchi (approved):

PR #344, opened 2022-07-18
Modifying modes of migration in HarbourBridge

shreyakhajanchi:
1. Added dry run mode of migration
2. Modified schema only and data only mode of migration
3. UI fixes to handle the updated structure
4. Modified the unit and integration tests

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -1,0 +1,84 @@
+/* Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.*/
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	sp "cloud.google.com/go/spanner"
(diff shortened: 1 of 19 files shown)


cmd/utils.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2022 Google LLC
```
manitgupta, line 1: We already have a `common/utils/utils.go` class, can we rename this and place this outside the `cmd` folder. The `cmd` folder is the first-level folder for all the sub-commands HB exposes. Let's preserve that.

common/utils/utils.go
```diff
@@ -506,6 +506,60 @@ func ReadSpannerSchema(ctx context.Context, conv *internal.Conv, client *sp.Clie
...
+			spannerCol := spannerTable.ColDefs[col]
+			if colDef.Name != spannerCol.Name || colDef.NotNull != spannerCol.NotNull ||
+				colDef.T.IsArray != spannerCol.T.IsArray || colDef.T.Len != spannerCol.T.Len || colDef.T.Name != spannerCol.T.Name {
+				return fmt.Errorf("column detail for table %v don't match", sessionTable.Name)
+			}
+		}
+		for i := range sessionTable.Indexes {
+			sessionDbIndex := sessionTable.Indexes[i]
+			found := 0
+			for j := range spannerTable.Indexes {
```
manitgupta, line 538: I am not sure if it is a `golang` thing, but can we use replace existing variables `t, i, j` with something more descriptive? 
It is hard to read the code where they are being referred to at once.

manitgupta (commented):
It will be great if we can also document in an open issue what error was not captured in the unit and integration tests we have which led to reverting the PR last time.
We can come back to it later.

cmd/utils.go
shreyakhajanchi: Putting these methods in utils/utils.go was creating an import cycle

manitgupta (approved):
LGTM

PR #348, opened 2022-07-19
UI backlogs 

pratikdhanavesearce:
synth_id column display in UI
Update Issues and Suggestions count bug fix
Update object detail on changing tab between source db and spanner db
alignment of the edit columns under index
changed source db from pg_dump to Postgres during pg_dump load


ui/src/app/components/object-explorer/object-explorer.component.html
```diff
@@ -1,5 +1,9 @@
 <div class="container">
-  <mat-tab-group [ngClass]="[isLeftColumnCollapse ? 'hidden' : 'display']" selectedIndex="1">
+  <mat-tab-group
```
shreyakhajanchi, line 2: https://drive.google.com/file/d/1tkeikLv0GzkPcv5kGWjukmwwR__KTiPT/view?usp=sharing&resourcekey=0-FOm_fyVTWLHFjm5z4qUYxA: the object explorer is showing click on the object name on the Spanner draft panel instead of source database

ui/src/app/services/conversion/conversion.service.ts
```diff
@@ -147,25 +147,47 @@ export class ConversionService {
...
-        spDataType: spannerColDef ? spannerColDef.T.Name : '',
-        srcColName: name,
-        srcDataType: data.SrcSchema[srcTableName].ColDefs[name].Type.Name,
-        spIsPk: spannerColDef
-          ? data.SpSchema[tableName].Pks.map((p) => p.Col).indexOf(spColName) !== -1
-          : false,
-        srcIsPk: srcPks ? srcPks.map((p) => p.Column).indexOf(name) !== -1 : false,
-        spIsNotNull: spannerColDef ? data.SpSchema[tableName].ColDefs[spColName].NotNull : false,
-        srcIsNotNull: data.SrcSchema[srcTableName].ColDefs[name].NotNull,
+    const res: IColumnTabData[] = data.SrcSchema[srcTableName].ColNames.map(
```
shreyakhajanchi, line 150: This is part of which fix?

ui/src/app/components/session-listing/session-listing.component.html
```diff
@@ -11,7 +11,8 @@ <h3>Session history</h3>
           <div class="table-header-container">
             <span>Session Name</span>
             <button mat-icon-button (click)="toggleFilterDisplay('sessionName')">
-              <mat-icon>filter_list</mat-icon>
+              <mat-icon *ngIf="displayFilter['sessionName']">keyboard_arrow_up</mat-icon>
```
shreyakhajanchi, line 14: In case no rows match the filter the alignment of header changes because Resume & Download buttons are not visible. We need to fix the css

ui/src/app/components/object-explorer/object-explorer.component.html
lyazii22: @shreyakhajanchi **"To view an object details, click on the object name on the [source database] panel."**

lyazii22: Should we change it too something like this when we are in source panel and didn't select any object?

ui/src/app/services/conversion/conversion.service.ts
lyazii22: @shreyakhajanchi This part is modified for adding synthetic key in spanner column.

ui/src/app/components/object-explorer/object-explorer.component.html
shreyakhajanchi: yes

shreyakhajanchi (approved):

PR #354, opened 2022-07-22
Fixing report file names for consistency with db name

shreyakhajanchi:
Currently, the generated database name and file names mismatch in case of target db name provided. This PR fixes the issue by generating the name only once.


manitgupta (approved):
LGTM

PR #357, opened 2022-08-04
Added migrate page in HarbourBridge UI

shreyakhajanchi:
- Added a new prepare migration page to run end to end migration from UI
- Added support for both bulk and low downtime migration
- Added respective support for direct connect, dump file and session file mode


cmd/utils.go
```diff
@@ -85,3 +88,45 @@ func PrepareMigrationPrerequisites(sourceProfileString, targetProfileString, sou
 	}
 	return sourceProfile, targetProfile, ioHelper, dbName, nil
 }
+
+// MigrateData creates database and populates data in it.
+func MigrateData(ctx context.Context, targetProfile profiles.TargetProfile, sourceProfile profiles.SourceProfile, dbName string, ioHelper *utils.IOStreams, writeLimit int64, conv *internal.Conv, skipForeignKeys, schemaOnly, dataOnly bool) (*writer.BatchWriter, error) {
```
manitgupta, line 93: This method signature mixes together entities which aren't on the same level. `skipForeignKeys` is a parameter of the schema and data commands, but is passed in together with `schemaOnly` and `dataOnly` booleans.

I will recommend passing in a key-value pair to this method with the `key` being the `command` name (schema, data etc.) and the value being an object which provides the user configurations.
That should enhance readability.

cmd/utils.go
```diff
@@ -85,3 +88,45 @@ func PrepareMigrationPrerequisites(sourceProfileString, targetProfileString, sou
...
+			err = fmt.Errorf("can't finish data conversion for db %s: %v", dbURI, err)
+			return nil, err
+		}
+
+		if !skipForeignKeys {
+			if err = conversion.UpdateDDLForeignKeys(ctx, adminClient, dbURI, conv, ioHelper.Out); err != nil {
+				err = fmt.Errorf("can't perform update schema on db %s with foreign keys: %v", dbURI, err)
+				return bw, err
+			}
+		}
```
manitgupta, line 129: This seems like too much branching, it will be hard to unit test and cover the branches for this method, especially if we add more subcommands. Can you think of a way to solve this?

webv2/web.go
```diff
@@ -90,6 +92,27 @@ type driverConfig struct {
...
+	MigrationMode string        `json:MigrationMode`
+	MigrationType string        `json:MigrationType`
+}
+
+type targetDetails struct {
+	TargetDB     string `json:"TargetDB"`
+	StreamingCfg string `json:"StreamingConfig"`
+}
+
 // databaseConnection creates connection with database when using
```
manitgupta, line 116: Is this the correct place for these structs?
Consider if they are local to `web.go` or if they will have future usability in other locations.

webv2/web.go
```diff
@@ -963,6 +1001,101 @@ func addIndexes(w http.ResponseWriter, r *http.Request) {
...
+		sourceIndexCount = sourceIndexCount + len(sourceSchema.Indexes)
+	}
+	sessionSummary.SourceIndexCount = sourceIndexCount
+	sessionSummary.SpannerIndexCount = spannerIndexCount
+	w.WriteHeader(http.StatusOK)
+	json.NewEncoder(w).Encode(sessionSummary)
+}
+
+func migrate(w http.ResponseWriter, r *http.Request) {
+
```
manitgupta, line 1031: Method is too long, consider refactoring to smaller units.

cmd/utils.go
shreyakhajanchi: updated to use interface structure

cmd/utils.go
shreyakhajanchi: updated

webv2/web.go
shreyakhajanchi: done

webv2/web.go
shreyakhajanchi: structs defined here can be used within package webv2. and they have use limited to web.go

cmd/data.go
```diff
@@ -44,10 +44,10 @@ type DataCmd struct {
 	targetProfile   string
 	sessionJSON     string
 	filePrefix      string // TODO: move filePrefix to global flags
-	writeLimit      int64
+	WriteLimit      int64
 	dryRun          bool
 	logLevel        string
-	skipForeignKeys bool
+	SkipForeignKeys bool
```
manitgupta, line 50: Why are these capitalised? Lets keep them the same as before to avoid confusion.
(I know the earlier convention used might be wrong but changing case for only some variables decreases readability)

cmd/utils.go
```diff
@@ -85,3 +88,60 @@ func PrepareMigrationPrerequisites(sourceProfileString, targetProfileString, sou
...
+			return nil, err
+		}
+		if !v.SkipForeignKeys {
+			if err = conversion.UpdateDDLForeignKeys(ctx, adminClient, dbURI, conv, ioHelper.Out); err != nil {
+				err = fmt.Errorf("can't perform update schema on db %s with foreign keys: %v", dbURI, err)
+				return bw, err
+			}
+		}
+	}
+	return bw, nil
```
manitgupta, line 146: Thanks for refactoring this, this is much more readable now! :)
One more suggestion on this - refactor the code inside the individual switch cases to private method to further improve the flow!
(The reader probably does not want to understand the details of how **_MigrateDatabase_** is performed for each of the different commands while reading this method.)

manitgupta (commented):
Minor comments but largely LGTM now.

cmd/data.go
shreyakhajanchi: Made them capitalised because we want to refer them in web.go file

webv2/routes.go
```diff
@@ -68,5 +68,10 @@ func getRoutes() *mux.Router {
 	router.HandleFunc("/GetConfig", config.GetConfig).Methods("GET")
 	router.HandleFunc("/SetSpannerConfig", config.SetSpannerConfig).Methods("POST")
 
+	// Run migration
+	router.HandleFunc("/Migrate", migrate).Methods("POST")
```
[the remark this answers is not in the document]
bharadwaj-aditya: Let it go in this PR, but need to cleanup across the file

bharadwaj-aditya (commented):
Great job overall. Just some minor comments.

cmd/schema.go
```diff
@@ -118,22 +117,10 @@ func (cmd *SchemaCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interfa
 	conv.Audit.MigrationRequestId = "HB-" + uuid.New().String()
 	conv.Audit.MigrationType = migration.MigrationData_SCHEMA_ONLY.Enum()
 
-	var (
-		adminClient *database.DatabaseAdminClient
-		dbURI       string
-	)
-
 	if !cmd.dryRun {
-		adminClient, _, dbURI, err = CreateDatabaseClient(ctx, targetProfile, sourceProfile.Driver, dbName, ioHelper)
```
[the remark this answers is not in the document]
shreyakhajanchi: creation of foreign key happens after data population currently so its not a part of schema cmd action in MigrateDatabase

ui/src/app/components/target-details-form/target-details-form.component.ts
```diff
@@ -0,0 +1,42 @@
...
+
+@Component({
+  selector: 'app-target-details-form',
+  templateUrl: './target-details-form.component.html',
+  styleUrls: ['./target-details-form.component.scss'],
+})
+export class TargetDetailsFormComponent implements OnInit {
+  targetDetailsForm: FormGroup
+  constructor(
+    private fb: FormBuilder,
```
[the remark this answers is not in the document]
shreyakhajanchi: done

ui/src/app/components/prepare-migration/prepare-migration.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    private dialog: MatDialog,
+    private fetch: FetchService,
+    private snack: SnackbarService,
+    private targetDetailService: TargetDetailsService
+  ) {}
+
+  isTargetDetailSet: boolean = false
+  isStreamingCfgSet: boolean = false
+  isSchemaMigration: boolean = true
+  isBulkMigration: boolean = true
```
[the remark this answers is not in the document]
shreyakhajanchi: modified the name to make better sense and set the default to false

ui/src/app/components/prepare-migration/prepare-migration.component.scss
```diff
@@ -0,0 +1,99 @@
+.header {
```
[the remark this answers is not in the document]
shreyakhajanchi: moved to styles.css

cmd/utils.go
```diff
@@ -101,46 +101,72 @@ func MigrateDatabase(ctx context.Context, targetProfile profiles.TargetProfile,
...
-		if !v.SkipForeignKeys {
-			if err = conversion.UpdateDDLForeignKeys(ctx, adminClient, dbURI, conv, ioHelper.Out); err != nil {
-				err = fmt.Errorf("can't perform update schema on db %s with foreign keys: %v", dbURI, err)
-				return bw, err
-			}
+	}
+	return bw, nil
+}
+
+func migrateSchemaAndData(ctx context.Context, targetProfile profiles.TargetProfile, sourceProfile profiles.SourceProfile,
```
bharadwaj-aditya, line 154: basic question - Why is this step not just calling migrateSchema and migrateData ?

bharadwaj-aditya (approved):
LGTM

cmd/utils.go
shreyakhajanchi: because in migrateData we validate the existing db which is not the case in schemaAndData

PR #365, opened 2022-08-17
Add releaser script and update Makefile

manitgupta:
- Adds a releaser script to create new releases for Harbourbridge.
- Updates the Makefile to remove redundant build targets and add relevant ones.


bharadwaj-aditya (approved):
LGTM

PR #367, opened 2022-08-25
Supporting changes for low downtime migrations

Deep1998:
This PR finishes the low downtime workflow by adding support to Write the session file to GCS after launching datastream.

This PR also takes care of the following:

- Multiplexing the snapshot migration logic correctly for all use cases, non streaming, DynamoDB streaming and MySQL/Oracle Streaming
- It also sets the numWorkers parameters now when launching the dataflow job.

### DIFF
--- a/common/utils/utils.go
+++ b/common/utils/utils.go
@@ -170,6 +170,43 @@ func PreloadGCSFiles(tables []ManifestTable) ([]ManifestTable, error) {
 	return tables, nil
 }
 
+func WriteToGCS(filePath, fileName, data string) error {
+	ctx := context.Background()
+
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		fmt.Printf("Failed to create GCS client")
+		return err
+	}
+	defer client.Close()
+	if filePath[len(filePath)-1] != '/' {
+		filePath = filePath + "/"
+	}
+	u, err := url.Parse(filePath)
+	if err != nil {
+		fmt.Printf("parseFilePath: unable to parse file path %s", filePath)
+		return err
+	}
+	if u.Scheme != "gs" {
+		fmt.Printf("not a valid GCS path: %s, should start with 'gs'", filePath)
+		return err
+	}
+	bucketName := u.Host
+	bucket := client.Bucket(bucketName)
+	obj := bucket.Object(u.Path[1:] + fileName)
+
+	w := obj.NewWriter(ctx)
+	if _, err := fmt.Fprint(w, data); err != nil {
+		fmt.Printf("Failed to write to Cloud Storage: %s", filePath)
+		return err
+	}
+	if err := w.Close(); err != nil {
+		fmt.Printf("Failed to close GCS file: %s", filePath)
+		return err
+	}
+	return nil
+}
+
 // GetProject returns the cloud project we should use for accessing Spanner.
 // Use environment variable GCLOUD_PROJECT if it is set.
 // Otherwise, use the default project returned from gcloud.
--- a/common/utils/utils.go
+++ b/common/utils/utils.go
@@ -73,7 +73,7 @@ func NewIOStreams(driver string, dumpFile string) IOStreams {
 		fmt.Printf("\nLoading dump file from path: %s\n", dumpFile)
 		var f *os.File
 		var err error
-		if u.Scheme == "gs" {
+		if u.Scheme == constants.GCS_SCHEME {
 			bucketName := u.Host
 			filePath := u.Path[1:] // removes "/" from beginning of path
 			f, err = DownloadFromGCS(bucketName, filePath, "harbourbridge.gcs.data")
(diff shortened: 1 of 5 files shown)


Deep1998:
Need to support custom DLQs

streaming/streaming.go
```diff
@@ -100,11 +104,37 @@ func VerifyAndUpdateCfg(streamingCfg *StreamingCfg, dbName string) error {
...
 		jobName, err := utils.GenerateName("hb-dataflow-" + dbName)
+		jobName = strings.Replace(jobName, "_", "-", -1)
 		if err != nil {
 			return fmt.Errorf("error generating stream name: %v", err)
 		}
 		streamingCfg.DataflowCfg.JobName = jobName
 	}
+
+	filePath := streamingCfg.TmpDir
+	if filePath[len(filePath)-1] != '/' {
```
shreyakhajanchi, line 115: This could break if the length of streamingCfg.TmpDir is 0 , would be better to handle that case as well?

common/utils/utils.go
```diff
@@ -170,6 +170,43 @@ func PreloadGCSFiles(tables []ManifestTable) ([]ManifestTable, error) {
...
+func WriteToGCS(filePath, fileName, data string) error {
+	ctx := context.Background()
+
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		fmt.Printf("Failed to create GCS client")
+		return err
+	}
+	defer client.Close()
+	if filePath[len(filePath)-1] != '/' {
```
shreyakhajanchi, line 182: This could break if the length of filePath is 0 , would be better to handle that case as well?

common/utils/utils.go
```diff
@@ -170,6 +170,43 @@ func PreloadGCSFiles(tables []ManifestTable) ([]ManifestTable, error) {
...
+		filePath = filePath + "/"
+	}
+	u, err := url.Parse(filePath)
+	if err != nil {
+		fmt.Printf("parseFilePath: unable to parse file path %s", filePath)
+		return err
+	}
+	if u.Scheme != "gs" {
```
shreyakhajanchi, line 190: Better to define `gs` as constant and use here

conversion/conversion.go
```diff
@@ -187,6 +187,18 @@ func performSnapshotMigration(config writer.BatchWriterConfig, conv *internal.Co
 	return batchWriter, nil
```
shreyakhajanchi, line 187: Why does this method have err in return params but only returns nil? Am I missing something here ?

streaming/streaming.go
```diff
@@ -100,11 +104,37 @@ func VerifyAndUpdateCfg(streamingCfg *StreamingCfg, dbName string) error {
...
+		filePath = filePath + "/"
+		streamingCfg.TmpDir = filePath
+	}
+	u, err := url.Parse(filePath)
+	if err != nil {
+		return fmt.Errorf("parseFilePath: unable to parse file path %s", filePath)
+	}
+	if u.Scheme != "gs" {
+		return fmt.Errorf("not a valid GCS path: %s, should start with 'gs'", filePath)
+	}
```
shreyakhajanchi, line 125: Most of the code here seems duplicated with utils.go would it make sense to move it into a method?

streaming/streaming.go
```diff
@@ -265,6 +297,11 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
 			"streamName":       fmt.Sprintf("projects/%s/locations/%s/streams/%s", project, datastreamCfg.StreamLocation, datastreamCfg.StreamId),
 			"instanceId":       instance,
 			"databaseId":       dbName,
+			"sessionFilePath":  streamingCfg.TmpDir + "session.json",
+		},
+		Environment: &dataflowpb.FlexTemplateRuntimeEnvironment{
+			NumWorkers: 60,
+			MaxWorkers: 60,
```
shreyakhajanchi, line 304: Better to define these as constants as use

shreyakhajanchi (commented):
Overall looks good, left few minor comments.

streaming/streaming.go
Deep1998: Good catch, fixed!

common/utils/utils.go
Deep1998: Done

streaming/streaming.go
Deep1998: Agreed, and eventually we should allow this parameters to be taken as input from the UI as well.

conversion/conversion.go
Deep1998: Looks like a refactoring we missed [here](https://github.com/cloudspannerecosystem/harbourbridge/pull/323) in this PR, we can remove the error being returned.

Deep1998: Fixed

shreyakhajanchi (approved):
LGTM

PR #368, opened 2022-08-25
Use String(50) type instead of INT64 for synthetic PK.

Deep1998:
We plan on using UUIDs in the Datastream to Spanner template for synthetic PK generation. However, UUIDs cannot fit in an INT64 hence we start using STRING(50) type for synthetic PK columns generated via HarbourBridge.


sources/mysql/data.go
```diff
@@ -84,7 +84,7 @@ func ConvertData(conv *internal.Conv, srcTable string, srcCols []string, srcSche
 	}
 	if aux, ok := conv.SyntheticPKeys[spTable]; ok {
 		c = append(c, aux.Col)
-		v = append(v, int64(bits.Reverse64(uint64(aux.Sequence))))
+		v = append(v, fmt.Sprintf("%d", int64(bits.Reverse64(uint64(aux.Sequence)))))
```
manitgupta, line 87: We seem to be bit reversing and casting back to `int64`. From the PR description it seemed that we are moving from `int64 -> string(50)`. Is this right?

Deep1998: So the fmt.Sprintf() part is what casts the int64 as a string. Alternatives include the strconv library but this works too.

manitgupta (approved):
LGTM

PR #370, opened 2022-08-29
Fix case sensitivity issue with Spangres target

manitgupta:
Fixes case sensitivity issue with Spangres migration. PostgreSQL is case sensitive if double quotes are added around identifiers. This was the default behaviour.

Read up about case sensitivity in Postgres:

1. https://stackoverflow.com/questions/7651417/escaping-keyword-like-column-names-in-postgres
2. https://dev.to/lefebvre/dont-get-bit-by-postgresql-case-sensitivity--457i

In order to protect Ids for transformation, we add backticks for Spanner target, and quotes for Spangres target. Adding quotes results in the schema in the comment above, and makes the CREATE TABLE and other related DDL statements case-sensitive.

As per the PostgreSQL docs, I did not find a direct equivalent for backticks. It looks like PostgreSQL allows creation of table names, columns with reserved keywords and it is during querying where it requires special handling of these keywords.

### Testing details:
1. Verified that a MySQL Db with created with `camelCase` fields does not preserve case sensitivity while migrating to Spangres.
2. Verified foreign key constraints are added properly.

### DIFF
--- a/spanner/ddl/ast.go
+++ b/spanner/ddl/ast.go
@@ -72,8 +72,9 @@ const (
 )
 
 // Type represents the type of a column.
-//     type:
-//        { BOOL | INT64 | FLOAT64 | STRING( length ) | BYTES( length ) | DATE | TIMESTAMP | NUMERIC }
+//
+//	type:
+//	   { BOOL | INT64 | FLOAT64 | STRING( length ) | BYTES( length ) | DATE | TIMESTAMP | NUMERIC }
 type Type struct {
 	Name string
 	// Len encodes the following Spanner DDL definition:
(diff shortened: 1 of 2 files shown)


manitgupta:
Adding test files:

[mysqldump_2022-08-29_1882-7db3.report.txt](https://github.com/cloudspannerecosystem/harbourbridge/files/9442470/mysqldump_2022-08-29_1882-7db3.report.txt)
[mysqldump_2022-08-29_1882-7db3.schema.ddl.txt](https://github.com/cloudspannerecosystem/harbourbridge/files/9442471/mysqldump_2022-08-29_1882-7db3.schema.ddl.txt)
[mysqldump_2022-08-29_188
[Dump20220829-fk.sql.zip](https://github.com/cloudspannerecosystem/harbourbridge/files/9442477/Dump20220829-fk.sql.zip)
2-7db3.schema.txt](https://github.com/cloudspannerecosystem/harbourbridge/files/9442472/mysqldump_2022-08-29_1882-7db3.schema.txt)

shreyakhajanchi (approved):
LGTM

PR #378, opened 2022-09-23
Create connection profile via UI

shreyakhajanchi:
Includes changes for:

1. Creation of connection profiles from HarbourBridge UI
2. Populating migartion request id and other details for migrations via spanner
3. Creation of streaming config json file

P.S. : Please ignore the number of commits, they are due to multiple rebasing


ui/src/app/components/connection-profile-form/connection-profile-form.component.scss
```diff
@@ -0,0 +1,18 @@
+.icon {
```
bharadwaj-aditya, line 1: check what can be moved to the central CSS

ui/src/app/components/prepare-migration/prepare-migration.component.ts
```diff
@@ -133,6 +150,30 @@ export class PrepareMigrationComponent implements OnInit {
     localStorage.removeItem(MigrationDetails.DataProgressMessage)
     localStorage.removeItem(MigrationDetails.SchemaProgressMessage)
   }
+  openConnectionProfileForm(isSource: boolean) {
+    let dialogRef = this.dialog.open(ConnectionProfileFormComponent, {
+      width: '30vw',
```
bharadwaj-aditya, line 155: Try to avoid and use common components defined in the styles/common CSS

ui/src/app/components/target-details-form/target-details-form.component.ts
```diff
@@ -1,15 +1,50 @@
...
-
+import { MigrationDetails, TargetDetails } from 'src/app/app.constants'
 @Component({
   selector: 'app-target-details-form',
   templateUrl: './target-details-form.component.html',
   styleUrls: ['./target-details-form.component.scss'],
 })
 export class TargetDetailsFormComponent implements OnInit {
   targetDetailsForm: FormGroup
+  regionList = [
```
bharadwaj-aditya, line 12: Can we get this from an API ?

webv2/profile/profile.go
```diff
@@ -0,0 +1,201 @@
...
+	for {
+		resp, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			http.Error(w, fmt.Sprintf("Error while getting list of connection profiles: %v", err), http.StatusBadRequest)
+			return
+		}
+		if source && databaseType == constants.MYSQL && resp.GetMysqlProfile().GetHostname() != "" {
```
bharadwaj-aditya, line 66: check for null pointers

bharadwaj-aditya (approved):
LGTM with some minor comments

webv2/profile/profile.go
shreyakhajanchi: in case parent is null it returns ""

ui/src/app/components/target-details-form/target-details-form.component.ts
shreyakhajanchi: in locations package the API is only available for REST and it requires auth token. Datastream has GetLocations API but it doesn't return the response expected. Will circle back on this later and see if the implementation is done for go locations API

ui/src/app/components/prepare-migration/prepare-migration.component.ts
shreyakhajanchi: will move the CSS bits in follow up PR

PR #379, opened 2022-09-26
Handle mysql booleans for direct connect

Deep1998:
The current code does not handle mysql bools correctly. MySQL stores bools as tinyint(1). This numeric length parameter is populated by the CHARACTER_MAXIMUM_LENGTH column in the infoschema query for other types. However for tinyints, this shows up as null and has to be instead parsed from the COLUMN_TYPE field.
```
+-------------+-----------+-------------+-------------+----------------+--------------------------+-------------------+---------------+-------+
| COLUMN_NAME | DATA_TYPE | COLUMN_TYPE | IS_NULLABLE | COLUMN_DEFAULT | CHARACTER_MAXIMUM_LENGTH | NUMERIC_PRECISION | NUMERIC_SCALE | EXTRA |
+-------------+-----------+-------------+-------------+----------------+--------------------------+-------------------+---------------+-------+
| col         | tinyint   | tinyint     | YES         | NULL           |                     NULL |                 3 |             0 |       |
+-------------+-----------+-------------+-------------+----------------+--------------------------+-------------------+---------------+-------+
```


shreyakhajanchi (approved):
LGTM

sources/mysql/infoschema.go
```diff
@@ -383,6 +383,13 @@ func toType(dataType string, columnType string, charLen sql.NullInt64, numericPr
 		return schema.Type{Name: dataType, Mods: []int64{numericPrecision.Int64, numericScale.Int64}}
 	case dataType == "decimal" && numericPrecision.Valid:
 		return schema.Type{Name: dataType, Mods: []int64{numericPrecision.Int64}}
+	case dataType == "tinyint" && len(columnType) > 7:
```
manitgupta, line 386: What is the significance of `> 7`?
Seems like a magic number. Please refactor it into a constant with appropriate naming.

PR #385, opened 2022-10-10
create bucket and generate resources

shreyakhajanchi:
Added changes to create a default bucket for migration and use it for connection profile, placing session file, etc. This PR also has code changes to display generated resources at the end of migration. These resources include spanner database, gcs bucket created and dataflow and datastream jobs.

### DIFF
--- a/common/utils/utils.go
+++ b/common/utils/utils.go
@@ -237,6 +237,7 @@ func CreateGCSBucket(bucketName, projectID string) error {
 			if err := bucket.Create(ctx, projectID, nil); err != nil {
 				return fmt.Errorf("failed to create bucket: %v", err)
 			}
+			fmt.Printf("Created new GCS bucket: %v\n", bucketName)
 			return nil
 		}
 		if err != nil {
--- a/common/utils/utils.go
+++ b/common/utils/utils.go
@@ -216,6 +216,38 @@ func WriteToGCS(filePath, fileName, data string) error {
 	return nil
 }
 
+func CreateGCSBucket(bucketName, projectID string) error {
+	ctx := context.Background()
+
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create GCS client: %v", err)
+	}
+	defer client.Close()
+	bucket := client.Bucket(bucketName)
+	buckets := client.Buckets(ctx, projectID)
+	for {
+		if bucketName == "" {
+			return fmt.Errorf("bucketName entered is empty %v", bucketName)
+		}
+		attrs, err := buckets.Next()
+		// Assume bucket not found if at Iterator end and create
+		if err == iterator.Done {
+			// Create bucket
+			if err := bucket.Create(ctx, projectID, nil); err != nil {
+				return fmt.Errorf("failed to create bucket: %v", err)
+			}
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("issues setting up Bucket(%q).Objects(): %v. Double check project id", attrs.Name, err)
+		}
+		if attrs.Name == bucketName {
+			return nil
+		}
+	}
+}
+
 // GetProject returns the cloud project we should use for accessing Spanner.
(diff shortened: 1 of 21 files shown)


common/utils/utils.go
```diff
@@ -216,6 +216,38 @@ func WriteToGCS(filePath, fileName, data string) error {
...
+	buckets := client.Buckets(ctx, projectID)
+	for {
+		if bucketName == "" {
+			return fmt.Errorf("bucketName entered is empty %v", bucketName)
+		}
+		attrs, err := buckets.Next()
+		// Assume bucket not found if at Iterator end and create
+		if err == iterator.Done {
+			// Create bucket
+			if err := bucket.Create(ctx, projectID, nil); err != nil {
```
bharadwaj-aditya, line 237: add in the logs that bucket created

ui/src/app/components/prepare-migration/prepare-migration.component.html
```diff
@@ -115,6 +115,19 @@ <h3>Target Details</h3>
         <mat-progress-bar mode="determinate" [value]="dataMigrationProgress"></mat-progress-bar>
         <span> {{this.dataProgressMessage}}</span>
     </div>
+    <div *ngIf="isResourceGenerated">
+        <h3> Generated Resources:</h3>
+        <li>Spanner database: <a [href]="resourcesGenerated.DatabaseUrl">{{resourcesGenerated.DatabaseName}}</a></li>
+        <li>GCS bucket: <a [href]="resourcesGenerated.BucketUrl">{{resourcesGenerated.BucketName}}</a></li>
```
bharadwaj-aditya, line 121: nit: BucketName maps to BucketUrl ?
Same for all other fields. If possible use the same names ?

ui/src/app/components/prepare-migration/prepare-migration.component.ts
```diff
@@ -237,10 +249,19 @@ export class PrepareMigrationComponent implements OnInit {
...
                 this.markMigrationComplete()
+              } else if (this.selectedMigrationType == MigrationTypes.lowDowntimeMigration && !displayStreamingMsg) {
+                this.markSchemaMigrationComplete()
+                this.hasDataMigrationStarted = false
+                localStorage.setItem(MigrationDetails.HasDataMigrationStarted, this.hasDataMigrationStarted.toString())
+                this.snack.openSnackBarWithoutTimeout('Setting up dataflow and datastream jobs', 'Close')
+                displayStreamingMsg = true
               }
             }
+            else if (res.Message.startsWith('Data migration complete')) {
```
bharadwaj-aditya, line 262: Performing logic like this on the basis of text is a bad idea. It is always better to pass a status field and text separately and use the status field for these checks, and text to just be shown.

webv2/profile/profile.go
```diff
@@ -179,7 +190,7 @@ func setConnectionProfile(isSource bool, sessionState session.SessionState, req
 	} else {
 		req.ConnectionProfile.Profile = &datastreampb.ConnectionProfile_GcsProfile{
 			GcsProfile: &datastreampb.GcsProfile{
-				Bucket:   bucket,
+				Bucket:   strings.ToLower(sessionState.Conv.Audit.MigrationRequestId),
```
bharadwaj-aditya, line 193: the bucket name is set as the request id ? Is this the case even if the name is passed ?

bharadwaj-aditya (changes requested):

ui/src/app/components/prepare-migration/prepare-migration.component.html
shreyakhajanchi: That's because in UI it will show name of these elements with links which can be used to go the asset

ui/src/app/components/prepare-migration/prepare-migration.component.ts
shreyakhajanchi: okay will add a status field in the progress object

PR #388, opened 2022-10-19
Add harbourbridge web command

manitgupta:
This PR adds `./harbourbridge web` CLI subcommand which would run the webV2 UI. 
This is done by committing the compiled `ng build` artifacts into the source and then embedding them within the binary using `embed` library in Go.
This ensures that the harbourbridge binary can be distributed as a stand-alone binary without reliance on the source.

A reference to read more about this -
https://shibumi.dev/posts/go-embed-and-angular/


webv2/web.go
```diff
@@ -1657,8 +1657,8 @@ func init() {
 
 // App connects to the web app v2.
 func App() {
-	addr := ":8080"
+	addr := ":4200"
```
bharadwaj-aditya, line 1660: This might break existing functionality of the UI running at 8080. We should check if there is anywhere else where 8080 is assumed

bharadwaj-aditya (changes requested):
Please include changes to documentation as required around build and running process

webv2/web.go
manitgupta: Previously:
1. webv2 UI used to run on 4200 port only since the UI was served by the `ng serve` (Angular Development server).
2. HB http server used to run on port 8080.

Now:
1. The HTTP webserver now serves the UI as well. S

Since the users are _used to_ opening localhost:4200 to open the webV2 UI, I have kept the experience same.


> Please include changes to documentation as required around build and running process

I have currently added comments on how to build here (Line 52) -
https://github.com/cloudspannerecosystem/harbourbridge/pull/388/files#diff-7d422311b9d1565f2d86bf865102cb1fe0531a3e6b62aa6a67fdffc64b173d5a

If you have a different place in mind, I can move it there.

ui/README.md
```diff
@@ -2,6 +2,18 @@
...
 
+## Updating the UI
+
+The `dist/ui` directory is mounted inside the harbourbridge binary using the `go embed` library.
+After making any changes to the UI, perform the following steps to update Harbourbridge with the UI changes:
+
+1. Run `ng build` inside the `ui/` directory. This places the generated angular artifacts in the `ui/dist/ui` directory.
+2. Run `make build` from the root directory to generate the harbourbridge binary.
+3. Run `./harbourbridge web` to run the web UI.
+4. Navigate to `http://localhost:4200/` to access the UI.
```
bharadwaj-aditya, line 13: Will this continue to be 4200 with the port change ?
If not, lets pass through the code to see any other references to the port and remove it

manitgupta: Thanks for catching this! Forgot to update it, will change.

bharadwaj-aditya (approved):
LGTM

PR #394, opened 2022-11-02
Feature - Restore foreign key

lyazii22:
- [ ] Remove interleave and convert back to foreign key implementation.


ui/src/app/components/object-detail/object-detail.component.html
```diff
@@ -753,7 +742,7 @@ <h3 class="title">
           "
           mat-raised-button
           color="primary"
```
shreyakhajanchi, line 744: Modify the SQL in SQL tab after interleaving property is removed

ui/src/app/components/object-detail/object-detail.component.html
```diff
@@ -700,17 +700,6 @@ <h3 class="title">
...
-                      interleaveStatus.tableInterleaveStatus &&
-                      !interleaveStatus.tableInterleaveStatus.Possible
-                    "
-                    [disabled]="element.get('spName').value === ''"
-                    mat-menu-item
-                    (click)="convertToFk()"
-                  >
-                    <span>Convert to Foreign Key</span>
-                  </button>
                 </mat-menu>
```
shreyakhajanchi, line 703: In case of two foreign keys, after interleaving one the second foreign key appears non adjacent to the source one: https://drive.google.com/file/d/19YkXKfZCM-BLLh_EKM-GXsK3z9astJ52/view?usp=share_link&resourcekey=0-oQLq3JDk0YtDSc8dSXz1Lg

ui/src/styles.scss
```diff
@@ -53,13 +53,18 @@ body {
...
 
 .success {
   color: green;
 }
 
+.suggestion {
+  color: #f6be00;
+  transform: rotate(180deg);
+}
+
```
shreyakhajanchi, line 67: these css changes do not seem to be part of this PR same for restore index PR

Ronak-Searce: Updated

ui/src/app/components/object-detail/object-detail.component.html
Ronak-Searce: Updated

shreyakhajanchi (approved):
LGTM

PR #396, opened 2022-11-04
Use filepath.Join to concatinate files and paths

michity15169:
* files and paths concatenation with filepath.Join is more portable rather than with '+'

* os.TempDir() returns '/tmp' if TMPDIR is not defilned(https://github.com/golang/go/blob/master/src/os/file_unix.go#L339-L346). But, the current implementation expected the path to end with a '/'

- [x] Tests pass

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


manitgupta:
Thanks for submitting the PR @michity15169. If it's not too much hassle, can you also make the `filePath.join` change to other files as well with usages of `+`?
It shouldn't take too much and help us maintain consistency. A quick search for `os.TempDir()` showed 12 results.

michity15169:
Hi @manitgupta, Your point is make sense.
Similar changes for other parts using os.TempDir()

manitgupta (approved):

PR #397, opened 2022-11-07
Phase2 bughunt v5

mayank2913:
> Fixed Bugs 

- [x]  Still getting notes for the name must adhere to certain regex although we had removed all the notes.
- [x]  Suggestion icon bulb appears to be black and inverted.
- [x] Save session should have the Editor name as a compulsory field.
- [x]  Save Session - Database name should be auto-populated.
- [x] Unnecessary Horizontal Scroll on Object ExplorerTab Removed


ui/src/app/components/object-explorer/object-explorer.component.scss
```diff
@@ -38,9 +38,12 @@
 .selected-row {
   background-color: #61b3ff4a;
 }
+
 .explorer-table {
   height: calc(100vh - 320px);
   overflow-y: auto;
+  overflow-x: hidden;
```
shreyakhajanchi, line 45: this should not be hidden , scroll bar should not appear if contents do not overflow

mayank2913: Updated

shreyakhajanchi (approved):
LGTM

PR #405, opened 2022-11-10
Phase2 bughunt v6

Ronak-Searce:
- [ ] Empty block of space is shown upon editing the primary key after deleting a column from the table.(53)
- [ ] Save session form doesn't show Notes as compulsary field but on saving shows error.(66)
- [x] Add column not working in the columns tab(89).
- [x] Getting convert to foreign key option while editing foreign key.(93)


ui/src/app/components/object-detail/object-detail.component.html
```diff
@@ -689,7 +689,7 @@ <h3 class="title">
                   >
                     <span>Drop Foreign Key</span>
                   </button>
-                  <button
+                  <!-- <button
```
shreyakhajanchi, line 692: Why has this been commented if its intended to be removed?

ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -507,9 +511,13 @@ export class ObjectDetailComponent implements OnInit {
...
     })
     for (let i = 0; i < this.localTableData.length; i++) {
-      if (!currentPkColumns.includes(this.localTableData[i].spColName))
+      if (
+        !currentPkColumns.includes(this.localTableData[i].spColName) &&
+        this.localTableData[i].spColName !== ''
+      )
         this.pkColumnNames.push(this.localTableData[i].spColName)
     }
+    console.log(this.pkColumnNames)
```
shreyakhajanchi, line 520: remove the log

ui/src/app/components/object-detail/object-detail.component.html
shreyakhajanchi, line 692: Dropping a fk is not waiting for the user to click on save and dropping it instantly

ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -319,16 +323,16 @@ export class ObjectDetailComponent implements OnInit {
...
       .map((item) => item.srcColName)
       .indexOf(this.addedColumnName)
-    this.tableData[index].spColName = this.droppedColumns[addedRowIndex].spColName
-    this.tableData[index].spDataType = this.droppedColumns[addedRowIndex].spDataType
-    this.tableData[index].spOrder = -1
-    this.tableData[index].spIsPk = this.droppedColumns[addedRowIndex].spIsPk
-    this.tableData[index].spIsNotNull = this.droppedColumns[addedRowIndex].spIsNotNull
+    this.localTableData[index].spColName = this.droppedColumns[addedRowIndex].spColName
+    this.localTableData[index].spDataType = this.droppedColumns[addedRowIndex].spDataType
+    this.localTableData[index].spOrder = 3
```
shreyakhajanchi, line 333: why is the order being assigned as 3?

ui/src/app/components/object-detail/object-detail.component.ts
Ronak-Searce: Updated

ui/src/app/components/object-detail/object-detail.component.html
Ronak-Searce: Updated

ui/src/app/components/object-detail/object-detail.component.ts
Ronak-Searce: Updated

shreyakhajanchi (approved):
LGTM

PR #410, opened 2022-11-15
Phase2 bughunt v7

mayank2913:
- [x] JSON edits not working properly fixed
- [x] Data type for synth_id being shown as blank in column edit mode
- [x] Downloaded file name after save session and review changes fixed
- [x] Deletion of the primary key column from the columns tab in the table should remove the primary key 
- [x] Issue with Interleave Fixed
- [x] Two columns can be renamed to the same column and saved. This should not be allowed.
- [x] UI is allowing two indexes to be created with same name.


spanner/ddl/ast.go
```diff
@@ -292,6 +292,9 @@ func (ct CreateTable) PrintCreateTable(config Config) string {
 		}
 	}
 
+	if len(keys) == 0 {
+		return fmt.Sprintf("%sCREATE TABLE %s (\n%s) %s", tableComment, config.quote(ct.Name), cols, interleave)
```
shreyakhajanchi, line 296: getting syntax error due to extra `,` : Error parsing Spanner DDL statement: CREATE TABLE `test123` ( `name` STRING(32), `age` INT64, ) : Syntax error on line 4, column 2: Expecting 'PRIMARY' but found 'EOF'

webv2/typemap/mysql_typemap.go
```diff
@@ -125,10 +125,10 @@ func ToSpannerTypeMySQL(srcType string, spType string, mods []int64) (ddl.Type,
 		return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, nil
 	case "json":
 		switch spType {
-		case ddl.Bytes:
```
shreyakhajanchi, line 128: why are we removing this byte case?

webv2/typemap/postgres_typemap.go
```diff
@@ -152,6 +152,13 @@ func ToSpannerTypePostgres(srcType string, spType string, mods []int64) (ddl.Typ
 		default:
 			return ddl.Type{Name: ddl.Timestamp}, []internal.SchemaIssue{internal.Timestamp}
 		}
+	case "json":
```
shreyakhajanchi, line 155: add jsonb as well as in postgres/toddl.go

webv2/typemap/sqlserver_typemap.go
```diff
@@ -81,6 +81,13 @@ func ToSpannerTypeSQLserver(srcType string, spType string, mods []int64) (ddl.Ty
 			}
 			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, nil
 		}
+	case "json":
```
shreyakhajanchi, line 84: json datatype is not present in sql server

webv2/session/session_handler.go
```diff
@@ -230,6 +230,14 @@ func SaveRemoteSession(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	sessionMetaData := GetSessionState().SessionMetadata
+	
+	sessionMetaData.DatabaseName = sm.DatabaseName
+	sessionMetaData.DatabaseType = sm.DatabaseType
+	sessionMetaData.SessionName = sm.SessionName
+
+	GetSessionState().SessionMetadata = sessionMetaData
```
manitgupta, line 239: This looks weird 🤔
Why create a the temp variable `sessionMetaData`?
You can just do `GetSessionState().SessionMetadata.DatabaseName = sm.DatabaseName`

webv2/typemap/postgres_typemap.go
Ronak-Searce: Updated

webv2/typemap/sqlserver_typemap.go
Ronak-Searce: Updated

webv2/typemap/mysql_typemap.go
mayank2913: Updated

ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -314,6 +314,7 @@ export class ObjectDetailComponent implements OnInit {
             data: { message: res, type: 'error' },
             maxWidth: '500px',
           })
+          this.isEditMode = true
```
shreyakhajanchi, line 317: Interleaving suggestion doesn't go away if you delete pk column and then fk : https://drive.google.com/file/d/1KZP9ow_RFnGlTl8K2Mq7EK9_RpzJzRqw/view?usp=share_link&resourcekey=0-I4iwAqWiZynbMp3FMOhLsA

ui/src/app/components/object-detail/object-detail.component.html
```diff
@@ -225,7 +225,7 @@ <h3 class="title">
             <ng-container matColumnDef="spColName">
               <th mat-header-cell class="table_header" *matHeaderCellDef>Name</th>
               <td mat-cell *matCellDef="let element">
-                <div *ngIf="isEditMode && element.get('spOrder').value !== ''">
```
shreyakhajanchi, line 228: Still getting undefined_undefined_undefined.json after saving session then resuming and then interleaving the table

ui/src/app/components/object-detail/object-detail.component.ts
Ronak-Searce: Updated

ui/src/app/components/object-detail/object-detail.component.html
Ronak-Searce: Updated

shreyakhajanchi (commented):
LGTM

shreyakhajanchi (approved):

PR #414, opened 2022-11-22
Enable autoscaling for dataflow jobs used in streaming migrations.

Deep1998:


streaming/streaming.go
```diff
@@ -345,8 +344,9 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
 			"deadLetterQueueDirectory": inputFilePattern + "dlq",
 		},
 		Environment: &dataflowpb.FlexTemplateRuntimeEnvironment{
-			NumWorkers: numWorkers,
-			MaxWorkers: maxWorkers,
+			MaxWorkers:            maxWorkers,
+			AutoscalingAlgorithm:  2, // 2 corresponds to AUTOSCALING_ALGORITHM_BASIC
```
shreyakhajanchi, line 348: What will be the initial number of workers in this case ?

Deep1998: It starts with 1, then every 10 minutes, the autoscaler re-scales.

shreyakhajanchi (approved):

PR #417, opened 2022-11-24
DLQ retry limit increase

Deep1998:
Update the retry limit to 500 and retry time to 10 minutes to have a larger window for FK/Interleaved table failures.


manitgupta:
can you please add some testing details?

Deep1998:
Updated the defaults on the template directly.

PR #421, opened 2022-11-29
Postgres lowdowntime migration support using HarbourBridge

shreyakhajanchi:
1. Added support for postgres low downtime migration
2. Modified the datastream package because some options were deprecated in older one
3. Added UI options for specifying replication slot and publication

### DIFF
--- a/streaming/streaming.go
+++ b/streaming/streaming.go
@@ -176,23 +176,38 @@ func getOracleSourceStreamConfig(dbName string) *datastreampb.SourceConfig_Oracl
 	return &datastreampb.SourceConfig_OracleSourceConfig{OracleSourceConfig: oracleSrcCfg}
 }
 
-func getPostgreSQLSourceStreamConfig(dbName, properties string) (*datastreampb.SourceConfig_PostgresqlSourceConfig, error) {
-	postgresdb := &datastreampb.PostgresqlSchema{
-		Schema: dbName,
-	}
+func getPostgreSQLSourceStreamConfig(properties string) (*datastreampb.SourceConfig_PostgresqlSourceConfig, error) {
 	params, err := profiles.ParseMap(properties)
 	if err != nil {
 		return nil, fmt.Errorf("could not parse properties: %v", err)
 	}
-	_, replicationSlotExists := params["replicationSlot"]
-	_, publicationExists := params["publication"]
+	infoschemadb := &datastreampb.PostgresqlSchema{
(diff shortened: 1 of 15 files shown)


streaming/streaming.go
```diff
@@ -157,25 +158,61 @@ func ReadStreamingConfig(file, dbName string) (StreamingCfg, error) {
...
+		Schema: "pg_toast",
+	}
+	postgres := &datastreampb.PostgresqlSchema{
+		Schema: "postgres",
+	}
+	pg_temp_1 := &datastreampb.PostgresqlSchema{
+		Schema: "pg_temp_1",
+	}
+	pg_toast_temp_1 := &datastreampb.PostgresqlSchema{
+		Schema: "pg_toast_temp_1",
```
Deep1998, line 200: This bit can be refactored as a function that loops over the schema names and returns a []*datastreampb.PostgresqlSchema object. Then we simply call that instead of creating multiple objects.

Deep1998:
Can you also add the flow you tested?

shreyakhajanchi:
> 

Added test cases which were verified

streaming/streaming.go
```diff
@@ -22,9 +22,9 @@ import (
 	"time"
 
 	dataflow "cloud.google.com/go/dataflow/apiv1beta3"
-	datastream "cloud.google.com/go/datastream/apiv1alpha1"
```
bharadwaj-aditya, line 25: does this library change affect other streaming migrations also ?

streaming/streaming.go
```diff
@@ -157,25 +158,61 @@ func ReadStreamingConfig(file, dbName string) (StreamingCfg, error) {
 
 func getMysqlSourceStreamConfig(dbName string) *datastreampb.SourceConfig_MysqlSourceConfig {
 	mydb := &datastreampb.MysqlDatabase{
-		DatabaseName: dbName,
+		Database: dbName,
```
bharadwaj-aditya, line 161: Why is this renamed ?

bharadwaj-aditya (approved):
LGTM with minor comments

streaming/streaming.go
shreyakhajanchi: yes same library is used across all the streaming migrations, the older library didn't have postgres source support

streaming/streaming.go
shreyakhajanchi: Because the updated library has these fields renamed

PR #425, opened 2022-12-02
Phase2 bughunt v9

Ronak-Searce:
- [ ] Use typemap from source/*/toddl.go instead of defining them again in webv2/typemap , it is increasing the places where changes need to be made for a datatype.
- [x] 1 critical severity vulnerability - angular while running npm install.
- [x] After converting foreign key to interleaving, The interleaved table tab just contains button Convert to Foreign key. It should also contain info that this table is interleaved with "parent table”.
- [x] "Some columns have source DB type 'datetime' which is mapped to Spanner type timestamp e.g. column 'CurrencyRateDate'. Spanner timestamp is closer to MySQL timestamp" this should be a warning and not a suggestion.
- [x] Interleaved table suggestion -- """Candidate for Interleaved Table rename director_id primary key in table directors_genres to match the foreign key directors_genres_ibfk_1 refer column id """  parent column name should be quoted.
- [x] Suggestion for interleaving should specify the order of the primary key that needs to be edited.
- [x] Suggestion for interleaving should have the table in which the table will get interleaved.
- [x] Include name of the parent table in all suggestions related to interleaving to avoid confusion in case more than one interleaving options are available for a table.
- [x] There are a lot of variables in local storage for UI , its better to clear those on server start to avoid any unforseen issues.
- [x] Drop foreign key doesn't wait for the user to click on save & convert and applies the changes directly.


ui/dist/ui/index.html
```diff
@@ -7,7 +7,7 @@
...
-  <style>.mat-typography{font:400 14px/20px Roboto,Helvetica Neue,sans-serif;letter-spacing:normal}html,body{height:100%;box-sizing:border-box}body{margin:0;font-family:Roboto,Helvetica Neue,sans-serif}</style><link rel="stylesheet" href="styles.0e4dc0b6086788f1.css" media="print" onload="this.media='all'"><noscript><link rel="stylesheet" href="styles.0e4dc0b6086788f1.css"></noscript></head>
+  <style>.mat-typography{font:400 14px/20px Roboto,Helvetica Neue,sans-serif;letter-spacing:normal}html,body{height:100%;box-sizing:border-box}body{margin:0;font-family:Roboto,Helvetica Neue,sans-serif}</style><link rel="stylesheet" href="styles.7c1f85f49aa7680f.css" media="print" onload="this.media='all'"><noscript><link rel="stylesheet" href="styles.7c1f85f49aa7680f.css"></noscript></head>
```
shreyakhajanchi, line 10: We need to add the new generated styles. **.css & polyfill.**.js files as well otherwise harbourbridge web won't work

internal/report.go
```diff
@@ -515,17 +544,17 @@ var IssueDB = map[SchemaIssue]struct {
...
 	StringOverflow:          {Brief: "String overflow issue might occur as maximum supported length in Spanner is 2621440", severity: warning},
 	HotspotTimestamp:        {Brief: "Timestamp Hotspot Occured", severity: warning},
 	HotspotAutoIncrement:    {Brief: "Autoincrement Hotspot Occured", severity: warning},
-	InterleavedOrder:        {Brief: "can be converted as Interleaved Table", severity: suggestion},
+	InterleavedOrder:        {Brief: "can be converted as Interleaved with Table", severity: suggestion},
 	RedundantIndex:          {Brief: "Redundant Index", severity: warning},
 	AutoIncrementIndex:      {Brief: "Auto increment column in Index can create a Hotspot", severity: warning},
 	InterleaveIndex:         {Brief: "can be converted to an Interleave Index", severity: suggestion},
-	InterleavedNotInOrder:   {Brief: "Can be converted to interleaved table if primary key order parameter is changed for the table", severity: suggestion},
+	InterleavedNotInOrder:   {Brief: "Can be converted to interleaved table if primary key order parameter is changed to 1 for the table", severity: suggestion},
```
shreyakhajanchi, line 557: added parent interleave table in this case as well

ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -748,16 +795,18 @@ export class ObjectDetailComponent implements OnInit {
...
 
   saveFk() {
-    let updatedFkNames: Record<string, string> = {}
-
-    this.fkArray.value.forEach((fk: IFkTabData, i: number) => {
-      let oldFk = this.fkData[i]
-      if (oldFk.spName !== fk.spName) {
-        updatedFkNames[oldFk.spName] = fk.spName
-      }
+    let spFkArr: IForeignKey[] = []
```
shreyakhajanchi, line 798: getting following error on trying to delete multiple foreign keys: Found duplicate names in input : ,

ui/src/app/app.component.ts
```diff
@@ -12,6 +12,7 @@ export class AppComponent {
   sidenavComponent: string = ''
   constructor(private sidenavService: SidenavService) {}
   ngOnInit(): void {
+    localStorage.clear()
```
shreyakhajanchi, line 15: few localstorage elements are being used to store a state in case user refreshes or opens a new tab so we cannot clear it from here

internal/report.go
Ronak-Searce: Fixed

ui/dist/ui/index.html
Ronak-Searce: Fixed

internal/report.go
shreyakhajanchi: add for InterleavedAddColumn case as well

ui/dist/ui/index.html
shreyakhajanchi: getting different files on running ng build

internal/report.go
Ronak-Searce: Fixed

shreyakhajanchi (approved):
LGTM

PR #430, opened 2022-12-09
Add postgres data type limitation

Deep1998:


manitgupta (approved):
Thanks for taking up the upgrade!

PR #434, opened 2022-12-13
using older dataflow template version

shreyakhajanchi:
- Modified dataflow template path to point to older release
- Remove postgreSQL minimal downtime migration from UI


bharadwaj-aditya (approved):
LGTM

PR #436, opened 2022-12-16
Phase2 bughunt v10

shellymathew98:
**Bug Fixes:** 

1. If a table has multiple interleaving candidates suggestion is being shown only for one.
2. Use same rate schema method from report.go instead of the one in utilities/rate_schema.go and don't use colour as category name.


webv2/web.go
```diff
@@ -777,98 +777,95 @@ func setParentTable(w http.ResponseWriter, r *http.Request) {
 }
 
 func parentTableHelper(table string, update bool) *TableInterleaveStatus {
-	tableInterleaveStatus := &TableInterleaveStatus{Possible: true}
```
shreyakhajanchi, line 780: getting same issue twice after changing pk order : https://drive.google.com/open?id=1j0Og0EnjsO2-nfaVUGZfM5SOye3e4T3A&authuser=0&resourcekey=0-WbU2fin4SYQWyLxMSfp-OA&usp=drive_link

webv2/web.go
```diff
@@ -777,98 +777,95 @@ func setParentTable(w http.ResponseWriter, r *http.Request) {
...
+	tableInterleaveStatus := &TableInterleaveStatus{
```
shreyakhajanchi, line 780: Parent table name not showing up in the 2nd interleave suggestion here: https://drive.google.com/open?id=1Tjhp98QK0G-8KmzYq_vVoLPRPyiB6VWJ&authuser=0&resourcekey=0-xGclQBkYSTRjrk4YNoyBcQ&usp=drive_link

web/web.go
```diff
@@ -1117,23 +1117,23 @@ func rateSchema(cols, warnings int64, missingPKey bool) string {
...
 	case warnings == 0 && missingPKey:
-		return "BLUE"
+		return "GOOD"
 	case good(cols, warnings) && !missingPKey:
-		return "BLUE"
+		return "GOOD"
 	case good(cols, warnings) && missingPKey:
-		return "BLUE"
+		return "GOOD"
 	case ok(cols, warnings) && !missingPKey:
```
shreyakhajanchi, line 1129: What is yellow here ? We should replace that too

ui/src/app/components/object-detail/object-detail.component.html
```diff
@@ -412,7 +412,7 @@ <h3 class="title">
           </ng-container>
 
           <ng-container matColumnDef="srcOrder">
-            <th class="table_header" mat-header-cell *matHeaderCellDef>Order</th>
+            <th class="table_header" mat-header-cell *matHeaderCellDef>S No.</th>
```
shreyakhajanchi, line 415: Primary key tab shows `S. No` for source and `order` for spanner which is wrong. It should be order for both as column orders matter in PK.

webv2/web.go
```diff
@@ -777,98 +777,95 @@ func setParentTable(w http.ResponseWriter, r *http.Request) {
...
+		Possible: false,
```
shreyakhajanchi, line 781: Two different suggestions showing same parent names: https://drive.google.com/open?id=1Bq1PHW-y9gplCByUFJdP7nS6Hh3ggcjH&authuser=0&resourcekey=0-s3FuRHQq9MyZyjku6UX9cg&usp=drive_link

shreyakhajanchi (commented):
The interleaving suggestions are not working properly in case a table has multiple interleaving candidates

web/web.go
Ronak-Searce: Yellow was associated with "OK" according to the previous implementation

Ronak-Searce: Yellow is replaced with "OK"

ui/src/app/components/object-detail/object-detail.component.html
shellymathew98: Fixed.

webv2/web.go
lyazii22: Updated. Please review it.

webv2/web.go
lyazii22: Updated. Please review it.

webv2/web.go
lyazii22: Updated. Please review it.

webv2/web.go
```diff
@@ -940,98 +940,95 @@ func setParentTable(w http.ResponseWriter, r *http.Request) {
 }
 
 func parentTableHelper(table string, update bool) *TableInterleaveStatus {
-	tableInterleaveStatus := &TableInterleaveStatus{Possible: true}
```
shreyakhajanchi, line 943: Interleave tab showing wrong parent table name and the css for Interleaved: at the top in header is also messed up: https://drive.google.com/open?id=15ObNMC8aOSjnK12Rl4uDkattqNzgLZBU&authuser=0&resourcekey=0-He1TRQDQPE2ScXqT2wwntA&usp=drive_link

shreyakhajanchi: After changing the primary key order and interleaving the parent table is not as expected

mayank2913: Updated

shreyakhajanchi (commented):
Majorly LGTM, 2 issues:
1. After changing PK order then interleaving then converting it back to foreign key is changing PK order
2. ng build files are not updated

shreyakhajanchi (approved):
LGTM

PR #439, opened 2022-12-22
Add support for JSONB data type for PG Spanner

manitgupta:
Adds support for PG.JSONB data type for PG Spanner.
All Postgres JSONB data type columns are now mapped to PG.JSONB data type instead of `character varying`.

Please note that the Postgres JSON data type, and JSON/JSONB arrays still map to `character varying`. This change only handles the `JSONB` data type.

Testing:
Command Run - 

```
./harbourbridge schema-and-data -source=postgresql --target-profile="instance=manit-testing,dialect=postgresql" --source-profile="host=localhost,port=5432,user=postgres,password=postgres@123,dbName=harbourbridge"
```

Spanner DDL:
```
CREATE TABLE test (
  id bigint NOT NULL,
  data jsonb,
  data_json character varying,
  PRIMARY KEY(id)
);
```

Source DDL:
|table_name|column_name|data_type|
|----------|-----------|---------|
|test|id|integer|
|test|data|jsonb|
|test|data_json|json|

The table contains 2 rows which get written to Spanner.

PR #445, opened 2023-01-09
Extend conversion from source JSON to PG Spanner JSONB

manitgupta:
We will map `JSON` from postgres, mysql and oracle to the JSONB data type in PGSpanner, in addition to `JSONB` which was done previously.

#### MySQL 

```
create table json_test (
id int,
col json,
primary key(id));

INSERT INTO json_test(id, col) VALUES (1, '{"pid": 101, "name": "name1"}');
```

HB command - 
```
./harbourbridge schema-and-data -source=mysql --target-profile="instance=manit-testing,dialect=postgresql" --source-profile="host=localhost,port=3306,user=root,password=mysql@123,dbName=SHARD3"
```

Result - Schema and Data migrated to Spanner. 

Dest Schema
```
CREATE TABLE json_test (
  id bigint NOT NULL,
  col jsonb,
  PRIMARY KEY(id)
);
```

#### PostgreSQL
Already captured here in #439.

### DIFF
--- a/sources/mysql/data.go
+++ b/sources/mysql/data.go
@@ -118,7 +118,7 @@ func convScalar(conv *internal.Conv, spannerType ddl.Type, srcTypeName string, T
 		return val, nil
 	case ddl.Timestamp:
 		return convTimestamp(srcTypeName, TimezoneOffset, val)
-	case ddl.JSON:
+	case ddl.JSON, ddl.JSONB:
 		return val, nil
 	default:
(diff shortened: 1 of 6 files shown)


charvisingla (approved):
LGTM

JSONB is the closest data type to map all JSON like data types from all sources(including postgres) for Spanner PG Dialect.

shreyakhajanchi (approved):
LGTM

PR #448, opened 2023-01-20
code refactoring to modify ToSpannerType

shreyakhajanchi:
Changes:
1. Currently, other PostgreSQL data types that have GSQL equivalent are treated differently from JSONB. Add similar handling for JSONB and map it to JSON column of GSQL dialect.
2. Refactored the toddl interface to have singe ToSpannerType method for both UI and CLI.

Bug link: https://b.corp.google.com/issues/265607004

Tests Run:

1. End to end bulk migration from UI for PostgreSQL database with columns of all supported data type
4. End to end minimal downtime migration from UI for MySQL database with columns of all supported data type
5. CLI schema and data migration to postgres dialect from MySQL database with columns of all supported data type. Cmd used: `go run main.go schema-and-data -source=mysql -source-profile='host=35.222.246.87,user=root,dbName=postgres_test,password=xxx' -target-profile='project=span-cloud-testing, instance=shreya-test-1, dbName=h4, dialect=postgresql'`

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -138,3 +138,11 @@ func (ty Type) Print() string {
 	}
 	return s
 }
+
+func MakeType() Type {
+	return Type{
+		Name:        "",
+		Mods:        []int64{},
+		ArrayBounds: []int64{},
+	}
+}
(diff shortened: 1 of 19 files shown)


manitgupta (changes requested):
One suggested change in the `ToSpannerType` interface, other than that rest looks good.

manitgupta (approved):

PR #449, opened 2023-01-20
Add debugger configurations for web and schema-and-data commands

manitgupta:
Adds VS code configurations for debugging harbourbridge `web` (UI) and `schema-and-data` commands.


bharadwaj-aditya (approved):
LGTM

PR #466, opened 2023-02-16
default open to false for HB UI

shreyakhajanchi:
In cases where HarbourBridge is ran on GCE VM or machines without GUI and default as true for flag --open it returns  following logs on the screen which user might interpret as HarbourBridge not working whereas it only means that system couldn't open the browser window automatically :

Harbourbridge UI started at: http://localhost:8080
[2005413:2005413:0216/100012.921483:ERROR:ozone_platform_x11.cc(238)] Missing X server or $DISPLAY
[2005413:2005413:0216/100012.921558:ERROR:env.cc(255)] The platform failed to initialize.  Exiting.

We will set the default for this flag as false to avoid the above mentioned issue.


webv2/webCmd.go
```diff
@@ -53,7 +53,7 @@ func (cmd *WebCmd) Usage() string {
 
 func (cmd *WebCmd) SetFlags(f *flag.FlagSet) {
 	f.StringVar(&cmd.logLevel, "log-level", "INFO", "Configure the logging level for the command (INFO, DEBUG), defaults to INFO")
-	f.BoolVar(&cmd.open, "open", true, "Opens the Harbourbridge web interface in the default browser, defaults to true")
+	f.BoolVar(&cmd.open, "open", false, "Opens the Harbourbridge web interface in the default browser, defaults to true")
```
manitgupta, line 56: Change description - "defaults to false"

manitgupta (approved):

PR #473, opened 2023-02-20
fixing index issue for clustered column index in sql server

shreyakhajanchi:
Testing -
1. Migrated sql server database with clustered column index
2. Migrated sql server database with row index


manitgupta (approved):
Nit - add comment next to `IX.type` and update README for SQL server

PR #486, opened 2023-03-08
Parallel schema creation

bharadwaj-aditya:
Create a parallel task processor. Used it for processing tables in parallel.

- [x] Tests pass

### DIFF
--- a/common/utils/utils.go
+++ b/common/utils/utils.go
@@ -561,7 +561,7 @@ func GetLegacyModeSupportedDrivers() []string {
 // ReadSpannerSchema fills conv by querying Spanner infoschema treating Spanner as both the source and dest.
 func ReadSpannerSchema(ctx context.Context, conv *internal.Conv, client *sp.Client) error {
 	infoSchema := spanner.InfoSchemaImpl{Client: client, Ctx: ctx, TargetDb: conv.TargetDb}
-	err := common.ProcessSchema(conv, infoSchema)
+	err := common.ProcessSchema(conv, infoSchema, 0)
 	if err != nil {
 		return fmt.Errorf("error trying to read and convert spanner schema: %v", err)
 	}
--- a/common/utils/utils.go
+++ b/common/utils/utils.go
@@ -561,7 +561,7 @@ func GetLegacyModeSupportedDrivers() []string {
 // ReadSpannerSchema fills conv by querying Spanner infoschema treating Spanner as both the source and dest.
 func ReadSpannerSchema(ctx context.Context, conv *internal.Conv, client *sp.Client) error {
 	infoSchema := spanner.InfoSchemaImpl{Client: client, Ctx: ctx, TargetDb: conv.TargetDb}
-	err := common.ProcessSchema(conv, infoSchema, 0)
+	err := common.ProcessSchema(conv, infoSchema, common.DefaultWorkers)
 	if err != nil {
 		return fmt.Errorf("error trying to read and convert spanner schema: %v", err)
 	}
--- a/sources/common/infoschema.go
+++ b/sources/common/infoschema.go
@@ -26,6 +26,8 @@ import (
 	"github.com/cloudspannerecosystem/harbourbridge/spanner/ddl"
 )
 
+const DefaultWorkers = 20 // Default to 20 - observed diminishing returns above this value
+
 // InfoSchema contains database information.
 type InfoSchema interface {
 	GetToDdl() ToDdl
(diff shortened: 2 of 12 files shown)


sources/common/infoschema.go
```diff
@@ -58,18 +59,36 @@ type FkConstraint struct {
 // ProcessSchema performs schema conversion for source database
 // 'db'. Information schema tables are a broadly supported ANSI standard,
 // and we use them to obtain source database's schema information.
-func ProcessSchema(conv *internal.Conv, infoSchema InfoSchema) error {
+// numWorkers decides the parallelism while converting tables
+func ProcessSchema(conv *internal.Conv, infoSchema InfoSchema, numWorkers int) error {
 	tables, err := infoSchema.GetTables()
+	fmt.Println("fetched tables", tables)
```
shreyakhajanchi, line 65: Instead of these print statements wouldn't it be better to use logger.Log.Debug

sources/common/infoschema.go
```diff
@@ -58,18 +59,36 @@ type FkConstraint struct {
...
 	if err != nil {
 		return err
 	}
-	for _, t := range tables {
-		if err := processTable(conv, t, infoSchema); err != nil {
-			return err
-		}
+
+	if numWorkers < 1 {
+		numWorkers = 20 // Default to 20 - observed diminishing returns above this value
```
shreyakhajanchi, line 71: It would be better to define this as a constant

sources/common/infoschema.go
shreyakhajanchi: Same comment for other print statements

common/utils/utils.go
```diff
@@ -561,7 +561,7 @@ func GetLegacyModeSupportedDrivers() []string {
 // ReadSpannerSchema fills conv by querying Spanner infoschema treating Spanner as both the source and dest.
 func ReadSpannerSchema(ctx context.Context, conv *internal.Conv, client *sp.Client) error {
 	infoSchema := spanner.InfoSchemaImpl{Client: client, Ctx: ctx, TargetDb: conv.TargetDb}
-	err := common.ProcessSchema(conv, infoSchema)
+	err := common.ProcessSchema(conv, infoSchema, 0)
```
shreyakhajanchi, line 564: Better to define a constant of 20 workers and use that instead of 0 for better readability

sources/common/infoschema.go
```diff
@@ -58,18 +59,36 @@ type FkConstraint struct {
...
+		mutex.Lock()
+		conv.SrcSchema[table.Name] = table
+		mutex.Unlock()
+		res := TaskResult[SchemaAndName]{t, e}
+		return res
 	}
+
+	res, e := RunParallelTasks(tables, numWorkers, asyncProcessTable, true)
+	if e != nil {
+		fmt.Println("exiting due to error while processing schema for table", res)
```
shreyakhajanchi, line 85: do we want to print error here instead of res?

shreyakhajanchi (commented):
Left minor comments

sources/common/infoschema.go
bharadwaj-aditya: planning to take that up as a separate exercise across the file as the logs are currently not being printed to stdout and will be lost.

webv2/web.go
```diff
@@ -240,13 +240,13 @@ func convertSchemaSQL(w http.ResponseWriter, r *http.Request) {
 	var err error
 	switch sessionState.Driver {
 	case constants.MYSQL:
-		err = common.ProcessSchema(conv, mysql.InfoSchemaImpl{DbName: sessionState.DbName, Db: sessionState.SourceDB})
+		err = common.ProcessSchema(conv, mysql.InfoSchemaImpl{DbName: sessionState.DbName, Db: sessionState.SourceDB}, common.DefaultWorkers)
 	case constants.POSTGRES:
-		err = common.ProcessSchema(conv, postgres.InfoSchemaImpl{Db: sessionState.SourceDB})
+		err = common.ProcessSchema(conv, postgres.InfoSchemaImpl{Db: sessionState.SourceDB}, 0)
```
shreyakhajanchi, line 245: DefaultWorkers here instead of 0

shreyakhajanchi (commented):
LGTM. 2 minor open comments

shreyakhajanchi (approved):

webv2/web.go
bharadwaj-aditya: fixed

PR #488, opened 2023-03-10
feat: Allow Dataflow to run in a shared VPC

manitgupta:
Adds support for running Dataflow inside a shared VPC.


streaming/streaming.go
```diff
@@ -367,12 +368,18 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
 		inputFilePattern = inputFilePattern + "/"
 	}
 	fmt.Println("Reading files from datastream destination ", inputFilePattern)
+	var dataflowHostProjectId string
+	if dataflowCfg.HostProjectId == "" {
+		dataflowHostProjectId, _ = utils.GetProject()
```
shreyakhajanchi, line 373: I think it will fail in case if GetProject returns empty string which it might if project is not set in env variable and gcloud config

ui/src/app/components/prepare-migration/prepare-migration.component.html
```diff
@@ -118,7 +118,7 @@ <h3>Target details:</h3>
                             [disabled]="isMigrationInProgress || !isTargetDetailSet">
                             Configure
                             <mat-icon>edit</mat-icon>
-                            <mat-icon *ngIf="isDataflowConfigurationSet" class="success"
+                            <mat-icon class="success"
```
shreyakhajanchi, line 121: If we remove this if will it not show the icon always?

manitgupta: Yes, that's the experience we want (Want the user to think Dataflow is already configured unless they know what they are doing by tweaking the config)

streaming/streaming.go
manitgupta: Is returning an empty string from `GetProject()` a valid return value? Shouldn't it return a non-empty project string, or an error otherwise?

bharadwaj-aditya (approved):
LGTM

streaming/streaming.go
shreyakhajanchi: A bug has been created to handle the error with GetProject() separately : [b/273418828](https://b.corp.google.com/273418828)

shreyakhajanchi (approved):
LGTM

PR #497, opened 2023-03-24
Release fixes for 2.0.3

shreyakhajanchi:
Issues Fixed -
- Removed the usage of defer rows.Close() because connections were not closing properly resulting in timeout error
- Added handling for dialect in Display Type in rules
- Fixed small issues related to dialect

TODO:
- Add retry logic for connection timeout and connection limit errors

### DIFF
--- a/webv2/web.go
+++ b/webv2/web.go
@@ -237,7 +237,6 @@ func convertSchemaSQL(w http.ResponseWriter, r *http.Request) {
 	}
 	conv := internal.MakeConv()
 
-	// Setting target db to spanner by default.
 	conv.SpDialect = sessionState.Dialect
 	var err error
 	switch sessionState.Driver {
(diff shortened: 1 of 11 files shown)


webv2/web.go
```diff
@@ -435,7 +434,7 @@ func convertSchemaDump(w http.ResponseWriter, r *http.Request) {
 	sessionState.SourceDB = nil
 	sessionState.Dialect = dc.SpannerDetails.Dialect
 	sessionState.SourceDBConnDetails = session.SourceDBConnDetails{
-		Path:           dc.Config.FilePath,
+		Path:           "upload-file/" + dc.Config.FilePath,
```
bharadwaj-aditya, line 437: Is this according to the name convention used for all other folders ?

bharadwaj-aditya (approved):
LGTM

webv2/web.go
shreyakhajanchi: will look through code for other naming convention violations and bring them inline as a part of separate exercise since folder creation is not happening as a part of this PR

PR #505, opened 2023-04-06
cleaning up webv1 & global command line mode

shreyakhajanchi:
Tests run:
- Schema migration via CLI
- Data migration via CLI
- Schema and data migration via CLI
- Schema and data migration via UI


cmd/data.go
```diff
@@ -49,6 +49,7 @@ type DataCmd struct {
 	dryRun          bool
 	logLevel        string
 	SkipForeignKeys bool
+	verbose         bool
```
manitgupta, line 52: This is not needed since the `log-level=DEBUG` is going to handle printing logs enabled via verbose.

manitgupta (approved):

PR #510, opened 2023-04-10
CLI support for sharded migrations for MySQL

manitgupta:
Implements the following - 

1. Configuration interface for accepting sharded configurations.
2. Serial bulk migrations via HB CLI.
3. Parallel minimal downtime sharded migrations via HB CLI.

Bulk Migration - 

```
./harbourbridge schema-and-data -source=mysql --target-profile="project=span-cloud-testing,instance=manit-testing,dbName=bulkyboo" --source-profile="config=shardbulk.cfg" --log-level=DEBUG
```

Minimal Downtime Migration - 

```
./harbourbridge schema-and-data -source=mysql --target-profile="project=span-cloud-testing,instance=manit-testing,dbName=minimal" --source-profile="config=shardConfigNew.cfg" --log-level=DEBUG
```

Note: Currently sharded migrations are only supported for MySQL source.


conversion/conversion.go
```diff
@@ -165,12 +166,61 @@ func getDbNameFromSQLConnectionStr(driver, sqlConnectionStr string) string {
...
+	params["host"] = shardConnInfo.Host
+	params["user"] = shardConnInfo.User
+	params["dbName"] = shardConnInfo.DbName
+	params["port"] = shardConnInfo.Port
+	params["password"] = shardConnInfo.Password
+	sourceProfileConnectionMySQL, err := profiles.NewSourceProfileConnectionMySQL(params)
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse connection configuration for the primary shard")
+	}
+	sourceProfileConnection := profiles.SourceProfileConnection{Mysql: sourceProfileConnectionMySQL, Ty: profiles.SourceProfileConnectionTypeMySQL}
```
bharadwaj-aditya, line 180: Should add a not on what would change while adding other source databases

conversion/conversion.go
```diff
@@ -165,12 +166,61 @@ func getDbNameFromSQLConnectionStr(driver, sqlConnectionStr string) string {
...
-		return conv, err
+	//handle fetching schema differently for sharded migrations, we only connect to the primary shard to
+	//fetch the schema. We reuse the SourceProfileConnection object for this purpose.
+	var infoSchema common.InfoSchema
+	var err error
+	switch sourceProfile.Ty {
+	case profiles.SourceProfileTypeConfig:
+		//Find Primary Shard Name
+		if sourceProfile.Config.ConfigType == "bulk" {
+			schemaShard := sourceProfile.Config.ShardConfigurationBulk.SchemaShard
```
bharadwaj-aditya, line 203: Nit: Probably change name from Schema shard to primaryShard, or something else without shard in the name ?

conversion/conversion.go
```diff
@@ -200,27 +250,82 @@ func snapshotMigrationHandler(sourceProfile profiles.SourceProfile, config write
...
-			return nil, err
+	//handle migrating data for sharded migrations differently
+	//sharded migrations are identified via the config= flag, if that flag is not present
+	//carry on with the existing code path in the else block
+	switch sourceProfile.Ty {
+	case profiles.SourceProfileTypeConfig:
+		////There are three cases to cover here, bulk migrations and sharded migrations (and later DMS)
+		//We provide an if-else based handling for each within the sharded code branch
+		//This will be determined via the configType, which can be "bulk", "streaming" or "dms"
+		if sourceProfile.Config.ConfigType == "bulk" {
```
bharadwaj-aditya, line 261: might make sense to have enums/constants for the config types as they are re-used in multiple places

conversion/conversion.go
```diff
@@ -200,27 +250,82 @@ func snapshotMigrationHandler(sourceProfile profiles.SourceProfile, config write
...
+			var bw *writer.BatchWriter
+
+			//Migrate the data from the data shards, the schema shard needs to be specified here again.
+			for _, dataShard := range sourceProfile.Config.ShardConfigurationBulk.DataShards {
+				//Create a connection profile object for it
+				fmt.Printf("Migrating shard: %v\n", dataShard.DbName)
```
bharadwaj-aditya, line 267: nit: Initiating migration for shard

conversion/conversion.go
```diff
@@ -200,27 +250,82 @@ func snapshotMigrationHandler(sourceProfile profiles.SourceProfile, config write
...
+			return bw, nil
+		} else if sourceProfile.Config.ConfigType == "dataflow" {
+			//STEP - 1 - Create batch for each physical shard
+			asyncProcessShards := func(p *profiles.DataShard, mutex *sync.Mutex) common.TaskResult[*profiles.DataShard] {
+				//create streaming cfg from the config source type.
+				streamingCfg := streaming.CreateStreamingConfig(*p)
+				//update the cfg with the HB defaults
+				err := streaming.VerifyAndUpdateCfg(&streamingCfg, targetProfile.Conn.Sp.Dbname)
+				if err != nil {
+					return common.TaskResult[*profiles.DataShard]{Result: p, Err: err}
```
bharadwaj-aditya, line 285: The error should contain the shard against which the error has occurred for better reporting.

conversion/conversion.go
```diff
@@ -200,27 +250,82 @@ func snapshotMigrationHandler(sourceProfile profiles.SourceProfile, config write
...
-		err = infoSchema.StartStreamingMigration(ctx, client, conv, streamInfo)
-		if err != nil {
-			return nil, err
+		var streamInfo map[string]interface{}
+		if sourceProfile.Conn.Streaming {
+			streamInfo, err = infoSchema.StartChangeDataCapture(ctx, conv)
+			if err != nil {
+				return nil, err
+			}
+			bw, err := snapshotMigrationHandler(sourceProfile, config, conv, client, infoSchema)
```
bharadwaj-aditya, line 317: Why is the snapshot hander and streaming handler invoked here ?

profiles/source_profile.go
```diff
@@ -429,12 +431,80 @@ func NewSourceProfileConnection(source string, params map[string]string) (Source
...
+	Location      string `json:"location"`
+	Network       string `json:"network"`
+	Subnetwork    string `json:"subnetwork"`
+	HostProjectId string `json:"hostProjectId"`
+}
+
+type DataShard struct {
+	DataShardId           string                `json:"dataShardId"`
+	SrcConnectionProfile  DatastreamConnProfile `json:"srcConnectionProfile"`
+	DestConnectionProfile DatastreamConnProfile `json:"destConnectionProfile"`
```
bharadwaj-aditya, line 457: nit: what is the convention in HB. Is it to use Dst or Dest ?

profiles/source_profile.go
```diff
@@ -429,12 +431,80 @@ func NewSourceProfileConnection(source string, params map[string]string) (Source
...
+	DataflowConfig        DataflowConfig        `json:"dataflowConfig"`
+	TmpDir                string                `json:"tmpDir"`
+	StreamLocation        string                `json:"streamLocation"`
+	LogicalShards         []LogicalShard        `json:"databases"`
+}
+
+type LogicalShard struct {
+	DbName         string   `json:"dbName"`
+	TableInclude   []string `json:"tableInclude"`
```
bharadwaj-aditya, line 466: If we're excluding tables, we would probably not create schema for them as well. In this case, this belongs in the session file.

profiles/source_profile.go
```diff
@@ -429,12 +431,80 @@ func NewSourceProfileConnection(source string, params map[string]string) (Source
...
+	SchemaShard DirectConnectionConfig `json:"schemaShard"`
+	DataShards  []*DataShard   `json:"dataShards"`
+}
+
+type ShardConfigurationBulk struct {
+	SchemaShard DirectConnectionConfig   `json:"schemaShard"`
+	DataShards  []DirectConnectionConfig `json:"dataShards"`
+}
+
+type ShardConfigurationDMS struct {
```
bharadwaj-aditya, line 482: add TODO

streaming/streaming.go
```diff
@@ -446,20 +479,62 @@ func getStreamingConfig(sourceProfile profiles.SourceProfile, targetProfile prof
 	}
 }
 
+func CreateStreamingConfig(pl profiles.DataShard) StreamingCfg {
+	//create dataflowcfg from pl receiver object
+	dataflowCfg := DataflowCfg{}
```
bharadwaj-aditya, line 484: Do any inputs need to be validated here ?

bharadwaj-aditya (changes requested):

conversion/conversion.go
```diff
@@ -200,27 +250,82 @@ func snapshotMigrationHandler(sourceProfile profiles.SourceProfile, config write
...
-		if err != nil {
-			return nil, err
+	//handle migrating data for sharded migrations differently
+	//sharded migrations are identified via the config= flag, if that flag is not present
+	//carry on with the existing code path in the else block
+	switch sourceProfile.Ty {
+	case profiles.SourceProfileTypeConfig:
+		////There are three cases to cover here, bulk migrations and sharded migrations (and later DMS)
+		//We provide an if-else based handling for each within the sharded code branch
+		//This will be determined via the configType, which can be "bulk", "streaming" or "dms"
```
shreyakhajanchi, line 260: // This will be determined via the configType, which can be "bulk", "dataflow" or "dms".

profiles/source_profile.go
```diff
@@ -510,7 +580,14 @@ func (src SourceProfile) ToLegacyDriver(source string) (string, error) {
 			}
 		}
 	case SourceProfileTypeConfig:
-		return "", fmt.Errorf("specifying source-profile using config not implemented")
+		{
+			switch strings.ToLower(source) {
+			case "mysql":
```
shreyakhajanchi, line 585: nit: better to use constant here

test_data/mysql_shard_bulk.cfg
```diff
@@ -0,0 +1,21 @@
...
+    "configType":"bulk",
+    "shardConfigurationBulk":{
+       "schemaShard":{
+          "host":"test",
+          "user":"test",
+          "password":"test",
+          "port":"3306",
+          "dbName":"test"
+       },
+       "otherShards":[
```
shreyakhajanchi, line 11: Lets keep the naming inline with the dataflow config

conversion/conversion.go
manitgupta: Added.

conversion/conversion.go
manitgupta: Naming is hard.. 🥲
Renamed the variable and the object both to `schemaSource`.

conversion/conversion.go
manitgupta: Converted to constants.

conversion/conversion.go
manitgupta: Done

conversion/conversion.go
manitgupta: This is the existing flow of minimal downtime migrations, it is just moved to the `default` switch case.
Ref - 
https://github.com/cloudspannerecosystem/harbourbridge/blob/master/conversion/conversion.go#L214

profiles/source_profile.go
manitgupta: Its actually `Dst`. I've renamed this object to match the convention.

profiles/source_profile.go
manitgupta: Yes, I had initially thought of including this here as a stop gap solution for selecting tables to migrate until individual table migrations came, I have removed this now.
Will update this logic with information derived from the session file with the individual table migrations feature.

profiles/source_profile.go
manitgupta: Added TODO here and in other places as well.

streaming/streaming.go
manitgupta: Inputs are validated in the `VerifyAndUpdateCfg` method which is called after this one.
Code snippet - 

```
streamingCfg := streaming.CreateStreamingConfig(*p)
//verify the CFG and update it with HB defaults
err := streaming.VerifyAndUpdateCfg(&streamingCfg, targetProfile.Conn.Sp.Dbname)
```

profiles/source_profile.go
manitgupta: Done

test_data/mysql_shard_bulk.cfg
manitgupta: Done

conversion/conversion.go
manitgupta: Done

test_data/mysql_shard_bulk.cfg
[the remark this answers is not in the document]
manitgupta: Yes, we should. Added.

manitgupta:
Resolved all comments, and have also improved the unit tests a bit.

conversion/conversion.go
```diff
@@ -201,27 +255,85 @@ func snapshotMigrationHandler(sourceProfile profiles.SourceProfile, config write
...
-			return nil, err
+	//handle migrating data for sharded migrations differently
+	//sharded migrations are identified via the config= flag, if that flag is not present
+	//carry on with the existing code path in the else block
+	switch sourceProfile.Ty {
+	case profiles.SourceProfileTypeConfig:
+		////There are three cases to cover here, bulk migrations and sharded migrations (and later DMS)
+		//We provide an if-else based handling for each within the sharded code branch
+		//This will be determined via the configType, which can be "bulk", "dataflow" or "dms"
+		if sourceProfile.Config.ConfigType == constants.BULK_MIGRATION {
```
bharadwaj-aditya, line 266: nit: method is getting too large, break it into a method for each case.

conversion/conversion.go
```diff
@@ -201,27 +255,85 @@ func snapshotMigrationHandler(sourceProfile profiles.SourceProfile, config write
...
+			var bw *writer.BatchWriter
+
+			//Migrate the data from the data shards, the schema shard needs to be specified here again.
+			for _, dataShard := range sourceProfile.Config.ShardConfigurationBulk.DataShards {
+				//Create a connection profile object for it
+				fmt.Printf("Initiating migration for shard: %v\n", dataShard.DbName)
+				infoSchema, err := getInfoSchemaForShard(dataShard, sourceProfile.Driver, targetProfile)
```
bharadwaj-aditya, line 273: The infoschema would be similar for all the shards. Is there any reason this is being called for each shard independently ?

conversion/conversion.go
```diff
@@ -201,27 +255,85 @@ func snapshotMigrationHandler(sourceProfile profiles.SourceProfile, config write
...
+				//finally perform a snapshot migration for the shard
+				bw = performSnapshotMigration(config, conv, client, infoSchema)
+			}
+			//finally, once all shard migrations are complete, return the batch writer object
+			return bw, nil
+		} else if sourceProfile.Config.ConfigType == constants.DATAFLOW_MIGRATION {
+			//STEP - 1 - Create batch for each physical shard
+			asyncProcessShards := func(p *profiles.DataShard, mutex *sync.Mutex) common.TaskResult[*profiles.DataShard] {
+				//create streaming cfg from the config source type.
+				streamingCfg := streaming.CreateStreamingConfig(*p)
```
bharadwaj-aditya, line 286: How does this handle if the schema is different on various Datashards ? i.e. if there are tables which need to be configured as per schema source, but don't exist on the data shard ?

streaming/streaming.go
```diff
@@ -447,20 +452,62 @@ func getStreamingConfig(sourceProfile profiles.SourceProfile, targetProfile prof
...
+	dataflowCfg.HostProjectId = inputDataflowConfig.HostProjectId
+	dataflowCfg.Subnetwork = inputDataflowConfig.Subnetwork
+	//create src and dst datastream from pl receiver object
+	datastreamCfg := DatastreamCfg{}
+	//set stream config
+	datastreamCfg.StreamLocation = pl.StreamLocation
+	//set src connection profile
+	inputSrcConnProfile := pl.SrcConnectionProfile
+	srcConnCfg := SrcConnCfg{}
+	srcConnCfg.Location = inputSrcConnProfile.Location
```
bharadwaj-aditya, line 470: nit: Might be more concise to be put the parameter as a part of the creation of the struct
Eg:
SrcConnCfg{Location: "asd", Name: "pqr}

bharadwaj-aditya (approved):
Minor comments.
LGTM

bharadwaj-aditya:
Please mention in the description of the PR, the support is only for MySQL.

conversion/conversion.go
manitgupta: Done

conversion/conversion.go
manitgupta: `getInfoForSchemaForShard` returns the `db` cursor which inside the `infoSchema` object which is used to query the data from the database. `infoSchema` has different methods inside it which deal with both `schema` and `data`.

conversion/conversion.go
manitgupta: Since we do not connect to the data shards from Harbourbridge in the case of dataflow based migrations, there is no way to know if the tables which have been configured from the schema source are present in the data source. 
This would need to be a guarantee coming from the customer. 

That being said, let's list out the scenarios - 

1. Identical schema - Happy path, proceed ahead.
2. Table X exists in Data shard but does not exist in schema source - This table will not be part of the session file. Correspondingly, no table will be created for it in Spanner. The CDC events which will come for this table will be  ignored.
3. Table X does not exist in data shard but exist in schema source - It will be created in Spanner. For data shards, which have this table, it will be streamed and populated. Data shards which do not have this table will not generate CDC for it and not contribute to writing to Spanner for the Table X.

streaming/streaming.go
manitgupta: Hehe, brain fart 😛
Thanks for highlighting! Updated!

manitgupta:
> Please mention in the description of the PR, the support is only for MySQL.

Its part of the PR title 😛, also updated the description.

bharadwaj-aditya:
LGTM Thanks !

test_data/mysql_shard_bulk.cfg
```diff
@@ -0,0 +1,22 @@
...
+    "configType":"bulk",
+    "shardConfigurationBulk":{
+       "schemaSource":{
+          "host":"test",
+          "user":"test",
+          "password":"test",
+          "port":"3306",
+          "dbName":"test"
+       },
+       "dataShards":[
```
shreyakhajanchi, line 11: Will we not have logical shards in direct connection and have separate dataShard objects for each db although in same physical instance?

conversion/conversion.go
```diff
@@ -201,27 +255,103 @@ func snapshotMigrationHandler(sourceProfile profiles.SourceProfile, config write
...
 		if err != nil {
-			return nil, err
+			return common.TaskResult[*profiles.DataShard]{Result: p, Err: err}
 		}
-		err = infoSchema.StartStreamingMigration(ctx, client, conv, streamInfo)
+
+		err = streaming.StartDataflow(ctx, targetProfile, streamingCfg, conv)
+		return common.TaskResult[*profiles.DataShard]{Result: p, Err: err}
+	}
+	_, err := common.RunParallelTasks(sourceProfile.Config.ShardConfigurationDataflow.DataShards, 5, asyncProcessShards, true)
```
shreyakhajanchi, line 330: Better to define 5 as constant above so that its easier to find and modify the value of parallelisation

manitgupta: Done

conversion/conversion.go
manitgupta: Got it, that error has more to do with individual table migrations than sharding. Currently, the `dbName` coming from the sharding configuration is specified in the include list during datastream configuration.

Will have to check out individual table migrations + sharding work together.

test_data/mysql_shard_bulk.cfg
manitgupta: For direct connection, went with this approach instead its more closer to how bulk migration for a single instance works, but your comments also makes sense. Will come back to this later to see if we need to more closely align the configuration with streaming migration.

PR #514, opened 2023-04-17
feat: adding support to convert BIT(1) data-type in MySQL source DB into Bool type.

arpan14:
Adds support to convert `BIT(1)` type from MYSQL source DB into `BOOL` datatype in Spanner. Without this change, the type in spanner is `BYTES(MAX)` by default. The current default data-type makes it difficult to interpret the values on client apps leading to failures and app side changes. 

The fix will better parsing abilities on client apps (For ex - In a Java App, the BIT(1) type can now be assigned to a `boolean` variable out of the box without any tweaks on client app)

Assumption: Apart from DDL generation we don't require any change on data conversion since there are existing functions that will convert string values stored in src columns into BOOL.

- [ ] Tests pass
Units tests have been added. Haven't yet added an end to end integration test. Exploring ways to add/modify integration tests.

Don't think any specific change is needed in README.


manitgupta:
Please validate the following -

1. Conversion of `bit(1)` to bool for both `GoogleSQL` and `PostgreSQL` dialects - You can do this by selecting the appropriate dialect in the HB UI.
2. Validate that the default conversion shown by HB UI for `bit(1)` is `bool` after this change, and not `string`. 
3. Validate that both string and bool should appear in the dropdown as possible Spanner datatypes for `bit(1)`. 
4. Validate the default conversion through the CLI was well. 

I'd recommend doing the above via an end to end migration.

manitgupta:
@arpan14 Are you still working on this feature?

PR #518, opened 2023-04-24
Fix datastream generated name handling

Deep1998:
Datastream no longer allows underscores in the name. We replace the underscores in the generated name with hyphens like how we do for the dataflow job name.


bharadwaj-aditya (approved):
LGTM

PR #520, opened 2023-04-26
Added functionality to migrate MySQL tables in the backend via Datapr…

eenclona-google:
First backend PR from dataproc templates team to incorporate migration via Dataproc. This PR includes functionality to migrate MySQL tables with no schema changes and no primary key via Dataproc. Let us know if we're missing anything in the PR process / any feedback on our code changes.


profiles/target_profile.go
```diff
@@ -53,9 +53,17 @@ type TargetProfileConnection struct {
...
+	Hostname   string
+	Subnetwork string
+	Port       string
+	TargetDB   string
+}
+
 type TargetProfile struct {
 	Ty   TargetProfileType
 	Conn TargetProfileConnection
+	Dc   DataprocConfig
```
manitgupta, line 66: `TargetProfile` captures the configuration related to the destination, which would be Spanner during Dataproc based migrations. This doesn't look like the correct place to include `Dataproc` specific configurations.

conversion/conversion.go
```diff
@@ -242,6 +242,39 @@ func performSnapshotMigration(config writer.BatchWriterConfig, conv *internal.Co
...
+	dataprocConfig["instance"] = targetProfile.Conn.Sp.Instance
+	dataprocConfig["project"] = targetProfile.Conn.Sp.Project
+	if len(targetProfile.Dc.Hostname) > 1 {
+		dataprocConfig["hostname"] = targetProfile.Dc.Hostname
+	}
+	if len(targetProfile.Dc.Port) > 1 {
+		dataprocConfig["port"] = targetProfile.Dc.Port
+	}
+	if len(targetProfile.Dc.Subnetwork) > 1 {
+		dataprocConfig["subnet"] = targetProfile.Dc.Subnetwork
```
manitgupta, line 265: This seems odd. Why does `targetProfile` had `HostName` and other host related properties?

conversion/conversion.go
```diff
@@ -242,6 +242,39 @@ func performSnapshotMigration(config writer.BatchWriterConfig, conv *internal.Co
...
+
+	//Adding all dataproc configs to a map
+	dataprocConfig := map[string]string{}
+	dataprocConfig["hostname"] = sourceProfile.Conn.Mysql.Host
+	dataprocConfig["port"] = sourceProfile.Conn.Mysql.Port
+	dataprocConfig["user"] = sourceProfile.Conn.Mysql.User
+	dataprocConfig["pwd"] = sourceProfile.Conn.Mysql.Pwd
+	dataprocConfig["targetdb"] = targetProfile.Dc.TargetDB
+	dataprocConfig["instance"] = targetProfile.Conn.Sp.Instance
+	dataprocConfig["project"] = targetProfile.Conn.Sp.Project
```
manitgupta, line 257: What is the advantage of merging `sourceProfile` and `targetProfile` objects into a single map?
Can we carry source and target values in their own respective objects inside `ProcessDataWithDataproc` method? This would ensure better readability IMO

conversion/conversion.go
```diff
@@ -299,6 +332,9 @@ func dataFromDatabase(ctx context.Context, sourceProfile profiles.SourceProfile,
 			}
 			return bw, nil
 		}
+		if sourceProfile.Conn.Dataproc {
+			return performDataprocMigration(sourceProfile, targetProfile, config, conv, client, infoSchema)
+		}
```
manitgupta, line 337: I don't think this is the right integration point, we should had a new config Type to `SourceProfileTypeConfig`.

This will make the integration much more easier and readable. Add a new type for Dataproc similar to the existing ones here - 
https://github.com/cloudspannerecosystem/harbourbridge/blob/master/profiles/source_profile.go#L486

internal/convert.go
```diff
@@ -180,6 +181,11 @@ type streamingStats struct {
 	ShardToDataflowJobMap map[string]string
 }
 
+type dataprocStats struct {
```
manitgupta, line 184: `dataProcStats` is misleading, probably `dataProcMetadata` or `dataProcIdMap` is a better naming here

conversion/conversion.go
manitgupta: I am not sure how we are ensuring that this works only for `MySQL` source. Can you please validate?

sources/common/infoschema.go
```diff
@@ -133,6 +139,141 @@ func ProcessData(conv *internal.Conv, infoSchema InfoSchema) {
...
+			conv.DataFlush()
+		}
+
+		if !conv.Audit.DryRun {
+			progressCtr++
+			conv.Audit.Progress.MaybeReport(int64(progressCtr))
+		}
+
+		//TODO: eenclona@ will remove hardcoded us-central1 to be parametarized
+		url := fmt.Sprintf("https://pantheon.corp.google.com/dataproc/batches/us-central1/%s", id)
```
manitgupta, line 175: This an internal url, will most likely need to be replaced by a `googleapis` link in addition to parameterization

sources/common/infoschema.go
```diff
@@ -133,6 +139,141 @@ func ProcessData(conv *internal.Conv, infoSchema InfoSchema) {
...
+
+// Function to trigger dataproc template
+func TriggerDataprocTemplate(srcTable string, srcSchema string, primaryKeys string, dataprocConfig map[string]string) (string, error) {
+	ctx := context.Background()
+
+	println("Triggering Dataproc template for " + srcSchema + "." + srcTable)
+
+	// Extract location from subnet
+	subnet := dataprocConfig["subnet"]
+	region_string := subnet[0:strings.Index(subnet, "/subnetworks")]
```
manitgupta, line 192: This is error prone, region should be passed as its own parameter IMO

sources/common/infoschema.go
```diff
@@ -133,6 +139,141 @@ func ProcessData(conv *internal.Conv, infoSchema InfoSchema) {
...
+	location := subnet[strings.LastIndex(region_string, "/")+1 : strings.LastIndex(subnet, "/subnetworks")]
```
manitgupta, line 193: Same as above, this is error prone.

sources/common/infoschema.go
```diff
@@ -133,6 +139,141 @@ func ProcessData(conv *internal.Conv, infoSchema InfoSchema) {
...
+
+	// Create the batch controller cliermnt.
+	batchEndpoint := fmt.Sprintf("%s-dataproc.googleapis.com:443", location)
+	batchClient, err := dataproc.NewBatchControllerClient(ctx, option.WithEndpoint(batchEndpoint))
```
manitgupta, line 197: Does this mean a new client is created on every trigger of the dataproc template? 
Can we have a singleton client created once, which is used for triggering dataproc templates as they complete?

sources/common/infoschema.go
```diff
@@ -133,6 +139,141 @@ func ProcessData(conv *internal.Conv, infoSchema InfoSchema) {
 	}
 }
 
+func ProcessDataWithDataproc(conv *internal.Conv, infoSchema InfoSchema, dataprocConfig map[string]string) error {
+
+	orderTableNamesByID := ddl.GetSortedTableIdsBySpName(conv.SpSchema)
```
manitgupta, line 144: This would return a list of tables taking interleaving into account, but since this method only gives out a `List`, it won't be possible to parallelise Dataproc template execution.

We should instead do the following - 

1. Implement a method which would return a `List<List>`, each `List` would contain 1 or more elements, ordered from parent to children.
2. Create a fixed size worker pool to process these lists. Each worker takes a `List` as an input.
3. Each worker serially processes the List and orchestrates Dataproc template.

There are some code refs in HB for worker pools -

https://github.com/cloudspannerecosystem/harbourbridge/blob/master/conversion/conversion.go#L337

For the method implementation - We need to represent the DDL has an adjacency list with root nodes being parent tables.

profiles/target_profile.go
forksumit: Moved to `SourceProfile.SourceProfileConfig.ShardConfigurationDataproc` 
Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/d65e9ff652f5c9adf3145b8c9c860c02076a8ad6

conversion/conversion.go
forksumit: Added new `SourceProfileTypeConfig ` for storing dataproc config
Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/d65e9ff652f5c9adf3145b8c9c860c02076a8ad6

conversion/conversion.go
forksumit: Moved all dataproc api request handling to a new `dproc` package
Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/c93cd29b12cb4aa416ae6d4d5d8e90c629e017df

sources/common/infoschema.go
forksumit: Fixed, Moved client creation (in a new helper function `CreateDataprocBatchClient()`) outside of `TriggerDataprocTemplate()` and reuse them for each batch trigger
Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/c93cd29b12cb4aa416ae6d4d5d8e90c629e017df

sources/common/infoschema.go
forksumit: As discussed, We'll support only serial execution (respecting interleaving) as part of this PR, and add support for parallel execution in upcoming PR

conversion/conversion.go
forksumit: Moved to `SourceProfile.SourceProfileConfig.ShardConfigurationDataproc` 

TODO (future): Rename `HostName` to a distinctive name to differentiate with `DirectConnectionConfig.Host`

Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/d65e9ff652f5c9adf3145b8c9c860c02076a8ad6

internal/convert.go
forksumit: Done
Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/c27b72ed48a710da764d418d69f1f40ac30f5950

sources/common/infoschema.go
forksumit: Done, 
Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/9012732e7958d82f494e148d78387f8c8c7eb447

manitgupta:
Please rebase and resolve conflicts

sources/common/infoschema.go
forksumit: For now, just added validators in front-end to validate subnet uri before sending it to backend.
Exploring other options for future to simplify it further

sources/common/infoschema.go
forksumit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520#discussion_r1247696721
> For now, just added validators in front-end to validate subnet uri before sending it to backend. Exploring other options for future to simplify it further

dproc/dproc.go
```diff
@@ -0,0 +1,165 @@
...
+
+type DataprocRequestParams struct {
+	Project       string
+	TemplateArgs  []string
+	JarFileUris   []string
+	SubnetworkUri string
+	Location      string
+}
+
+func GetDataprocRequestParams(sourceProfile profiles.SourceProfile, targetProfile profiles.TargetProfile, srcSchema string, srcTable string, primaryKeys string, location string, subnet string) (DataprocRequestParams, error) {
```
manitgupta, line 44: Please add unit tests

forksumit: Added some base unit tests for method `TestGetDataprocRequestParams()`, there's still scope for more test cases for same and other methods, for now added TODO comments for same, will add them eventually
Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/e85e5b1db50678234e466fd266f5484e7aaae492

common/constants/constants.go
```diff
@@ -71,12 +71,16 @@ const (
...
+
 	//bulk migration type
 	BULK_MIGRATION = "bulk"
 	//dataflow migration type
 	DATAFLOW_MIGRATION = "dataflow"
 	//DMS migration type
-	DMS_MIGRATION       = "dms"
+	DMS_MIGRATION = "dms"
+	//dataproc migration type
+	DATAPROC_MIGRATION = "dproc"
```
bharadwaj-aditya, line 82: nit: better to use the whole name here

conversion/conversion.go
```diff
@@ -247,6 +249,76 @@ func performSnapshotMigration(config writer.BatchWriterConfig, conv *internal.Co
...
+		if conv.DataFlush != nil {
+			conv.DataFlush()
+		}
+
+		if !conv.Audit.DryRun {
+			progressCtr++
+			conv.Audit.Progress.MaybeReport(int64(progressCtr))
+		}
+
+		//TODO: eenclona@ will remove hardcoded us-central1
```
bharadwaj-aditya, line 302: nit: Suggest to remove ldaps as this is a public repo

dproc/dproc.go
```diff
@@ -0,0 +1,166 @@
...
+	ctx := context.Background()
+	batchEndpoint := fmt.Sprintf("%s-dataproc.googleapis.com:443", location)
+	batchClient, err := dataproc.NewBatchControllerClient(ctx, option.WithEndpoint(batchEndpoint))
+	return batchClient, err
+}
+
+func TriggerDataprocTemplate(batchClient *dataproc.BatchControllerClient, srcTable string, srcSchema string, primaryKeys string, dataprocRequestParams DataprocRequestParams) (string, error) {
+	ctx := context.Background()
+
+	fmt.Printf("Triggering Dataproc template for %s.%s\n", srcSchema, srcTable)
```
bharadwaj-aditya, line 121: please use the logger here

bharadwaj-aditya: if it would make sense the log the non-sensitive parameters of the job here

dproc/dproc.go
```diff
@@ -0,0 +1,166 @@
...
+					Args:        dataprocRequestParams.TemplateArgs,
+					JarFileUris: dataprocRequestParams.JarFileUris,
+				},
+			},
+		},
+	}
+
+	op, err := batchClient.CreateBatch(ctx, req)
+	if err != nil {
+		fmt.Printf("error creating the batch: %s\n", err.Error())
```
bharadwaj-aditya, line 149: please use the logger

dproc/dproc_test.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"fmt"
+	"path/filepath"
+	"testing"
+
+	"github.com/cloudspannerecosystem/harbourbridge/profiles"
+	"github.com/stretchr/testify/assert"
+)
+
+// TODO: add unit tests for otjer methods in dproc package
```
bharadwaj-aditya, line 12: Please add the unit tests along with the PR

profiles/source_profile.go
```diff
@@ -483,11 +494,19 @@ type ShardConfigurationBulk struct {
 type ShardConfigurationDMS struct {
 }
 
+// TODO: Define the sharding structure for dataproc migrations here.
+type ShardConfigurationDataproc struct {
+	SchemaSource DirectConnectionConfig `json:"schemaSource"`
+	// DataShards     []*DataShard           `json:"dataShards"`
```
bharadwaj-aditya, line 500: Even if we don't do the sharding in this PR, please ensure the structure is compatible with sharded migrations in the future.

bharadwaj-aditya (changes requested):

dproc/dproc.go
forksumit: Done, 
Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/a245b849a66861ec60e6d44c3294e6e052f233bb

dproc/dproc.go
forksumit: Done,
Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/a245b849a66861ec60e6d44c3294e6e052f233bb

common/constants/constants.go
forksumit: Done,
Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/c187c901898bd2f01b3f9aec41e787d93cb7aa12

profiles/source_profile.go
forksumit: `ShardConfigurationDataproc` will be compatible with sharding in future, only `DataShards` will have to be uncommented (no changes will be required in existing struct members).

Had it commented as it's being in use anywhere, can comment right now itself, so that it'll be auto initialized to null or empty array

Let me know, what you think is better.

dproc/dproc_test.go
forksumit: Added more test cases for `GetDataprocRequestParams()`, 

Didn't add any unit tests for others methods (`CreateDataprocBatchClient()`, `TriggerDataprocTemplate()`), these two methods are used used for only for making dataproc API calls to create dataproc client and  to submit the job via generated params from `GetDataprocRequestParams()` , to have unit tests for these two methods will need a lot of mocking, there won't be much to test after mocking, Plan is to test it in integration test.

Let me know what you think.

Commit: https://github.com/cloudspannerecosystem/harbourbridge/pull/520/commits/64946605e2140447fbb83f6bc9b323e8d4a59d4b

bharadwaj-aditya:
Closing as PR raised on dataproc branch #594

PR #531, opened 2023-05-12
Fix case sensitivity; reserved keyword handling and general stability improvements

manitgupta:
This PR fixes a bunch of bugs in Harbourbridge:

- Fix handling of case sensitive source databases when migrating to PGSQL. Now, quoting for PGSQL is source aware and only quotes when Postgres is the source. For non-case sensitive sources like MySQL, no quoting is done.
- Handling of reserved keywords: Irrespective of source database, reserved keywords in PGSQL need to be quoted when the DDL is being generated. For example, if a MySQL database has a table with column name `user`, it needs to be quoted.
- Fix handling of migration of non-public Postgres schemas. This is done by stripping the `tableName` with `schema` at the time of querying.
- QoL improvement - If the postgres database contains only one schema, its name is stripped. If it contains multiple schemas, `public` schema name is stripped.
- Switch default log level to `DEBUG` from `INFO`. This is useful for customers since most errors encountered by the customers can be detected via DEBUG logs.
- Fix log handling for web command. If the web server does not get up, it now logs the error correctly on the UI.
- Add a port flag for the `web` command. Now the user can override the default port from `8080` to a port of their choice.

Note: `InfoSchemaImpl` is passed by value as a receiver in all the method calls inside postgres `infoschema.go`, because of which a bool pointer had to be used to support detection of unique schemas. We would need to refactor the interface + implementation to move all the methods to use pass by reference.

### DIFF
--- a/sources/postgres/infoschema.go
+++ b/sources/postgres/infoschema.go
@@ -103,7 +103,7 @@ func (isi InfoSchemaImpl) GetRowsFromTable(conv *internal.Conv, tableId string) 
 	// PostgreSQL schema and name can be arbitrary strings.
 	// Ideally we would pass schema/name as a query parameter,
 	// but PostgreSQL doesn't support this. So we quote it instead.
-	isSchemaNamePrefixed := strings.HasPrefix(conv.SrcSchema[tableId].Name, conv.SrcSchema[tableId].Schema)
+	isSchemaNamePrefixed := strings.HasPrefix(conv.SrcSchema[tableId].Name+".", conv.SrcSchema[tableId].Schema)
 	var tableName string
 	if isSchemaNamePrefixed {
 		tableName = strings.TrimPrefix(conv.SrcSchema[tableId].Name, conv.SrcSchema[tableId].Schema+".")
--- a/sources/postgres/infoschema.go
+++ b/sources/postgres/infoschema.go
@@ -92,6 +92,8 @@ func (isi InfoSchemaImpl) GetToDdl() common.ToDdl {
 func (isi InfoSchemaImpl) GetTableName(schema string, tableName string) string {
 	if *isi.IsSchemaUnique { // Drop schema name as prefix if only one schema is detected.
 		return tableName
+	} else if schema == "public" {
+		return tableName
 	}
 	return fmt.Sprintf("%s.%s", schema, tableName)
 }
--- a/sources/postgres/infoschema.go
+++ b/sources/postgres/infoschema.go
@@ -103,7 +103,7 @@ func (isi InfoSchemaImpl) GetRowsFromTable(conv *internal.Conv, tableId string) 
 	// PostgreSQL schema and name can be arbitrary strings.
 	// Ideally we would pass schema/name as a query parameter,
 	// but PostgreSQL doesn't support this. So we quote it instead.
-	isSchemaNamePrefixed := strings.HasPrefix(conv.SrcSchema[tableId].Name+".", conv.SrcSchema[tableId].Schema)
+	isSchemaNamePrefixed := strings.HasPrefix(conv.SrcSchema[tableId].Name, conv.SrcSchema[tableId].Schema+".")
 	var tableName string
 	if isSchemaNamePrefixed {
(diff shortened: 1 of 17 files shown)


sources/postgres/infoschema.go
```diff
@@ -87,7 +103,14 @@ func (isi InfoSchemaImpl) GetRowsFromTable(conv *internal.Conv, tableId string)
 	// PostgreSQL schema and name can be arbitrary strings.
 	// Ideally we would pass schema/name as a query parameter,
 	// but PostgreSQL doesn't support this. So we quote it instead.
-	q := fmt.Sprintf(`SELECT * FROM "%s"."%s";`, conv.SrcSchema[tableId].Schema, conv.SrcSchema[tableId].Name)
+	isSchemaNamePrefixed := strings.HasPrefix(conv.SrcSchema[tableId].Name, conv.SrcSchema[tableId].Schema)
```
shreyakhajanchi, line 106: Will it be better to validate prefix for schemaName + "." for our case to avoid cases were a column name might genuinely have schema name as prefix?

manitgupta:
Addresses issues highlighted in #532, also adds some more stability improvements.

spanner/ddl/ast.go
```diff
@@ -173,12 +183,35 @@ type Config struct {
...
+		if strings.EqualFold(KEYWORD, identifier) {
+			return true
+		}
+	}
+	return false
+}
+
+func isSourceCaseSensitive(source string) bool {
+	switch source {
+	case constants.POSTGRES:
```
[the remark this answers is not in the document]
manitgupta: Added.

webv2/web.go
```diff
@@ -2272,16 +2273,16 @@ func init() {
 }
 
 // App connects to the web app v2.
-func App(logLevel string, open bool) {
+func App(logLevel string, open bool, port int) error {
 	err := logger.InitializeLogger(logLevel)
 	if err != nil {
-		log.Fatal("Error initialising webapp, did you specify a valid log-level? [DEBUG, INFO, WARN, ERROR, FATAL]")
+		return fmt.Errorf("error initialising webapp, did you specify a valid log-level? [DEBUG, INFO, WARN, ERROR, FATAL]")
```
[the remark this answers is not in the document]
manitgupta: Makes sense, updated.

sources/postgres/infoschema.go
manitgupta: Fair point! Updated!

sources/postgres/infoschema_test.go
```diff
@@ -562,3 +562,8 @@ func mkMockDB(t *testing.T, ms []mockSpec) *sql.DB {
 	}
 	return db
 }
+
+func newFalsePtr() *bool {
```
bharadwaj-aditya, line 566: this is funny :)

bharadwaj-aditya (approved):
LGTM

sources/postgres/infoschema.go
```diff
@@ -103,7 +103,7 @@ func (isi InfoSchemaImpl) GetRowsFromTable(conv *internal.Conv, tableId string)
 	// PostgreSQL schema and name can be arbitrary strings.
 	// Ideally we would pass schema/name as a query parameter,
 	// but PostgreSQL doesn't support this. So we quote it instead.
-	isSchemaNamePrefixed := strings.HasPrefix(conv.SrcSchema[tableId].Name, conv.SrcSchema[tableId].Schema)
+	isSchemaNamePrefixed := strings.HasPrefix(conv.SrcSchema[tableId].Name+".", conv.SrcSchema[tableId].Schema)
```
shreyakhajanchi, line 106: Isn't the "." appended opposite ? Shouldn't it be appended to Schema name for verification?

shreyakhajanchi (commented):
LGTM. Just 1 comment

shreyakhajanchi (approved):

sources/postgres/infoschema.go
manitgupta: Oops. I've updated it now.

PR #539, opened 2023-05-24
Error dialog box display for prepare migration button

shreyakhajanchi:


ui/src/app/components/workspace/workspace.component.ts
```diff
@@ -16,6 +16,8 @@ import { extractSourceDbName } from 'src/app/utils/utils'
 import { ClickEventService } from 'src/app/services/click-event/click-event.service'
 import IViewAssesmentData from 'src/app/model/view-assesment'
 import IDbConfig from 'src/app/model/db-config'
+import { InfodialogComponent } from '../infodialog/infodialog.component'
```
manitgupta, line 19: Rename to something more meaningful -  E.g `ConfirmSchemaConversionComponent`

webv2/web.go
```diff
@@ -653,6 +653,20 @@ func getTypeMap(w http.ResponseWriter, r *http.Request) {
...
+	for id, issues := range sessionState.Conv.SchemaIssues {
+		if len(issues.TableLevelIssues) != 0 {
+			tableNames = append(tableNames, sessionState.Conv.SpSchema[id].Name)
+		}
+	}
+	w.WriteHeader(http.StatusOK)
+	json.NewEncoder(w).Encode(tableNames)
+}
+
 // applyRule allows to add rules that changes the schema
```
manitgupta, line 670: API name is `getErrors` but it returns a list of tables - This is confusing, consider renaming to `getTablesWithErrors()`.

Both the `Id` and the `Name` pair should be returned in this API for future extensibility.

ui/src/app/components/workspace/workspace.component.ts
```diff
@@ -301,7 +304,25 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
     return false
   }
   prepareMigration() {
-    this.router.navigate(['/prepare-migration'])
+    this.fetch.getErrors().subscribe({
+      next: (res: any) => {
+        if (res != null && res.length !=0)
+        {
+          let errMsg = 'Please fix the errors for the following tables to move ahead: '+ res.join(',')
```
manitgupta, line 311: Can you add a screenshot of how this looks please? 
For a large number of tables this might go out of bounds of the dialog box width of 500 px. 🤔

manitgupta (changes requested):

ui/src/app/components/workspace/workspace.component.ts
shreyakhajanchi: It is an existing component and same component is reused in multiple places therefore a generic name makes more sense IMO

webv2/web.go
shreyakhajanchi: Makes sense. Done

manitgupta (approved):

PR #543, opened 2023-06-02
Update reserved keywords

manitgupta:
Updated Reserved keywords used in Postgres - https://www.postgresql.org/docs/current/sql-keywords-appendix.html

This list is different from the last time since earlier only keywords marked with "reserved" were included, but we also want to include keywords with types - `reserved (can be function or type)`, `reserved, requires AS`


bharadwaj-aditya (approved):
LGTM

But please clarify in the comment why the list is different from last time

shreyakhajanchi (approved):
LGTM. Validated the reserved words against the list shared.

PR #547, opened 2023-06-05
Fix: Skip PG.NUMERIC index creation, update reserved keywords and improve FK creation throughput

manitgupta:
PG.NUMERIC cannot be indexed in Spanner. Harbourbridge currently does not account for this and converts an index on a NUMERIC field in PG to an index on PG.NUMERIC in PGSQL dialect. This change will skip the creation of this index.

Note: This PR does not take into account the UI experience of this change. This is to be treated as a hot fix to ensure correctness, since HB generated DDL cannot be used without this change.


sources/common/toddl.go
```diff
@@ -251,6 +255,13 @@ func CvtIndexHelper(conv *internal.Conv, tableId string, srcIndex schema.Index,
 		for _, v := range spColIds {
 			if v == k.ColId {
 				isPresent = true
+				if (conv.SpDialect == constants.DIALECT_POSTGRESQL) {
+					if(spColDef[v].T.Name == ddl.Numeric) {
+						//index on NUMERIC is not supported in PGSQL Dialect currently.
+						//Indexes which contains a NUMERIC column in it will need to be skipped.
+						return ddl.CreateIndex{}
```
bharadwaj-aditya, line 262: Is the purpose of this passing back an empty index ?
meaning this is a placeholder for a constant ?

bharadwaj-aditya (approved):
LGTM

shreyakhajanchi (commented):
Shouldn't we include a msg of some sort that why was this index deleted ? Maybe a tooltip? It might confuse the users.

shreyakhajanchi (approved):
LGTM

shreyakhajanchi (approved):
LGTM

spanner/ddl/ast.go
```diff
@@ -94,15 +94,16 @@ var PGSQL_TO_STANDARD_TYPE_TYPEMAP = map[string]string{
 	PGJSONB:       JSON,
 }
 
-// List taken from - https://www.postgresql.org/docs/current/sql-keywords-appendix.html
+// PGDialect keyword list
```
bharadwaj-aditya, line 97: Can you add the source ?

bharadwaj-aditya (approved):
LGTM

spanner/ddl/ast.go
manitgupta: Its an internal code source link so I skipped adding it here.

PR #556, opened 2023-06-16
Create ForeignKey earlier in low downtime migration

aksharauke:
During low downtime migration - In the Create database - also enable the foreign key creation
During low downtime migration - If the database already exists and there are foreign keys to be applied - give an exception. Since applying foreign keys on existing database takes long time, even if tables are empty. 

Note that for PostgreSQL dialect, applying DDL as part of database creation is not supported, hence the behavior for PostgreSQL dialect will remain like current - the foreign keys will be applied once the Dataflow jobs start

### DIFF
--- a/conversion/conversion.go
+++ b/conversion/conversion.go
@@ -663,7 +663,7 @@ func ValidateDDL(ctx context.Context, adminClient *database.DatabaseAdminClient,
 }
 
 // CreatesOrUpdatesDatabase updates an existing Spanner database or creates a new one if one does not exist.
-func CreateOrUpdateDatabase(ctx context.Context, adminClient *database.DatabaseAdminClient, dbURI, driver string, conv *internal.Conv, out *os.File) error {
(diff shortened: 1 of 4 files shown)


aksharauke:
#review @shreyakhajanchi  @manitgupta

conversion/conversion.go
```diff
@@ -676,12 +676,15 @@ func CreateOrUpdateDatabase(ctx context.Context, adminClient *database.DatabaseA
 		ctx = metadata.AppendToOutgoingContext(ctx, constants.MigrationMetadataKey, migrationMetadataValue)
 	}
 	if dbExists {
+		if conv.SpDialect != constants.DIALECT_POSTGRESQL && migrationType == constants.DATAFLOW_MIGRATION {
+			return fmt.Errorf("Low downtime schema migration cannot work on existing database")
```
manitgupta, line 680: Lets reword this to "Harbourbridge does not support minimal downtime schema/schema-and-data migrations to an existing database".

manitgupta (approved):

conversion/conversion.go
aksharauke: Done

shreyakhajanchi (approved):
LGTM

PR #562, opened 2023-06-23
Misc: Bug Fixes and QoL Improvements

manitgupta:
Misc Bug Fixes and QoL Improvement features

- Download JSON option should only be visible for sharded migrations.
- Persist form values in the direct connection page in case of a failed connection.
- Add a test connection button on the direct connection page.
- Change `Spanner Draft` URL to use a pointer instead of a text cursor
- Remove confusing error message around environment variable.
- Add Helpful info tooltips at various places.
- All hyperlinks will now open in a new tab by default.
- Documentation updates
- Refactor Source and Target details into `<mat-card>`
- Add database loader with custom messaging for testing connection
- Float connection errors during direct connect to the UI from the backend API
- Make existing connection profile as the default


ui/src/app/components/database-loader/database-loader.component.html
```diff
@@ -4,7 +4,12 @@
     alt="database-loader-gif"
     class="loader-gif"
   />
-  <h3 *ngIf="loaderType === 'direct'">Connecting to {{ databaseName }} database...</h3>
+  <h3 *ngIf="loaderType === 'direct'">Populating schema for {{ databaseName }} database. Please wait...</h3>
```
shreyakhajanchi, line 7: Populating schema sounds misleading , should we say Reading infoschema for databse ...

shreyakhajanchi (approved):
LGTM. Left 1 minor comment

ui/src/app/components/database-loader/database-loader.component.html
manitgupta: Done

PR #565, opened 2023-06-27
Disable public IP when network is specified

aksharauke:
Dataflow recommendation is to always disable public IP for the Dataflow workers.
Hence keeping public IPs only for internal testing when there is no network/subnetwork specified.

### DIFF
--- a/streaming/streaming.go
+++ b/streaming/streaming.go
@@ -181,7 +181,7 @@ func getMysqlSourceStreamConfig(dbList []profiles.LogicalShard, tableList []stri
 	for _, db := range dbList {
 		//create include db object
 		includeDb := &datastreampb.MysqlDatabase{
-			Database: db.DbName,
+			Database:    db.DbName,
 			MysqlTables: mysqlTables,
 		}
 		includeDbList = append(includeDbList, includeDb)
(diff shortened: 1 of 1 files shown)


streaming/streaming.go
```diff
@@ -445,16 +445,22 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
 	} else {
 		dataflowHostProjectId = dataflowCfg.HostProjectId
 	}
+
 	dataflowSubnetwork := ""
+
+	// if custom network is not selected, use public IP. Typical internal testing flow
```
Deep1998, line 451: nit: golang comment style should be full sentences aka start with capital letter and end with full stop.

streaming/streaming.go
```diff
@@ -445,16 +445,22 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
...
+	workerIpAddressConfig := dataflowpb.WorkerIPAddressConfiguration_WORKER_IP_PUBLIC
+
 	if dataflowCfg.Network != "" {
```
Deep1998, line 454: and also != nil?

streaming/streaming.go
```diff
@@ -445,16 +445,22 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
 	} else {
 		dataflowHostProjectId = dataflowCfg.HostProjectId
 	}
+
 	dataflowSubnetwork := ""
+
+	// if custom network is not selected, use public IP. Typical internal testing flow
+	workerIpAddressConfig := dataflowpb.WorkerIPAddressConfiguration_WORKER_IP_PUBLIC
```
Deep1998, line 452: should this be PUBLIC or UNSPECIFIED

Deep1998 (approved):

streaming/streaming.go
aksharauke: PUBLIC. This gives the default behavior that exists when the public IP is not explicitly disabled.

streaming/streaming.go
aksharauke: This was existing check, tested it with and without specifying VPC. So seems fine as is.

manitgupta (approved):

PR #574, opened 2023-07-03
Script to setup reverse replication pipeline

Deep1998:
Script to setup reverse replication pipeline. This script creates all the GCP resources required by the pipeline (if they do not already exists).

Namely, it creates the following, skipping the resources that already exist:
1.  The Spanner changestream
2. Changestream metadata database
3. Pub/Sub topic
4. Pub/Sub subscriptions based on the sourceShardsFile.
5. Dataflow ordering job
6. Dataflow writer job

This script requires you have the session file and source shards file already present in GCS.

Sample launch command in README.


reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+		return fmt.Errorf("please specify a valid pubSubDataTopicId of the form project/project-name/topics/topic-name")
+	}
+	if pubSubEndpoint == "" {
+		pubSubEndpoint = fmt.Sprintf("https://%s-pubsub.googleapis.com:443", dataflowRegion)
+	}
+	return nil
+}
+
+func main() {
+	ORDERING_TEMPLATE := "gs://deepchowdhury-gsql/templates/flex/Spanner_Change_Streams_to_Sink"
```
Deep1998, line 95: TODO: Update paths after template release

reverse_replication/README.md
```diff
@@ -0,0 +1,60 @@
...
+## Resources
+The pipeline requires a few GCP resources to be setup. The launcher script creates these resources for you, skipping creation if they already exist. The resources are:
+- `Change Stream`: The target spanner database should have a changestream setup with value_capture_type = 'NEW_ROW'. This helps stream CDC events from Spanner.
+- `Ordering Dataflow Job`: This dataflow job reads from Spanner CDC, orders the data and pushes it to a PubSub topic.
+- `PubSub Topic & Subscriptions`: The topic that the ordering job pushes to needs to be created beforehand. For each shard, a subscription needs to be created, with the subscription name as the corresponding logicalShardId. These names are fetched from the source shards file mentioned later.
+- `Writer Dataflow Job`: This reads messages from the PubSub subscriptions, translates them to SQL and writes to the source shards.
+
+This launcher script will create
+
+## Arguments
```
aksharauke, line 16: Would be good to mention defaults

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+	pubSubEndpoint       string
+	sourceShardsFilePath string
+	sessionFilePath      string
+)
+
+func setupGlobalFlags() {
+	flag.StringVar(&projectId, "projectId", "", "projectId")
+	flag.StringVar(&dataflowRegion, "dataflowRegion", "", "region for dataflow jobs")
+	flag.StringVar(&jobNamePrefix, "jobNamePrefix", "reverse-rep", "job name prefix for the dataflow jobs")
+	flag.StringVar(&changeStreamName, "changeStreamName", "", "change stream name")
```
aksharauke, line 42: Can we have a default name, like ReverseReplicationStream

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+	flag.StringVar(&instanceId, "instanceId", "", "spanner instance id")
+	flag.StringVar(&dbName, "dbName", "", "spanner database name")
+	flag.StringVar(&metadataInstance, "metadataInstance", "", "spanner instance name to store changestream metadata")
```
aksharauke, line 45: Here also we can default this to the Spanner instance id if not provided

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+	flag.StringVar(&metadataDatabase, "metadataDatabase", "", "spanner database name to store changestream metadata")
```
aksharauke, line 46: We could give a default database name like change_stream_metadata

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+	flag.StringVar(&pubSubDataTopicId, "pubSubDataTopicId", "", "pub/sub data topic id, should be of the form projects/my-project/topics/my-topic")
```
aksharauke, line 47: A default name like reverse-replication-topic

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+		return fmt.Errorf("please specify a valid dbName")
+	}
+	if metadataInstance == "" {
+		return fmt.Errorf("please specify a valid metadataInstance")
+	}
+	if metadataDatabase == "" {
+		return fmt.Errorf("please specify a valid metadataDatabase")
+	}
+	if pubSubDataTopicId == "" {
+		return fmt.Errorf("please specify a valid pubSubDataTopicId of the form project/project-name/topics/topic-name")
```
aksharauke, line 76: Can we add this description of the topic name format in the Readme and flags definition

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+	if pubSubEndpoint == "" {
+		pubSubEndpoint = fmt.Sprintf("%s-pubsub.googleapis.com:443", dataflowRegion)
+	}
+	return nil
+}
+
+func main() {
+	ORDERING_TEMPLATE := "gs://deepchowdhury-gsql/templates/flex/Spanner_Change_Streams_to_Sink"
+	WRITER_TEMPLATE := "gs://deepchowdhury-gsql/templates/flex/Ordered_Changestream_Buffer_to_Sourcedb"
+	//WRITER_TEMPLATE := "gs://aks-test-revrep/images/ordered-changestream-buffer-to-sourcedb-image-spec.json"
```
aksharauke, line 97: Can delete this line

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+	err := prechecks()
+	if err != nil {
+		fmt.Println("incorrect arguments passed:", err)
+		return
+	}
+	ctx := context.Background()
+	adminClient, _ := database.NewDatabaseAdminClient(ctx)
+	op, err := adminClient.UpdateDatabaseDdl(ctx, &adminpb.UpdateDatabaseDdlRequest{
+		Database:   fmt.Sprintf("projects/%s/instances/%s/databases/%s", projectId, instanceId, dbName),
+		Statements: []string{fmt.Sprintf("CREATE CHANGE STREAM %s FOR ALL OPTIONS (value_capture_type = 'NEW_ROW')", changeStreamName)},
```
aksharauke, line 110: We should have a TODO here to only create change streams of the tables that exist in the session file

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		fmt.Printf("Cannot submit request create change stream request: %v\n", err)
+		return
+	}
+	if err := op.Wait(ctx); err != nil {
+		if !strings.Contains(err.Error(), "Duplicate name in schema") {
+			fmt.Printf("Could not update database ddl: %v\n", err)
+			return
+		} else {
+			fmt.Printf("changestream with name '%s' already exists...skipping changestream creation\n", changeStreamName)
```
aksharauke, line 121: Is there a way to check that the existing change stream has the same characteristic that are needed - example value type is NEW_ROW

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+				Topic:                 client.Topic(topicName),
+				AckDeadline:           600 * time.Second,
+				EnableMessageOrdering: true,
+				Filter:                fmt.Sprintf("attributes.shardId=\"%s\"", shardId),
+			})
+			if err != nil {
+				if !(strings.Contains(err.Error(), "AlreadyExists")) {
+					fmt.Printf("could not create subscription: %v\n", err)
+				} else {
+					fmt.Printf("subscription '%s' already exists, skipping creation...\n", shardId)
```
aksharauke, line 172: Is there a way to check that the existing subscription is ordered and has the expected filter criteria?

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+				}
+				return
+			}
+			fmt.Println("Created sub: ", shardId)
```
aksharauke, line 176: Created subscription for PubSub

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+		JobName:  fmt.Sprintf("%s-ordering", jobNamePrefix),
+		Template: &dataflowpb.LaunchFlexTemplateParameter_ContainerSpecGcsPath{ContainerSpecGcsPath: ORDERING_TEMPLATE},
+		Parameters: map[string]string{
+			"changeStreamName":   changeStreamName,
+			"instanceId":         instanceId,
+			"databaseId":         dbName,
+			"spannerProjectId":   projectId,
+			"metadataInstance":   metadataInstance,
+			"metadataDatabase":   metadataDatabase,
+			"startTimestamp":     "",
```
aksharauke, line 198: I believe this is defaulted to now in the Dataflow template, might as well skip this here

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+			"sinkType":           "pubsub",
+			"pubSubDataTopicId":  pubSubDataTopicId,
+			"pubSubErrorTopicId": pubSubDataTopicId,
+			"pubSubEndpoint":     pubSubEndpoint,
+			"sessionFilePath":    sessionFilePath,
+		},
+		Environment: &dataflowpb.FlexTemplateRuntimeEnvironment{
+			NumWorkers:            5,
+			AdditionalExperiments: []string{"use_runner_v2"},
+			MachineType:           "n2-standard-4",
```
aksharauke, line 209: This I think we should parameterise and default to n2-standard-4. Reason being - if performance is not as expected on the customer project, we could experiment with larger machines.

reverse_replication/launcher.go
```diff
@@ -0,0 +1,273 @@
...
+		Template: &dataflowpb.LaunchFlexTemplateParameter_ContainerSpecGcsPath{ContainerSpecGcsPath: WRITER_TEMPLATE},
+		Parameters: map[string]string{
+			"sourceShardsFilePath": sourceShardsFilePath,
+			"sessionFilePath":      sessionFilePath,
+			"bufferType":           "pubsub",
+			"pubSubProjectId":      projectId,
+		},
+		Environment: &dataflowpb.FlexTemplateRuntimeEnvironment{
+			NumWorkers: 5,
+			//	AdditionalExperiments: []string{"use_runner_v2"},
```
aksharauke, line 238: We can uncomment this now

reverse_replication/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+The script takes in multiple arguments to orchestrate the pipeline. They are:
```
aksharauke, line 18: One good feature is if we can give some run mode type which can:

1. Do the full setup and launch
2. Launch only the Dataflow jobs and skip the ChangeStream and PubSub
3. Launch only the Ordering Dataflow job
4. Launch only the Writer Dataflow job

These will help in case of reruns after failure.
You could add a TODO and do this later as well.

manitgupta:
Can you please add to description what does launching this look like?
Seems to be a completely unrelated file right now..

reverse_replication/launcher.go
Deep1998: Currently, it does not auto create a db if one doesnt exist. Will add that feature and default to this.

reverse_replication/launcher.go
Deep1998: Yes there is a way, but keeping it beyond the scope of this script for now.

reverse_replication/launcher.go
Deep1998: Same as above

reverse_replication/launcher.go
Deep1998: It is but keeping it for now, as if someone wants to modify this param and rerun the script, its one lesser step for them.

aksharauke: There is no input parameter that takes the startTimestamp. Either remove the startTimestamp from here or take it as input parameter.

reverse_replication/launcher.go
aksharauke: Reverse replication will fail incase the subscriber is not ordered and does not have the shard filter. If we are checking the existence of the subscriber in this script, then it should also check the properties of the same. It can be a TODO, but I believe the scope should be this script.

aksharauke:
LGTM

aksharauke (approved):

PR #589, opened 2023-07-17
added a rule to add shard id column as primary key

shreyakhajanchi:
1. Added a new rule to add shard id column as a part of primary key
2. The rule will only show up for sharded database migration
3. The rule will be applied by default and shard id column will be added at the start of the PK


common/constants/constants.go
```diff
@@ -76,7 +76,8 @@ const (
 	//dataflow migration type
 	DATAFLOW_MIGRATION = "dataflow"
 	//DMS migration type
-	DMS_MIGRATION       = "dms"
-	EditColumnMaxLength = "edit_column_max_length"
-	SESSION_FILE = "sessionFile"
+	DMS_MIGRATION        = "dms"
```
bharadwaj-aditya, line 79: nit: This is not a type of rule. Please add this along with the rule types

ui/src/app/components/add-shard-id-primary-key/add-shard-id-primary-key.component.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  setViewRuleData(data: IRule) {
+    this.ruleId = data?.Id
+    this.addShardIdPrimaryKeyForm.controls['primaryKeyOrder'].setValue(data?.Data?.AddedAtTheStart)
+  }
+
+  deleteRule() {
+    this.data.dropRule(this.ruleId)
+    this.resetRuleType.emit('')
```
bharadwaj-aditya, line 77: does this remove the primary rule from all tables ?

ui/src/app/components/sidenav-rule/sidenav-rule.component.ts
```diff
@@ -22,8 +23,20 @@ export class SidenavRuleComponent implements OnInit {
   ruletype: string = ''
   viewRuleData: any = []
   viewRuleFlag: boolean = false
+  shardedMigration: boolean = false
 
   ngOnInit(): void {
+    this.data.conv.subscribe({
+      next: (res: IConv) => {
+        console.log(1, res)
```
bharadwaj-aditya, line 31: Residual log ?

ui/src/app/components/sidenav-rule/sidenav-rule.component.ts
```diff
@@ -22,8 +23,20 @@ export class SidenavRuleComponent implements OnInit {
...
+        const keys = Object.keys(res.SpSchema) as Array<string>
+        if (res.SpSchema[keys[0]].ShardIdColumn != '') {
```
bharadwaj-aditya, line 33: is this the only way to find if a migration is sharded ? There should be a flag which explicitly calls it out .

ui/src/app/model/conv.ts
```diff
@@ -162,13 +162,7 @@ export interface IInterleaveStatus {
 
 export interface IPrimaryKey {
   TableId: string
-  Columns: IPkColumnDefs[]
-}
-
-export interface IPkColumnDefs {
```
bharadwaj-aditya, line 168: where has this moved to ?

webv2/web.go
```diff
@@ -988,6 +1036,67 @@ func dropRule(w http.ResponseWriter, r *http.Request) {
...
+			primarykey.UpdatePrimaryKeyAndSessionFile(pkRequest)
+		}
+	} else {
+		for _, table := range sessionState.Conv.SpSchema {
+			size := len(table.PrimaryKeys)
+			pkRequest := primarykey.PrimaryKeyRequest{
+				TableId: table.Id,
+				Columns: table.PrimaryKeys,
+			}
+			pkRequest.Columns = append(pkRequest.Columns, ddl.IndexKey{ColId: table.ShardIdColumn, Order: size + 1})
```
bharadwaj-aditya, line 1060: seems like a bunch of code in the 2 blocks around iterating over tables etc is common. Can we use common code and move the check inside the loop ?

webv2/web.go
```diff
@@ -988,6 +1036,67 @@ func dropRule(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+
+}
+
+func revertShardIdColumnAsPrimaryKey(isAddedAtFirst bool) {
+	sessionState := session.GetSessionState()
+	if isAddedAtFirst {
+		for _, table := range sessionState.Conv.SpSchema {
+			pkRequest := primarykey.PrimaryKeyRequest{
```
bharadwaj-aditya, line 1071: Same as above

bharadwaj-aditya (changes requested):

ui/src/app/components/add-shard-id-primary-key/add-shard-id-primary-key.component.ts
shreyakhajanchi: Yes its a global rule, addition or deletion of this rule will impact all the tables

ui/src/app/model/conv.ts
shreyakhajanchi: The IPkColumnDefs[] interface was same as IIndexKey[] so removed it

ui/src/app/components/sidenav-rule/sidenav-rule.component.ts
shreyakhajanchi: Added IsSharded flag

bharadwaj-aditya (approved):
LGTM

PR #591, opened 2023-07-18
Added support for data transformations BE

shreyakhajanchi:
TODO:

1. Add transformation_test file
2. Add support for numeric and JSON datatypes in transformation function wherever applicable

### DIFF
--- a/internal/convert.go
+++ b/internal/convert.go
@@ -28,24 +28,25 @@ import (
 
 // Conv contains all schema and data conversion state.
 type Conv struct {
-	mode           mode                     // Schema mode or data mode.
-	SpSchema       ddl.Schema               // Maps Spanner table name to Spanner schema.
-	SyntheticPKeys map[string]SyntheticPKey // Maps Spanner table name to synthetic primary key (if needed).
-	SrcSchema      map[string]schema.Table  // Maps source-DB table name to schema information.
-	SchemaIssues   map[string]TableIssues   // Maps source-DB table/col to list of schema conversion issues.
-	ToSpanner      map[string]NameAndCols   `json:"-"` // Maps from source-DB table name to Spanner name and column mapping.
-	ToSource       map[string]NameAndCols   `json:"-"` // Maps from Spanner table name to source-DB table name and column mapping.
-	UsedNames      map[string]bool          `json:"-"` // Map storing the names that are already assigned to tables, indices or foreign key contraints.
-	dataSink       func(table string, cols []string, values []interface{})
-	DataFlush      func()              `json:"-"` // Data flush is used to flush out remaining writes and wait for them to complete.
-	Location       *time.Location      // Timezone (for timestamp conversion).
-	sampleBadRows  rowSamples          // Rows that generated errors during conversion.
-	Stats          stats               `json:"-"`
-	TimezoneOffset string              // Timezone offset for timestamp conversion.
-	SpDialect      string              // The dialect of the spanner database to which HarbourBridge is writing.
-	UniquePKey     map[string][]string // Maps Spanner table name to unique column name being used as primary key (if needed).
-	Audit          Audit               `json:"-"` // Stores the audit information for the database conversion
-	Rules          []Rule              // Stores applied rules during schema conversion
+	mode            mode                     // Schema mode or data mode.
+	SpSchema        ddl.Schema               // Maps Spanner table name to Spanner schema.
+	SyntheticPKeys  map[string]SyntheticPKey // Maps Spanner table name to synthetic primary key (if needed).
+	SrcSchema       map[string]schema.Table  // Maps source-DB table name to schema information.
+	SchemaIssues    map[string]TableIssues   // Maps source-DB table/col to list of schema conversion issues.
+	ToSpanner       map[string]NameAndCols   `json:"-"` // Maps from source-DB table name to Spanner name and column mapping.
+	ToSource        map[string]NameAndCols   `json:"-"` // Maps from Spanner table name to source-DB table name and column mapping.
+	UsedNames       map[string]bool          `json:"-"` // Map storing the names that are already assigned to tables, indices or foreign key contraints.
+	dataSink        func(table string, cols []string, values []interface{})
+	DataFlush       func()              `json:"-"` // Data flush is used to flush out remaining writes and wait for them to complete.
+	Location        *time.Location      // Timezone (for timestamp conversion).
+	sampleBadRows   rowSamples          // Rows that generated errors during conversion.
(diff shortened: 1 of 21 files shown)


internal/convert.go
```diff
@@ -204,6 +205,27 @@ type Rule struct {
...
+	Enabled           bool
+	AddedOn           datetime.DateTime
+	Function          string
+	Input             []Input
+	Action            string
+	ActionConfig      map[string]interface{}
+}
+
+type Input struct {
+	Type     string
```
bharadwaj-aditya, line 224: What type is this ? Suggest you add sample values here to clarify what these fields are for

sources/mysql/data_test.go
```diff
@@ -72,7 +71,7 @@ func TestProcessDataRow(t *testing.T) {
 	conv.SetDataSink(func(table string, cols []string, vals []interface{}) {
 		rows = append(rows, spannerData{table: table, cols: cols, vals: vals})
 	})
-	ProcessDataRow(conv, tableId, colIds, conv.SrcSchema[tableId], conv.SpSchema[tableId], []string{"4.2", "6", "prisoner zero"}, internal.AdditionalDataAttributes{})
+	ProcessDataRow(conv, tableId, colIds, conv.SrcSchema[tableId], conv.SpSchema[tableId], []string{"4.2", "6", "prisoner zero"}, internal.AdditionalDataAttributes{}, nil)
```
bharadwaj-aditya, line 74: when will this field be nil ?

sources/oracle/data_test.go
```diff
@@ -201,31 +143,31 @@ func TestConvertMultiColData(t *testing.T) {
 			cols:   []string{"a", "b", "c"},
 			colIds: []string{"c1", "c2", "c3"},
 			vals:   []string{"6", "6.6", "1"},
-			ecols:  []string{"a", "b", "c"},
+			ecols:  []string{"c1", "c2", "c3"},
```
bharadwaj-aditya, line 146: Has the definition of the test generator changed to the take the column id's instead of the values ?

sources/oracle/infoschema.go
```diff
@@ -129,7 +129,11 @@ func (isi InfoSchemaImpl) ProcessData(conv *internal.Conv, tableId string, srcSc
 			conv.CollectBadRow(srcTableName, srcCols, values)
 			continue
 		}
-		ProcessDataRow(conv, tableId, commonColIds, srcSchema, spSchema, newValues)
+		mapSrcColIdToVal := make(map[string]string)
```
bharadwaj-aditya, line 132: nit:move to parent method if possible

sources/postgres/data.go
```diff
@@ -47,7 +47,17 @@ func ProcessDataRow(conv *internal.Conv, tableId string, colIds, vals []string)
 		conv.Unexpected(fmt.Sprintf("Error while converting data: %s\n", err))
 		conv.StatsAddBadRow(srcTableName, conv.DataMode())
 		conv.CollectBadRow(srcTableName, srcCols, vals)
-	} else {
+		return
+	}
+	toddl := InfoSchemaImpl{}.GetToDdl()
+	spCols, spVals, err = transformation.ProcessTransformation(conv, tableId, spCols, spVals, mapSrcColIdToVal, toddl, additionalAttributes)
```
bharadwaj-aditya, line 53: This should probably be movable to the parent class as it is common for all sources

transformation/constants.go
```diff
@@ -0,0 +1,186 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package transformation
+
+import "github.com/cloudspannerecosystem/harbourbridge/spanner/ddl"
+
+const (
+	MathOperation         = "mathOp"
```
bharadwaj-aditya, line 20: nit: Move the functions/operators to different enum so that it is clear

transformation/transformation.go
```diff
@@ -0,0 +1,703 @@
...
+}
+
+// Logical OR operator
+func logicalOr(a, b interface{}) (bool, error) {
+	switch a := a.(type) {
+	case bool:
+		if b, ok := b.(bool); ok {
+			return a || b, nil
+		}
+	case int64:
```
bharadwaj-aditya, line 442: Logical operators on non-boolean - are they meant to check if exists ?

transformation/transformation.go
```diff
@@ -0,0 +1,703 @@
...
+		return a == civil.Date{}, nil
+	case time.Time:
+		return a.IsZero(), nil
+	}
+	return false, fmt.Errorf("unsupported type for logical operation: %T", a)
+}
+
+func applyCompare(firstInput, operator, secondInput inputValue) (interface{}, error) {
+	switch operator.value {
+	case EqualToOperator:
```
bharadwaj-aditya, line 528: Can we emulate <= as > value false ?

transformation/transformation.go
```diff
@@ -0,0 +1,703 @@
...
+	} else if firstInput.dataType == secondInput.dataType && firstInput.dataType == ddl.Timestamp {
+		return firstInput.value.(time.Time).Equal(secondInput.value.(time.Time)), nil
+	} else if firstInput.dataType == secondInput.dataType && firstInput.dataType == ddl.String {
+		return firstInput.value == secondInput.value, nil
+	}
+	return false, fmt.Errorf("unsupported data types for comparison: %T, %T", firstInput.value, secondInput.value)
+}
+
+func compareGreaterThan(firstInput, secondInput inputValue) (bool, error) {
+	if firstInput.dataType == secondInput.dataType && firstInput.dataType == ddl.Int64 {
```
bharadwaj-aditya, line 558: this would need to be a number based comparison, not type based. 
Eg:
10 > 9.5

transformation/transformation.go
```diff
@@ -0,0 +1,703 @@
...
+	return false, fmt.Errorf("unsupported data types for comparison: %T, %T", firstInput.value, secondInput.value)
+}
+
+func applyMathOp(firstInput, operator, secondInput inputValue) (interface{}, error) {
+	switch operator.value.(string) {
+	case AddOperator:
+		switch firstInput.value.(type) {
+		case int64:
+			switch secondInput.value.(type) {
+			case int64:
```
bharadwaj-aditya, line 589: same as above - the output of this should be based on the expected output type.
So we can run int + float type operations.

bharadwaj-aditya (changes requested):

bharadwaj-aditya:
I think we can close this PR as not implementing.

PR #594, opened 2023-07-24
feat(backend): add postgres support for dataproc migration

forksumit:
This PR includes functionality to migrate Postgres tables via dataproc.
 
 * update `getInfoSchemaForShard()` to set `sourceProfileConnection` based on source types (instead of forcing to mysql)
 * update jdbcParams for postgres source
 * add unit test for postgres config
  
Sample CLI Command: `go run main.go schema-and-data -source=postgres --target-profile="project=demo-project,instance=demo-instance,dbName=demo_db" --source-profile="config=postgres_shard_dataproc.cfg" --log-level=DEBUG`

Related unit tests (dproc package): `go test -v github.com/cloudspannerecosystem/harbourbridge/dproc`


shreyakhajanchi (approved):
LGTM

dproc/dproc.go
```diff
@@ -71,6 +71,16 @@ func GetDataprocRequestParams(sourceProfile profiles.SourceProfile, targetProfil
...
+	} else if sourceProfile.Driver == constants.POSTGRES {
+		jdbcParams["url"] = fmt.Sprintf("jdbc:postgresql://%s:%s/%s?user=%s&password=%s", host, port, srcDb, user, pwd)
+		jdbcParams["driver"] = "org.postgresql.Driver"
+		if primaryKeys == "" {
+			jdbcParams["sql"] = fmt.Sprintf("select gen_random_uuid() as synth_id, t.* from %s.%s as t", srcSchema, srcTable)
+			primaryKeys = "synth_id"
+		} else {
+			jdbcParams["sql"] = fmt.Sprintf("select * from %s.%s", srcSchema, srcTable)
+		}
+		jdbcParams["jar"] = "gs://dataproc-templates/jars/postgresql-42.2.6.jar"
```
bharadwaj-aditya, line 83: it seems like every time there's a release this will need to be changed. Can we have a common location where the release version will be updated so that it gets picked up automatically without code change in HB ?
It seems like the mysql-connector is doing that.

dproc/dproc_test.go
```diff
@@ -108,13 +112,13 @@ func TestGetDataprocRequestParams(t *testing.T) {
 			errorExpected: false,
 		},
 		{
-			name: "valid mysql source profile and target profile with missing primary key in source",
```
bharadwaj-aditya, line 111: is the mysql test being replaced with the postgres test here ?

bharadwaj-aditya (changes requested):

dproc/dproc_test.go
forksumit: MySQL test for missing primary key got lost on remote branch merge. 
Fixed it , Commit: https://github.com/GoogleCloudPlatform/spanner-migration-tool/pull/594/commits/2caf17d33348b36cab183961f592bd52f878283d

dproc/dproc.go
forksumit: Valid Point!
Do we have any public bucket for harbourbridge or spanner, which we can reuse.
I'll also check with dataproc team, if they have any alternatives.

eenclona-google: Spoke with Surjit about this; so the jar file for the database connector should actually be supplied by the user. The expectation is they download whatever connector jar file they need and reference that as they run dataproc template. This also applies to mysql-connector jar file. The one we were referencing earlier is not guaranteed to be the most up to date.

eenclona-google: We should probably add another input box in the Dataproc config to allow user to pass in this parameter.

bharadwaj-aditya (approved):
LGTM

PR #595, opened 2023-07-25
Fix index conversion from source to Spanner

manitgupta:
Fixes  - 

1. Bug in the `name to id` implementation which added the columnIds in the `STORING` clause instead of the names.
2. In Spanner, the `STORING` clause of an `INDEX` cannot contain columns which are part of the primary of the base table. The base table's keys are already part of the keys of the index's underlying table and hence should be omitted from the `STORING` clause. This is a database agnostic conversion.


shreyakhajanchi (approved):
LGTM

bharadwaj-aditya (approved):
LGTM

PR #597, opened 2023-07-25
Update template paths to the latest release

Deep1998:
Update the template paths for reverse replication and streaming templates.


shreyakhajanchi (approved):
LGTM

PR #598, opened 2023-07-26
Renaming HB to Spanner migration tool - metadataDB name and move data…

darshan-sj:
Renaming HB to Spanner migration tool - metadata DB name and move data from old metadata DB to new metadata DB

- On init after the checkOrCreateMetadataDB method is called 
    - check if old metadata DB exists
    - If exists read all the data from that and write it to new Metadata DB
    - Drop old metadata DB if migration is successful

Tested:
- Launch SMT web, configure project and instance ID in UI. Successfully created and migrated the metadata to new DB
- Launch SMT web with project and instance ID preconfigured. Successfully created and migrated the metadata to new DB
- Launch SMT web, configure project and instance ID with doesn't have old metadata DB. Skipped the migration step

- [x] Tests pass


bharadwaj-aditya (approved):
LGTM

webv2/session/session_service.go
```diff
@@ -61,10 +65,105 @@ func SetSessionStorageConnectionState(projectId string, spInstanceId string) (bo
...
+		fmt.Printf("could not write to newMetadataDB. error=%v\n", err)
+		return
+	}
+
+	fmt.Println("Successfully wrote data to new metadata DB.")
+
+	err = adminClient.DropDatabase(ctx, &databasepb.DropDatabaseRequest{
+		Database: oldMetadataDbUri,
+	})
+	if err != nil {
```
shreyakhajanchi, line 164: In case the drop command fails, what will happen when user tried to set the instance next time? Will it not fail always because of failed insertion due to existing records?

webv2/helpers/helpers.go
```diff
@@ -37,7 +37,7 @@ const (
 	GOOGLE_SQL_DIALECT     = "Google Standard SQL"
 )
 
-const metadataDbName string = "harbourbridge_metadata"
+const metadataDbName string = "spannermigrationtool_metadata"
```
shreyakhajanchi, line 40: nit: will it be better to rename the table as spanner_migration_tool_metadata for better readability?

darshan-sj: `spanner_migration_tool_metadata` exceeds the allowed database name length in Spanner.

webv2/session/session_service.go
darshan-sj: Yes. Correct. 
This data movement could fail at any stage. 
1. If just drop database is failed, In the next attempts insertion will fail and this method will simply log the error and proceed. We cannot change the mutation to InsertOrUpdate mutation because we might overwrite new data in smt_metadata in subsequent init of SMT. 
2. And similarly, we cannot drop oldMetadataDB after the Insert into new metadataDB fails, We might lose old saved sessions.

The logic is consciously written like this to solve for a full happy path scenario. The priority is in any scenario there should be no data loss, and no overwriting of new data with old data.

shreyakhajanchi (approved):
LGTM

PR #603, opened 2023-07-27
Improve logging and audit metadata for dataproc migration mode

forksumit:
* Reuse same Context object for all dataproc jobs
* Move job waiting outside `TriggerDataprocTemplate()`
* Add logging/auditing for dataproc jobs right after triggering
* Add more context to log message
* Remove hardcoded `us-central1` from jobUrl
* Add project id param in dataproc job url
* Change `DataprocMetadata` from array to map with table id as key
* Add `SrcTable` and`DataprocJobStatus`  to audit metadata


bharadwaj-aditya (approved):
LGTM

PR #607, opened 2023-07-28
fixed bugs and friction log issues

shreyakhajanchi:
Fixes:

1. Array types are not supported for lowdowntime migrations, we should raise this earlier. Also make this nullable
2. Test connection error not shown for non sharded migration
3. When creating a connection profile, after testing the connection, when clicking on the 'Create Profile' , the test connection spinner keeps circling - After this I cannot add more shards to the connection profile
4. Resuming from saved session for a sharded migration becomes non sharded migration
5. Spanner migration tool gave unexpected message to 'confirm deletion' when clickin on 'Skip Table'. Post that the tool crashed 
6. Remove Prerequisites
7. The instruction panel on UI is not relevant - not updated data - P0
8. Download Session File - Rename in Download Artifacts - P2
9. Save session - Notes must not be null - bug - P0
10. Add a regex to handle valid inputs in save session
11. Add loader/spinner for test/connect profile
12. Add validator on connection profile name 
13. Add i icon on Configure Shard id and database name.


bharadwaj-aditya (approved):
LGTM

bharadwaj-aditya:
Only concern here is that some tests the column coverage has reduced. I'm guessing that is because the array type has been marked as Nullable.

PR #610, opened 2023-08-01
add casting in temp.query to support schema conversion for dataproc migration mode

forksumit:
Add spark sql casting in temp.query based on target datatypes in spanner to support schema conversion for dataproc migration mode

* generate dynamic jdbcSql based on column definations
* move jdbcSql logic to new function `getJdbcSql()`
* remove redundant params for dproc package functions
* remove hardcoded synth_id for missing primary keys
* append new temp.table and temp.query arguments generated from `getTempViewArgs()` to do data conversion compatible to target datatype
* modify unit tests

NOTE: currently it doesn't support complex datatypes likes (ARRAY/MAP/STRUCT)


dproc/dproc.go
```diff
@@ -57,22 +127,22 @@ func GetDataprocRequestParams(sourceProfile profiles.SourceProfile, targetProfil
...
-		} else {
-			jdbcParams["sql"] = fmt.Sprintf("select * from %s.%s", srcSchema, srcTable)
-		}
 		jdbcParams["jar"] = "gs://dataproc-templates/jars/mysql-connector-java.jar"
 	} else {
 		return DataprocRequestParams{}, fmt.Errorf("dataproc migration for source %s not supported", sourceProfile.Driver)
 	}
+	jdbcParams["sql"] = getJdbcSql(conv, sourceProfile.Driver, spTableID, primaryKeys)
+
+	outputPrimaryKeys := primaryKeys
```
shreyakhajanchi, line 142: why is this required in the template? What happens if user edits the primary key ?

forksumit: Currently, 
* In case of schema+data migration via template: Dataproc template uses primary key parameter to define primary key in target spanner database when doing 
* **[OUR CASE]** In case of data migration only (schema created already before triggering template) this parameter is not used (it even works if given any random value to the parameter)

BUT, In both cases, this parameter is set as required parameter, (throws error: `ValidationUtil: Violation: primaryKey jdbctospanner.output.primaryKey; must not be empty` if not given)

@enclona-google, Please add if i missed anything

forksumit: Looking at current dataproc template code, it's not a very simple change to identify migration type in the validation phase, I suggest until that change is done, We keep on passing the actual primary key as we're doing now in harbourbridge.

forksumit: Added comment explaining same
Commit: https://github.com/GoogleCloudPlatform/spanner-migration-tool/pull/610/commits/956615ac6f5c5c5ad57ffc972404a3702092a3e2

dproc/dproc.go
```diff
@@ -41,8 +42,77 @@ type DataprocRequestParams struct {
...
+		return "BOOLEAN"
+	case "STRING":
+		return "STRING"
+	case "TIMESTAMP":
+		return "TIMESTAMP"
+	case "DATE":
+		return "DATE"
+	case "NUMERIC":
+		return "DECIMAL"
+	default:
```
shreyakhajanchi, line 63: JSON type ?

shreyakhajanchi (commented):
LGTM, just 1 comment

dproc/dproc.go
forksumit: Currently complex datatypes (like JSON, ARRAY, SET) are not supported by the dataproc template

PR #613, opened 2023-08-01
bug fixes

shreyakhajanchi:
Fixes :
1. Removed data shard id
2. Increased parallelization to 20
3. JSON column only allows JSON and BYTES as spanner types, should allow STRING instead
4. Add more information about what target database requires - is it an existing db or a new db, etc
5. Added error log for - Loading schema in Hb misses tables intermittently
6. Foreign Key issue in sharded migration


sources/common/infoschema.go
```diff
@@ -63,10 +63,17 @@ type FkConstraint struct {
...
+	tableCount, err := GenerateSrcSchema(conv, infoSchema, numWorkers)
+	if err != nil {
+		return err
+	}
 	initPrimaryKeyOrder(conv)
 	initIndexOrder(conv)
 	SchemaToSpannerDDL(conv, infoSchema.GetToDdl())
+	if tableCount != len(conv.SpSchema) {
+		fmt.Printf("Failed to load all the source tables, source table count: %v, processed tables:%v\n", tableCount, len(conv.SpSchema))
+		return fmt.Errorf("failed to load all the source tables, source table count: %v, processed tables:%v", tableCount, len(conv.SpSchema))
```
manitgupta, line 75: We should also add the action expected from the user on seeing this error. 
Example - Please retry connecting to the source database to load tables.

ui/src/app/components/sharded-dataflow-migration-details-form/sharded-dataflow-migration-details-form.component.html
```diff
@@ -83,11 +78,6 @@
                     <input matInput placeholder="Connection profile name" type="text" formControlName="newSourceProfile"
                         required="true" />
                 </mat-form-field>
-                <mat-form-field appearance="outline">
-                    <mat-label>Data Shard Id</mat-label>
-                    <input matInput placeholder="Data ShardId" type="text" formControlName="dataShardId"
-                        required="true" />
-                </mat-form-field>
```
manitgupta, line 90: Is it possible to generate a `dataShardId` and display it on the UI  in a label instead of removing it? We can also add a toolTip on what this randomly generated string means. 

For the UI, the generated `datashardId` will be passed from the UI, and in the case of CLI, if it is empty, one will be auto-generated using the `generateName` changes you have made above. That sounds reasonable.

This will help the user in understanding the generated resources result later on, since otherwise a bunch of jobs will be randomly listed with no direct way to understanding which job contains which set of logical shards.

ui/src/app/components/target-details-form/target-details-form.component.html
```diff
@@ -1,6 +1,9 @@
 <div mat-dialog-content>
   <form [formGroup]="targetDetailsForm" class="target-detail-form">
-    <h2>Spanner Database Details</h2>
+    <h2>Spanner Database Details<mat-icon class="configure"
+      matTooltip='Schema & Schema and data migration: Specify new database name or existing database with empty schema. Data migration: Specify existing database with tables already created'>
```
manitgupta, line 4: Is there a toolTip available on the prepare-migration-page's configure database button itself? That seems like a more suitable place to add this.

conversion/conversion.go
```diff
@@ -332,6 +332,14 @@ func dataFromDatabaseForDataflowMigration(targetProfile profiles.TargetProfile,
 		for _, l := range p.LogicalShards {
 			dbNameToShardIdMap[l.DbName] = l.LogicalShardId
 		}
+		if p.DataShardId == "" {
+			dataShardId, err := utils.GenerateName("smt-datashard")
+			dataShardId = strings.Replace(dataShardId, "_", "-", -1)
```
bharadwaj-aditya, line 337: why is this required ?

conversion/conversion.go
```diff
@@ -332,6 +332,14 @@ func dataFromDatabaseForDataflowMigration(targetProfile profiles.TargetProfile,
 		for _, l := range p.LogicalShards {
 			dbNameToShardIdMap[l.DbName] = l.LogicalShardId
 		}
+		if p.DataShardId == "" {
+			dataShardId, err := utils.GenerateName("smt-datashard")
```
bharadwaj-aditya, line 336: Log that the data shard id is being generated

bharadwaj-aditya (approved):
LGTM

ui/src/app/components/sharded-dataflow-migration-details-form/sharded-dataflow-migration-details-form.component.html
shreyakhajanchi: done

PR #623, opened 2023-08-09
Typecast array datatype to string in all the migrations

shreyakhajanchi:
Implementing short term fix. Details below:

Array datatype

- UI doesn’t show if a column has array datatype in spanner column equivalent but shows in SQL, this is because backend sends only datatype string to frontend in typemap instead of type object
- We don’t support Array<Primitive type> to string conversions
- MYSQL
     1. Datstream reads set datatype of MySQL as STRING : https://cloud.google.com/datastream/docs/unified-types#mapping_of_mysql_data_types_to_unified_types
     2. - Dataflow tries to write this as array because of array datatype in session file but fails because array is unsupported datatype in dataflow template

- PostgreSQL:
     1. Postgresql has array datatype support in source
     2. Datstream reads array datatype of PostgreSQL as UnSupported type and maps that value to NULL 
     3. Dataflow tries to write this as array because of array datatype in session file but fails because array is unsupported 
     datatype in dataflow template

- **Short term Solution**: Map array types to string
- Mid-term solution: 
   1. Add UI support for array datatypes display and edit
   2. Add support for Array<Primitive type> to string conversions and revert the short term fix

- Long term fix - Add support for array data types in datastream & dataflow templates


internal/reports/report_helpers.go
```diff
@@ -418,7 +418,7 @@ var IssueDB = map[internal.SchemaIssue]struct {
 	internal.RowLimitExceeded:            {Brief: "Non key columns exceed the spanner limit of 1600 MB. Please modify the column sizes", severity: errors},
 	internal.ShardIdColumnAdded:          {Brief: "column was added because this is a sharded migration and this column cannot be dropped", severity: note},
 	internal.ShardIdColumnPrimaryKey:     {Brief: "column is not a part of primary key. You may go to the Primary Key tab and add this column as a part of Primary Key", severity: suggestion},
-	internal.ArrayTypeNotSupported:       {Brief: "Array datatype is not supported in minimal downtime migration", severity: warning},
+	internal.ArrayTypeNotSupported:       {Brief: "Array datatype is not supported in minimal downtime migration and they will be mapped to STRING", severity: warning},
```
manitgupta, line 421: If we are mapping array to string for all migrations, should be remove "minimal downtime migration" from this text?

shreyakhajanchi: done

manitgupta (approved):

bharadwaj-aditya (approved):
LGTM

PR #628, opened 2023-08-28
fixed few bugs and QOL improvements

shreyakhajanchi:
Fixed the following bugs:
1. Removing the unique column as primary key and manually adding some other column as primary key in UI should remove the missing primary key warning from the warnings section, but it does not
2. Interleaving a table then deleting the PK rule for shard id column is creating two entries of the interleaved table in case of sharded migration
3. For POC migration - note about minimal downtime migration shows up
4. Enable buttons only when profiles are created in the sharded migration - datastream config form
5. Usability Suggestion: Able to skip all tables from Spanner migration, yet it allowed to proceed with later migration stages like Schema and Data. We should disable prepare migration in case there are no tables to be migrated or disable skipping all tables.
6. Typo: Add Shard id rule


ui/src/app/components/workspace/workspace.component.ts
```diff
@@ -397,6 +397,11 @@ export class WorkspaceComponent implements OnInit, OnDestroy {
             data: { message: "Please configure spanner project id and instance id to proceed", type: 'error', title: 'Configure Spanner' },
             maxWidth: '500px',
           })
+        } else if (Object.keys(this.conv.SpSchema).length == 0) {
+          this.dialog.open(InfodialogComponent, {
+            data: { message: "Please restore some table(s) to proceed with the migration", type: 'error', title: 'All tables skipped' },
+            maxWidth: '500px',
+          })
```
manitgupta, line 404: Should we be disabling the `Prepare migration button` instead of showing a popup error?

shreyakhajanchi: We show this pop-up for few other scenarios like error in the schema conversion, spanner instance not connected etc. To keep the behaviour similar across all the scenarios added a popup instead of disabling

internal/reports/report_helpers.go
```diff
@@ -351,7 +328,7 @@ func buildTableReportBody(conv *internal.Conv, tableId string, issues map[string
 					}
 
 				case internal.AutoIncrementIndex:
-					str := fmt.Sprintf(" %s for Table %s and Column  %s", IssueDB[i].Brief, spSchema.Name, spColName)
+					str := fmt.Sprintf(" %s for Table '%s' and Column '%s'", IssueDB[i].Brief, spSchema.Name, spColName)
```
bharadwaj-aditya, line 331: wondering why the quotes have been added on all the logs. Is this for readability ? Has this been taken up while logging tables throughput the code base?

bharadwaj-aditya (approved):
Minor comments, but LGTM

internal/reports/report_helpers.go
shreyakhajanchi: This has been added for readability and to make it consistent across all the issues/warnings. It is not used for logging

internal/convert.go
```diff
@@ -411,24 +412,24 @@ func (conv *Conv) AddPrimaryKeys() {
...
 	var columnLevelIssues map[string][]SchemaIssue
 	if tableIssues, ok := conv.SchemaIssues[tableId]; ok {
 		columnLevelIssues = tableIssues.ColumnLevelIssues
 	} else {
 		columnLevelIssues = make(map[string][]SchemaIssue)
-	} 
+	}
 	issues := columnLevelIssues[colId]
-	issues = append(issues, MissingPrimaryKey)
+	issues = append(issues, schemaIssue)
```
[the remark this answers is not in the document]
shreyakhajanchi: done

manitgupta (approved):

PR #630, opened 2023-09-11
Configure MaxWorkers, NumWorkers and ServiceAccount params

Deep1998:
Configure MaxWorkers, NumWorkers and ServiceAccount params for dataflow jobs via SMT.
Context: This pr does not cover the end to end suppor for dataflow params, this is a stopgap to unblock zee because they need these params for migration. The detailed support will be added as a followup.

Tested:
- Dataflow form: NumWorkers, Maxworkers only accepting positive integers. Tool tips working correctly.
- Non-sharded Flow: NumWorkers, MaxWorkers and ServiceAccountEmail set correctly.  Without configuring, defaults set correctly.
- Sharded Flow:  NumWorkers, MaxWorkers and ServiceAccountEmail set correctly.
Screenshot 2023-09-13 at 5 50 07 PM


ui/src/app/components/prepare-migration/prepare-migration.component.ts
```diff
@@ -437,10 +443,14 @@ export class PrepareMigrationComponent implements OnInit {
...
-        hostProjectId: localStorage.getItem(Dataflow.HostProjectId) as string
+        hostProjectId: localStorage.getItem(Dataflow.HostProjectId) as string,
+        maxWorkers: localStorage.getItem(Dataflow.MaxWorkers) as string,
+        numWorkers: localStorage.getItem(Dataflow.NumWorkers) as string,
+        serviceAccountEmail: localStorage.getItem(Dataflow.ServiceAccountEmail) as string
       }
+      console.log(this.dataflowConfig)
       this.isDataflowConfigurationSet = localStorage.getItem(Dataflow.IsDataflowConfigSet) as string === 'true'
-      if (this.isSharded) {
+      if (this.isDataflowConfigurationSet) {
```
Deep1998, line 453: @manitgupta switched this if condition to handle both sharded and non sharded flow, any thing that I could be missing?

ui/src/app/services/fetch/fetch.service.ts
```diff
@@ -23,7 +23,7 @@ import IStructuredReport from 'src/app/model/structured-report'
   providedIn: 'root',
 })
 export class FetchService {
-  private url: string = window.location.origin
+  private url: string = 'http://localhost:8080'//window.location.origin
```
manitgupta, line 26: Revert

ui/src/app/components/prepare-migration/prepare-migration.component.ts
manitgupta: This doesn't look right since `setDataflowDetailsForShardedMigrations` call would be made for non-sharded flow as well with this case.

manitgupta (commented):
1. Please add testing details on the PR with description of the jobs launched by setting the different parameters (`MaxWorkers`, `NumWorkers` etc)
2. We have recently add tooltips in Harbourbridge in a number of places to explain the different settings. Please add the same tool tips for the Dataflow form as well.
3. Please update unit tests for the newly added parameters. If the unit tests do not exist for the file, please create an internal taskflow bug for tracking add to the following hotlist: b/hotlists/5291617

ui/src/app/components/prepare-migration/prepare-migration.component.ts
Deep1998: Right, reverted.

Deep1998:
@manitgupta created b/300068690 to track unit testing for dataflow params.

streaming/streaming.go
```diff
@@ -61,12 +58,15 @@ type DatastreamCfg struct {
...
-	Network            string
-	Subnetwork         string
-	DbNameToShardIdMap map[string]string
+	JobName             string
+	Location            string
+	HostProjectId       string
+	Network             string
+	Subnetwork          string
+	MaxWorkers          string
+	NumWorkers          string
```
bharadwaj-aditya, line 67: is this the minimum workers or default ?

streaming/streaming.go
```diff
@@ -61,12 +58,15 @@ type DatastreamCfg struct {
...
-	HostProjectId      string
-	Network            string
-	Subnetwork         string
-	DbNameToShardIdMap map[string]string
+	JobName             string
+	Location            string
+	HostProjectId       string
+	Network             string
+	Subnetwork          string
+	MaxWorkers          string
```
bharadwaj-aditya, line 66: why is this a string instead of an integer ?

streaming/streaming.go
Deep1998: the initial number of workers

streaming/streaming.go
```diff
@@ -459,9 +459,46 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
...
+			return fmt.Errorf("could not parse MaxWorkers parameter %s, please provide a positive integer as input", dataflowCfg.MaxWorkers)
+		}
+		maxWorkers = int32(intVal)
+	}
+	if dataflowCfg.NumWorkers != "" {
+		intVal, err := strconv.ParseInt(dataflowCfg.NumWorkers, 10, 64)
+		if err != nil {
+			return fmt.Errorf("could not parse NumWorkers parameter %s, please provide a positive integer as input", dataflowCfg.NumWorkers)
+		}
+		numWorkers = int32(intVal)
```
bharadwaj-aditya, line 477: please add checks on max/min value of these numbers

streaming/streaming.go
Deep1998: It is easier to handle unspecified cases (via '') in string as opposed to treating 0 as unspecified in ints. 
Plus for the CLI, the streaming cfg is manually specified, where the user can leave unspecified with string as opposed to providing 0 for integer.

We can use ints as well but it will require some additional code changes to cover CLI and UI both.

streaming/streaming.go
```diff
@@ -459,9 +459,52 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
 			dataflowSubnetwork = fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/%s/regions/%s/subnetworks/%s", dataflowHostProjectId, dataflowCfg.Location, dataflowCfg.Subnetwork)
 		}
 	}
-
-	launchParameters := createLaunchParameters(dataflowCfg, inputFilePattern, project, datastreamCfg, instance, dbName, streamingCfg, dataflowSubnetwork, workerIpAddressConfig)
-
+	fmt.Printf("Working with dfCfg: %v\n", dataflowCfg)
```
manitgupta, line 462: I think this was for testing, do we want to print this?

streaming/streaming.go
```diff
@@ -459,9 +459,52 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
...
+	maxWorkers := int32(50)
+	numWorkers := int32(1)
```
manitgupta, line 464: Specify max and num worker defaults as global constants

streaming/streaming.go
```diff
@@ -459,9 +459,52 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
...
+	if dataflowCfg.MaxWorkers != "" {
+		intVal, err := strconv.ParseInt(dataflowCfg.MaxWorkers, 10, 64)
+		if err != nil {
+			return fmt.Errorf("could not parse MaxWorkers parameter %s, please provide a positive integer as input", dataflowCfg.MaxWorkers)
+		}
```
manitgupta, line 469: From the UI/CLI flow, can't the validations be performed upfront instead of at the orchestration stage? 
Seems like sanitised values should be passed to this layer.

streaming/streaming.go
```diff
@@ -459,9 +459,52 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
...
+		maxWorkers = int32(intVal)
+		if maxWorkers < 1 || maxWorkers > 1000 {
+			return fmt.Errorf("maxWorkers should be a positive integer (> 0) that does not exceed 1000")
+		}
```
manitgupta, line 473: `1` and `1000` should be declared as global constants so that its easy to understand what does the check mean

ui/src/app/components/dataflow-form/dataflow-form.component.html
```diff
@@ -16,6 +16,24 @@ <h2>Dataflow Details</h2>
...
+      [matTooltipPosition]="'right'" matTooltipHideDelay="100000">
+        <mat-label>Number of Workers</mat-label>
+        <input matInput placeholder="1" type="text" formControlName="numWorkers"/>
+      </mat-form-field>
+      <br>
+      <mat-form-field class="full-width" appearance="outline" matTooltip="Set the service account to run the dataflow job(s) as"
+      [matTooltipPosition]="'right'" matTooltipHideDelay="100000">
+        <mat-label>Service Account Email</mat-label>
+        <input matInput placeholder="Service Account Email" type="text" formControlName="serviceAccountEmail"/>
+      </mat-form-field>
```
manitgupta, line 36: Can you please add a screenshot of this form to the PR description?

ui/src/app/components/dataflow-form/dataflow-form.component.ts
```diff
@@ -28,10 +31,15 @@ export class DataflowFormComponent implements OnInit {
 
   updateDataflowDetails() {
     let formValue = this.dataflowForm.value
-    localStorage.setItem(Dataflow.Network, formValue.network)
-    localStorage.setItem(Dataflow.Subnetwork, formValue.subnetwork)
-    localStorage.setItem(Dataflow.HostProjectId, formValue.hostProjectId)
-    localStorage.setItem(Dataflow.IsDataflowConfigSet, "true")
+    if (formValue.network||formValue.subnetwork||formValue.maxWorkers||formValue.numWorkers||formValue.serviceAccountEmail) {
```
manitgupta, line 34: Why is the `if` condition required?

streaming/streaming.go
```diff
@@ -459,9 +459,52 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
...
+			"transformationContextFilePath": streamingCfg.TmpDir + "transformationContext.json",
+		},
+		Environment: &dataflowpb.FlexTemplateRuntimeEnvironment{
+			MaxWorkers:            maxWorkers,
+			NumWorkers:            numWorkers,
+			ServiceAccountEmail:   dataflowCfg.ServiceAccountEmail,
+			AutoscalingAlgorithm:  2, // 2 corresponds to AUTOSCALING_ALGORITHM_BASIC
+			EnableStreamingEngine: true,
+			Network:               dataflowCfg.Network,
+			Subnetwork:            dataflowSubnetwork,
```
manitgupta, line 504: We should add the check for validating that `subnet` region is the same as the `dataflowRegion` that we were discussing in the reverse replication PR here.

manitgupta (commented):
I think this is a good time to evaluate if any other aspect of a Dataflow API call should be made configurable. 

Specifying that a Dataflow job uses Dataflow `Prime` comes at the top of mind for me. 

Can you look through the Dataflow API and assess this? Maybe talking to someone in Dataflow to get a sense of most commonly used configurations will also be a good exercise. 

Finally, please test these flows via CLI as well. Thanks for creating a tracking bug for the missing unit tests!

ui/src/app/components/dataflow-form/dataflow-form.component.ts
shreyakhajanchi, line 34: please format this file

ui/src/app/components/dataflow-form/dataflow-form.component.ts
```diff
@@ -17,8 +17,11 @@ export class DataflowFormComponent implements OnInit {
     private dialofRef: MatDialogRef<DataflowFormComponent>
   ) {
     this.dataflowForm = new FormGroup({
-      network: new FormControl('', Validators.required),
-      subnetwork: new FormControl('', Validators.required),
+      network: new FormControl(''),
+      subnetwork: new FormControl(''),
```
shreyakhajanchi, line 21: since, we have removed the required validators will it work in all cases where either network or sub network is specified and the other is not?

Deep1998: There is still a backend check which will fail if only 1 is specified. However, we eventually need ot remove that check as well because network and subnetwork dont need to be specified together

ui/src/app/components/dataflow-form/dataflow-form.component.ts
```diff
@@ -17,8 +17,11 @@ export class DataflowFormComponent implements OnInit {
...
+      numWorkers: new FormControl('1',[Validators.required, Validators.pattern('^[1-9][0-9]*$')]),
+      maxWorkers: new FormControl('50', [Validators.required, Validators.pattern('^[1-9][0-9]*$')]),
```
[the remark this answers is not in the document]
Deep1998: Yes but this way it ensures user is aware of what they are setting via the UI. We autofill 1 and 50 anyways, so empty means the user specifically chose to remove these values, which I think is better they fill correctly than to allow empty.

ui/src/app/components/dataflow-form/dataflow-form.component.ts
Deep1998: Not needed anymore, removed

Deep1998: removed the if

streaming/streaming.go
Deep1998: We do perform some level of validations at the UI layer. But the backend should still be robust and have handling, especially given we have so many ways to reach here (UI sharded, non sharded, cli). Having prechecks help fail fast but think its best to keep our backend fullproof i.e. ensure job launch means params are right as much as possible.

streaming/streaming.go
Deep1998: Will take up during network config handling as a followup since this is a stopgap for zee.

Deep1998:
@manitgupta since this is a stopgap, not adding other dataflow params right now. The follow up excercise is coming up with a proper way (maybe a new UI box) for dataflow params, cover all the relevant params and handl ethe prechecks etc.

streaming/streaming.go
manitgupta: I don't think thats a good way to handle this, since it would mean every layer of the code needs to have these type of validations.
Validating input parameters at the top request layer and passing sanitised inputs to the underlying execution layers is a pretty standard practice. 

Since we are parsing integers here, this is the first instance of parsing validation logic we have had to add here. I suggest we think deeper about this.

ui/src/app/components/dataflow-form/dataflow-form.component.html
manitgupta: Please follow the HB standard methodlogy of adding tooltips - 
https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/9a38536a918edb3b5da4125f88ea59a35dd86f2a/ui/src/app/components/prepare-migration/prepare-migration.component.html#L53

Screenshot - https://screenshot.googleplex.com/4ToDsifpc82DKdP

webv2/web.go
```diff
@@ -415,15 +421,19 @@ func setDataflowDetailsForShardedMigrations(w http.ResponseWriter, r *http.Reque
 	}
 	var dataflowLocation DataflowLocation
 	err = json.Unmarshal(reqBody, &dataflowLocation)
+	fmt.Printf("dataflowLocation: %v\n", dataflowLocation)
```
manitgupta, line 424: Remove

manitgupta (approved):
1. Please create tracker bugs for the comments/discussions which couldn't be addressed in this PR due to the customer need for this feature. 
2. Please link the created bugs on this PR.

shreyakhajanchi (approved):
LGTM. Please rebase once and run ng build to update the build files

bharadwaj-aditya (approved):
LGTM

Please ensure all the feedback from this PR not immediately picked up is captured as follow up work.

PR #631, opened 2023-09-11
Added support to specify vpcNetwork and vpcSubnetwork for reverse replication

Deep1998:
User can now provide the network and subnetwork for the reverse replication jobs. Both parameters need to be provided, if only one is provided, the script errors out.

The subnetwork region used is the same as the region specified by the dataflowRegion parameter.

### DIFF
--- a/reverse_replication/launcher.go
+++ b/reverse_replication/launcher.go
@@ -45,6 +45,10 @@ var (
 	sourceShardsFilePath string
 	sessionFilePath      string
 	machineType          string
+	vpcNetwork           string
+	vpcSubnetwork        string
+	vpcHostProjectId     string
+	serviceAccountEmail  string
 	orderingWorkers      int
 	writerWorkers        int
 )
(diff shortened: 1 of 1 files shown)


aksharauke (approved):

reverse_replication/launcher.go
```diff
@@ -120,6 +124,12 @@ func prechecks() error {
 	if pubSubEndpoint == "" {
 		pubSubEndpoint = fmt.Sprintf("%s-pubsub.googleapis.com:443", dataflowRegion)
 	}
+	if vpcNetwork != "" && vpcSubnetwork == "" {
+		return fmt.Errorf("if VPC network is specified, VPC subnetwork cannot be empty")
+	}
+	if vpcNetwork == "" && vpcSubnetwork != "" {
```
manitgupta, line 130: We should also add a check which validates that the `subnetwork` provided in the input has the same region as the `dataflowRegion` parameter and fail fast.

manitgupta (changes requested):
I don't think the present set of changes will work for launching Dataflow inside a Shared VPC, please add work for it as well - 

https://cloud.google.com/dataflow/docs/guides/specifying-networks#shared

reverse_replication/launcher.go
Deep1998: Will add in next iteration

reverse_replication/launcher.go
```diff
@@ -120,6 +124,12 @@ func prechecks() error {
...
+		return fmt.Errorf("provide both VPC network and subnetwork to use private connectivity")
```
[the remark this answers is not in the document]
Deep1998: Removed checks

reverse_replication/launcher.go
```diff
@@ -45,6 +45,10 @@ var (
 	sourceShardsFilePath string
 	sessionFilePath      string
 	machineType          string
+	vpcNetwork           string
+	vpcSubnetwork        string
+	vpcProjectId         string
```
manitgupta, line 50: Nit: Suggest renaming to `vpcHostProjectId` to keep parity with GCP docs - 
https://cloud.google.com/dataflow/docs/guides/specifying-networks#shared:~:text=that%20HOST_PROJECT_ID

manitgupta (approved):

manitgupta:
Please merge this after validating `shared VPC` is working

PR #638, opened 2023-09-21
docs: Add documentation for dataflow config

Deep1998:
Add documentation for dataflow config.

Jekyll link: https://deep1998.github.io/harbourbridge/ui/prepare-migration/dataflow-tuning.html

### DIFF
--- a/docs/ui/prepare-migration/dataflow-tuning.md
+++ b/docs/ui/prepare-migration/dataflow-tuning.md
@@ -1,0 +1,107 @@
+---
+layout: default
+title: Tune Dataflow (Optional)
+parent: Prepare Migration Page
+grand_parent: SMT UI
+nav_order: 4
+---
+
+# Tune Dataflow (Optional)
+{: .no_toc }
+
+{: .important }
+
+This is relevant especially if you want to run Dataflow inside a VPC.
+
+In case of minimal downtime migration, the dataflow jobs launched by the Spanner Migration Tool can be **optionally** tuned with custom runtime environment variables such as MaxWorkers, NumWorkers, specifying [networks and subnetworks](https://cloud.google.com/dataflow/docs/guides/specifying-networks) etc. Tuning refers to tweaking these parameters to run dataflow is a custom configuration.
+
+## Tuning use cases
+
+SMT by default launches dataflow with a preset configuration. However, this may not be applicable to all use cases. 
+Some use cases when the user would want to tweak the jobs are:
+- Dataflow machines should run inside a VPC.
+- Dataflow and Spanner should run in separate projects for cost tracking.
+- Use a custom service account to launch the Dataflow job.
+- Apply labels for better cost tracking for the jobs.
+
+
+{: .highlight }
+
+To configure dataflow, first specify the target database in the 'Configure Spanner Database' step. This enables the configure button for the remaining steps.
+
+![](https://services.google.com/fh/gumdrop/preview/misc/dataflow-form.png)
+
+<details open markdown="block">
+  <summary>
+    Table of contents
+  </summary>
+  {: .text-delta }
+1. TOC
+{:toc}
+</details>
+
+{: .highlight }
+
+SMT exposes the most frequently changed dataflow configurations to the user. Please reach out to us if you have a use-case that is not satisfied by the provided configurations.
+
+### VPC Host ProjectId
+Specify the project id of the VPC that you want to use. This is required in order to use private connectivity. By default, this is assumed to be the same as Spanner project. Ensure this is specified if also specifying a network and subnetwork.
+
+If using a shared VPC, a common practice is to have it in a separate project. Ensure this field specifies the correct host project for shared VPC use cases.
+
+Present under the Networking section of the form. 
+
+### VPC Network
+Specify the name of the VPC network to use. For private connectivity, specify both the VPC network and subnetwork. If no network and subnetwork is provided, the [default](https://cloud.google.com/dataflow/docs/guides/specifying-networks#specifying_a_network_and_a_subnetwork) network is used.
+
+Present under the Networking section of the form. 
+
+### VPC Subnetwork
+Specify the name of the VPC subnetwork to use. For private connectivity, specify both the VPC network and subnetwork. If no network and subnetwork is provided, the [default](https://cloud.google.com/dataflow/docs/guides/specifying-networks#specifying_a_network_and_a_subnetwork) network is used.
+
+Present under the Networking section of the form. 
+
+{: .highlight }
+
+SMT sets the IP configuration based on VPC network and subnetwork. If either network or subnetwork is provided (running inside a VPC), the public IPs are disabled (IPConfiguration is private). If neither are provided, the IP configuration is set to PUBLIC.
+
+### Max Workers
+Specify the max workers for the dataflow job(s). By default, set to 50.
+
+Present under the Performance section of the form. 
+
+### Number of Workers
+Specify the initial number of workers for the dataflow job(s). By default, set to 1.
+
+Present under the Performance section of the form. 
+
+### Machine Type
+The machine type to use for the job, eg: n1-standard-2. Use default machine type if not specified.
+
+Present under the Performance section of the form. 
+
+### Service Account Email
+Specify a custom service account email to run the job as. Uses the default compute engine service account if not specified. For more details, click [here](https://cloud.google.com/dataflow/docs/reference/pipeline-options#security_and_networking).
+
+### Additional User Labels
+Additional user labels to be specified for the job via a JSON string. Example: { "name": "wrench", "mass": "1kg", "count": "3" }.
+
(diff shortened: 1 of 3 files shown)


docs/ui/prepare-migration/dataflow-config.md
```diff
@@ -0,0 +1,47 @@
+---
+layout: default
+title: Configure Dataflow (Optional)
+parent: Prepare Migration Page
+grand_parent: SMT UI
+nav_order: 4
+---
+
+# Setting up connection profiles
```
manitgupta, line 9: Change title

docs/ui/prepare-migration/dataflow-config.md
```diff
@@ -0,0 +1,47 @@
...
+{: .no_toc }
+
+In case of minimal downtime migration, the dataflow jobs launched by the Spanner Migration Tool can be OPTIONALLY configured with custom runtime environment variables such as MaxWorkers, NumWorkers etc.
```
manitgupta, line 12: Nit: Optionally as bold instead of in capitals

docs/ui/prepare-migration/dataflow-config.md
```diff
@@ -0,0 +1,47 @@
...
+{: .no_toc }
+
+In case of minimal downtime migration, the dataflow jobs launched by the Spanner Migration Tool can be OPTIONALLY configured with custom runtime environment variables such as MaxWorkers, NumWorkers etc. 
+
+{: .highlight }
+The step shows configured (via a green tick) by default, denoting it is optional to configure.
+
+To configure dataflow, first specify the target database in the 'Configure Spanner Database' step. This enables the configure button for the remaining steps. The configurable fields for Dataflow are mentioned below.
+
+![](https://services.google.com/fh/files/misc/dataflow-config.png)
```
manitgupta, line 19: The screenshot looks too small to understand the different fields in the form. Would suggest snipping and having only the form in the screenshot for clarity

docs/ui/prepare-migration/dataflow-config.md
```diff
@@ -0,0 +1,47 @@
...
+  <summary>
+    Table of contents
+  </summary>
+  {: .text-delta }
+1. TOC
+{:toc}
+</details>
+
+## Host Project Id
+Specify the project id of the VPC that you want to use. This is required in order to use private connectivity. By default, this is assumed to be the same as Spanner project.
```
manitgupta, line 31: We need to specify this is to be used for shared VPC, and the project used here will be the host project

docs/ui/prepare-migration/dataflow-config.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Setting up connection profiles
+{: .no_toc }
+
+In case of minimal downtime migration, the dataflow jobs launched by the Spanner Migration Tool can be OPTIONALLY configured with custom runtime environment variables such as MaxWorkers, NumWorkers etc. 
+
+{: .highlight }
+The step shows configured (via a green tick) by default, denoting it is optional to configure.
+
+To configure dataflow, first specify the target database in the 'Configure Spanner Database' step. This enables the configure button for the remaining steps. The configurable fields for Dataflow are mentioned below.
```
manitgupta, line 17: We should add the link of the Dataflow documentation which talks about networks and subnetworks.

manitgupta: We should also add reference to general Dataflow configuration parameters (which may not be part of the form) here for completeness.

docs/ui/prepare-migration/dataflow-config.md
```diff
@@ -0,0 +1,47 @@
...
+Specify the name of the VPC subnetwork to use. For private connectivity, specify both the VPC network and subnetwork,
+
+## Max Workers
+Specify the max workers for the dataflow job(s). By default, set to 50.
+
+## Number of Workers
+Specify the initial number of workers for the dataflow job(s). By default, set to 1.
+
+## Service Account Email
+Specify a custom service account email to run the job as. Uses the default compute engine service account if not specified.
```
manitgupta, line 46: Need to add link to further documentation here

docs/ui/prepare-migration/dataflow-config.md
Deep1998: This needs to be specified for all cases: normal as well as shared VPCs. Added a line for shared VPC as well though.

docs/ui/prepare-migration/dataflow-config.md
```diff
@@ -0,0 +1,53 @@
...
+  {: .text-delta }
+1. TOC
+{:toc}
+</details>
+
+{: .highlight }
+Currently, we only allow configuring a subset of options available in Dataflow via SMT. For the full list of options available, click [here](https://cloud.google.com/dataflow/docs/reference/pipeline-options).
+
+## Host Project Id
+Specify the project id of the VPC that you want to use. This is required in order to use private connectivity. By default, this is assumed to be the same as Spanner project.
```
bharadwaj-aditya, line 35: Let's mention what are the defaults, i.e. what happens if these fields are not provided

bharadwaj-aditya (changes requested):

docs/ui/prepare-migration/dataflow-config.md
Deep1998: Done

docs/ui/prepare-migration/dataflow-tuning.md
```diff
@@ -0,0 +1,88 @@
...
+layout: default
+title: Tune Dataflow (Optional)
+parent: Prepare Migration Page
+grand_parent: SMT UI
+nav_order: 4
+---
+
+# Tune Dataflow (Optional)
+{: .no_toc }
+
```
manitgupta, line 11: Let's an `{ important }` tag that the top saying "This configuration is relevant especially you want to run Dataflow inside a VPC."

docs/ui/prepare-migration/dataflow-tuning.md
```diff
@@ -0,0 +1,88 @@
...
+    Table of contents
+  </summary>
+  {: .text-delta }
+1. TOC
+{:toc}
+</details>
+
+{: .highlight }
+
+Currently, we only allow configuring a subset of options available in Dataflow via SMT. For the full list of options available, click [here](https://cloud.google.com/dataflow/docs/reference/pipeline-options).
```
manitgupta, line 32: Use `Spanner migration tool` instead of "we".

Lets also rephrase this to - SMT exposes the most frequently changed dataflow configurations to the user. Please reach out to us <insert github issue ref> if you have a use-case that is not satisfied by the provided configurations.

IMO thinking the URL for full pipeline options is not helpful here since in the customer will not be use those configurations via HB, so it can be misleading.

docs/ui/prepare-migration/dataflow-tuning.md
```diff
@@ -0,0 +1,88 @@
...
+
+### VPC Network
+Specify the name of the VPC network to use. For private connectivity, specify both the VPC network and subnetwork. If no network and subnetwork is provided, the [default](https://cloud.google.com/dataflow/docs/guides/specifying-networks#specifying_a_network_and_a_subnetwork) network is used.
+
+Present under the Networking section of the form. 
+
+### VPC Subnetwork
+Specify the name of the VPC subnetwork to use. For private connectivity, specify both the VPC network and subnetwork. If no network and subnetwork is provided, the [default](https://cloud.google.com/dataflow/docs/guides/specifying-networks#specifying_a_network_and_a_subnetwork) network is used.
+
+Present under the Networking section of the form.
```
manitgupta, line 49: Let's add a note here about what we do under which type of configuration as far as enabling/disabling public IPs for the Dataflow VMs are concerned.

For example - We disable public IPs when running inside a VPC. On that note, I think we should also expose enabling/disabling public IPs directly on the UI. Krishna faced this issue and I think its a valid configuration to add.

Deep1998: Added a note explaining this. 

As for the option to configure this, we decided to remove this based on recommendations for the dataflow team.

docs/ui/prepare-migration/dataflow-tuning.md
```diff
@@ -0,0 +1,105 @@
...
+![](https://services.google.com/fh/gumdrop/preview/misc/preset-flags.png)
+
+### Dataflow Project
+Specify the project to run the dataflow job in.
+
+### Dataflow Location
+Specify the region to run the dataflow job in. It is recommended to keep the region same as Spanner region for performance. Example: us-central1
+
+### GCS Template Path
+Cloud Storage path to the template spec. Use this to run launch dataflow with custom templates. Example: `gs://my-bucket/path/to/template`
```
bharadwaj-aditya, line 105: It might makes sense to add a link here to the dataflow page here so that folks can read up more in detail there if required.

bharadwaj-aditya (approved):

PR #640, opened 2023-09-25
(chore): Process SQLServer CDC tables and UI bug fixes

manitgupta:
1. Addresses: b/300771260 - SQLServer CDC enabled tables were not processed due to being excluded in the SQL query used to query INFORMATION_SCHEMA. Updated the query and tested schema migration and POC `schema-and-data` migration for SQLServer for both CDC enabled and disabled tables.
2. Fixed a bug where Minimal downtime migration would appear in the dropdown for `SQLServer`.
3. `MigrationType` has no significance when a customer is doing a `schema` migration. Removed the `migrationType` dropdown when a customer selects `schema` migration.


sources/common/utils.go
```diff
@@ -172,7 +172,7 @@ func RunParallelTasks[I any, O any](input []I, numWorkers int, f func(i I, mutex
 	}()
 
 	mutex := &sync.Mutex{}
-	logger.Log.Debug(fmt.Sprint("initiating workers", numWorkers))
+	logger.Log.Debug(fmt.Sprint("initiating workers = ", numWorkers))
```
bharadwaj-aditya, line 175: nit: having special characters in logs makes it hard to parse or apply regexes

sources/sqlserver/infoschema.go
```diff
@@ -199,7 +199,7 @@ func (isi InfoSchemaImpl) GetTables() ([]common.SchemaAndName, error) {
 	FROM sys.tables AS TBL
 	INNER JOIN sys.schemas AS SCH 
 	ON SCH.schema_id = TBL.schema_id
-	WHERE TBL.type = 'U' AND TBL.is_tracked_by_cdc = 0 AND TBL.is_ms_shipped = 0 AND TBL.name <> 'sysdiagrams'
+	WHERE TBL.type = 'U' AND TBL.is_ms_shipped = 0 AND TBL.name <> 'sysdiagrams'
```
bharadwaj-aditya, line 202: Wondering why this check was added in the first place. Is there any repercussion of removing this check ?

bharadwaj-aditya (approved):
LGTM

sources/common/utils.go
manitgupta: Changed to a sentence

sources/sqlserver/infoschema.go
manitgupta: Not sure about that, I have tested a POC migration for both CDC and non-CDC enabled tables though to test that the fix works.

PR #641, opened 2023-09-25
Display equivalent gcloud cli command for Dataflow

Deep1998:
Display the equivalent gcloud cli command for Dataflow for streaming migrations.
Added an icon next to the generated dataflow job link for the equivalent gcloud command.

Tested:
- Non sharded flow: Command generation
- Sharded flow: Command generation
- Command generation with df config params changed

Screenshot 2023-09-25 at 1 46 19 PM


common/utils/dataflow_utils.go
```diff
@@ -0,0 +1,108 @@
...
+	"encoding/json"
+	"fmt"
+	"sort"
+	"strings"
+
+	"cloud.google.com/go/dataflow/apiv1beta3/dataflowpb"
+	"golang.org/x/exp/maps"
+)
+
+func GetGcloudDataflowCommand(req *dataflowpb.LaunchFlexTemplateRequest) string {
```
manitgupta, line 29: Nit: Add docs to the top

internal/convert.go
```diff
@@ -185,6 +185,12 @@ type Audit struct {
 	SkipMetricsPopulation    bool                                   `json:"-"` // Flag to identify if outgoing metrics metadata needs to skipped
 }
 
+// Stores information related to resources.
+type JobIdAndGcloudCmd struct {
+	JobId     string `json:"JobId"`
+	GcloudCmd string `json:"GcloudCmd"`
+}
+
```
manitgupta, line 193: I think we should rename this to something more generic to allow for future `shard + dataflow` specific stuff to live here

common/utils/dataflow_utils.go
manitgupta, line 29: Should we also call this method and print the equivalent command in the CLI flow?

ui/src/app/components/equivalent-gcloud-command/equivalent-gcloud-command.component.html
```diff
@@ -0,0 +1,12 @@
+<div mat-dialog-content>
+  <h2>Equivalent gcloud command line<mat-icon class="configure"
+    matTooltip='This is the gcloud command to launch the dataflow job with the same parameters.'>
+  info</mat-icon></h2>
+  
+  <span class="left-text">{{ gcloudCmd }}</span>
```
manitgupta, line 6: Lets use something like - 
https://www.npmjs.com/package/ngx-highlightjs  to display code snippets in Harbourbridge.

Demo of the package - https://ngx-highlight.netlify.app/

ui/src/app/components/equivalent-gcloud-command/equivalent-gcloud-command.component.html
```diff
@@ -0,0 +1,12 @@
+<div mat-dialog-content>
+  <h2>Equivalent gcloud command line<mat-icon class="configure"
+    matTooltip='This is the gcloud command to launch the dataflow job with the same parameters.'>
+  info</mat-icon></h2>
+
```
manitgupta, line 5: We should guidance here on what this gCloud command is for i.e can be used to re-run Dataflow pipeline manually. We should also link the corresponding gCloud CLI documentation for running dataflow templates.

ui/src/app/components/prepare-migration/prepare-migration.component.html
```diff
@@ -324,6 +324,10 @@ <h3> Generated Resources:</h3>
         <span *ngIf="resourcesGenerated.DataflowJobName!=='' && (selectedMigrationType ==='lowdt') && !isSharded">
             <li>Dataflow job: <a [href]="resourcesGenerated.DataflowJobUrl"
                     target="_blank">{{resourcesGenerated.DataflowJobName}}</a>
+                    <span><button mat-button class="configure" (click)="openGcloudPopup(resourcesGenerated.DataflowGcloudCmd)">
+                        <mat-icon matTooltip="Equivalent gCloud command" matTooltipPosition="above">
+                            code</mat-icon>
+                    </button></span>
```
manitgupta, line 330: For the screenshot, it doesn't look very obvious what the `< >` button would do...I am not sure what is the best UX experience here.
@bharadwaj-aditya @shreyakhajanchi thoughts?

manitgupta (commented):
Thanks for taking this up, added some minor comments.

ui/src/app/components/prepare-migration/prepare-migration.component.html
Deep1998: I have added a tool tip as well which on hover which might help, could'nt find a better icon.

Screenshot 2023-09-26 at 3 33 26 PM

ui/src/app/components/equivalent-gcloud-command/equivalent-gcloud-command.component.html
Deep1998: Updated.

For the documentation link, I dont think it should go in a tool tip. Plus there isn't any apt documentation I can find for this, given there isnt anything for the user to do other than copy paste this. Plus I don't see any pre-reqs to running this command that the user would need to configure if they have reached this point.

common/utils/dataflow_utils.go
Deep1998: I was thinking that, and it makes sense for non-sharded flow, but for sharded we need to come up with a good way to output it for all shards (in a file maybe? but we already produce like 5 files).

internal/convert.go
Deep1998: Switched to ShardedDataflowJobResources for the lack of a better word.

common/utils/dataflow_utils.go
manitgupta: I was thinkng lets just write them as debug logs for now. We have `--log-level` flag, if the customer does not want they can set the log level to INFO

manitgupta (approved):
LGTM

PR #645, opened 2023-09-27
Fixed dump file migration issue for PostgreSQL dialect

shreyakhajanchi:
Tests run:
UI - schema anad data migration for dump file to PostgreSQL dialect

### DIFF
--- a/webv2/web.go
+++ b/webv2/web.go
@@ -600,8 +600,7 @@ func convertSchemaDump(w http.ResponseWriter, r *http.Request) {
 	// We don't support Dynamodb in web hence no need to pass schema sample size here.
 	sourceProfile, _ := profiles.NewSourceProfile("", dc.Config.Driver)
 	sourceProfile.Driver = dc.Config.Driver
-	targetProfile, _ := profiles.NewTargetProfile(fmt.Sprintf("dialect=%s", dc.SpannerDetails.Dialect))
-	conv, err := conversion.SchemaConv(sourceProfile, targetProfile, &utils.IOStreams{In: f, Out: os.Stdout})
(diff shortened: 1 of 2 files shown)


webv2/web.go
```diff
@@ -600,8 +600,7 @@ func convertSchemaDump(w http.ResponseWriter, r *http.Request) {
 	// We don't support Dynamodb in web hence no need to pass schema sample size here.
 	sourceProfile, _ := profiles.NewSourceProfile("", dc.Config.Driver)
 	sourceProfile.Driver = dc.Config.Driver
-	targetProfile, _ := profiles.NewTargetProfile(fmt.Sprintf("dialect=%s", dc.SpannerDetails.Dialect))
-	conv, err := conversion.SchemaConv(sourceProfile, targetProfile, &utils.IOStreams{In: f, Out: os.Stdout})
+	conv, err := conversion.SchemaFromDump(sourceProfile.Driver, dc.SpannerDetails.Dialect, &utils.IOStreams{In: f, Out: os.Stdout})
```
manitgupta, line 603: Didn't understand why `conversion.SchemaConv` can't be used here as done previously..
Do we need to open to package access for `schemaFromDump` for this to work?  I liked the previous approach where `SchemaConv` was the interface to the library and `schemaFromDatabase` and `schemaFromDump` were internal implementations..

manitgupta: Makes sense, and since the user can configure spanner instance later (say on the schema conversion workspace), not having target profile available is a valid use-case, right?

shreyakhajanchi: Yes, correct

manitgupta (approved):

PR #646, opened 2023-09-28
added some more details in troubleshooting section

aksharauke:
added some more details in troubleshooting section related to source database connection retry

### DIFF
--- a/docs/reverse-replication/ReverseReplicationUserGuide.md
+++ b/docs/reverse-replication/ReverseReplicationUserGuide.md
@@ -164,6 +164,12 @@ In this case, check if you observe the following:
 
 - ***There is data in Pub/Sub yet not present in source database***
 
+    Check for  logs to ensure that records read being read from PubSub. Logs should have messages like below, which indicate records are read from the buffer.
+    ![DataflowLog](https://services.google.com/fh/files/misc/recordsreadfrompubsub.png)
+
+    Check for logs to see if there are any Connection exception warnings like below. This means that the source database is not reachable and the job keeps retrying to connect, hence nothing gets written to the source database.
+    ![DataflowLog](https://services.google.com/fh/files/misc/connectionretry.png)
+
   Records of below nature are dropped from reverse replication. Check the Dataflow logs to see if they are dropped.
 
      1.Records for which primary key cannot be determined on the source database.This can happen when the source database table does not have a primary key, or the primary key value was not present in the change stream data, or the record was deleted on Cloud Spanner and the deleted record was removed from Cloud Spanner due to lapse of retention period by the time the record was to be reverse replicated.


docs/reverse-replication/ReverseReplicationUserGuide.md
```diff
@@ -164,6 +164,12 @@ In this case, check if you observe the following:
 
 - ***There is data in Pub/Sub yet not present in source database***
 
+    Check for  logs to ensure that records read being read from PubSub. Logs should have messages like below, which indicate records are read from the buffer.
+    ![DataflowLog](https://services.google.com/fh/files/misc/recordsreadfrompubsub.png)
+
+    Check for logs to see if there are any Connection exception warnings like below. This means that the source database is not reachable and the job keeps retrying to connect, hence nothing gets written to the source database.
```
Deep1998, line 170: Should we also point to the section saying if not reachable, ensure your Dataflow IPs are allowlisted on the MySQL instance, credentials are correct, MySQL is up etc.

docs/reverse-replication/ReverseReplicationUserGuide.md
```diff
@@ -164,6 +164,12 @@ In this case, check if you observe the following:
...
+    ![DataflowLog](https://services.google.com/fh/files/misc/connectionretry.png)
+
   Records of below nature are dropped from reverse replication. Check the Dataflow logs to see if they are dropped.
```
Deep1998, line 173: Not part of this change but we should reformat this, we should not make points if there is just one:

Check the Dataflow logs to see if records are being dropped. This can happen for records for which primary key cannot be determined on the source database. This can happen when:
-  The source database table does not have a primary key
-  The primary key value was not present in the change stream data
-  The record was deleted on Cloud Spanner and the deleted record was removed from Cloud Spanner due to lapse of retention period by the time the record was to be reverse replicated.


Side note: If MYSQL primary key does not exist, we wouldnt drop recorcds right? It would just insert a duplicate probably.

aksharauke: Done. We drop the record - since we do UPSERTS and that needs a primary key

docs/reverse-replication/ReverseReplicationUserGuide.md
aksharauke: done, added a reference to the prerequisite.

docs/reverse-replication/ReverseReplicationUserGuide.md
```diff
@@ -45,7 +45,7 @@ These steps are achieved by two Dataflow jobs, along with an interim buffer whic
 
 A few prerequisites must be considered before starting with reverse replication.
 
-1. Make sure that there is network connectivity between source database and your GCP project on which the Dataflow jobs will run.Ensure the Dataflow worker IPs can access the MySQL IPs.
+1. Make sure that there is network connectivity between source database and your GCP project on which the Dataflow jobs will run.Ensure the Dataflow worker IPs can access the MySQL IPs, meaning Dataflow IPs are allowlisted on the MySQL instance. The MySQL credentials are correctly specidied in the [source shards file](./RunnigReverseReplication.md#sample-sourceshards-file) and MySQL server  is up.
```
Deep1998, line 48: How about this:
1. Ensure network connectivity between the source database and your GCP project, where your Dataflow jobs will run.
    - Allowlist Dataflow worker IPs on the MySQL instance so that they can access the MySQL IPs.
    - Check that the MySQL credentials are correctly specified in the [source shards file](./RunningReverseReplication.md#sample-sourceshards-file).
    - Check that the MySQL server is up.

Deep1998 (approved):

Deep1998:
LGTM, minor spellchecks and formatting errors in the prereqs, suggested a rewrite

docs/reverse-replication/ReverseReplicationUserGuide.md
aksharauke: Done

PR #648, opened 2023-10-03
Updated the UI test files

shreyakhajanchi:
Updated test files for all the UI components for successful run.

Command executed:
`ng test`


.github/workflows/ui-test-coverage.yaml
```diff
@@ -0,0 +1,40 @@
+# Copyright 2023 Google LLC
```
manitgupta, line 1: Do we need this as a separate action? I would suggest having this part of one of the existing test workflows. We can update/enhance those workflows to be a single unit testing workflow.

More actions = More workers = More money = More spin up time = Longer pre-flight checks

ui/angular.json
```diff
@@ -106,7 +106,8 @@
               "./node_modules/@angular/material/prebuilt-themes/indigo-pink.css",
               "src/styles.scss"
             ],
-            "scripts": []
+            "scripts": [],
+            "codeCoverage": true
```
manitgupta, line 110: Please also include `npm test` as part of the `make test` and `make test-coverage` targets in the `Makefile`

.github/workflows/ui-test-coverage.yaml
shreyakhajanchi: Moved it to single github action

ui/angular.json
shreyakhajanchi: added

ui/src/app/components/direct-connection/direct-connection.component.spec.ts
```diff
@@ -80,7 +80,7 @@ describe('DirectConnectionComponent', () => {
 
     let pass = component.connectForm.get('password')
     pass?.setValue('')
-    expect(pass?.invalid).toBeTruthy()
+    expect(pass?.invalid).toBeFalsy()
```
bharadwaj-aditya, line 83: Is this flip intentional ?

bharadwaj-aditya (approved):
LGTM

ui/src/app/components/direct-connection/direct-connection.component.spec.ts
shreyakhajanchi: Yes because password is optional field

PR #652, opened 2023-10-05
Enable coverage reports upload to CodeCov

manitgupta:
Enables uploading coverage reports to CodeCov. 

The CodeCov configuration itself will be a follow up PR once the data starts getting synced to CodeCov.


shreyakhajanchi (approved):
LGTM

PR #654, opened 2023-10-05
Add CodeCov configuration

manitgupta:
Adds the `CodeCov` configuration.

It splits the repo into 4 parts - 

1. `backend-library` - contains the non-cli and the non-api portions of SMT.
2. `backend-api` - contains the backend apis implemented in Golang for the Angular frontend.
3. `frontend` - contains the UI frontend. This will currently be empty since we do not run code coverage for this automatically today. Once we do, this should get populated.
4. `cli` - contains the SMT CLI code.

The threshold percentage has been kept as 5%, which means that the PR check will fail if the code coverage has dropped by 5%. 

For more details on the configuration, go through the following - 

1. https://docs.codecov.com/docs/commit-status#basic-configuration
2. https://docs.codecov.com/docs/components
3. https://docs.codecov.com/docs/pull-request-comments

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -41,7 +41,7 @@ component_management:
     - component_id: frontend
       name: frontend # only frontend components
       paths:
-        - ui/**
+        - "ui/**"
       statuses: # the core component has its own statuses
         - type: project
           target: auto
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -6,6 +6,7 @@ component_management:
       - type: project # in this case every component that doens't have a status defined will have a project type
         target: auto
         threshold: 5%
+        informational: true
         branches:
           - master
   individual_components:
@@ -17,6 +18,7 @@ component_management:
       statuses: # each component has its own status and corresponding configuration
         - type: project
           target: auto
+          informational: true
     - component_id: backend-library
       name: backend-library # only library code
       paths:
@@ -26,6 +28,7 @@ component_management:
       statuses: # each component has its own status and corresponding configuration
         - type: project
           target: auto
+          informational: true
     - component_id: cli 
       name: cli # only the CLI
       paths:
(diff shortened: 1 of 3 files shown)


.github/codecov.yml
```diff
@@ -0,0 +1,52 @@
...
+          target: auto
+          informational: true
+    - component_id: frontend
+      name: frontend # only frontend components
+      paths:
+        - ui/**
+      statuses: # the core component has its own statuses
+        - type: project
+          target: auto
+          informational: true
```
shreyakhajanchi, line 48: Frontend end tests will refer to a report different from backend end tests , it might make sense to use flags: https://docs.codecov.com/docs/flags instead of components

manitgupta: As per https://docs.codecov.com/docs/components I looks me to that the `codecov` uploader will automatically find relevant coverage reports and send them to `CodeCov` backend.
After that, the virtual filters in the `codecav.yml` would be applicable for all the collected reports.

.github/workflows/test-coverage.yaml
```diff
@@ -31,23 +31,7 @@ jobs:
         run: |
           go test ./... -coverprofile coverage.out -covermode count
           go tool cover -func coverage.out
-
-      - name: Quality Gate - Test coverage should be above threshold
-        env:
-          TESTCOVERAGE_THRESHOLD: 50
-        run: |
-          echo "Quality Gate: checking if test coverage is above threshold ..."
```
aksharauke, line 39: Will these be enabled later? Or do we want to enable this coverage threshold with a low value - say 30%?

manitgupta: This has now moved to `CodeCov`. It is doing to use the base commit coverage as the base and use the specified threshold for leniency. In our current config, the PR should not reduce the coverage percentage, however, the actual enforcement will not be done since everything is `informational`. 

We will enable informational with time as the team gets comfortable with `CodeCov`.

More details here - https://docs.codecov.com/docs/commit-status#target

aksharauke: coo, thanks

aksharauke (approved):

PR #672, opened 2023-10-25
Dataflow Tuning

Deep1998:
This PR add support to tune the most commonly used params for dataflow.
The changes include:
- New dataflow form for configuration that groups networking and performance parameters.
- Add a preset section that is not editable by default. The user must explicitly click edit to modify these parameters.
- Remove the green tick next to tune dataflow step by default. It now appears only if configured

Pending improvements:
- Add more validation in the dataflow form
- Change KMS key name to a radio button

Dataflow Form:
Screenshot 2023-10-25 at 1 52 48 PM

Networking Section:
Screenshot 2023-10-25 at 1 53 12 PM

Performance Section:
Screenshot 2023-10-25 at 1 53 27 PM

Presets section:
Screenshot 2023-10-25 at 1 53 50 PM

Tested:
UI:
- Sharded: datastream workers passed correctly
- NonSharded: datastream workers passed correctly
- Sharded: dataflow params VPC (via subnet and autonetwork), numWorkers, maxWorkers, machineType, userLabels, region, customPath set correctly.
- NonSharded: dataflow params VPC (via subnet and auto network), numWorkers, maxWorkers, machineType, userLabels, region, customPath set correctly.

CLI:
Non-sharded: parameters passed correctly
Sharded:  parameters passed correctly

### DIFF
--- a/profiles/source_profile.go
+++ b/profiles/source_profile.go
@@ -446,13 +446,19 @@ type DatastreamConnProfile struct {
 }
 
 type DataflowConfig struct {
-	Location            string `json:"location"`
-	Network             string `json:"network"`
-	Subnetwork          string `json:"subnetwork"`
-	HostProjectId       string `json:"hostProjectId"`
-	MaxWorkers          string `json:"maxWorkers"`
-	NumWorkers          string `json:"numWorkers"`
-	ServiceAccountEmail string `json:"serviceAccountEmail"`
+	ProjectId            string `json:"ProjectId"`
+	Location             string `json:"location"`
+	Network              string `json:"network"`
+	Subnetwork           string `json:"subnetwork"`
+	HostProjectId        string `json:"hostProjectId"`
+	MaxWorkers           string `json:"maxWorkers"`
+	NumWorkers           string `json:"numWorkers"`
+	ServiceAccountEmail  string `json:"serviceAccountEmail"`
+	JobName              string `json:"JobName"`
+	MachineType          string `json:"MachineType"`
+	AdditionalUserLabels string `json:"AdditionalUserLabels"`
+	KmsKeyName           string `json:"KmsKeyName"`
+	GcsTemplatePath      string `json:"GcsTemplatePath"`
 }
 
 type DataShard struct {
--- a/profiles/source_profile.go
+++ b/profiles/source_profile.go
@@ -446,19 +446,19 @@ type DatastreamConnProfile struct {
 }
 
 type DataflowConfig struct {
-	ProjectId            string `json:"ProjectId"`
+	ProjectId            string `json:"projectId"`
 	Location             string `json:"location"`
 	Network              string `json:"network"`
 	Subnetwork           string `json:"subnetwork"`
-	HostProjectId        string `json:"hostProjectId"`
+	VpcHostProjectId     string `json:"vpcHostProjectId"`
 	MaxWorkers           string `json:"maxWorkers"`
 	NumWorkers           string `json:"numWorkers"`
 	ServiceAccountEmail  string `json:"serviceAccountEmail"`
-	JobName              string `json:"JobName"`
-	MachineType          string `json:"MachineType"`
-	AdditionalUserLabels string `json:"AdditionalUserLabels"`
-	KmsKeyName           string `json:"KmsKeyName"`
-	GcsTemplatePath      string `json:"GcsTemplatePath"`
+	JobName              string `json:"jobName"`
+	MachineType          string `json:"machineType"`
+	AdditionalUserLabels string `json:"additionalUserLabels"`
+	KmsKeyName           string `json:"kmsKeyName"`
+	GcsTemplatePath      string `json:"gcsTemplatePath"`
 }
 
 type DataShard struct {
--- a/profiles/source_profile.go
+++ b/profiles/source_profile.go
@@ -450,7 +450,7 @@ type DataflowConfig struct {
 	Location             string `json:"location"`
 	Network              string `json:"network"`
 	Subnetwork           string `json:"subnetwork"`
-	VpcHostProjectId     string `json:"vpcHostProjectId"`
+	VpcHostProjectId     string `json:"hostProjectId"`
 	MaxWorkers           string `json:"maxWorkers"`
 	NumWorkers           string `json:"numWorkers"`
 	ServiceAccountEmail  string `json:"serviceAccountEmail"`
--- a/streaming/streaming.go
+++ b/streaming/streaming.go
@@ -776,6 +776,7 @@ func StoreGeneratedResources(conv *internal.Conv, streamingCfg StreamingCfg, dfJ
 	dataflowCfg := streamingCfg.DataflowCfg
 	conv.Audit.StreamingStats.DataStreamName = datastreamCfg.StreamId
 	conv.Audit.StreamingStats.DataflowJobId = dfJobId
+	conv.Audit.StreamingStats.DataflowLocation = streamingCfg.DataflowCfg.Location
 	conv.Audit.StreamingStats.DataflowGcloudCmd = gcloudDataflowCmd
 	conv.Audit.StreamingStats.PubsubCfg = streamingCfg.PubsubCfg
 	conv.Audit.StreamingStats.MonitoringDashboard = dashboardName
--- a/streaming/streaming.go
+++ b/streaming/streaming.go
@@ -49,6 +49,8 @@ var (
 	MAX_WORKER_LIMIT int32 = 1000
 	// Min allowed value for maxWorkers and numWorkers.
 	MIN_WORKER_LIMIT int32 = 1
+	// Default gcs path of the Dataflow template.
+	DEFAULT_TEMPLATE_PATH string = "gs://dataflow-templates-southamerica-west1/2023-09-12-00_RC00/flex/Cloud_Datastream_to_Spanner"
 )
 
 type SrcConnCfg struct {
@@ -73,15 +75,20 @@ type DatastreamCfg struct {
 }
 
 type DataflowCfg struct {
-	JobName             string            `json:"jobName"`
-	Location            string            `json:"location"`
-	HostProjectId       string            `json:"hostProjectId"`
-	Network             string            `json:"network"`
-	Subnetwork          string            `json:"subnetwork"`
-	MaxWorkers          string            `json:"maxWorkers"`
-	NumWorkers          string            `json:"numWorkers"`
-	ServiceAccountEmail string            `json:"serviceAccountEmail"`
-	DbNameToShardIdMap  map[string]string `json:"dbNameToShardIdMap"`
+	ProjectId            string            `json:"projectId"`
+	JobName              string            `json:"jobName"`
+	Location             string            `json:"location"`
+	HostProjectId        string            `json:"hostProjectId"`
+	Network              string            `json:"network"`
+	Subnetwork           string            `json:"subnetwork"`
+	MaxWorkers           string            `json:"maxWorkers"`
+	NumWorkers           string            `json:"numWorkers"`
+	ServiceAccountEmail  string            `json:"serviceAccountEmail"`
+	MachineType          string            `json:"machineType"`
+	AdditionalUserLabels string            `json:"additionalUserLabels"`
+	KmsKeyName           string            `json:"kmsKeyName"`
+	GcsTemplatePath      string            `json:"gcsTemplatePath"`
+	DbNameToShardIdMap   map[string]string `json:"dbNameToShardIdMap"`
 }
 
 type StreamingCfg struct {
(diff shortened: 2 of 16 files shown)


profiles/source_profile.go
```diff
@@ -446,13 +446,19 @@ type DatastreamConnProfile struct {
...
-	Subnetwork          string `json:"subnetwork"`
-	HostProjectId       string `json:"hostProjectId"`
-	MaxWorkers          string `json:"maxWorkers"`
-	NumWorkers          string `json:"numWorkers"`
-	ServiceAccountEmail string `json:"serviceAccountEmail"`
+	ProjectId            string `json:"projectId"`
+	Location             string `json:"location"`
+	Network              string `json:"network"`
+	Subnetwork           string `json:"subnetwork"`
+	VpcHostProjectId     string `json:"vpcHostProjectId"`
```
manitgupta, line 453: Can renaming this lead to backward compatibility issues in the CLI?

streaming/streaming.go
```diff
@@ -71,15 +73,20 @@ type DatastreamCfg struct {
...
+	Subnetwork           string            `json:"subnetwork"`
+	MaxWorkers           string            `json:"maxWorkers"`
+	NumWorkers           string            `json:"numWorkers"`
+	ServiceAccountEmail  string            `json:"serviceAccountEmail"`
+	MachineType          string            `json:"machineType"`
+	AdditionalUserLabels string            `json:"additionalUserLabels"`
+	KmsKeyName           string            `json:"kmsKeyName"`
+	GcsTemplatePath      string            `json:"gcsTemplatePath"`
+	DbNameToShardIdMap   map[string]string `json:"dbNameToShardIdMap"`
 }
```
manitgupta, line 90: Just so that I remember the usage here - what is the difference between `dataflowCfg` and `DataflowConfig`?

streaming/streaming.go
```diff
@@ -635,24 +642,40 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
...
-
-	if dataflowCfg.Network != "" {
+	// Initiate runtime environment flags and overrides.
+	var (
+		dataflowProjectId        = project
+		dataflowVpcHostProjectId = project
+		gcsTemplatePath          = DEFAULT_TEMPLATE_PATH
+		dataflowSubnetwork       = ""
+		workerIpAddressConfig    = dataflowpb.WorkerIPAddressConfiguration_WORKER_IP_PUBLIC
+		dataflowUserLabels       = make(map[string]string)
```
manitgupta, line 653: Didnt see user labels in the config defined above, where does the user enter them?

streaming/streaming.go
```diff
@@ -635,24 +642,40 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
...
-			dataflowSubnetwork = fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/%s/regions/%s/subnetworks/%s", dataflowHostProjectId, dataflowCfg.Location, dataflowCfg.Subnetwork)
+		if dataflowCfg.Subnetwork != "" {
+			dataflowSubnetwork = fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/%s/regions/%s/subnetworks/%s", dataflowVpcHostProjectId, dataflowCfg.Location, dataflowCfg.Subnetwork)
+		}
+	}
+
+	if dataflowCfg.AdditionalUserLabels != "" {
+		err = json.Unmarshal([]byte(dataflowCfg.AdditionalUserLabels), &dataflowUserLabels)
+		if err != nil {
+			return internal.DataflowOutput{}, fmt.Errorf("could not unmarshal AdditionalUserLabels json %s : error = %v", dataflowCfg.AdditionalUserLabels, err)
```
manitgupta, line 678: Should we fail on improperly configured User labels or should this be a warning?

ui/src/app/components/dataflow-form/dataflow-form.component.html
```diff
@@ -1,38 +1,109 @@
...
       </mat-form-field>
+    </form>
+      <hr>
+    <form [formGroup]="presetFlagsForm" class="dataflow-form">
+      <h2>
+        Preset Flags 
+        <span><button mat-button class="configure" (click)="enablePresetFlags()" [disabled]="!disablePresetFlags">
+          <mat-icon>edit</mat-icon>
+          EDIT
+      </button></span>
```
manitgupta, line 87: Use the standard color in HB for this - blue, as done in other places

ui/src/app/components/dataflow-form/dataflow-form.component.html
```diff
@@ -1,38 +1,109 @@
...
+      </h2> 
+      <h5>These flags are set by SMT by default and <b>SHOULD NOT BE</b> modified unless running Dataflow in a non-standard configuration.</h5>
```
manitgupta, line 89: These should look disabled, not ever clear from the screenshot if they are disabled or not. Also, when we hover or them, does the cursor change to disabled?

ui/src/app/components/dataflow-form/dataflow-form.component.html
```diff
@@ -1,38 +1,109 @@
...
+      [matTooltipPosition]="'right'">
+        <mat-label>Dataflow Project Id</mat-label>
+        <input matInput placeholder="Dataflow Project Id" type="text" formControlName="dataflowProjectId"/>
       </mat-form-field>
       <br>
-      <mat-form-field class="full-width" appearance="outline" matTooltip="Set initial number of workers for the dataflow job(s). Default value: 1"
-      [matTooltipPosition]="'right'" matTooltipHideDelay="100000">
-        <mat-label>Number of Workers</mat-label>
-        <input matInput placeholder="1" type="text" formControlName="numWorkers"/>
+      <mat-form-field class="full-width" appearance="outline" matTooltip="Specify the region to run the dataflow job in. It is recommended to keep the region same as Spanner region for performance."
```
manitgupta, line 97: We should provide an example of what type of string needs to be input here.
Same goes for the template location as well.

manitgupta (commented):
I think this looks good overall, minor comments.
We do need the following though - 

1. We need to test against different combinations of these parameters are ensure that they work. I will recommend using the CLI here to run these tests.
2. Documentation updates
3. Please talk to @darshan-sj about these test scenarios so that he is aware of the need of this kind of integration testing.

profiles/source_profile.go
Deep1998: If someone is using an old streamingcfg/sharded config and configuring vpc, then yes this could cause an issue. But this renaming is important as hostprojectid is only relevant to vpc.

streaming/streaming.go
Deep1998: They are largely similar with DataflowConfig being a subset of dataflowCfg. 
- DataflowConfig is used in web.go primarily to unmarshall request bodies and set session state. 
- DataflowCfg is whats used in the backend for orchestrating streaming migrations.

They can be deduped but is a slightly more involved changed so have'nt taken that as part of this.

streaming/streaming.go
Deep1998: THey are defined under AdditionalUserLabels, the variable scope is just this function hence used the name dataflowUserLabels.

streaming/streaming.go
Deep1998: Thats a good question, normally I would keep a warning, but dataflow does not allow the user to add a label once the job has started (unlike other resources). They would either miss it and the job would run without a label or they notice labels werent applied, cancel the job and manually retrigger. 

I chose to fail but open to alternatives.

ui/src/app/components/dataflow-form/dataflow-form.component.html
Deep1998: This is what the presets look like:
Screenshot 2023-10-25 at 1 53 50 PM

The font color for disabled fields is slightly different (not too contrasting I agree). The user anyways cannot edit these fields even if they try till they click on the edit button (since they are disabled by default). So there is no way for the user to accidentally modify these fields.

Deep1998:
> I think this looks good overall, minor comments. We do need the following though -
> 
> 1. We need to test against different combinations of these parameters are ensure that they work. I will recommend using the CLI here to run these tests.
> 2. Documentation updates
> 3. Please talk to @darshan-sj about these test scenarios so that he is aware of the need of this kind of integration testing.

Added testing details
Doc updates covered here: https://github.com/GoogleCloudPlatform/spanner-migration-tool/pull/638

ui/src/app/components/dataflow-form/dataflow-form.component.ts
```diff
@@ -11,6 +11,8 @@ import ISpannerConfig from 'src/app/model/spanner-config';
 })
 export class DataflowFormComponent implements OnInit {
   dataflowForm: FormGroup
```
manitgupta, line 13: Let's rename this form to something more meaningful now that we have two other forms all of which contribute to dataflow configuration.

ui/src/app/components/dataflow-form/dataflow-form.component.ts
```diff
@@ -22,8 +24,19 @@ export class DataflowFormComponent implements OnInit {
       numWorkers: new FormControl('1',[Validators.required, Validators.pattern('^[1-9][0-9]*$')]),
       maxWorkers: new FormControl('50', [Validators.required, Validators.pattern('^[1-9][0-9]*$')]),
       serviceAccountEmail: new FormControl(''),
-      hostProjectId: new FormControl(data.GCPProjectID, Validators.required),
+      vpcHostProjectId: new FormControl(data.GCPProjectID, Validators.required),
+      machineType: new FormControl(''),
+      // With the json validator, the form still allows nested json values, which is not a valid entry for user labels.
+      // Wwe rely on the Dataflow validation for nested json validation.
+      additionalUserLabels: new FormControl('', [this.jsonValidator]),
```
manitgupta, line 31: Why can't we write a custom validator?

profiles/source_profile.go
```diff
@@ -446,13 +446,19 @@ type DatastreamConnProfile struct {
...
+	MaxWorkers           string `json:"maxWorkers"`
+	NumWorkers           string `json:"numWorkers"`
+	ServiceAccountEmail  string `json:"serviceAccountEmail"`
+	JobName              string `json:"jobName"`
+	MachineType          string `json:"machineType"`
+	AdditionalUserLabels string `json:"additionalUserLabels"`
```
manitgupta, line 459: I think `AdditionalUserLabels` itself should be struct with `key,value` and this would be a list of that struct. It will avoid a bunch of issues around validations and deserialisation

ui/src/app/components/dataflow-form/dataflow-form.component.html
```diff
@@ -1,38 +1,109 @@
 <div mat-dialog-content>
     <form [formGroup]="dataflowForm" class="dataflow-form">
-      <h2>Dataflow Details</h2>
-      <mat-form-field class="full-width" appearance="outline" matTooltip="Edit to run Dataflow in a shared VPC"
-      [matTooltipPosition]="'right'" matTooltipHideDelay="100000">
-        <mat-label>Host ProjectID</mat-label>
-        <input matInput placeholder="Host ProjectID" type="text" formControlName="hostProjectId"/>
+      <h2>Tune Dataflow (Optional)</h2>
+      <h5>This form is optional and should only be edited to tune runtime environment for Dataflow.</h5>
```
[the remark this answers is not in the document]
Deep1998: Yes thats because of the delay that was set which I have already removed.

ui/src/app/components/dataflow-form/dataflow-form.component.ts
Deep1998: NA since found the regex for additional user labels.

profiles/source_profile.go
Deep1998: The data is encoded into a string when we read from the API, so whatever conversions we are doing currently are unavoidable. Changing the struct field to key value would require doing this processing right after reading in web.go, but we would have to do it anyways.

webv2/web.go
```diff
@@ -2279,7 +2289,7 @@ func getGeneratedResources(w http.ResponseWriter, r *http.Request) {
 	}
 	if sessionState.Conv.Audit.StreamingStats.DataflowJobId != "" {
 		generatedResources.DataflowJobName = sessionState.Conv.Audit.StreamingStats.DataflowJobId
-		generatedResources.DataflowJobUrl = fmt.Sprintf("https://console.cloud.google.com/dataflow/jobs/%v/%v?project=%v", sessionState.Region, sessionState.Conv.Audit.StreamingStats.DataflowJobId, sessionState.GCPProjectID)
+		generatedResources.DataflowJobUrl = fmt.Sprintf("https://console.cloud.google.com/dataflow/jobs/%v/%v?project=%v", sessionState.Conv.Audit.StreamingStats.DataflowLocation, sessionState.Conv.Audit.StreamingStats.DataflowJobId, sessionState.GCPProjectID)
```
manitgupta, line 2292: We should also store `region` separately inside the generatedResources for Dataflow in addition to the URL. Similarly for other configurations such as the `hostProjectId`. 

 Once we start persisting generatedResources in the metadata database, we will need these values individually to perform operations on them - and interpreting the URL to get these values will be prone to error.

Deep1998: We should take this refactoring up separately along with other resources like pubsub datastream etc. Filed b/311343651 to track.

manitgupta (approved):
LGTM

streaming/streaming.go
Deep1998: created b/312106652

profiles/source_profile.go
Deep1998: Done

streaming/streaming.go
Deep1998: With the new regex check, we wont need this anymore

PR #679, opened 2023-11-02
added flag for filtration mode

aksharauke:
added flag for filtration mode

### DIFF
--- a/reverse_replication/launcher.go
+++ b/reverse_replication/launcher.go
@@ -52,6 +52,7 @@ var (
 	orderingWorkers      int
 	writerWorkers        int
 	networkTags          string
+	filtrationMode       string
 )
 
 const (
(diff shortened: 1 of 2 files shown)


reverse_replication/launcher.go
```diff
@@ -80,6 +81,7 @@ func setupGlobalFlags() {
 	flag.IntVar(&orderingWorkers, "orderingWorkers", 5, "number of workers for ordering job")
 	flag.IntVar(&writerWorkers, "writerWorkers", 5, "number of workers for writer job")
 	flag.StringVar(&networkTags, "networkTags", "", "Network tags addded to the Dataflow jobs worker and launcher VMs")
+	flag.StringVar(&filtrationMode, "filtrationMode", "forward_migration", "Whether to filter forward migrated data or not. Supported values are forward_migration and none, defaults to 'forward_migration'")
```
manitgupta, line 84: Do we expect this to be an `ENUM` like field? If not, can we go with a boolean? Maybe something like `filterForwardMigration`?

aksharauke: Yes in the Dataflow template it's an enum, just to support more than two values if needed in future.

manitgupta (approved):

PR #682, opened 2023-11-06
feat: Monitoring metrics for sharded and non-sharded migration

asthamohta:
- The feature creates monitoring dashboard for sharded and non sharded migration. 
- For sharded migrations it does not show an aggregated dashboard. That will be part of a separate PR
- If dashboard creation fails then the flow logs and error and continues.

The following use cases were tested for sharded and non-sharded (mySql) migration:

1. Successful dashboard creation
2. Unsuccessful dashboard creation with empty dashboard string
2. Unsuccessful dashboard creation with an error thrown during dashboard creation


conversion/conversion.go
```diff
@@ -305,7 +305,26 @@ func dataFromDatabase(ctx context.Context, sourceProfile profiles.SourceProfile,
...
+				ShardId:              "",
+			}
+			respDash, dashboardErr := metrics.CreateDataflowShardMonitoringDashboard(ctx, monitoringResources)
+			var dashboardName string
+			if dashboardErr != nil {
+				dashboardName = ""
+				logger.Log.Error(fmt.Sprintf("Creation of the monitoring dashboard failed, please create the dashboard manually\n error=%v\n", dashboardErr))
+			} else {
+				dashboardName = strings.Split(respDash.Name, "/")[3]
+				fmt.Printf("Monitoring Dashboard: %+v\n", dashboardName)
```
manitgupta, line 325: Curious - what does `respDash` contain?

manitgupta (approved):
LGTM. I have not run the UI myself, but the screenshots and the dashboard look fine!

conversion/conversion.go
asthamohta: Dashboard response. The result of dashboard create request

manitgupta (changes requested):
I had missed a file - added a couple of comments

conversion/conversion.go
manitgupta: Hah! I meant what does it contain that we have to `strings.Split` 🙂, I assumed that they would return some sort of ID only.

common/metrics/metrics.go
```diff
@@ -102,3 +173,182 @@ func getMigrationDataSourceDetails(driver string, migrationData *migration.Migra
 		return migration.MigrationData_SOURCE_CONNECTION_MECHANISM_UNSPECIFIED.Enum(), migration.MigrationData_SOURCE_UNSPECIFIED.Enum()
 	}
 }
+
+// CreateDataflowShardMonitoringDashboard returns a monitoring dashboard for a single shard
+func CreateDataflowShardMonitoringDashboard(ctx context.Context, resourceIds MonitoringMetricsResources) (*dashboardpb.Dashboard, error) {
```
manitgupta, line 178: This is a hot topic, but I personally think that this method would be more readable using `MonitoringMetricsResources` as a function receiver instead of a parameter

conversion/conversion.go
asthamohta: Yes, they return it in form of projects/<project-name>/dashboards/<dashboard-name>

common/metrics/metrics.go
```diff
@@ -102,3 +138,188 @@ func getMigrationDataSourceDetails(driver string, migrationData *migration.Migra
...
+	// create pubsub metrics tiles
+	pubsubMetricsTiles := createShardPubsubMetrics(resourceIds)
+	pubsubMetricsGroupTile, heightOffset := createCollapsibleGroupTile(TileInfo{Title: fmt.Sprintf("Pubsub: %s", resourceIds.PubsubSubscriptionId)}, pubsubMetricsTiles, heightOffset)
+	mosaicLayoutTiles = append(append(mosaicLayoutTiles, pubsubMetricsTiles...), pubsubMetricsGroupTile)
+	// create spanner metrics tiles
+	spannerMetricsTiles := createShardSpannerMetrics(resourceIds)
+	spannerMetricsGroupTile, heightOffset := createCollapsibleGroupTile(TileInfo{Title: fmt.Sprintf("Spanner: instances/%s/databases/%s", resourceIds.SpannerInstanceId, resourceIds.SpannerDatabaseId)}, spannerMetricsTiles, heightOffset)
+	mosaicLayoutTiles = append(append(mosaicLayoutTiles, spannerMetricsTiles...), spannerMetricsGroupTile)
+	mosaicLayout := dashboardpb.MosaicLayout{
+		Columns: 48,
```
shreyakhajanchi, line 181: Better to define this value as a constant as well

ui/src/app/components/prepare-migration/prepare-migration.component.html
```diff
@@ -313,6 +313,20 @@ <h3>Source database details:</h3>
         Note: Spanner migration tool is creating datastream and dataflow resources. Please look at the terminal logs to
         check the progress of resource creation. All created resources will be displayed here once they are generated.
     </div>
+    <div *ngIf="isResourceGenerated && resourcesGenerated.MonitoringDashboardName!==''">
+      <h3> Monitoring Dashboards:</h3>
+      <span *ngIf="resourcesGenerated.MonitoringDashboardName!=='' && (selectedMigrationType ==='lowdt') && !isSharded">
```
shreyakhajanchi, line 318: This condition seems to be repetitive since we have already checked in line 316 : resourcesGenerated.MonitoringDashboardName!==''

shreyakhajanchi, line 318: We can move (selectedMigrationType ==='lowdt') to line 316 since both the ifs have same condition

ui/src/app/components/prepare-migration/prepare-migration.component.html
```diff
@@ -313,6 +313,20 @@ <h3>Source database details:</h3>
         Note: Spanner migration tool is creating datastream and dataflow resources. Please look at the terminal logs to
         check the progress of resource creation. All created resources will be displayed here once they are generated.
     </div>
+    <div *ngIf="isResourceGenerated && resourcesGenerated.MonitoringDashboardName!==''">
```
shreyakhajanchi, line 316: Will MonitoringDashboardName not be null in case of sharded migration? Since aggregated dashboard is not a part of current PR, I am a bit confused

shreyakhajanchi (commented):
Largely LGTM, left minor comments

manitgupta:
The documentation for this feature needs to be written. Please decide if you want to take that up in this PR, the following one or a separate PR.
Let's aim to get the documentation and the release of this feature in one HB release version.

common/metrics/metrics.go
```diff
@@ -1,14 +1,50 @@
...
 	"github.com/GoogleCloudPlatform/spanner-migration-tool/common/constants"
 	"github.com/GoogleCloudPlatform/spanner-migration-tool/internal"
 	"github.com/GoogleCloudPlatform/spanner-migration-tool/internal/reports"
 	"github.com/GoogleCloudPlatform/spanner-migration-tool/proto/migration"
 )
 
+// Defines dimensions for Monitoring Dashboard Metrics
+const (
+	// Default height of a tile in the monitoring dashboard
+	defaultMonitoringMetricHeight int32 = 16
```
bharadwaj-aditya, line 21: I think this should be a relative height as opposed to absolute

common/metrics/metrics.go
```diff
@@ -102,3 +138,188 @@ func getMigrationDataSourceDetails(driver string, migrationData *migration.Migra
...
+	dataflowMetricsTiles := createShardDataflowMetrics(resourceIds)
+	dataflowMetricsGroupTile, heightOffset := createCollapsibleGroupTile(TileInfo{Title: fmt.Sprintf("Dataflow Job: %s", resourceIds.DataflowJobId)}, dataflowMetricsTiles, heightOffset)
+	mosaicLayoutTiles = append(append(mosaicLayoutTiles, dataflowMetricsTiles...), dataflowMetricsGroupTile)
+	// create datastream metrics tiles
+	datastreamMetricsTiles := createShardDatastreamMetrics(resourceIds)
+	datastreamMetricsGroupTile, heightOffset := createCollapsibleGroupTile(TileInfo{Title: fmt.Sprintf("Datastream: %s", resourceIds.DatastreamId)}, datastreamMetricsTiles, heightOffset)
+	mosaicLayoutTiles = append(append(mosaicLayoutTiles, datastreamMetricsTiles...), datastreamMetricsGroupTile)
+	// create gcs bucket metrics tiles
+	gcsMetricsTiles := createShardGcsMetrics(resourceIds)
+	gcsMetricsGroupTile, heightOffset := createCollapsibleGroupTile(TileInfo{Title: fmt.Sprintf("GCS Bucket: %s", strings.Split(resourceIds.GcsBucketId, "/")[2])}, gcsMetricsTiles, heightOffset)
```
bharadwaj-aditya, line 170: nit: it seems like the same few methods called repeatedly. There may be scope to simplify this or extract into a method to clean up the code

bharadwaj-aditya (commented):
There is some code repetition and some scope to improve tests. PTAL

ui/src/app/components/prepare-migration/prepare-migration.component.html
asthamohta: Following the current convention we always set this value even for sharded migrations: https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/master/streaming/streaming.go#L721-L725

common/metrics/metrics.go
asthamohta: Hi Aditya, I am not sure I understand. Relative to what?

manitgupta (approved):

shreyakhajanchi (approved):

PR #684, opened 2023-11-07
angular update to v16

shreyakhajanchi:


ui/src/app/components/direct-connection/direct-connection.component.ts
```diff
@@ -75,55 +75,63 @@
...
-          localStorage.setItem(StorageKeys.Type, InputType.DirectConnect)
-          localStorage.setItem(StorageKeys.SourceDbName, extractSourceDbName(dbEngine))
+    if (typeof dialect === "string" && typeof dbEngine === "string") {
+      this.connectRequest =this.fetch.connectTodb(config, dialect).subscribe({
+        next: () => {
+          this.getSchemaRequest = this.data.getSchemaConversionFromDb()
+          this.data.conv.subscribe((res) => {
+            localStorage.setItem(
+              StorageKeys.Config,
+              JSON.stringify({ dbEngine, hostName, port, userName, password, dbName })
```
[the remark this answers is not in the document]
manitgupta: I don't think this is a false positive, we should be using a crypto lib to encrypt and store passwords in the `localStorage`. 

Can you create a backlog bug to track this - we need to take this up on priority.

ui/src/app/components/direct-connection/direct-connection.component.ts
```diff
@@ -75,55 +75,63 @@ export class DirectConnectionComponent implements OnInit {
   }
 
   testConn() {
-    this.clickEvent.openDatabaseLoader('test-connection', this.connectForm.value.dbName)
+    if (typeof this.connectForm.value.dbName === "string") {
+      this.clickEvent.openDatabaseLoader('test-connection', this.connectForm.value.dbName)
+    }
```
manitgupta, line 80: Very interesting, what is the error if we don't do the `if` check? Can't we simply force a cast to string for `dbName` to avoid the `if` check?

ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -1263,22 +1264,25 @@ export class ObjectDetailComponent implements OnInit {
...
-      srcColId: undefined,
-      spColId: this.currentObject
-        ? this.conversion.getColIdFromSpannerColName(
-          this.addIndexKeyForm.value.columnName,
-          this.currentObject.parentId,
-          this.conv
-        )
-        : '',
-    })
+    if (typeof this.addIndexKeyForm.value.columnName === "string") {
```
manitgupta, line 1267: I think we are having to do this because the time of defining the form we did not define the datatype? Is there a way to define the data type at the time of form declaration?

```
addIndexKeyForm = new FormGroup({
    columnName: new FormControl('', [Validators.required]),
    ascOrDesc: new FormControl('', [Validators.required]),
  })
```

I think you need to do this - 
https://angular.io/guide/typed-forms#specifying-an-explicit-type

This should take care of all the `if` checks added in the code.

ui/src/app/model/db-config.ts
```diff
@@ -1,11 +1,11 @@
...
-  userName: string
-  password: string
-  dbName: string
+  dbEngine: string| null | undefined
+  isSharded: boolean| null | undefined
+  hostName: string | null | undefined
+  port: string | null | undefined
+  userName: string | null | undefined
+  password: string | null | undefined
+  dbName: string | null | undefined
```
manitgupta, line 8: What are the usecases where they can be `null` or `undefined`? 
(I know this is probably due to how the code is written in the first place but an example of how the `dbEngine` can be null can provide a clue what is written incorrectly..)

manitgupta (changes requested):

ui/src/app/components/direct-connection/direct-connection.component.ts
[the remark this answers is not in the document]
shreyakhajanchi: Created a bug with id: 312378508

ui/src/app/components/direct-connection/direct-connection.component.ts
shreyakhajanchi: Removed the if check and added ! - non null assertion operator

ui/src/app/components/object-detail/object-detail.component.ts
shreyakhajanchi: Removed the if check and added ! - non null assertion operator

ui/src/app/model/db-config.ts
shreyakhajanchi: These are required fields and wouldn't be null but these were added because typescript upgrade required stronger type assertions. Added ! non null assertion operator to overcome this

ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -219,9 +220,9 @@ export class ObjectDetailComponent implements OnInit {
             fb.get('spColMaxLength')?.setValue('MAX')
           }
           else if (row.spColMaxLength !== 'MAX') {
-            if ((row.spDataType === 'STRING' || row.spDataType === 'VARCHAR') && row.spColMaxLength > ColLength.StringMaxLength) {
+            if ((row.spDataType === 'STRING' || row.spDataType === 'VARCHAR') && typeof row.spColMaxLength === "number" && row.spColMaxLength > ColLength.StringMaxLength) {
               fb.get('spColMaxLength')?.setValue('MAX')
-            } else if (row.spDataType === 'BYTES' && row.spColMaxLength > ColLength.ByteMaxLength) {
+            } else if (row.spDataType === 'BYTES' && typeof row.spColMaxLength === "number" && row.spColMaxLength > ColLength.ByteMaxLength) {
```
manitgupta, line 225: Why is this comparison needed? Can't the data type of `spColMaxLength` we declared upfront?

shreyakhajanchi: the datatype of spColMaxLength is number | string because 'MAX' is also a valid value for max length

manitgupta (approved):

Issue #350, opened 2022-07-19
Fix Logging in HarbourBridge


manitgupta:
As part of #322 `zap` structured logging library has been integrated into HarbourBridge.
We now need to fix the logging in Harbourbridge which is currently done through three mechanisms:

1. Using the Golang `log` package.
2. Using fmt.Printf statements.
3. Using custom `verbose` wrapper.

We need to change all the three instances above and streamline it to use `zap` logger everywhere.

PR #686, opened 2023-11-17
feat: Adding support for an Aggregated monitoring dashboard in sharded migration

asthamohta:
1. Add support for an Aggregated monitoring dashboard in sharded migration
2. Store GCS Buckets associated with a stream
2. Revise certain queries for sharded and unsharded migration based on discussions

Testing
1. Sharded migration dashboard creation success
2. Non sharded migration dashboard creation success
3. Dashboard creation when fetching gcs bucket fails
4. Testing dashboard creation failure


common/metrics/queries.go
```diff
@@ -2,48 +2,94 @@ package metrics
...
+	pubsubAggSubscriptionSentMessageCountQuery = "fetch pubsub_subscription | metric 'pubsub.googleapis.com/subscription/sent_message_count' " +
+		"| filter (%s ) | align rate(1m) | every 1m | group_by [], " +
+		"[value_sent_message_count_aggregate: aggregate(value.sent_message_count)]"
+	pubsubAggOldestUnackedMessageAgeQuery = "fetch pubsub_subscription | metric 'pubsub.googleapis.com/subscription/oldest_unacked_message_age' " +
+		"| filter (%s) | group_by 1m, " +
+		"[value_oldest_unacked_message_age_max: max(value.oldest_unacked_message_age)] | every 1m | group_by [], " +
+		"[value_oldest_unacked_message_age_max_max: max(value_oldest_unacked_message_age_max)]"
+	gcsAggTotalBytesQuery = "fetch gcs_bucket | metric 'storage.googleapis.com/storage/total_bytes' | " +
+		"filter (%s) | group_by 1m, [value_total_bytes_mean: mean(value.total_bytes)] " +
+		"| every 1m | group_by [], [value_total_bytes_mean_aggregate: aggregate(value_total_bytes_mean)]"
```
manitgupta, line 94: NIt: Add a comment here between the constants to segregate out aggregated queries and single shard queries. I wouldn't mind putting them in their own files either..

conversion/conversion.go
```diff
@@ -346,6 +367,7 @@ func dataFromDatabaseForDataflowMigration(targetProfile profiles.TargetProfile,
 	conv.Audit.StreamingStats.ShardToDataStreamNameMap = make(map[string]string)
 	conv.Audit.StreamingStats.ShardToPubsubIdMap = make(map[string]internal.PubsubCfg)
 	conv.Audit.StreamingStats.ShardToDataflowInfoMap = make(map[string]internal.ShardedDataflowJobResources)
+	conv.Audit.StreamingStats.ShardToGcsResources = make(map[string]internal.GcsResources)
```
manitgupta, line 370: Nit: `GCS` is an acronym and should be in capitals.

conversion/conversion.go
```diff
@@ -387,11 +409,20 @@ func dataFromDatabaseForDataflowMigration(targetProfile profiles.TargetProfile,
 		if err != nil {
 			return common.TaskResult[*profiles.DataShard]{Result: p, Err: err}
 		}
+
+		// Fetch and store the GCS bucket associated with the datastream
+		dsClient := getDatastreamClient(ctx)
+		gcsBucket, _, fetchGcsErr := streaming.FetchTargetBucketAndPath(ctx, dsClient, targetProfile.Conn.Sp.Project, streamingCfg.DatastreamCfg.DestinationConnectionConfig)
+		if fetchGcsErr != nil {
+			logger.Log.Error(fmt.Sprintf("Could not fetch GCS Bucket for Shard %s hence Monitoring Dashboard will not contain Metrics for the gcs bucket\n error=%v\n", p.DataShardId, fetchGcsErr))
+		}
```
manitgupta, line 418: In my own experimentation I recently realised that the zap logger (that we use in HB), prints the whole stack trace when logging at the `ERROR` level.  This results in nasty looking stack traces when an error is encountered by the customer. 
Because we allow customers to set log-levels, and they probably don't want to see these stack traces unless they are running at `DEBUG` level, I would suggest something like the following - 

1. Log all errors at the `INFO` level. This will just print the error for the customer and not HB internal stack trace.
2. Add a second log below the `INFO` log which prints errr stack trace only if the log level is set to `DEBUG`.

I understand that this might require some tweaking (for example, make `DEBUG` behave the same way as `ERROR` ?) which might be out of scope of this PR and more relevant to "fix logging in HB" that we already have in our plate. Just noting this here down for posterity.

Mentioning #350 here to link back to this comment whenever we get to it. @bharadwaj-aditya something to think about..

conversion/conversion.go
```diff
@@ -387,11 +409,20 @@ func dataFromDatabaseForDataflowMigration(targetProfile profiles.TargetProfile,
...
+		// create monitoring dashboard for a single shard
 		monitoringResources := metrics.MonitoringMetricsResources{
 			ProjectId:            targetProfile.Conn.Sp.Project,
 			DataflowJobId:        dfOutput.JobID,
 			DatastreamId:         streamingCfg.DatastreamCfg.StreamId,
-			GcsBucketId:          streamingCfg.TmpDir,
+			GcsBucketId:          gcsBucket,
 			PubsubSubscriptionId: streamingCfg.PubsubCfg.SubscriptionId,
 			SpannerInstanceId:    targetProfile.Conn.Sp.Instance,
 			SpannerDatabaseId:    targetProfile.Conn.Sp.Dbname,
```
manitgupta, line 428: Curious - we don't need `region` to chart some of these resources?

internal/convert.go
```diff
@@ -203,12 +207,15 @@ type streamingStats struct {
...
 	DataflowGcloudCmd             string
+	GcsResources                  GcsResources
 	ShardToDataStreamNameMap      map[string]string
 	ShardToDataflowInfoMap        map[string]ShardedDataflowJobResources
 	PubsubCfg                     PubsubCfg
 	ShardToPubsubIdMap            map[string]PubsubCfg
+	ShardToGcsResources           map[string]GcsResources
 	MonitoringDashboard           string
 	ShardToMonitoringDashboardMap map[string]string
+	AggMonitoringDashboard        string
```
manitgupta, line 218: Similar to `GcsResources`, lets create an object for Dashboard instead of it being a string.

ui/src/app/components/prepare-migration/prepare-migration.component.html
```diff
@@ -321,6 +321,12 @@ <h3> Monitoring Dashboards:</h3>
             </li>
       </span>
       <span *ngIf="isSharded">
+        <span *ngIf="resourcesGenerated.AggMonitoringDashboardName !== '' && isSharded">
+          <li>
+            Aggregated Monitoring Dashboard:
+            <a [href]="resourcesGenerated.AggMonitoringDashboardUrl" target="_blank">{{resourcesGenerated.AggMonitoringDashboardName}}</a>
+          </li>
+        </span>
```
manitgupta, line 329: Nit: Since you have now added a new section for generated resources, I think we should display them in a tabular format 🙂

webv2/web.go
```diff
@@ -2984,17 +2988,19 @@ type GeneratedResources struct {
...
+	DataflowJobUrl             string `json:"DataflowJobUrl"`
+	DataflowGcloudCmd          string `json:"DataflowGcloudCmd"`
+	PubsubTopicName            string `json:"PubsubTopicName"`
+	PubsubTopicUrl             string `json:"PubsubTopicUrl"`
+	PubsubSubscriptionName     string `json:"PubsubSubscriptionName"`
+	PubsubSubscriptionUrl      string `json:"PubsubSubscriptionUrl"`
+	MonitoringDashboardName    string `json:"MonitoringDashboardName"`
+	MonitoringDashboardUrl     string `json:"MonitoringDashboardUrl"`
+	AggMonitoringDashboardName string `json:"AggMonitoringDashboardName"`
+	AggMonitoringDashboardUrl  string `json:"AggMonitoringDashboardUrl"`
```
manitgupta, line 3003: This should be done via a struct for `MonitoringResources`.

common/metrics/metrics.go
```diff
@@ -31,21 +30,30 @@ var dashboardClient *dashboard.DashboardsClient
...
+	GcsBucketId              string
+	PubsubSubscriptionId     string
+	SpannerInstanceId        string
+	SpannerDatabaseId        string
+	ShardToDataStreamNameMap map[string]string
+	ShardToDataflowInfoMap   map[string]internal.ShardedDataflowJobResources
+	ShardToPubsubIdMap       map[string]internal.PubsubCfg
+	ShardToGcsMap            map[string]internal.GcsResources
+	ShardId                  string
 }
```
manitgupta, line 45: I think it probably makes sense have structs in separate file, utility functions in a separate file, and shard-level and aggregated dashboard creation stuff in their own files. 

`metrics.go` is already a long file and this will soon become hard to manage if we add more metrics in the future.

conversion/conversion.go
asthamohta: https://cloud.google.com/storage/docs/buckets
It mentions in this documentation that the name needs to be "globally unique"

conversion/conversion.go
asthamohta: I think keeping gcs in capital reduces readability
Also if we look into how other golang libraries have done it, example datastream, we can see they also use Gcs not GCS as given [here](https://pkg.go.dev/cloud.google.com/go/datastream@v1.10.3/apiv1/datastreampb#ConnectionProfile.GetGcsProfile)

webv2/web.go
asthamohta: I am not sure I understand this comment. Can you elaborate?

manitgupta (approved):

shreyakhajanchi (approved):
LGTM

common/metrics/dashboard_components.go
```diff
@@ -0,0 +1,360 @@
+package metrics
```
shreyakhajanchi, line 1: Please add license header to all the files

PR #691, opened 2023-11-22
Introduce migrationJobIds and cleanup CLI command

manitgupta:
### Migration Jobs

This feature introduces the concept of a `migrationJobId`. A migration job Id encapsulates the details associated with a migration run by Harbourbridge. It is a central identifier that can be used to refer to a given migration orchestration at any point of time.

To begin with, we implement `migrationJobIds` to keep track of the resources created during an orchestration. On each orchestration, we create and persist `JobExecutionData` and `ShardExecutionData`. A non-sharded migration is handled as a special case of a sharded migration in the backend, though the user experience remains unchanged. 

During a job orchestration, a job is first created and persisted. This contains such related metadata, for example if the job is sharded or not. Then the shard level orchestration begins. In this case, orchestration creates resources and then created resources are persisted. 
Finally, any resources generated at the job level (such as an aggregated monitoring dashboard) are persisted, at the job level. 

The `migrationJobId` will be printed on the console logs. For the UI flow, it will also be shown at the top of `Generated Resources` section. Revamping the experience of displaying resources to the user is out of scope of this PR.

### Cleanup

This feature uses the `migrationJobs` above to provide a CLI to the user to cleanup a migration. A cleanup cleans up the temporary resources used for performing the migration e.g Datastream stream, Dataflow, Pubsub, Monitoring dashboard. It does not clean the `output` of a migratio job - e.g it does not drop the Spanner database. 
It also does not cleanup GCS buckets, due to GCS specific limitations around bulk deletes. This will be handled via lifecycle policies (TBD later).

Customers have the option to select which all resources to delete via flags. An example run of the CLI is - 

```
./spanner-migration-tool cleanup --jobId "SMT-8f92c3a2-bf6b-4047-a5b1-f9b586be411c" --target-profile="project=span-cloud-testing,instance=manit-testing" --datastream --dataflow --pubsub --monitoring
```
This will delete the following resources for each shard - 
1. Datastream stream
2. Dataflow job
3. Pubsub topic and subscription
4. GCS bucket notification
5. Monitoring dashboard

Finally, it will also delete the job level aggregate monitoring dashboard.

### Testing

1. Tested sharded and non-sharded migrations via UI and CLI.
6. Tested sharded and non-sharded cleanup via CLI.


cmd/cleanup.go
```diff
@@ -0,0 +1,116 @@
+package cmd
```
asthamohta, line 1: Please add license to all the files

common/metrics/dashboard_components.go
```diff
@@ -42,18 +42,18 @@ var dashboardClient *dashboard.DashboardsClient
...
-	SpannerDatabaseId        string
-	ShardToDataStreamNameMap map[string]string
-	ShardToDataflowInfoMap   map[string]internal.ShardedDataflowJobResources
-	ShardToPubsubIdMap       map[string]internal.PubsubCfg
-	ShardToGcsMap            map[string]internal.GcsResources
-	ShardId                  string
+	ProjectId                     string
+	DataflowJobId                 string
+	DatastreamId                  string
+	MetadataGcsBucketId           string
```
asthamohta, line 48: Maybe this is not super clear to me but isn't the metadata gcs bucket the TmpDir passed by users? This isn't the same gcs bucket. This refers to the gcs bucket associated with each shard

conversion/conversion.go
```diff
@@ -318,6 +322,7 @@ func dataFromDatabase(ctx context.Context, sourceProfile profiles.SourceProfile,
 			dfJobId := dfOutput.JobID
 			gcloudCmd := dfOutput.GCloudCmd
 			streamingCfg, _ := streamInfo["streamingCfg"].(streaming.StreamingCfg)
+			// store the generated resources locally in conv, this is used as source of truth for persistence and the UI (should change to persisted values)
```
asthamohta, line 325: Should this comment be on line 356?

streaming/cleanup.go
```diff
@@ -0,0 +1,221 @@
...
+		err = CleanupDataflowJob(ctx, c, dataflowResources.JobId, project, dataflowResources.Region)
+		if err != nil {
+			logger.Log.Debug(fmt.Sprintf("Cleanup of the dataflow job: %s was unsuccessful, please clean up the job manually. err = %v\n", dataflowResources.JobId, err))
+		} else {
+			logger.Log.Info("Dataflow job cleaned up successfully.\n")
+		}
+	}
+	if options.Datastream {
+		logger.Log.Debug("Attempting to delete datastream stream...\n")
+		dsClient, err := datastream.NewClient(ctx)
```
asthamohta, line 70: Wouldn't this client be created everytime? There is a common getDatastreamClient method in conversion.go that can be used

streaming/cleanup.go
```diff
@@ -0,0 +1,221 @@
...
+		if e == iterator.Done {
+			break
+		}
+		if e != nil {
+			err = e
+			break
+		}
+		var shardExecutionData ShardExecutionData
+		row.ToStruct(&shardExecutionData)
+		if filterbyDataShardId(shardExecutionData.DataShardId, dataShardIds) {
```
asthamohta, line 184: Why not use the contains function in golang here?

common/metrics/dashboard_components.go
manitgupta: Naming things is hard I guess, they are both Metadata bucket in a sense, but this bucket is more _meta_ than the one used in Datastream since that one actually contains customer data, while this one will persist HB local stuff.

streaming/cleanup.go
manitgupta: There is no contains for a list in golang

streaming/store.go
```diff
@@ -0,0 +1,202 @@
...
+	err = writeJobExecutionMetadata(ctx, migrationJobId, isSharded, dbName, time.Now(), client)
+	if err != nil {
+		err = fmt.Errorf("can't store generated resources for datashard: %v", err)
+		return err
+	}
+	fmt.Printf("Generated resources stored successfully for migration jobId: %s. You can also look at the 'spannermigrationtool_metadata' database in your spanner instance to get this jobId at a later point of time.\n", migrationJobId)
+	return nil
+}
+
+func UpdateJobExecutionDataWithAggregateMonitoringResources(ctx context.Context, targetProfile profiles.TargetProfile, sourceProfile profiles.SourceProfile, conv *internal.Conv, migrationJobId string) error {
```
asthamohta, line 75: Should this be a more generic function in case in the future we want to update something else as well so this can be done in a single transaction?

streaming/cleanup.go
manitgupta: Yeah it would, I just followed the existing approach in HB of creating client on demand.

conversion/conversion.go
manitgupta: Thanks, fixed!

streaming/cleanup.go
manitgupta: I had to use a bunch of other clients for cleanup (as done in other options) so I thought might as well keep it uniform for now and then refactor later

asthamohta: Okay sure

cmd/cleanup.go
manitgupta: Missed this, added now

streaming/cleanup.go
asthamohta: https://pkg.go.dev/golang.org/x/exp/slices#Contains

streaming/store.go
manitgupta: This is the reason why I made it super specific. We don't know what we will need to update in the future and its hard to define the interface of what all parameters we may need for that update. For example, the new information that needs to go in might not be in `conv`. 

I think we can refactor once we have a usecase but I am also okay with renaming this if you think that is better. Let me know what you think.

streaming/cleanup.go
manitgupta: Ahh I see, I was trying  `dataShardIds.contains()`. Thanks will update

manitgupta: Very interesting, the native `slices` package is available in Go 1.21 onwards. 
We are on go Go 1.19, and in order to consume this package, I will have to `go get` it. Seems like a lot for checking an array 😀

conversion/conversion.go
```diff
@@ -347,10 +350,22 @@ func dataFromDatabase(ctx context.Context, sourceProfile profiles.SourceProfile,
 				dashboardName = strings.Split(respDash.Name, "/")[3]
 				fmt.Printf("Monitoring Dashboard: %+v\n", dashboardName)
 			}
-
+			// store the generated resources locally in conv, this is used as source of truth for persistence and the UI (should change to persisted values)
 			streaming.StoreGeneratedResources(conv, streamingCfg, dfJobId, gcloudCmd, targetProfile.Conn.Sp.Project, "", internal.GcsResources{BucketName: gcsBucket}, dashboardName)
+			//persist job and shard level data in the metadata db
+			err = streaming.PersistJobExecutionData(ctx, targetProfile, sourceProfile, conv, migrationJobId, false)
+			if err != nil {
+				fmt.Println("Error storing generated resources in SMT metadata store...the migration job will still continue as intended.", err)
```
bharadwaj-aditya, line 358: nit: Please log rather than just print.

profiles/common.go
```diff
@@ -61,6 +61,23 @@ func ParseMap(s string) (map[string]string, error) {
 	return params, nil
 }
 
+func ParseList(s string)([]string, error) {
```
bharadwaj-aditya, line 64: I guess this should be in utils.go

streaming/cleanup.go
```diff
@@ -0,0 +1,234 @@
...
+		} else {
+			logger.Log.Info("Dataflow job cleaned up successfully.\n")
+		}
+	}
+	if options.Datastream {
+		logger.Log.Debug("Attempting to delete datastream stream...\n")
+		dsClient, err := datastream.NewClient(ctx)
+		logger.Log.Debug("Created datastream client...")
+		if err != nil {
+			return fmt.Errorf("datastream client can not be created: %v", err)
```
bharadwaj-aditya, line 86: same comment as above. We should log a failure and proceed, not fail if we encounter an error.

streaming/cleanup.go
```diff
@@ -0,0 +1,234 @@
...
+	Datastream bool
+	Pubsub     bool
+	Monitoring bool
+}
+
+type JobCleanupOptions struct {
+	Monitoring bool
+}
+
+func InitiateJobCleanup(ctx context.Context, options JobCleanupOptions, jobExecutionData JobExecutionData, project string) error {
```
bharadwaj-aditya, line 45: It is unclear from the name how JobCleanup is different from ShardCleanup. 
ShardCleanup should be a subset of this functionality

webv2/helpers/helpers.go
```diff
@@ -79,6 +79,24 @@ func createDatabase(ctx context.Context, uri string) error {
 				SchemaConversionObject JSON NOT NULL,
 				CreateTimestamp TIMESTAMP NOT NULL,
 			  ) PRIMARY KEY(VersionId)`,
+			  `CREATE TABLE JobExecutionData (
+				MigrationJobId STRING(50) NOT NULL,
+				SpannerDatabaseName STRING(50) NOT NULL,
+				AggMonitoringResources JSON,
```
bharadwaj-aditya, line 85: What all is stored in this field ? Please add a comment

webv2/helpers/helpers.go
```diff
@@ -79,6 +79,24 @@ func createDatabase(ctx context.Context, uri string) error {
...
+				IsShardedMigration BOOL NOT NULL,
+				CreateTimestamp TIMESTAMP NOT NULL,
```
bharadwaj-aditya, line 87: We should maintain a state of the job here

bharadwaj-aditya: We should also soft delete jobs, not really clear them out. The state can double up for this if required

webv2/helpers/helpers.go
```diff
@@ -79,6 +79,24 @@ func createDatabase(ctx context.Context, uri string) error {
...
+			  ) PRIMARY KEY(MigrationJobId)`,
+			  `CREATE TABLE ShardExecutionData (
+				MigrationJobId STRING(50) NOT NULL,
+				DataShardId STRING(50) NOT NULL,
```
bharadwaj-aditya, line 91: If we change any part of this execution pipeline, the design of the table would change. What is the thought process of handling such scenarios.

webv2/helpers/helpers.go
bharadwaj-aditya: Probably also need an updated time here

bharadwaj-aditya (changes requested):
I think the storage structure needs some more thought.

conversion/conversion.go
manitgupta: Done

profiles/common.go
manitgupta: This file also implements `ParseMap` which is also used for parsing the input profiles in the CLI so I've kept it here for consistency.

streaming/cleanup.go
manitgupta: I understand the reason for the confusion. Restructured code to make `ShardCleanup` a subset of the `JobCleanup` functionality.

webv2/helpers/helpers.go
manitgupta: Not possible to put a comment here inline since it is a string. Added a comment at the top of the method.

webv2/helpers/helpers.go
manitgupta: This is not a `JobDetails` table. It only contains the execution data for job i.e artefacts that are relevant for job execution. 

Job status and associated semantics will be relevant to a `JobDetails` and will add more details such as `status`, `type`, `config` etc and is out of scope of the current PR.

webv2/helpers/helpers.go
manitgupta: My thought process is evolve the schema as we have a line of sight on what different types of jobs would look like. 
Short of defining everything in a NoSQL like fashion and storing job related information as a JSON, it will be hard to design a schema which fits very varied possible job designs. Even storing it in JSON would mean shifting the complexity to the application rather than solving the problem itself.

streaming/cleanup.go
manitgupta: Makes sense, I realised that because of the way clients are currently created, this will require some amount of refactoring to do properly. I will make the changes and update the PR!

manitgupta: Refactored the code to skip any failures, and also merged UI code flow for cleaning up.

webv2/helpers/helpers.go
manitgupta: Added `UpdatedAt` field to both the tables

streaming/cleanup.go
```diff
@@ -0,0 +1,318 @@
...
+		if err != nil {
+			logger.Log.Debug("Unable to read aggregate monitoring metadata for deletion\n")
+		} else {
+			cleanupMonitoringDashboard(ctx, aggMonitoringResources, project)
+		}
+	}
+}
+
+func GetJobDetails(ctx context.Context, migrationJobId string, dataShardIds []string, project string, instance string) (JobExecutionData, []ShardExecutionData, error) {
+	dbURI := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, "spannermigrationtool_metadata")
```
shreyakhajanchi, line 78: better to define spannermigrationtool_metadata as a constant in utils/constants to avoids spell errors

conversion/conversion.go
```diff
@@ -347,10 +350,22 @@ func dataFromDatabase(ctx context.Context, sourceProfile profiles.SourceProfile,
...
-
+			// store the generated resources locally in conv, this is used as source of truth for persistence and the UI (should change to persisted values)
 			streaming.StoreGeneratedResources(conv, streamingCfg, dfJobId, gcloudCmd, targetProfile.Conn.Sp.Project, "", internal.GcsResources{BucketName: gcsBucket}, dashboardName)
+			//persist job and shard level data in the metadata db
+			err = streaming.PersistJobExecutionData(ctx, targetProfile, sourceProfile, conv, migrationJobId, false)
+			if err != nil {
+				logger.Log.Info(fmt.Sprintf("Error storing generated resources in SMT metadata store...the migration job will still continue as intended. %v", err))
+			} else {
+				//only attempt persisting shard level data if the job level data is persisted
+				err = streaming.PersistShardExecutionData(ctx, targetProfile, sourceProfile, conv, migrationJobId, "smt-default")
```
shreyakhajanchi, line 362: Store "smt-default" as constant too for better readability and usage

conversion/conversion.go
```diff
@@ -347,10 +350,22 @@ func dataFromDatabase(ctx context.Context, sourceProfile profiles.SourceProfile,
 				dashboardName = strings.Split(respDash.Name, "/")[3]
 				fmt.Printf("Monitoring Dashboard: %+v\n", dashboardName)
 			}
-
+			// store the generated resources locally in conv, this is used as source of truth for persistence and the UI (should change to persisted values)
 			streaming.StoreGeneratedResources(conv, streamingCfg, dfJobId, gcloudCmd, targetProfile.Conn.Sp.Project, "", internal.GcsResources{BucketName: gcsBucket}, dashboardName)
+			//persist job and shard level data in the metadata db
+			err = streaming.PersistJobExecutionData(ctx, targetProfile, sourceProfile, conv, migrationJobId, false)
+			if err != nil {
+				logger.Log.Info(fmt.Sprintf("Error storing generated resources in SMT metadata store...the migration job will still continue as intended. %v", err))
```
shreyakhajanchi, line 359: Will it be better to modify the log msg to say job level data and shard level data respectively?

cmd/cleanup.go
```diff
@@ -0,0 +1,116 @@
...
+	if !(cmd.datastream || cmd.dataflow || cmd.pubsub || cmd.monitoring) {
+		logger.Log.Error("At least one of datastream, dataflow, pubsub or monitoring must be specified, we recommend cleaning up all resources!\n")
+		return subcommands.ExitUsageError
+	}
+	// all input parameters have been validated
+	if cmd.validate {
+		logger.Log.Info("All input parameters are valid.")
+		return subcommands.ExitSuccess
+	}
+	jobExecutionData, shardExecutionDataList, err := streaming.GetJobDetails(ctx, cmd.jobId, dataShardIds, project, instance)
```
shreyakhajanchi, line 103: We don't seem to be storing anywhere the details of migration type i.e. POC or minimal . And the job details are only populated for streaming migrations, in future when we aim to add statefulness to these migrations we might need this detail captured somewhere

manitgupta: No statefulness for POC migrations

conversion/conversion.go
manitgupta: Done

streaming/cleanup.go
manitgupta: Done

cmd/cleanup.go
```diff
@@ -0,0 +1,116 @@
...
+		logger.Log.Debug(fmt.Sprintf("Could not parse datashardIds: %v\n", err))
+		return subcommands.ExitFailure
+	}
+	if !(cmd.datastream || cmd.dataflow || cmd.pubsub || cmd.monitoring) {
+		logger.Log.Error("At least one of datastream, dataflow, pubsub or monitoring must be specified, we recommend cleaning up all resources!\n")
+		return subcommands.ExitUsageError
+	}
+	// all input parameters have been validated
+	if cmd.validate {
+		logger.Log.Info("All input parameters are valid.")
```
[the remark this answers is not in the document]
manitgupta: Done

conversion/conversion.go
manitgupta: Done

asthamohta (approved):
Lgtm

cmd/cleanup.go
shreyakhajanchi: sure, makes sense

shreyakhajanchi (approved):
LGTM

cmd/cleanup.go
```diff
@@ -0,0 +1,112 @@
...
+	"os"
+	"path"
+
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/logger"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/profiles"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/streaming"
+	"github.com/google/subcommands"
+)
+
+type CleanupCmd struct {
```
bharadwaj-aditya, line 29: Please add comments/docs on the purpose of this file and command

cmd/cleanup.go
```diff
@@ -0,0 +1,112 @@
...
+
+Cleanup GCP resources generated as part of setting up a migration pipeline by providing a 
+jobId generated during the job creation.
+`, path.Base(os.Args[0]))
+}
+
+// SetFlags sets the flags.
+func (cmd *CleanupCmd) SetFlags(f *flag.FlagSet) {
+	f.StringVar(&cmd.jobId, "jobId", "", "Flag for specifying the migration jobId")
+	f.StringVar(&cmd.dataShardIds, "dataShardIds", "", "Flag for specifying a comma separated list of dataShardIds to be cleaned up. Defaults to ALL shards. Optional flag, and only valid for a sharded migration.")
```
bharadwaj-aditya, line 63: nit: It will be easier to read if the flags are grouped as required and optional. It is unclear which are required currently.

cmd/cleanup.go
```diff
@@ -0,0 +1,112 @@
...
+		logger.Log.Info("All required parameters are present, validated that the command is syntactically correct.\n")
+		return subcommands.ExitSuccess
+	}
+	jobCleanupOptions := streaming.JobCleanupOptions{
+		Datastream: cmd.datastream,
+		Dataflow:   cmd.dataflow,
+		Pubsub:     cmd.pubsub,
+		Monitoring: cmd.monitoring,
+	}
+	logger.Log.Info(fmt.Sprintf("Initiating job cleanup for jobId: %v, dataShardIds: %v \n", cmd.jobId, dataShardIds))
```
bharadwaj-aditya, line 109: nit: It will help if we log what all is going to be cleaned up here.

cmd/data.go
```diff
@@ -116,7 +116,8 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 		banner string
 	)
 	// Populate migration request id and migration type in conv object.
-	conv.Audit.MigrationRequestId = "SMT-" + uuid.New().String()
+	conv.Audit.MigrationRequestId, _ = utils.GenerateName("smt-job")
+	conv.Audit.MigrationRequestId = strings.Replace(conv.Audit.MigrationRequestId, "_", "-", -1)
```
bharadwaj-aditya, line 120: why do we need to do this replacement ? Can we not just edit the job id to comply with this structure if required ?

cmd/schema_and_data.go
```diff
@@ -129,7 +129,8 @@ func (cmd *SchemaAndDataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...
 	conv.Audit.SchemaConversionDuration = schemaCoversionEndTime.Sub(schemaConversionStartTime)
 
 	// Populate migration request id and migration type in conv object.
-	conv.Audit.MigrationRequestId = "SMT-" + uuid.New().String()
+	conv.Audit.MigrationRequestId, _ = utils.GenerateName("smt-job")
```
bharadwaj-aditya, line 132: seems like this is being done in 3 or more places. Maybe worth moving to a common method.

streaming/cleanup.go
```diff
@@ -0,0 +1,315 @@
...
+	defer client.Close()
+	txn := client.ReadOnlyTransaction()
+	defer txn.Close()
+
+	//fetch all resources
+	var resourceQuery spanner.Statement
+	if dataShardIds != nil {
+		//query the provided data shards only
+		resourceQuery = spanner.Statement{
+			SQL: `SELECT
```
bharadwaj-aditya, line 139: nit: both the queries can probably be merged adding the data shards where required. This will avoid changes in 2 places going forward.

streaming/cleanup.go
```diff
@@ -0,0 +1,315 @@
...
+				WHERE JobId = @migrationJobId and ResourceType = @resourceType`,
+			Params: map[string]interface{}{
+				"migrationJobId": migrationJobId,
+				"resourceType":   resourceType,
+			},
+		}
+	}
+	iter := txn.Query(ctx, resourceQuery)
+	jobResourcesList := []JobResources{}
+	for {
```
bharadwaj-aditya, line 176: A more "Go" way of doing this would be 

> for row,e := txn.Query(..) {
> ....
> }

or

>for iter.Next() {
>}

streaming/store.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/common/constants"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/common/utils"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/internal"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/logger"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/profiles"
+)
+
+// This file contains functions specific to storing state for a minimal downtime migration.
+
+// Struct to store minimal downtime migration specific metadata in the JobMetadata field in the JobDetails table
```
bharadwaj-aditya, line 33: nit: Should this file be called store or job ? or something else ?

streaming/store.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		err = fmt.Errorf("can't create database client: %v", err)
+		return err
+	}
+	err = writeJobDetails(ctx, migrationJobId, isSharded, conv, dbName, time.Now(), client)
+	if err != nil {
+		err = fmt.Errorf("can't store generated resources for datashard: %v", err)
+		return err
+	}
+	logger.Log.Info(fmt.Sprintf("Generated resources stored successfully for migration jobId: %s. You can also look at the 'spannermigrationtool_metadata' database in your spanner instance to get this jobId at a later point of time.\n", migrationJobId))
```
bharadwaj-aditya, line 61: nit: Do we need \n while using Log.Info - I think it automatically adds a new line anyway

streaming/store.go
```diff
@@ -0,0 +1,248 @@
...
+	})
+	if err != nil {
+		err = fmt.Errorf("can't store aggregate monitoring resources for migration job %s: %v", migrationJobId, err)
+		return err
+	}
+	return nil
+}
+
+// PersistResources stores all the metadata associated with a shard orchestration for a minimal downtime migration in the metadata db. An example of this metadata is generated resources.
+func PersistResources(ctx context.Context, targetProfile profiles.TargetProfile, sourceProfile profiles.SourceProfile, conv *internal.Conv, migrationJobId string, dataShardId string) (err error) {
```
bharadwaj-aditya, line 103: Why are there different persistResources and writeJobResources method ? It sounds very similar

streaming/store.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+	logger.Log.Debug(fmt.Sprintf("Storing generated resources for data shard %s...\n", dataShardId))
+	_, err = client.ReadWriteTransaction(ctx, func(ctx context.Context, txn *spanner.ReadWriteTransaction) error {
+		datastreamMutation, err := createResourceMutation(migrationJobId, dataShardId, datastreamResources.DatastreamName, constants.DATASTREAM_RESOURCE, datastreamResources.DatastreamName, string(datastreamResourcesBytes))
+		if err != nil {
+			return err
+		}
+		dataflowMutation, err := createResourceMutation(migrationJobId, dataShardId, dataflowResources.JobId, constants.DATAFLOW_RESOURCE, dataflowResources.JobId, string(dataflowResourcesBytes))
+		if err != nil {
+			return err
```
bharadwaj-aditya, line 203: We should do a best effort at writing the jobs. Even if one fails, we should log and proceed with the remaining.

webv2/config.json
```diff
@@ -1,5 +1,4 @@
 {
- "GCPProjectID": "",
- "SpannerInstanceID": ""
-}
-
+ "GCPProjectID": "span-cloud-testing",
```
bharadwaj-aditya, line 2: please revert.

webv2/helpers/helpers.go
```diff
@@ -37,16 +37,13 @@ const (
...
-}
-
 func GetSpannerUri(projectId string, instanceId string) string {
-	return fmt.Sprintf("projects/%s/instances/%s/databases/%s", projectId, instanceId, GetMetadataDbName())
+	return fmt.Sprintf("projects/%s/instances/%s/databases/%s", projectId, instanceId, constants.METADATA_DB)
 }
 
+// Creates the schema for the internal metadata database
+// JobDetails contains the execution data associated with a job, such as the Id of the aggregated monitoring dashboard
+// ShardExecutionData contains the execution data associated with a data shard, such as the Id of the dataflow job, datastream stream etc.
```
bharadwaj-aditya, line 46: These comments are stale. Please clean them up.

Nit: it would be good if we can add the comments closer to each table rather than on top.

bharadwaj-aditya (changes requested):
Please also check for the test on these flows.

cmd/data.go
manitgupta: `utils.GenerateName` is used very extensively through the code base, and requires this replacement to work. I have opted for not refactoring that other method currently.

cmd/schema_and_data.go
manitgupta: That would become a wrapper over the `utils.GenerateName`. We should refactor that method itself so that `string.Replace` itself would be called inside it.  Have not done that refactoring as part of this PR.

streaming/cleanup.go
manitgupta: Didn't get you, do you mean extracting out the common part of the query in a separate string? 

I have currently kept it this way because extracting out the common part would decrease readability IMO. Open to suggestions!

streaming/cleanup.go
manitgupta: Interesting, I have just used the existing reference in the code here -

https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/2fa927e9dd835261eb2899bb24ec13bd18467bf1/webv2/session/remote_session_store.go#L60

streaming/store.go
manitgupta: I think this should be called `store.go` since its in the `streaming` package and is responsible for storing job and resource details.

Can rename it to `store_job.go` as well but decided against because we store both `job` and `resource` details. Open to suggestions!

streaming/store.go
manitgupta: It only does so when - 

```
Log.Info("some text")
```

When we use `fmt.Sprintf` it prints whatever the `Sprintf` returns, to a `\n` is needed.

streaming/store.go
manitgupta: `PersistResources` is an externally visible API interface for this package, `writeJobResources` is an internal, package-local method.

streaming/store.go
manitgupta: Why? Currently, all resources are persisted together inside a single transaction (multiple mutations) (refer to Line 217)

Note that we don't fail the migration itself, but we either fail or persist all resources created for a migration as one atomic write.

webv2/config.json
manitgupta: Done

webv2/helpers/helpers.go
manitgupta: Done

manitgupta:
Created a tracking bug to improve orchestration flow: b/314106108

bharadwaj-aditya (approved):
LGTM

main.go
```diff
@@ -52,6 +52,7 @@ func main() {
 	subcommands.Register(&cmd.SchemaCmd{}, "")
 	subcommands.Register(&cmd.DataCmd{}, "")
 	subcommands.Register(&cmd.SchemaAndDataCmd{}, "")
+	subcommands.Register(&cmd.CleanupCmd{}, "")
```
aksharauke, line 55: Is it possible that the cleanup be a subcommand of the flow type?

common/constants/constants.go
```diff
@@ -81,4 +81,21 @@ const (
...
+	MINIMAL_DOWNTIME_MIGRATION = "minimal_downtime"
+	//Job Resource Types
+	DATAFLOW_RESOURCE string = "dataflow"
+	PUBSUB_RESOURCE      string = "pubsub"
+	MONITORING_RESOURCE  string = "monitoring"
+	AGG_MONITORING_RESOURCE  string = "aggregated_monitoring"
+	DATASTREAM_RESOURCE  string = "datastream"
+	GCS_RESOURCE  string = "gcs"
+	// Metadats table names
+	SMT_JOB_TABLE string = "SMT_JOBS"
```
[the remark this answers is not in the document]
manitgupta: The other tables are named `SMT_RESOURCES` and `SMT_STATES` so I just went with plurals everywhere. 😅

main.go
manitgupta: Do you mean forward and reverse flow? 
Yes, ideally, all the 4 commands above should be context of the flow type...we only have one today though so I kept it at same level. 
Maybe we refactor all of these together once we introduce the concept of a "flow"? I don't want this to become a confusing experience until we have that..

streaming/cleanup.go
```diff
@@ -0,0 +1,321 @@
...
+			if err != nil {
+				logger.Log.Debug("Unable to read monitoring metadata for deletion\n")
+			} else {
+				cleanupMonitoringDashboard(ctx, monitoringResources, project)
+			}
+		}
+	}
+}
+
+func FetchResources(ctx context.Context, migrationJobId string, resourceType string, dataShardIds []string, project string, instance string) ([]SmtResources, error) {
```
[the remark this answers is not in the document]
manitgupta: Agreed! It should - some comment on restructuring as the previous one though.

We probably need a `dao` package which flow based DAO implementations in it - forward, reverse etc.

PR #696, opened 2023-11-24
Support GCS tuning from SMT

Deep1998:
Added support to tune gcs from SMT.
Currently, we only support a single tuning option, that is, enabling delete lifecycle rule on the Datastream destination GCS bucket. The rule conditions accept a ttl in days and apply on the data directory prefix.

Additional minor fixes:
- Datastream tuning option no longer stays highlighted before specifying target db
- Added default worker machine type as n1-standard-2.
- Removed Unmarshallers for dataflow and datastream
- Hide tune cdc workers field for postges source type

GCS tuning form:
Screenshot 2023-11-24 at 4 14 54 PM

On check:
Screenshot 2023-11-24 at 4 15 29 PM

Tested:
- Sharded: rule, ttl set correctly with correct prefix
- Sharded: unset ttl does not add lifecycle rule by default
- NonSharded: rule, ttl set correctly with correct prefix
- NonSharded: unset ttl does not add lifecycle rule by default
- CLI sharded ttl gets set


ui/src/app/services/fetch/fetch.service.ts
```diff
@@ -103,6 +103,10 @@ export class FetchService {
     })
   }
 
+  setGcsDetailsForShardedMigrations(payload: IGcsConfig) {
+    return this.http.post(`${this.url}/SetGcsDetailsForShardedMigrations`, payload)
+  }
+
   setDataflowDetailsForShardedMigrations(payload: IDataflowConfig) {
     return this.http.post(`${this.url}/SetDataflowDetailsForShardedMigrations`, {
       DataflowConfig: payload
```
manitgupta, line 112: We should think about refactoring this to create a generic API.

common/utils/utils.go
```diff
@@ -679,3 +680,43 @@ func sortKeysByOrder(pks []ddl.IndexKey) {
...
+	}
+
+	attrs, err := bucket.Update(ctx, bucketAttrsToUpdate)
+	if err != nil {
+		return fmt.Errorf("could not bucket with lifecycle: %w", err)
+	}
+	logger.Log.Info(fmt.Sprintf("Added lifecycle rule to bucket %v\n. Rule Action: %v\t Rule Condition: %v\n",
+		bucketName, attrs.Lifecycle.Rules[0].Action, attrs.Lifecycle.Rules[0].Condition))
+	return nil
+}
```
manitgupta, line 722: This, along with other methods related to GCS should be in own file which creates and reuses a singleton client for gcs.

conversion/conversion.go
```diff
@@ -327,12 +323,22 @@ func dataFromDatabase(ctx context.Context, sourceProfile profiles.SourceProfile,
...
+			gcsBucket, gcsDestPrefix, fetchGcsErr := streaming.FetchTargetBucketAndPath(ctx, dsClient, targetProfile.Conn.Sp.Project, streamingCfg.DatastreamCfg.DestinationConnectionConfig)
 			if fetchGcsErr != nil {
 				logger.Log.Info("Could not fetch GCS Bucket, hence Monitoring Dashboard will not contain Metrics for the gcs bucket\n")
 				logger.Log.Debug("Error", zap.Error(fetchGcsErr))
 			}
 
+			// Try to apply lifecycle rule to Datastream destination bucket.
+			gcsConfig := streamingCfg.GcsCfg
+			if gcsConfig.TtlInDaysSet {
+				err = utils.EnableBucketLifecycleDeleteRule(ctx, gcsBucket, []string{gcsDestPrefix}, gcsConfig.TtlInDays)
```
manitgupta, line 335: `FetchTargetBucketAndPath` is defined in `streaming` package, while the bucket lifecycle method is defined in `util`. Please refactor this for a more coherent structure.

ui/src/app/components/prepare-migration/prepare-migration.component.ts
```diff
@@ -503,6 +513,33 @@ export class PrepareMigrationComponent implements OnInit {
...
+    })
+    dialogRef.afterClosed().subscribe(() => {
+      this.gcsConfig = {
+        ttlInDays: localStorage.getItem(Gcs.TtlInDays) as string,
+        ttlInDaysSet: (localStorage.getItem(Gcs.TtlInDaysSet) as string === 'true')    
+      }
+      this.isGcsConfigurationSet = localStorage.getItem(Gcs.IsGcsConfigSet) as string === 'true'
+      // We only call setDatastreamDetailsForShardedMigrations for sharded flows which is fetched later to create sharding configs for the shards. 
+      // Non-sharded flows write a streaming config file during the migrate() operation using data passed in the payload on the fly.
+      // This should be refactored to keep the flow same for both sharded and non-sharded.
```
manitgupta, line 530: Stray comment

conversion/conversion.go
Deep1998: moved FetchTargetBucketAndPath to utils. Ideally we would want a datastream accessor and dataflow accessor (like gcs) but that will be a bigger task.

ui/src/app/services/fetch/fetch.service.ts
Deep1998: We can probably have a common method definition but we will need some switch cases inside backend to differentiate the resources.

Its 3 small methods vs 1 large method.

Created this for tracking anyways: b/313957500

common/utils/utils.go
Deep1998: Yeah we currently create multiple clients as and when needed for gcs and datastream in a bunch of places.
We should move to an accessor model like we discussed earlier.
Tracking: b/313958874

conversion/conversion.go
```diff
@@ -272,15 +272,11 @@ func snapshotMigrationHandler(sourceProfile profiles.SourceProfile, config write
...
-		dataShard.DatastreamConfig = shardedDataflowConfig.DatastreamConfig
-	}
-}
-
-func updateShardsWithDataflowConfig(shardedDataflowConfig profiles.ShardConfigurationDataflow) {
-	for _, dataShard := range shardedDataflowConfig.DataShards {
-		dataShard.DataflowConfig = shardedDataflowConfig.DataflowConfig
+func updateShardsWithTuningConfigs(shardedTuningConfig profiles.ShardConfigurationDataflow) {
+	for _, dataShard := range shardedTuningConfig.DataShards {
+		dataShard.DatastreamConfig = shardedTuningConfig.DatastreamConfig
```
bharadwaj-aditya, line 277: nit: Where are the defaults for these defined ?

bharadwaj-aditya (commented):
The defaults for these values should be set from the backend. I can't seem to find that in this PR.

manitgupta (approved):
LGTM.

I see there is some refactoring done, so please test the related flows associated with it.

PR #709, opened 2023-12-05
Added support for Postgresql include objects config for datastream job

shreyakhajanchi:


streaming/streaming.go
```diff
@@ -292,19 +294,35 @@ func getPostgreSQLSourceStreamConfig(datastreamCfg DatastreamCfg) (*datastreampb
...
 		return nil, fmt.Errorf("could not parse properties: %v", err)
 	}
-	var excludeObjects []*datastreampb.PostgresqlSchema
-	for _, s := range []string{"information_schema", "postgres", "pg_catalog", "pg_temp_1", "pg_toast", "pg_toast_temp_1"} {
-		excludeObjects = append(excludeObjects, &datastreampb.PostgresqlSchema{
-			Schema: s,
-		})
+	postgreSQLSchema := []*datastreampb.PostgresqlSchema{}
+	for schema, tableList := range datastreamCfg.TableSchemaMap {
+		postgreSQLTables := []*datastreampb.PostgresqlTable{}
```
bharadwaj-aditya, line 299: can you add a unit test case to cover this ?

bharadwaj-aditya (approved):
LGTM

conversion/conversion.go
```diff
@@ -385,10 +385,10 @@ func dataFromDatabaseForDataflowMigration(targetProfile profiles.TargetProfile,
 	conv.Audit.StreamingStats.ShardToDataflowInfoMap = make(map[string]internal.ShardedDataflowJobResources)
 	conv.Audit.StreamingStats.ShardToGcsResources = make(map[string]internal.GcsResources)
 	conv.Audit.StreamingStats.ShardToMonitoringResourcesMap = make(map[string]internal.MonitoringResources)
-	tableList, err := common.GetIncludedSrcTablesFromConv(conv)
+	tableListWithSchema, err := common.GetIncludedSrcTablesFromConv(conv)
```
manitgupta, line 388: Nit: Maybe rename to `schemaToTableMap`

conversion/conversion.go
```diff
@@ -385,10 +385,10 @@ func dataFromDatabaseForDataflowMigration(targetProfile profiles.TargetProfile,
...
 	if err != nil {
 		fmt.Printf("unable to determine tableList from schema, falling back to full database")
-		tableList = []string{}
+		tableListWithSchema = map[string][]string{}
```
manitgupta, line 391: I think the following interface will be better - 

```
map[string]SchemaDetails

struct SchemaDetails {
tableDetails TableDetails[]
}

TableDetails struct {
tableName string
}

```

This will help in capturing details at the schema level and future properties which need to be passed at the table level as well.

Note: Please re-use existing structs as appropriate - this is just an indicative guidance.

streaming/streaming.go
```diff
@@ -73,15 +73,15 @@ type DstConnCfg struct {
...
-	TableList                   []string   `json:"tableList"`
-	MaxConcurrentBackfillTasks  string     `json:"maxConcurrentBackfillTasks"`
-	MaxConcurrentCdcTasks       string     `json:"maxConcurrentCdcTasks"`
+	StreamId                    string              `json:"streamId"`
+	StreamLocation              string              `json:"streamLocation"`
+	StreamDisplayName           string              `json:"streamDisplayName"`
+	SourceConnectionConfig      SrcConnCfg          `json:"sourceConnectionConfig"`
+	DestinationConnectionConfig DstConnCfg          `json:"destinationConnectionConfig"`
+	Properties                  string              `json:"properties"`
+	TableSchemaMap              map[string][]string `json:"tableList"`
```
manitgupta, line 82: Is this backward compatible? Will this cause existing streaming JSON configurations to break?

shreyakhajanchi: It is backward compatible

manitgupta: This is part of `streamingCfg` which is used exposed to the customer in case of a non-sharded migration right?

conversion/conversion.go
shreyakhajanchi: Modified to use structs

PR #719, opened 2023-12-11
Fixing issues with validateDb flow in case of data only migration for PostgreSQL dialect

shreyakhajanchi:
Fixes :
- isUnique column in `information_schema.indexes` is of `STRING` datatype in PG dialect whereas its of `BOOL` datatype in GSQL dialect
- Added toSpannerType conversions for PG Dialect datatypes
- In case of PG dialect Spanner by default adds is_nullable = false to all the columns that are a part of primary key.


sources/spanner/infoschema.go
```diff
@@ -349,19 +350,31 @@ func (isi InfoSchemaImpl) GetIndexes(conv *internal.Conv, table common.SchemaAnd
...
-		if err != nil {
-			fmt.Println(err)
-			conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
-			continue
+		if isi.SpDialect == constants.DIALECT_POSTGRESQL {
+			err = row.Columns(&name, &column, &sequence, &ordering, &isPgUnique)
+			if err != nil {
+				fmt.Println(err)
+				conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
+				continue
```
manitgupta, line 358: This structure is not correct, we need to think of a better way to write this code where PGSQL specific nuances are not implemented as if statements - this is very error prone.

Please create a tech debt item with all the details.

sources/spanner/infoschema.go
```diff
@@ -349,19 +350,31 @@ func (isi InfoSchemaImpl) GetIndexes(conv *internal.Conv, table common.SchemaAnd
...
-		err = row.Columns(&name, &column, &sequence, &ordering, &isUnique)
-		if err != nil {
-			fmt.Println(err)
-			conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
-			continue
+		if isi.SpDialect == constants.DIALECT_POSTGRESQL {
+			err = row.Columns(&name, &column, &sequence, &ordering, &isPgUnique)
+			if err != nil {
+				fmt.Println(err)
+				conv.Unexpected(fmt.Sprintf("Can't scan: %v", err))
```
manitgupta, line 357: If an `err` is encountered, we still go ahead with constructing the `conv` object.

Can this lead to issues? For example, if an error is encountered while reading indexes from the source database, that index will never get migrated during the schema conversion.

I wonder what should be our strategy to handle these failures...should we -

1. Completely fail and prevent user from moving forward.
2. Should we do a best effort conversion and inform the user that there were partial errors in reading the schema.
3. If there are partial errors, should we store this information in `conv` or someplace else - this can be used has a hook later to a subsequent action (example - refuse validations if the spanner schema was created using a partial error conv?)

We need to think more on this front. Please create a tech debt item with all the relevant details.

sources/spanner/toddl.go
```diff
@@ -40,23 +40,28 @@ func (tdi ToDdlImpl) ToSpannerType(conv *internal.Conv, spType string, srcType s
...
+	case "JSON", "jsonb":
 		return ddl.Type{Name: ddl.JSON}, nil
-	case "NUMERIC":
+	case "NUMERIC", "numeric":
 		return ddl.Type{Name: ddl.Numeric}, nil
-	case "STRING":
+	case "STRING", "character varying":
+		if len(srcType.Mods) == 0 {
+			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, nil
+		}
```
manitgupta, line 62: Is this relevant for `STRING` datatype of GSQL as well? If not, I would suggest moving `character varying` to its own case.

sources/spanner/infoschema.go
shreyakhajanchi: Created a ticket with id 315784385

sources/spanner/infoschema.go
shreyakhajanchi: created a ticket with id : 315784390

manitgupta (approved):

PR #722, opened 2023-12-15
feat: Add IAM based authentication for CloudSQL MySQL and PostgreSQL databases

manitgupta:
### Feature description

This feature allows Harbourbridge to use the [Cloud SQL Go connector](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector) to authenticate to CloudSQL MySQL and PostgreSQL instances via GCP IAM based authentication.

**This simplifies the connectivity experience and removes an adoption hurdle when using SMT to migrate to Spanner from a source CloudSQL instance. There is no need to whitelist SMT IP on the CloudSQL instance for connection. For a customer with IAM authenticated user setup and public IP enabled, it just works...**

**For customers that have a CloudSQL instance with private IP, SMT would need to run in the same VPC network. This ensures network connectivity is possible to the CloudSQL instance. From that point, IAM based auth will take of the authentication**.

Instead, Harbourbridge will use the IAM credentials of the user/service account it is being run as (via ADC credentials), and will reuse the same credentials to authenticate to the CloudSQL instance.

A mandatory requirement for this to work is to have IAM based authentication enabled on the CloudSQL instance. CloudSQL provides a simple, 1-time setup guide to setup IAM based authentication. Moreover, an existing CloudSQL customer would likely already be using IAM based authentication as part of their daily operations.

---

### CLI experience

This feature works for both POC and minimal downtime migrations. It works for `schema`, `data` and `schema-and-data` commands.  Currently the feature is only exposed via the CLI.  The UI implementation of this feature will be in a follow up PR.

The CLI experience looks as follows - 

#### MySQL

```
./spanner-migration-tool schema-and-data -source=mysql 
--target-profile="project=span-cloud-testing,instance=manit-testing,dbName=testgoconn2" 
--source-profile="user=manitgupta,dbName=db11,instance=manit-mysql-shard1,region=asia-south2,project=span-cloud-testing"
```

#### PostgreSQL

```
./spanner-migration-tool schema -source=pg 
--target-profile="project=span-cloud-testing,instance=manit-testing,dbName=testgoconn7" 
--source-profile="user=XXXX@google.com,dbName=postgres,instance=shreya-postgres,region=us-central1,project=span-cloud-testing"
```

The `--source-profile` flag has been enhanced to accept both direction connection parameters and CloudSQL specific parameters.

The CloudSQL specific parameters are - 

1. `user` - This is the IAM authenticated user in CloudSQL to be used for connection.
2. `dbName` - Database name (mandatory)
3. `instance` - CloudSQL instance (mandatory)
4. `region` - The region of the CloudSQL instance (mandatory)
5. `project` - The project of the CloudSQL instance (optional). If not specified, the `project` configured in `gCloud` will be used.

---

### Testing details

1. Tested for `schema`, `data` and `schema-and-data` for POC for MySQL.
2. Tested for `schema`, `data` and `schema-and-data` for POC for PostgreSQL.

Pending: Testing for minimal downtime migrations.


conversion/conversion.go
```diff
@@ -64,10 +66,13 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	dydb "github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/dynamodbstreams"
+	mysqldriver "github.com/go-sql-driver/mysql"
 	"go.uber.org/zap"
 	adminpb "google.golang.org/genproto/googleapis/spanner/admin/database/v1"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/proto"
+	"github.com/jackc/pgx/v5"
+    "github.com/jackc/pgx/v5/stdlib"
```
bharadwaj-aditya, line 75: formatting differences. Please run go mod tidy

conversion/conversion.go
```diff
@@ -1280,6 +1300,67 @@ func ProcessDump(driver string, conv *internal.Conv, r *internal.Reader) error {
...
+                return nil, fmt.Errorf("cloudsqlconn.NewDialer: %w", err)
+        }
+        var opts []cloudsqlconn.DialOption
+		instanceName := fmt.Sprintf("%s:%s:%s", sourceProfile.ConnCloudSQL.Mysql.Project, sourceProfile.ConnCloudSQL.Mysql.Region, sourceProfile.ConnCloudSQL.Mysql.InstanceName)
+        mysqldriver.RegisterDialContext("cloudsqlconn",
+                func(ctx context.Context, addr string) (net.Conn, error) {
+                        return d.Dial(ctx, instanceName, opts...)
+                })
+
+        dbURI := fmt.Sprintf("%s:empty@cloudsqlconn(localhost:3306)/%s?parseTime=true",
```
bharadwaj-aditya, line 1318: is the port always going to be 3306 ? Can this be fetched from the instance ? 
Is it called out in the documentation ?

bharadwaj-aditya (changes requested):
Minor changes. But the code coverage has dropped. Please add tests for these flows.

conversion/conversion.go
manitgupta: Not sure why it is showing like that here, on my IDE it looks fine. Ran `go mod tidy` once to be sure.

conversion/conversion.go
manitgupta: This port cannot be changed in CloudSQL and is always 3306 for MySQL -
https://stackoverflow.com/questions/32354991/change-tcp-port3306-for-google-cloud-sql-instance

manitgupta:
I looked at `conversion.go` - a class of methods here use `databases/sql` driver to open `sql.open` connections to the source database. This file does not have unit tests written for it. 

Portions of this code need to be rewritten so that APIs from the `databases/sql` lib and other associated library are passed into the methods, so that we can mock them later. 
We may also need to declare methods in the context of a function receiver as well. 

Since this will require a bit more thought, I am skipping creating a new UT file for this class for now..

Probably we should also break this file up as well..

bharadwaj-aditya (approved):

PR #723, opened 2023-12-15
fk error on empty name

aksharauke:
When there are two foreign keys, on deleting them one after other - there was error being issued.


bharadwaj-aditya (approved):
LGTM

PR #724, opened 2023-12-15
support to convert bit to boolean

bharadwaj-aditya:
Added support to convert from bit(1) to boolean.
The default mapping is that bit(1) will migrate to boolean.

- [x] Tests pass


asthamohta:
Do we add a similar issue for bit(1)?
https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/2da4d00ef7dbe37a2618dfd04eefb98bd8cf5618/webv2/web.go#L3093

sources/mysql/data.go
```diff
@@ -137,9 +137,22 @@ func convScalar(conv *internal.Conv, spannerType ddl.Type, srcTypeName string, T
...
+func convBool(conv *internal.Conv, spannerType ddl.Type, srcTypeName string, val string) (bool, error) {
 	b, err := strconv.ParseBool(val)
 	if err != nil {
+
+		if srcTypeName == "bit" {
+			// To handle scenarios where bit is used to store boolean
+			// zero treated as false
+			// one treated as true
+			switch val {
+			case "\x00":
```
Deep1998, line 149: Will this byte string be guaranteed to be 2 digit - either "\x00" or "\x01"? Can source not be "\x1" or "\x001" etc?

Should we instead parse to number if "starts with \x" and compare to 1 or 0?

bharadwaj-aditya: This is how 0 and 1 are represented in bytes in go

Deep1998: I see, bit(1) data is always read as these 2 string we should be good

Deep1998:
LGTM

Deep1998 (approved):

PR #737, opened 2024-01-02
Update the metadata table schemas

Deep1998:
This PR updates the metadata table schema. Changes include:
1) StateData column moved to SMT_JOB and SMT_RESOURCE as JobStateData and ResourceStateData columns.
2) History tables track entire row. New PK is (JobId, Version) / (ResourceId, Version)
3) Removed StateId column since it is not relevant anymore.


webv2/helpers/helpers.go
```diff
@@ -76,31 +76,48 @@ func createDatabase(ctx context.Context, uri string) error {
...
 				ResourceName STRING(100) NOT NULL,
 				ResourceType STRING(100) NOT NULL,
+				ResourceStateData JSON,
 				ResourceData JSON,
-				CreatedAt TIMESTAMP NOT NULL,
-			  ) PRIMARY KEY(ResourceId)`,
-			`CREATE TABLE SMT_STATES (
-				StateId STRING(100) NOT NULL,
-				StateVersion INT64 NOT NULL,
+				UpdatedAt TIMESTAMP NOT NULL OPTIONS (allow_commit_timestamp=true),
```
manitgupta, line 108: This should have both `createdAt` and `updatedAt` right?

streaming/store.go
```diff
@@ -127,14 +127,15 @@ func writeJobDetails(ctx context.Context, migrationJobId string, isShardedMigrat
 		logger.Log.Error(fmt.Sprintf("internal error occurred while persisting metadata for migration job %s: %v\n", migrationJobId, err))
 		return err
 	}
-	jobDetails := SmtJobs{
+	jobDetails := SmtJob{
 		JobId:               migrationJobId,
 		JobName:             migrationJobId,
 		JobType:             constants.MINIMAL_DOWNTIME_MIGRATION,
 		Dialect:             conv.SpDialect,
+		JobStateData:        "{\"state\": \"RUNNING\"}",
```
manitgupta, line 135: Weren't `CREATING` and `CREATED` the only states we were using right now? 
Wouldn't this mean all the jobs would be in `RUNNING` state forever?

Deep1998: For resources, we have created and Created, deleting deleted.
for jobs, creating->running->pausing->paused->resuming->running-> Deleting->deleted.

Deep1998: Since the concept of jobs is not fully introduced to fwd flows, we can keep some temporary state for them. AFAIK its not being used anywhere else its just a placeholder.

webv2/helpers/helpers.go
Deep1998: Based on the discussion, we decided UpdatedAt will act as the "CreatedAt" value at first, but subsequent updates will overwrite it. The history table would would be used if we need the first created at timestamp.

manitgupta: I think I meant that CreatedAt and UpdatedAt were both part of the live table. I understand we can find the earliest timestamp using the history but storing both of these together is typically a pretty standard optimization/schema design.

Deep1998: Done

manitgupta (approved):

PR #738, opened 2024-01-03
Update metadata db schema if table not exists

manitgupta:
If the metadata db exists, it creates new tables if they do not exist, else noop.
If metadata db is not exist, it creates the db and the required new tables.

### DIFF
--- a/webv2/helpers/helpers.go
+++ b/webv2/helpers/helpers.go
@@ -84,7 +84,7 @@ func GetSpannerUri(projectId string, instanceId string) string {
 }
 
 // Creates the schema for the internal metadata database
-func createDatabase(ctx context.Context, uri string, isExist bool) error {
+func createDatabase(ctx context.Context, uri string, dbExists bool) error {
 
 	// Spanner uri will be in this format 'projects/project-id/instances/spanner-instance-id/databases/db-name'
 	matches := regexp.MustCompile("^(.*)/databases/(.*)$").FindStringSubmatch(uri)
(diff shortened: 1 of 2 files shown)


webv2/helpers/helpers.go
```diff
@@ -136,10 +148,9 @@ func CheckOrCreateMetadataDb(projectId string, instanceId string) (isExist bool,
 	}
 	if dbExists {
 		isExist = true
-		return
 	}
 
-	err = createDatabase(ctx, uri)
+	err = createDatabase(ctx, uri, isExist)
```
aksharauke, line 153: Can we pass dbExists only instead for isExist?

manitgupta: Done

webv2/helpers/helpers.go
```diff
@@ -37,12 +37,54 @@ const (
...
+	) PRIMARY KEY(StateId, StateVersion)`,
+}
+
 func GetSpannerUri(projectId string, instanceId string) string {
 	return fmt.Sprintf("projects/%s/instances/%s/databases/%s", projectId, instanceId, constants.METADATA_DB)
 }
 
 // Creates the schema for the internal metadata database
-func createDatabase(ctx context.Context, uri string) error {
+func createDatabase(ctx context.Context, uri string, isExist bool) error {
```
shreyakhajanchi, line 87: nit: better to rename isExist to dbExists

shreyakhajanchi (approved):

webv2/helpers/helpers.go
manitgupta: Done

aksharauke (approved):

PR #739, opened 2024-01-03
Enhance schema validation logic for data only migrations

manitgupta:
Updates the schema validation logic used during data only migrations to care only about things that are relevant to the actual migration -

1. Table names
2. Column names (number and substance)
3. Primary keys (length and substance)

Improved error logging in this layer so that its easier for the customer to understand where the schema validation is failing instead of a generic error.

Flows tested -

Schema only migration followed by 

1. Data only migration with no changes - success.
2. Data only migration with index added directly on spanner - success.
3. Data only migration with data type changed on spanner - failure (expected).

Please comment if any other flow needs to be tested


common/utils/utils.go
```diff
@@ -588,91 +588,68 @@ func ReadSpannerSchema(ctx context.Context, conv *internal.Conv, client *sp.Clie
 }
 
 // CompareSchema compares the spanner schema of two conv objects and returns specific error if they don't match
-func CompareSchema(conv1, conv2 *internal.Conv) error {
-	if conv1.SpDialect != conv2.SpDialect {
+func CompareSchema(sessionFileConv, actualSpannerConv *internal.Conv) error {
+	if sessionFileConv.SpDialect != actualSpannerConv.SpDialect {
 		return fmt.Errorf("spanner dialect don't match")
```
aksharauke, line 593: Will help to print the expected and actual dialects

aksharauke (approved):

common/utils/utils.go
manitgupta: Done

common/utils/utils.go
```diff
@@ -588,91 +588,68 @@ func ReadSpannerSchema(ctx context.Context, conv *internal.Conv, client *sp.Clie
...
+			return fmt.Errorf("table name don't match: session table %v, spanner table %v", sessionTable.Name, spannerTable.Name)
+		}
+
+		//parent table names should match
+		if sessionTableParentName != spannerTableParentName {
+			return fmt.Errorf("parent table name don't match: session table %v, parent session table name: %v, spanner table %v, parent spanner table name: %v", sessionTable.Name, sessionTableParentName, spannerTable.Name, spannerTableParentName)
+		}
 
-		if sessionTable.Name != spannerTable.Name || sessionTableParentName != spannerTableParentName ||
-			len(sessionTable.PrimaryKeys) != len(spannerTable.PrimaryKeys) || len(sessionTable.ColDefs) != len(spannerTable.ColDefs) ||
```
shreyakhajanchi, line 603: Shouldn't we keep the check len(sessionTable.ColDefs) != len(spannerTable.ColDefs)? What if an extra column is added in spanner ?

manitgupta: Right, I must have missed this during refactoring. Thanks adding it back

shreyakhajanchi (approved):
LGTM

PR #740, opened 2024-01-03
feat: allow resource generation using json in the CLI flow

asthamohta:
Test Scenarios Verified:
- Successfully generate resources(connection profiles and gcs bucket) with all information such as id, location, port, etc provided
- Successfully generate Resources(connection profiles and gcs bucket) with information missing such as id and location
- Error thrown when Validation of resource generation fails
- Error thrown and already created resources deleted when validation passes but one of the resource creation fails
- No affects on UI and non sharded migration flow


manitgupta:
Please add unit tests wherever applicable. If changes are made to an existing class for which UTs are not present, create the UT file to cover the portion of the code that has been changed in this PR

internal/convert.go
```diff
@@ -29,26 +29,29 @@ import (
...
+	TimezoneOffset 		string              // Timezone offset for timestamp conversion.
+	SpDialect      		string              // The dialect of the spanner database to which Spanner migration tool is writing.
+	UniquePKey     		map[string][]string // Maps Spanner table name to unique column name being used as primary key (if needed).
+	Audit          		Audit               `json:"-"` // Stores the audit information for the database conversion
+	Rules          		[]Rule              // Stores applied rules during schema conversion
+	IsSharded      		bool                // Flag denoting if the migration is sharded or not
+	ConvLock       		sync.RWMutex        `json:"-"` // ConvLock prevents concurrent map read/write operations. This lock will be used in all the APIs that either read or write elements to the conv object.
+	SpRegion			string				// Region for Spanner Instance
+	ResourceValidation	bool				// Flag denoting if validation for resources to generated is complete
+	UI					bool				// Flag if UI interface was used for migration. ToDo: Remove flag after resource generation is introduced to UI
```
manitgupta, line 54: Are these internal only fields? We should ensure that they are not deserialized.
Please also validate that this does not break backward compatibility of `conv` object.

conversion/conversion.go
```diff
@@ -67,25 +67,63 @@ import (
...
+	Port string
+	// For source connection profile password of MySql instance
+	Password string
+	// For source connection profile user name of MySql instance
+	User string
+	// Region of connection profile to be created
+	Region string
+	// For target connection profile name of gcs bucket to be created
+	BucketName string
+}
```
manitgupta, line 112: I think a similar struct exists elsewhere, can you check if there is an opportunity to reuse?

conversion/resource_generation.go
```diff
@@ -0,0 +1,413 @@
...
+	getConnProfilesRegion(ctx context.Context, projectId string, region string, dsClient *datastream.Client) ([]string, error)
+	connectionProfileExists(ctx context.Context, projectId string, profileName string, profileLocation string, connectionProfiles map[string][]string, dsClient *datastream.Client, s resourceGenerationInterface) (bool, error)
+	getResourcesForCreation(ctx context.Context, projectId string, sourceProfile profiles.SourceProfile, region string, validateOnly bool, dsClient *datastream.Client) ([]*ConnectionProfileReq, []*ConnectionProfileReq, error)
+	prepareMinimalDowntimeResources(createResourceData *ConnectionProfileReq, mutex *sync.Mutex) common.TaskResult[*ConnectionProfileReq]
+	setConnectionProfileFromRequest(details *ConnectionProfileReq, req *datastreampb.CreateConnectionProfileRequest) error
+	GetSpannerRegion(ctx context.Context, projectId string, instanceName string) (string, error)
+	connectionProfileCleanUp(ctx context.Context, profiles []*ConnectionProfileReq) error
+	ValidateResourceGeneration(ctx context.Context, projectId string, instanceId string, sourceProfile profiles.SourceProfile, conv *internal.Conv) error
+}
```
manitgupta, line 57: Should the interface contain private methods?
Even in unit testing we do not typically mock private methods.

conversion/resource_generation.go
```diff
@@ -0,0 +1,413 @@
...
+				Bucket:   details.ConnectionProfile.BucketName,
+				RootPath: "/",
+			},
+		}
+		return nil
+	}
+	return nil
+}
+
+func (r ResourceGenerationStruct) GetSpannerRegion(ctx context.Context, projectId string, instanceName string) (string, error) {
```
manitgupta, line 176: Nit: Add docs for all public methods. 

Nit: Move public methods to the top of the file and the private methods to the bottom.

conversion/resource_generation.go
```diff
@@ -0,0 +1,413 @@
...
+	// Check if connection profiles for the given region are fetched. if not, fetch them
+	profiles, ok := connectionProfiles[profileLocation]
+	var err error = nil
+	if !ok {
+		profiles, err = s.getConnProfilesRegion(ctx, projectId, profileLocation, dsClient)
+		if err != nil {
+			return false, err
+		}
+		connectionProfiles[profileLocation] = profiles
+	}
```
manitgupta, line 225: Looks like some of this code already be implemented when the UI feature of "listing connection profiles" was implemented. We should reuse/refactor code so that both flows use the same logic for common operations such as listing connection profiles from Datastream.

conversion/resource_generation.go
```diff
@@ -0,0 +1,413 @@
...
+			// Check if destination connection profile exists
+			dstProfileExists, err = s.connectionProfileExists(ctx, projectId, profile.DstConnectionProfile.Name, profile.DstConnectionProfile.Location, connectionProfiles, dsClient, &s)
+			if err != nil {
+				return sourceProfilesToCreate, dstProfilesToCreate, err
+			}
+		}
+		if !sourceProfileExists {
+			id := profile.SrcConnectionProfile.Name
+			if id == "" {
+				id = "CN-" + uuid.New().String()
```
manitgupta, line 305: I think we use `hb-` prefix everywhere like `hb-job-*` , `hb-pub-*`. We also don't use the full uuid if I recall correctly. Please use the naming convention used in other parts of the code.

conversion/resource_generation.go
```diff
@@ -0,0 +1,413 @@
...
+			if err := gcsClient.Bucket(profile.ConnectionProfile.BucketName).Delete(ctx); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Returns source and destination connection profiles to be created
+func (r *ResourceGenerationStruct) getResourcesForCreation(ctx context.Context, projectId string, sourceProfile profiles.SourceProfile, region string, validateOnly bool, dsClient *datastream.Client) ([]*ConnectionProfileReq, []*ConnectionProfileReq, error) {
```
manitgupta, line 270: Very long method - please refactor.

Suggested refactoring - source and destination connection profiles in separate methods.

conversion/resource_generation.go
```diff
@@ -0,0 +1,413 @@
...
+
+// 1. For each datashard, check if source and destination connection profile exists or not
+// 2. If source connection profile doesn't exists create it or validate if creation is possible.
+// 3. If validation is false and destination connection profile doesn't exists create a corresponding gcs bucket and then a destination connection profile
+func (r ResourceGenerationStruct) CreateResourcesForShardedMigration(ctx context.Context, projectId string, instanceName string, validateOnly bool, region string, sourceProfile profiles.SourceProfile, dsClient *datastream.Client) error {
+	var sourceProfilesToCreate []*ConnectionProfileReq
+	var dstProfilesToCreate []*ConnectionProfileReq
+
+	// Fetches list with resources which do not exist and need to be created
+	s := ResourceGenerationStruct{}
```
manitgupta, line 359: `s` is not readable. Please rename appropriately, applicable to other places

conversion/resource_generation.go
```diff
@@ -0,0 +1,413 @@
...
+		fastExit = true
+	}
+
+	var errorsList []error = []error{}
+
+	// Create or validate source connection profiles in parallel threads
+	resSourceProfiles, resCreationErr := common.RunParallelTasks(sourceProfilesToCreate, 20, r.prepareMinimalDowntimeResources, fastExit)
+	// If creation failed, perform cleanup of resources
+	if resCreationErr != nil && !validateOnly {
+		err = r.connectionProfileCleanUp(ctx, resourcesForCleanup)
```
manitgupta, line 377: 1. Please check with @VardhanThigle if this needs to be rate limited. We have this cause issues in the past.
2. Why isn't `connectionProfileCleanUp` also parallelised?

conversion/resource_generation.go
```diff
@@ -0,0 +1,413 @@
...
+		_, resCreationErr := common.RunParallelTasks(dstProfilesToCreate, 20, r.prepareMinimalDowntimeResources, fastExit)
+		if resCreationErr != nil {
+			err = r.connectionProfileCleanUp(ctx, resourcesForCleanup)
+			if err != nil {
+				return fmt.Errorf("resource generation failed due to %s, resources created could not be cleaned up, please cleanup manually: %s", resCreationErr.Error(), err.Error())
+			} else {
+				return resCreationErr
+			}
+		}
+	}
```
manitgupta, line 404: This flow is a bit confusing to read tbh - `ValidateResource` calls into `CreateResource` with a boolean flag.

Shouldn't `ValidateResource` call into the underlying methods that `CreateResource` would use to create resources directly, and passing the validation flag to it directly, instead of hopping through `CreateResource`?

conversion/conversion.go
```diff
@@ -67,25 +67,63 @@ import (
...
 	// This number should not be too high so as to not hit the AdminQuota limit.
 	// AdminQuota limits are mentioned here: https://cloud.google.com/spanner/quotas#administrative_limits
 	// If facing a quota limit error, consider reducing this value.
-	MaxWorkers       = 50
-	once             sync.Once
-	datastreamClient *datastream.Client
+	MaxWorkers          = 50
+	once                sync.Once
+	datastreamClient    *datastream.Client
+	resourcesForCleanup []*ConnectionProfileReq
```
manitgupta, line 86: Why do we need this global variable?

conversion/resource_generation_test.go
```diff
@@ -0,0 +1,103 @@
+// // Copyright 2023 Google LLC
```
manitgupta, line 1: Nit: Change year

conversion/resource_generation.go
manitgupta: This would enhance the readability of `CreateResource` since I shouldn't have to think about `validation` or not when reading that code.

internal/convert.go
asthamohta: I have checked. Nothing breaks

conversion/conversion.go
asthamohta: We do, but it is in UI and has some missing fields. That would get removed when the two sharded and non-sharded flows will merge

conversion/resource_generation.go
asthamohta: Breaking this up in the new PR

conversion/resource_generation.go
asthamohta: It is, it'll get removed and this function will be called in the UI PR

conversion/resource_generation.go
asthamohta: changed in new PR

conversion/resource_generation.go
asthamohta: Perhaps a better naming convention should be used to make it less confusing. Let me change this

conversion/conversion.go
asthamohta: Because parallel tasks are used which cannot take any extra params and this seemed like the best way to add resources that were actually created.

PR #741, opened 2024-01-04
Adding basic rate limiting for DataStream and DataFlow and idempotenet retries for dataStream

VardhanThigle:
## Note:
This does not cover adding fine grained rate limiting (like leaky bucket) or adding retries for all API calls that SMT makes (like Pub/Sub). It fixes the issue encountered in migrations that test scale of the order of 100 streams. If datastream and dataflow calls are rate limited at a coarse level, others like pub/sub get naturally ratelimited as they are part of the same code flow.
Rate limiting is kept to meet the default API quota of DataStream and DataFlow.

A more fuller version of retries and rate limiting will have to be taken as a separate task

- [X] Tests pass
   - Large scale sharded migration with 97 streams runs w/o quota errors.

### DIFF
--- a/streaming/streaming.go
+++ b/streaming/streaming.go
@@ -38,6 +38,10 @@ import (
 	"github.com/GoogleCloudPlatform/spanner-migration-tool/internal"
 	"github.com/GoogleCloudPlatform/spanner-migration-tool/logger"
 	"github.com/GoogleCloudPlatform/spanner-migration-tool/profiles"
+	"github.com/google/uuid"
+	"github.com/googleapis/gax-go/v2"
+	"go.uber.org/ratelimit"
+	"google.golang.org/grpc/codes"
 )
 
 var (
@@ -59,6 +63,32 @@ var (
 	MIN_WORKER_LIMIT int32 = 1
 	// Default gcs path of the Dataflow template.
 	DEFAULT_TEMPLATE_PATH string = "gs://dataflow-templates-southamerica-west1/2023-09-12-00_RC00/flex/Cloud_Datastream_to_Spanner"
+
+	DEFAULT_DATASTREAM_CLIENT_BACKOFF_BASE_DELAY time.Duration = 1.0 * time.Second
+	DEFAULT_DATASTREAM_CLIENT_BACKOFF_MAX_DELAY  time.Duration = 900 * time.Second
+	DEFAULT_DATASTREAM_CLIENT_BACKOFF_MULTIPLIER float64       = 1.6
+	DEFAULT_DATASTREAM_RETRY_CODES               []codes.Code  = []codes.Code{
+		codes.DeadlineExceeded,
+		codes.Unavailable,
+		codes.ResourceExhausted,
+		codes.Unknown,
+	}
+
+	// DataStream by default provides 20 API calls per second.
+	// Each launch operation calls datastream twice, ignoring `op.wait`, hence keeping it at 10 per second.
+	DEFAULT_DATASTREAM_LAUNCH_RATE_PER_SEC int = 10
+	// TODO(vardhanvthigle): Caliberate this.
+	// Keeping it less for now since the dataFlow launch operation makes outbound calls to various clients.
+	// Keeping it at 1 per second will not impact the actual time it takes to provision resources for a large
+	// scale migration, as that depends a lot on actual time a batch of 20 (current parallelization in code) tkaes
+	// to complete it's provisioning which includes waiting for various operations.
+	DEFAULT_DATAFLOW_LAUNCH_RATE_PER_SEC int = 1
+	// Rate Limiters
+	// A coarse delay based rate limiting for launching datastream.
+	DATA_STREAM_RL ratelimit.Limiter = ratelimit.New(DEFAULT_DATASTREAM_LAUNCH_RATE_PER_SEC)
+
+	// A coarse delay based rate limiting for launching DataFlow.
+	DATA_FLOW_RL ratelimit.Limiter = ratelimit.New(DEFAULT_DATAFLOW_LAUNCH_RATE_PER_SEC)
 )
 
 type SrcConnCfg struct {
@@ -113,6 +143,15 @@ type StreamingCfg struct {
 	TmpDir        string                   `json:"tmpDir"`
 	PubsubCfg     internal.PubsubResources `json:"pubsubCfg"`
 	DataShardId   string                   `json:"dataShardId"`
+}
+
+// Returns the retry error codes and backoff policy to the GCP client retry logic.
+func dataStreamGaxRetrier() gax.Retryer {
+	return gax.OnCodes(DEFAULT_DATASTREAM_RETRY_CODES, gax.Backoff{
+		Initial:    DEFAULT_DATASTREAM_CLIENT_BACKOFF_BASE_DELAY,
+		Max:        DEFAULT_DATASTREAM_CLIENT_BACKOFF_MAX_DELAY,
+		Multiplier: DEFAULT_DATASTREAM_CLIENT_BACKOFF_MULTIPLIER,
+	})
 }
 
 // VerifyAndUpdateCfg checks the fields and errors out if certain fields are empty.
@@ -472,6 +511,8 @@ func LaunchStream(ctx context.Context, sourceProfile profiles.SourceProfile, dbL
 		return fmt.Errorf("datastream client can not be created: %v", err)
 	}
 	defer dsClient.Close()
+	// Rate limit this function to match DataStream API Quota.
+	DATA_STREAM_RL.Take()
 	fmt.Println("Created client...")
 	prefix := datastreamCfg.DestinationConnectionConfig.Prefix
 	prefix = utils.ConcatDirectoryPath(prefix, "data")
@@ -503,11 +544,13 @@ func LaunchStream(ctx context.Context, sourceProfile profiles.SourceProfile, dbL
 		Parent:   fmt.Sprintf("%s/locations/%s", projectNumberResource, datastreamCfg.StreamLocation),
 		StreamId: datastreamCfg.StreamId,
 		Stream:   streamInfo,
+		// Setting a RequestId makes idempotent retries possible.
+		RequestId: uuid.New().String(),
 	}
 
 	fmt.Println("Created stream request..")
 
-	dsOp, err := dsClient.CreateStream(ctx, createStreamRequest)
+	dsOp, err := dsClient.CreateStream(ctx, createStreamRequest, gax.WithRetry(dataStreamGaxRetrier))
 	if err != nil {
 		fmt.Printf("cannot create stream: createStreamRequest: %+v\n", createStreamRequest)
 		return fmt.Errorf("cannot create stream: %v ", err)
@@ -525,8 +568,10 @@ func LaunchStream(ctx context.Context, sourceProfile profiles.SourceProfile, dbL
 	updateStreamRequest := &datastreampb.UpdateStreamRequest{
 		UpdateMask: &fieldmaskpb.FieldMask{Paths: []string{"state"}},
 		Stream:     streamInfo,
+		// Setting a RequestId makes idempotent retries possible.
+		RequestId: uuid.New().String(),
 	}
-	upOp, err := dsClient.UpdateStream(ctx, updateStreamRequest)
+	upOp, err := dsClient.UpdateStream(ctx, updateStreamRequest, gax.WithRetry(dataStreamGaxRetrier))
 	if err != nil {
 		return fmt.Errorf("could not create update request: %v", err)
 	}
@@ -543,6 +588,10 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
 	project, instance, dbName, _ := targetProfile.GetResourceIds(ctx, time.Now(), "", nil)
 	dataflowCfg := streamingCfg.DataflowCfg
 	datastreamCfg := streamingCfg.DatastreamCfg
+
+	// Rate limit this function to match DataFlow createJob Quota.
+	DATA_FLOW_RL.Take()
+
 	fmt.Println("Launching dataflow job ", dataflowCfg.JobName, " in ", project, "-", dataflowCfg.Location)
 
 	c, err := dataflow.NewFlexTemplatesClient(ctx)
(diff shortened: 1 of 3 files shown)


manitgupta:
Our testing effort can proceed ahead without merging this particular change in a rush (my understanding is that we are running that test from a different branch, please correct me if I am wrong). 

If my understanding is correct, I don't see a lot of benefit for merging this in its current state. I would recommend we complete the rate limiting/retry aspects you want to cover as part of this change itself. At the minimum, I think we should - 

- Cover rate limiting and retries of all GAX based clients. 
- Investigate if the current level of parallelization for shard execution needs to be tweaked. We shouldn't be using a larger pool of workers and then artificially rate limiting the clients - it sounds like an anti-pattern.

FWIW I think a simple TPS based rate limiting (what you have right now) is sufficient for the orchestration use-cases we have seen till now, we may not need more fine grained limiting in the near future (would be happy to discuss if you think otherwise).

streaming/streaming.go
```diff
@@ -59,6 +63,23 @@ var (
...
+
+	DEFAULT_DATASTREAM_CLIENT_BACKOFF_BASE_DELAY time.Duration = 1.0 * time.Second
+	DEFAULT_DATASTREAM_CLIENT_BACKOFF_MAX_DELAY  time.Duration = 900 * time.Second
+	DEFAULT_DATASTREAM_CLIENT_BACKOFF_MULTIPLIER float64       = 1.6
+	DEFAULT_DATASTREAM_RETRY_CODES               []codes.Code  = []codes.Code{
+		codes.DeadlineExceeded,
+		codes.Unavailable,
+		codes.ResourceExhausted,
+		codes.Unknown,
+	}
```
manitgupta, line 75: I wonder if we should just let the clients retry for all errors 🤔

HB is not a continuous load generating service where a smaller number of additional retries (even for non-retryable errors) would cause brownouts or outages in the dependent services.

On the other hand, we win in resiliency if we get an undeclared retryable exception from the downstream (what happened in the most recent run with Datastream).

WDYT @VardhanThigle and @bharadwaj-aditya?

streaming/streaming.go
```diff
@@ -115,6 +136,15 @@ type StreamingCfg struct {
 	DataShardId   string                   `json:"dataShardId"`
 }
 
+// Returns the retry error codes and backoff policy to the GCP client retry logic.
+func dataStreamGaxRetrier() gax.Retryer {
+	return gax.OnCodes(DEFAULT_DATASTREAM_RETRY_CODES, gax.Backoff{
+		Initial:    DEFAULT_DATASTREAM_CLIENT_BACKOFF_BASE_DELAY,
+		Max:        DEFAULT_DATASTREAM_CLIENT_BACKOFF_MAX_DELAY,
+		Multiplier: DEFAULT_DATASTREAM_CLIENT_BACKOFF_MULTIPLIER,
+	})
```
manitgupta, line 145: What are the total number of retries that will be performed?

manitgupta (commented):
Please add unit tests to cover the delta lines pointed out by CodeCov

streaming/streaming.go
VardhanThigle: Total Number of retries will be around 15 
$$log2(900) / log2(1.6)$$

VardhanThigle:
> Our testing effort can proceed ahead without merging this particular change in a rush (my understanding is that we are running that test from a different branch, please correct me if I am wrong).
>  The Recent test was ran on this branch which was rebased on top of my earlier projectNumber conversion, before raising the PR I just synced the branch to master (as it had the conversion checked in), so this is PR has been qualified on a large scale test once already.


> If my understanding is correct, I don't see a lot of benefit for merging this in its current state. I would recommend we complete the rate limiting/retry aspects you want to cover as part of this change itself. At the minimum, I think we should -
>  I think the current state is letting us reach a milestone of close to 100 streams and jobs. Which might be awaited by some users.

> * Cover rate limiting and retries of all GAX based clients.
    That's right. It needs to be done, ideally everytime we interface with an API method. 
    When we cover retries for all the API methods, I don't think the  code for retries of datastream calls will change. Correct me if that's not the case.
    This PR is incrementally taking us there and fixing issues that we have seen. Doing the entire change, as we mentioned in the beginning is desirable but we need to prioritize too. I can work on the entire change in parallel to the large scale migration fixes I am doing.

> * Investigate if the current level of parallelization for shard execution needs to be tweaked. We shouldn't be using a larger pool of workers and then artificially rate limiting the clients - it sounds like an anti-pattern.
>  Rate limiting and scalability can be endlessly improved :), for example, every call  (including operation.wait) can pass through a leaky bucket. The question comes to how much scale we need and what kinod of complexity we are willing to introduce. Currently we want to reach the milestone of spawning resources to support O(100 physical shards), for that the coarse level rate limiting should work w/o incurring un-necessary complexity.  For the order of 100s the current thread pool is good and keeps things simple and we are not getting starved on threads for any of the HB runs I have seen till now.

If we want to scale to say 10 times the number of shards, there are other quotas on total number of jobs or streams which either we need to increase or may be we might need to tweak our architecture a bit. 

Trying to weigh the tradeoff of complexity vs requirements here.

> FWIW I think a simple TPS based rate limiting (what you have right now) is sufficient for the orchestration use-cases we have seen till now, we may not need more fine grained limiting in the near future (would be happy to discuss if you think otherwise).
That's right. That's why I have gone with that for now.  Disussed this more in the previous point.

VardhanThigle:
> Please add unit tests to cover the delta lines pointed out by CodeCov

We don't yet have any UT for streaming.go.
This would involve mocking almost all the services and the scope will be more larger than this small change here.

If we just want to test the retrier, its a function that's just returning form that the gax likes.

manitgupta:
> The Recent test was ran on this branch which was rebased on top of my earlier projectNumber conversion, before raising the PR I just synced the branch to master (as it had the conversion checked in), so this is PR has been patiently qualified on a large scale test once already.

I don't mean that the change isn't manually tested - I mean that adding retries in parts to a section of clients at a time is an approach we shouldn't take in general. Unless it's an emergency fix, we should be addressing a specific problem holistically. With this check-in, its easy to form the perception that "we have added rate limiting and retries", when in-fact, we have only done it partially.

> I think the current state is letting us reach a milestone of close to 100 streams and jobs. Which might be awaited by some users.

I don't disagree, I just don't see the urgency of breaking up "configure rate limiting and retries" as a task that should be broken up into different stages.

> That's right. It needs to be done, ideally everytime we interface with an API method.
When we cover retries for all the API methods, I don't think the code for retries of datastream calls will change. Correct me if that's not the case.
This PR is incrementally taking us there and fixing issues that we have seen. Doing the entire change, as we mentioned in the beginning is desirable but we need to prioritize too. I can work on the entire change in parallel to the large scale migration fixes I am doing.

Going back to the same point earlier, I don't see the need to do this incrementally. I am not saying that the code for retries for datastream or dataflow will change - but that's kind of the point, we have around 5-6 clients what we should add the retries in all of them as part of a feature. I expect that code change would look identical to the one done now. Pleasew correct me if I am wrong. Breaking up the change in multiple PRs due to large number of changes is fine, but doing things incrementally without having a line of sight into when the other things would come is not. Can we discuss the total effort required to do this change across all clients? 

> Rate limiting and scalability can be endlessly improved :), for example, every call (including operation.wait) can pass through a leaky bucket. The question comes to how much scale we need and what kinod of complexity we are willing to introduce. Currently we want to reach the milestone of spawning resources to support O(100 physical shards), for that the coarse level rate limiting should work w/o incurring un-necessary complexity. For the order of 100s the current thread pool is good and keeps things simple and we are not getting starved on threads for any of the HB runs I have seen till now.
If we want to scale to say 10 times the number of shards, there are other quotas on total number of jobs or streams which either we need to increase or may be we might need to tweak our architecture a bit.
Trying to weigh the tradeoff of complexity vs requirements here.

I think we should try to keep things as simple as possible. I don't see the benefit of discussing leaky bucket at all when simple rate limiting is what we need. We have a simple problem in front of us - slow down the orchestration process. This can be done in one of two ways - reducing parallelization and implementing artificial throttling. We seem to have only done the latter. 
Consider this - if I reduce the thread pool to 2, only two shards will be orchestrated at a time - and we probably won't even need any artificial rate limiting. This is the line of thought I am talking about.
We don't seem to have attacking the root of the problem.

> We don't yet have any UT for streaming.go.
This would involve mocking almost all the services and the scope will be more larger than this small change here.
If we just want to test the retrier, its a function that's just returning form that the gax likes.

We do have a `streaming_test.go` file - though I agree that writing a unit test for the code path you have modified will be larger effort than the change made. We do need to start sometime though to move the needle on our testing - I am starting a thread on this.

VardhanThigle:
> > The Recent test was ran on this branch which was rebased on top of my earlier projectNumber conversion, before raising the PR I just synced the branch to master (as it had the conversion checked in), so this is PR has been patiently qualified on a large scale test once already.
> 
> I don't mean that the change isn't manually tested - I mean that adding retries in parts to a section of clients at a time is an approach we shouldn't take in general. Unless it's an emergency fix, we should be addressing a specific problem holistically. With this check-in, its easy to form the perception that "we have added rate limiting and retries", when in-fact, we have only done it partially.

I think the PR is titled correctly to mention that we have added idempotent retries for `DataStream`. Is there anything I could have done to title it right?

We can add retries for other clients too, in fact we should, and also we should evaluate it upfront for every next API call integration (some APIs are auto retried by GCP clients),  but we need to throughly test such change in a large scale test environment, since a wrong retry strategy can cause a heavy bill to the customer if there is any problem in the idempotency.

> > I think the current state is letting us reach a milestone of close to 100 streams and jobs. Which might be awaited by some users.
> 
> I don't disagree, I just don't see the urgency of breaking up "configure rate limiting and retries" as a task that should be broken up into different stages.

It's the bandwidth and time constraints for large scale testing. As we have discussed offline I will have a fast follow on with the UT and the retries for other clients which will be tested on our scale testing setup too.

> > That's right. It needs to be done, ideally everytime we interface with an API method.
> > When we cover retries for all the API methods, I don't think the code for retries of datastream calls will change. Correct me if that's not the case.
> > This PR is incrementally taking us there and fixing issues that we have seen. Doing the entire change, as we mentioned in the beginning is desirable but we need to prioritize too. I can work on the entire change in parallel to the large scale migration fixes I am doing.
> 
> Going back to the same point earlier, I don't see the need to do this incrementally. I am not saying that the code for retries for datastream or dataflow will change - but that's kind of the point, we have around 5-6 clients what we should add the retries in all of them as part of a feature. I expect that code change would look identical to the one done now. Pleasew correct me if I am wrong. Breaking up the change in multiple PRs due to large number of changes is fine, but doing things incrementally without having a line of sight into when the other things would come is not. Can we discuss the total effort required to do this change across all clients?

`we have around 5-6 clients what we should add the retries in all of them as part of a feature.` 

:) I completely agree, either retries should be part of the API calls from the beginning or we definitely need to add them for the existing code base and as we discussed I will work on a follow on for the same.

It does not have to be one PR per client, just that breaking it into one known issue that we need to fix and deliver to prod fast enough followed by the other changes that we need to do too.

> 
> > Rate limiting and scalability can be endlessly improved :), for example, every call (including operation.wait) can pass through a leaky bucket. The question comes to how much scale we need and what kinod of complexity we are willing to introduce. Currently we want to reach the milestone of spawning resources to support O(100 physical shards), for that the coarse level rate limiting should work w/o incurring un-necessary complexity. For the order of 100s the current thread pool is good and keeps things simple and we are not getting starved on threads for any of the HB runs I have seen till now.
> > If we want to scale to say 10 times the number of shards, there are other quotas on total number of jobs or streams which either we need to increase or may be we might need to tweak our architecture a bit.
> > Trying to weigh the tradeoff of complexity vs requirements here.
> 
> I think we should try to keep things as simple as possible. I don't see the benefit of discussing leaky bucket at all when simple rate limiting is what we need. We have a simple problem in front of us - slow down the orchestration process. This can be done in one of two ways - reducing parallelization and implementing artificial throttling. We seem to have only done the latter. Consider this - if I reduce the thread pool to 2, only two shards will be orchestrated at a time - and we probably won't even need any artificial rate limiting. This is the line of thought I am talking about. We don't seem to have attacking the root of the problem.

Agreed. That's what we are doing here. I think we both have mutually agreed on this part in multiple comments :)

> > We don't yet have any UT for streaming.go.
> > This would involve mocking almost all the services and the scope will be more larger than this small change here.
> > If we just want to test the retrier, its a function that's just returning form that the gax likes.
> 
> We do have a `streaming_test.go` file - though I agree that writing a unit test for the code path you have modified will be larger effort than the change made. We do need to start sometime though to move the needle on our testing - I am starting a thread on this.
+1

streaming/streaming.go
```diff
@@ -59,6 +63,23 @@ var (
...
+	DEFAULT_DATASTREAM_RETRY_CODES               []codes.Code  = []codes.Code{
+		codes.DeadlineExceeded,
+		codes.Unavailable,
+		codes.ResourceExhausted,
+		codes.Unknown,
+	}
+	// Rate Limiters
+	// A coarse delay based rate limiting for launching datastream.
+	// Each launch operation calls datastream twice, hence keeping it at 10 per second.
+	DATA_STREAM_RL ratelimit.Limiter = ratelimit.New(10)
```
bharadwaj-aditya, line 79: The reasoning does not clarify the choice of 10 here. Can clarify on the comment ?

streaming/streaming.go
```diff
@@ -59,6 +63,23 @@ var (
...
+
+	// A coarse delay based rate limiting for launching DataFlow.
+	DATA_FLOW_RL ratelimit.Limiter = ratelimit.New(1)
```
bharadwaj-aditya, line 82: Can you specify the function of this ? What does the value of 1 mean here ?

bharadwaj-aditya (approved):
LGTM - Please improve the docs describing some of these settings though.

streaming/streaming.go
VardhanThigle: Thanks.
Made the static rate-limiter definitions more expressive by layering the integer literals differently. 
Added more comments.

VardhanThigle:
> LGTM - Please improve the docs describing some of these settings though.

`Please improve the docs describing some of these settings though.`
As of now they are not user tunables. Do we want to make them tunable (that would be a fair ask as differnt users could have increased quotas).
If yes, I will create a tracking item for the same.

streaming/streaming.go
VardhanThigle: Added more comments. Thanks.

PR #743, opened 2024-01-08
[feat] API: Reverse Replication Create workflow

Deep1998:
[Design Doc](https://docs.google.com/document/d/174vIqsins85CN8CA6MWl_m5okbPnzQEU5lKfzSdBpxY/edit?pli=1&tab=t.0#heading=h.8bda1hvfoz2l)
The reverse replication create workflow does the following:
1) Create a gcs bucket if needed
2) Create a change stream if required
3) Create a metadatadb for reverse replication if required.
4) Launch Reader dataflow job
5) Launch writer dataflow job

The entry point of the API is via `create.go`, where each of these steps is denoted via an `Activity`.  Each of these activities implement the interface in `IActivity.go`.

**Other major changes include:**

1. Introduction of `accessors` (like dataflow accessor, spanner accessor, storage accessor) and singleton clients. All functions that access a client should from now on go through here.
2. Introduction of `DAO` which is used to interact with the SMT metadata tables.
3. Zap logger for the reverse replication workflow.


manitgupta:
**TL;DR - Please break up this PR.** 

Please break up this PR into separate logical stacked PRs.  I will suggest breaking this up into atleast 4+ PRs, something like this - 

1. One Accessor implementation
2. Rest of the accessor implementation
3. Activity interface and one implementation
4. Rest of the activities
5. Workflow
6. Changes to web layer (I see some changes made in `web.go`)

This will also help you write code iteratively and test as you go, as you write tests for each PR. 

A good rule of thumb for PR size is less than 1000 lines or 10 files - whichever comes first.
https://thestartupfactory.tech/pull-request-best-practices

Stacked PRs blogs -
1. https://blog.logrocket.com/using-stacked-pull-requests-in-github/
2. https://timothya.com/blog/git-stack/

PR #745, opened 2024-01-09
[feat] RR Create API 1: Add dataflow accessor

Deep1998:
Adds a dataflow accessor object for interacting with the client. All usage of dataflow client should now go via the accessor.

All util files like common/utils/utils.go and common/utils/dataflow_utils.go contain methods that do not use a dataflow client. These files should not import any SMT package.


accessors/dataflow/dataflow_accessor.go
```diff
@@ -0,0 +1,94 @@
...
+	ServiceAccountEmail   string            `json:"serviceAccountEmail"`
+	MachineType           string            `json:"machineType"`
+	AdditionalUserLabels  map[string]string `json:"additionalUserLabels"`
+	KmsKeyName            string            `json:"kmsKeyName"`
+	GcsTemplatePath       string            `json:"gcsTemplatePath"`
+	AdditionalExperiments []string          `json:"additionalExperiments"`
+	EnableStreamingEngine bool              `json:"enableStreamingEngine"`
+}
+
+func GetDataflowLaunchRequest(parameters map[string]string, cfg DataflowTuningConfig) (*dataflowpb.LaunchFlexTemplateRequest, error) {
```
aksharauke, line 43: What is the use case for a public method that just generates a request? The consumer of the client would be interested in either launching,updating,fetching attributes of or stopping the dataflow job. What would the consumer do with a request object?

accessors/dataflow/dataflow_accessor.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+	}
+	request := &dataflowpb.LaunchFlexTemplateRequest{
+		ProjectId: cfg.ProjectId,
+		LaunchParameter: &dataflowpb.LaunchFlexTemplateParameter{
+			JobName:    cfg.JobName,
+			Template:   &dataflowpb.LaunchFlexTemplateParameter_ContainerSpecGcsPath{ContainerSpecGcsPath: cfg.GcsTemplatePath},
+			Parameters: parameters,
+			Environment: &dataflowpb.FlexTemplateRuntimeEnvironment{
+				MaxWorkers:            cfg.MaxWorkers,
```
aksharauke, line 64: What happens when these are not supplied by the user - does this code always assume non-empty values?

accessors/dataflow/dataflow_accessor.go
manitgupta: +1

accessors/dataflow/dataflow_accessor.go
Deep1998: For unspecified values, dataflow automatically sets defaults since these are optional. The service will be responsible for passing the right values, today reverse tomorrow forward.

accessors/dataflow/dataflow_accessor.go
Deep1998: Makes sense, will move the launchRequest function to the caller, only functions calling a client should reside here.

accessors/dataflow/dataflow_accessor.go
```diff
@@ -0,0 +1,109 @@
...
+	ServiceAccountEmail   string            `json:"serviceAccountEmail"`
+	MachineType           string            `json:"machineType"`
+	AdditionalUserLabels  map[string]string `json:"additionalUserLabels"`
+	KmsKeyName            string            `json:"kmsKeyName"`
+	GcsTemplatePath       string            `json:"gcsTemplatePath"`
+	AdditionalExperiments []string          `json:"additionalExperiments"`
+	EnableStreamingEngine bool
+}
+
+func GetDataflowLaunchRequest(parameters map[string]string, cfg DataflowTuningConfig) (*dataflowpb.LaunchFlexTemplateRequest, error) {
```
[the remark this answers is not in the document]
Deep1998: Agreed will move to caller, only methods doing interactions with client directly should happen here.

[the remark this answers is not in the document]
manitgupta: Maybe to a `dataflow_helper` or similar instead? Two callers wanting to call the same function will probably want to reuse what you have written here..

accessors/dataflow/dataflow_accessor_test.go
Deep1998, line 1: @manitgupta should this file be removed then since tests are not expected here.

streaming/streaming.go
```diff
@@ -722,7 +723,8 @@ func LaunchDataflowJob(ctx context.Context, targetProfile profiles.TargetProfile
 		fmt.Printf("flexTemplateRequest: %+v\n", req)
 		return internal.DataflowOutput{}, fmt.Errorf("unable to launch template: %v", err)
 	}
-	gcloudDfCmd := utils.GetGcloudDataflowCommand(req)
+	// Refactor to use accessor return value.
```
Deep1998, line 726: @manitgupta when this is refactored to use accessors, the new LaunchFlexTemplateClient() will return jobid, gcloud command and error.

accessors/dataflow/dataflow_accessor.go
```diff
@@ -0,0 +1,182 @@
...
+
+type DataflowAccessor interface {
+	// This function takes the template parameters (@parameters) and runtime environment config (@cfg) as input, and returns
+	// the generated jobId, equivalentGcloudCommand and error if any.
+	LaunchFlexTemplate(ctx context.Context, parameters map[string]string, cfg DataflowTuningConfig) (string, string, error)
+}
+
+type DataflowAccessorImpl struct{}
+
+func (dfA *DataflowAccessorImpl) LaunchFlexTemplate(ctx context.Context, parameters map[string]string, cfg DataflowTuningConfig) (string, string, error) {
```
manitgupta, line 36: Can you please add coverage for this method?

https://screenshot.googleplex.com/6mXNkg3huwrxgPd

Deep1998: Accessor methods like these calling third party methods cannot be mocked as we discussed. These would require some form integration testing using emulator for spanner accessor, but for dataflow accessor there is no good way. (without pushing the dataflowClient out and take that as an interface input, which goes against wanting to keep client management within accessors.)

manitgupta: Thanks, I understand this better now. Two principles seem to be at odds with each other - wanting to keep the client management within accessors, and wanting to mock out clients to test the accessors. 

This is generally solved by injection. However, since we are not using injection, if I have to chose between the two, I would rather pass the client object by requesting it from `dataflow_client` and passing it to `dataflow_accessor` than leave this piece of code untested. 

I know there has been a bit of churn on this - we're figuring this out as we go - but I do think testing supersedes convention here.

manitgupta (approved):
74ZKM6hJWgamMxF

🥳🥳 @bharadwaj-aditya @rkmoorthy-github

Awesome work @Deep1998.

PR #756, opened 2024-01-30
[feat] RR Create API 2: Add storage and spanner accessor 

Deep1998:
This PR adds accessors, clients and mocks for some of the resource classes used by Reverse Replication. 
The accessor classes contains methods that internally use these resource clients (ex: adminClient.CreateDatabase()). Moving forward, all resource client calls should be made via accessors, no resource client method should be called directly.
- The `accessors/client/<resource>/` contains interface and mock definitions for the corresponding <resource> client.
- The `accessors/<resource>/` contains interface and mock definitions for the corresponding <resource> accessors.

For unit testing purposes, the accessor implementors are created just before the call wherever refactoring is done [Example](https://github.com/GoogleCloudPlatform/spanner-migration-tool/pull/756/files#diff-c56312e9279245ad1b77841b732d9ccdba0d4585ea7cca56433c48158b28a5f1R865-R874). 
When the caller methods are unit tested, the accessor implementors constructor should be moved up and taken as an input. In the linked example, when unit testing `StartDataflow`, `storageaccessor.StorageAccessor` will become a function parameter and `StorageAccessorImpl{}` will be passed in its place wherever `StartDataflow()` is called.

### DIFF
--- a/accessors/spanner/spanner_accessor.go
+++ b/accessors/spanner/spanner_accessor.go
@@ -29,16 +29,27 @@ import (
 	"google.golang.org/api/iterator"
 )
 
+// The SpannerAccessor provides methods that internally use a spanner client (can be adminClient/databaseclient/instanceclient etc).
+// Methods should only contain generic logic here that can be used by multiple workflows.
 type SpannerAccessor interface {
+	// Fetch the dialect of the spanner database.
 	GetDatabaseDialect(ctx context.Context, adminClient spanneradmin.AdminClient, dbURI string) (string, error)
+	// CheckExistingDb checks whether the database with dbURI exists or not.
+	// If API call doesn't respond then user is informed after every 5 minutes on command line.
 	CheckExistingDb(ctx context.Context, adminClient spanneradmin.AdminClient, dbURI string) (bool, error)
+	// Create a database with no schema.
 	CreateEmptyDatabase(ctx context.Context, adminClient spanneradmin.AdminClient, dbURI string) error
+	// Fetch the leader of the Spanner instance.
 	GetSpannerLeaderLocation(ctx context.Context, instanceClient spinstanceadmin.InstanceAdminClient, instanceURI string) (string, error)
+	// Check if a change stream already exists.
 	CheckIfChangeStreamExists(ctx context.Context, changeStreamName, dbURI string) (bool, error)
+	// Validate that change stream option 'VALUE_CAPTURE_TYPE' is 'NEW_ROW'.
 	ValidateChangeStreamOptions(ctx context.Context, changeStreamName, dbURI string) error
+	// Create a change stream with default options.
 	CreateChangeStream(ctx context.Context, adminClient spanneradmin.AdminClient, changeStreamName, dbURI string) error
 }
 
+// This implements the SpannerAccessor interface. This is the primary implementation that should be used in all places other than tests.
 type SpannerAccessorImpl struct{}
 
 func (sp *SpannerAccessorImpl) GetDatabaseDialect(ctx context.Context, adminClient spanneradmin.AdminClient, dbURI string) (string, error) {
@@ -49,8 +60,6 @@ func (sp *SpannerAccessorImpl) GetDatabaseDialect(ctx context.Context, adminClie
 	return strings.ToLower(result.DatabaseDialect.String()), nil
 }
 
-// CheckExistingDb checks whether the database with dbURI exists or not.
-// If API call doesn't respond then user is informed after every 5 minutes on command line.
 func (sp *SpannerAccessorImpl) CheckExistingDb(ctx context.Context, adminClient spanneradmin.AdminClient, dbURI string) (bool, error) {
 	gotResponse := make(chan bool)
 	var err error
@@ -108,6 +117,7 @@ func (sp *SpannerAccessorImpl) GetSpannerLeaderLocation(ctx context.Context, ins
 	return "", fmt.Errorf("no leader found for spanner instance %s while trying fetch location", instanceURI)
 }
 
+// Consider using a CreateChangestream operation and check for alreadyExists error. That uses adminClient which can be unit tested.
 func (sp *SpannerAccessorImpl) CheckIfChangeStreamExists(ctx context.Context, changeStreamName, dbURI string) (bool, error) {
 	spClient, err := spannerclient.GetOrCreateClient(ctx, dbURI)
 	if err != nil {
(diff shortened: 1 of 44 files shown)


common/utils/storage_utils.go
```diff
@@ -0,0 +1,43 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/*
+Package utils contains common helper functions used across multiple other packages.
+Utils should not import any Spanner migration tool packages.
+*/
+package utils
```
manitgupta, line 19: There is a `dataflow_utils` inside the `accessor` and this one is in `common`. Consider restructuring to have a single, top-level `utils` package.

manitgupta (commented):
Overall looks good. Please add UTs for the accessors as done with the earlier PR for Dataflow.

common/utils/storage_utils.go
Deep1998: The common/utils package contain methods which do not import any other SMT dependency, only third party, which is how util packages should be. Hence, storage_utils does stuff like parseGCSURI etc. which do not import any other SMT dependency.

The dataflow_utils is a layer that contains methods which call accessors, but are generic enough to be re-used by multiple flows (for ex: unmarshalTuningConfig() will reads local or gcs file, and creates the tuning struct, which will be used by both fwd and reverse so should be in some common package, but not utils since it imports storageaccessor)

The dataflow utils should be renamed/moved to another top level package, a layer of common util-like functions which use accessors, any suggestions?

accessors/storage/storage_accessor.go
```diff
@@ -0,0 +1,211 @@
...
+)
+
+type StorageAccessor interface {
+	CreateGCSBucket(ctx context.Context, bucketName, projectID, location string) error
+	CreateGCSBucketWithLifecycle(ctx context.Context, bucketName, projectID, location string, matchesPrefix []string, ttl int64) error
+	EnableBucketLifecycleDeleteRule(ctx context.Context, bucketName string, matchesPrefix []string, ttl int64) error
+	UploadLocalFileToGCS(ctx context.Context, filePath, fileName, localFilePath string) error
+	WriteDataToGCS(ctx context.Context, filePath, fileName, data string) error
+	ReadGcsFile(ctx context.Context, filePath string) (string, error)
+	ReadAnyFile(ctx context.Context, filePath string) (string, error)
```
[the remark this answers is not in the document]
Deep1998: ReadAnyFile can read a local (os.file()) or a gcs file, will add a comment.

common/utils/storage_utils.go
manitgupta: Interesting. I think that makes sense and is fair. 
Probably `utils` is overloaded here? Maybe call the package that houses `unmarshalTuningConfig` as `dataflow_helpers` or something similar. 

I've always viewed `utils` what you said here -

> The common/utils package contain methods which do not import any other SMT dependency, only third party, which is how util packages should be to freely be imported anywhere in SMT. Hence, storage_utils does stuff like parseGCSURI etc. which do not import any other SMT dependency.

accessors/spanner/spanner_accessor.go
```diff
@@ -0,0 +1,203 @@
...
+		return "", fmt.Errorf("cannot connect to database: %v", err)
+	}
+	return strings.ToLower(result.DatabaseDialect.String()), nil
+}
+
+func (sp *SpannerAccessorImpl) CheckExistingDb(ctx context.Context, adminClient spanneradmin.AdminClient, dbURI string) (bool, error) {
+	gotResponse := make(chan bool)
+	var err error
+	go func() {
+		_, err = adminClient.GetDatabase(ctx, &databasepb.GetDatabaseRequest{Name: dbURI})
```
asthamohta, line 67: why do we need to launch a goroutine for this? why not just normally fetchif the database exists?

accessors/storage/storage_accessor.go
```diff
@@ -0,0 +1,193 @@
...
+	// Read a Gcs file path and returns the contents as a string.
+	ReadGcsFile(ctx context.Context, sc storageclient.StorageClient, filePath string) (string, error)
+	// Read a local or gcs file path. Files starting with a 'gs://' are treated as GCS files.
+	ReadAnyFile(ctx context.Context, sc storageclient.StorageClient, filePath string) (string, error)
+}
+
+// This implements the StorageAccessor interface. This is the primary implementation that should be used in all places other than tests.
+type StorageAccessorImpl struct{}
+
+func (sa *StorageAccessorImpl) CreateGCSBucket(ctx context.Context, sc storageclient.StorageClient, bucketName, projectID, location string, ttl int64, matchesPrefix []string) error {
```
asthamohta, line 53: should we a create or update request to pass which can be shared among fxns?

conversion/conversion.go
```diff
@@ -520,11 +534,11 @@ func dataFromDatabaseForDataflowMigration(targetProfile profiles.TargetProfile,
 
 	// create monitoring aggregated dashboard for sharded migration
 	aggMonitoringResources := metrics.MonitoringMetricsResources{
-		ProjectId:                     targetProfile.Conn.Sp.Project,
```
asthamohta, line 523: nit: is this an accidental change?

Deep1998: This the gofmt which formats the file, there is no code change just the indentation. Im guessing the setting is different for whoever merged this.

accessors/spanner/spanner_accessor.go
Deep1998: I think this was done via a goroutine to print a warning after 5 mins of no response on ln73.
I have just lifted the function as is from conversion.go.

accessors/storage/storage_accessor.go
Deep1998: Done

accessors/spanner/spanner_accessor.go
```diff
@@ -0,0 +1,211 @@
...
+	result, err := sp.GetDatabase(ctx, dbURI)
+	if err != nil {
+		return "", fmt.Errorf("cannot connect to database: %v", err)
+	}
+	return strings.ToLower(result.DatabaseDialect.String()), nil
+}
+
+// CheckExistingDb checks whether the database with dbURI exists or not.
+// If API call doesn't respond then user is informed after every 5 minutes on command line.
+func (sp *SpannerAccessorImpl) CheckExistingDb(ctx context.Context, dbURI string) (bool, error) {
```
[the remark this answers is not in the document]
Deep1998: I have just moved this function from an existing conversion function, added a tracking bug for this. Don't want to modify this further since this is called in multiple places.

PR #763, opened 2024-02-06
chore: Refactor web.go into 3 files

manitgupta:
This refactors web.go into two files - `web.go`, `api/schema.go` and `api/rules.go`. Each group of APIs are put into their won file with associated UT.

This PR does not add new code, its just for refactoring.


webv2/api/rules.go
```diff
@@ -0,0 +1,411 @@
...
+	session.UpdateSessionFile()
+	convm := session.ConvWithMetadata{
+		SessionMetadata: sessionState.SessionMetadata,
+		Conv:            *sessionState.Conv,
+	}
+	w.WriteHeader(http.StatusOK)
+	json.NewEncoder(w).Encode(convm)
+}
+
+// setGlobalDataType allows to change Spanner type globally.
```
asthamohta, line 231: Should this go in rules?

manitgupta: This actually belongs in the rules file because it is a type of a rule that can be applied via the `ApplyRule` API.

asthamohta (approved):

webv2/config.json
```diff
@@ -1,4 +1,4 @@
 {
- "GCPProjectID": "",
- "SpannerInstanceID": ""
+ "GCPProjectID": "span-cloud-testing",
+ "SpannerInstanceID": "deep-heavy-100gb"
```
darshan-sj, line 3: Remove

webv2/api/rules.go
```diff
@@ -0,0 +1,411 @@
...
+			}
+		}
+		common.ComputeNonKeyColumnSize(sessionState.Conv, tableId)
+	}
+}
+
+// addIndex checks the new name for spanner name validity, ensures the new name is already not used by existing tables
+// secondary indexes or foreign key constraints. If above checks passed then new indexes are added to the schema else appropriate
+// error thrown.
+func addIndex(newIndex ddl.CreateIndex) (ddl.CreateIndex, error) {
```
darshan-sj, line 260: addIndex also comes under Rules?

darshan-sj (approved):
Looks Good!

One suggestion, In subsequent PRs, we can split api/schema.go even further. Can remove loadFromDump and dump related functions to a separate file and Conversion rate related functions to a separate file.

webv2/api/rules.go
manitgupta: Yeah, it is type of a rule.

PR #766, opened 2024-02-08
[feat] RR Create API 3: Add DAO to access metadatatables

Deep1998:
Add DAO to interact with the smt metadata tables. All metadata table interations should occur via this dao interface. Dao is structurally equivalent to an accessor for metadata tables.


dao/dao.go
```diff
@@ -0,0 +1,320 @@
...
+type StateData struct {
+	State string `json:"state"`
+}
+
+type DAO interface {
+	InsertSMTJobEntry(ctx context.Context, jobId, jobName, jobType, dialect, dbName string, jobData spanner.NullJSON) error
+	UpdateSMTJobState(ctx context.Context, jobId, state string) error
+	InsertSMTResourceEntry(ctx context.Context, resourceId, jobId, externalId, resourceName, resourceType string, resourceData spanner.NullJSON) error
+	UpdateSMTResourceState(ctx context.Context, resourceId, state string) error
+	UpdateSMTResourceExternalId(ctx context.Context, resourceId, externalId string) error
```
manitgupta, line 33: Nit: We can remove the `SMT` prefix, it is implicit

dao/dao_client.go
```diff
@@ -0,0 +1,48 @@
...
+	if spClient == nil {
+		once.Do(func() {
+			spClient, err = newClient(ctx, dbURI)
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to create spanner database client: %v", err)
+		}
+		return spClient, nil
+	}
+	return spClient, nil
```
manitgupta, line 42: We should add retries to the clients as well. 
We should add to other clients as well, but the Spanner client, being the DAO layer is specially important for resiliency - 
https://cloud.google.com/spanner/docs/custom-timeout-and-retry#go

dao/dao.go
```diff
@@ -0,0 +1,320 @@
...
+			"spannerDatabaseName": spannerDatabaseName,
+		},
+	}
+	return txn.Update(ctx, jobStmt)
+}
+
+func getLatestJobVersionWithinTxn(ctx context.Context, txn *spanner.ReadWriteTransaction, jobId string) (int64, error) {
+	// Fetch latest version for the job from history table.
+	stmt := spanner.Statement{SQL: `SELECT MAX(Version) FROM SMT_JOB_HISTORY WHERE JobId = @jobId;`,
+		Params: map[string]interface{}{"jobId": jobId},
```
manitgupta, line 239: This does not look like a good idea.

Monotonically increasing version numbers are a good idea when they can be generated via auto-increment by the DB itself. This kind of versioning (counting by 1) is an anti-pattern in Spanner, I think we should just make the `commit timestamp` as the other key in the PK instead of using a version number.

This will significantly simplify the DAO logic, avoid running `MAX` queries on the DB, will be human readable, and monotonic.


The same thing holds true for`resource history` as well.

dao/dao_client.go
```diff
@@ -0,0 +1,48 @@
...
+
+var once sync.Once
+var spClient *sp.Client
+
+// This function is declared as a global variable to make it testable. The unit
+// tests edit this function, acting like a double.
+var newClient = sp.NewClient
+
+func GetOrCreateClient(ctx context.Context, dbURI string) (*sp.Client, error) {
+	var err error
```
manitgupta, line 32: We already have a client implementation that takes in a `dbURI` and creates a client I think, is there need for another? Looks like a duplicate.
https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/master/accessors/clients/spanner/client/spanner_client.go

Deep1998: The spannerclient is for interactions with the targetDatabase of the customer and DAO is for the metadata database. Since we cannot reuse clients as one client is responsible for one db only, we 2 need separate ones for each.

There might also be cases in the future when a flow could need 3 databases: metadatadb, targetDatabase and some other metadatadb used by templates, in this case a new spanner client will need to be created and closed within the scope of the function where it is being called.

dao/dao.go
Deep1998: Even commit timestamp would cause hotspotting since that will also be monotonically increasing, but yes we can get rid of the fetch latest and increment logic in SMT.
Why did we decide to use an int version in the first place? I remember this was discussed i.e. having timestamp vs version col as PK.

dao/dao_client.go
Deep1998: For dao layer (and other spanner clients as well), the [default retry settings](https://github.com/googleapis/googleapis/blob/master/google/spanner/v1/spanner_grpc_service_config.json#L101-L107) already exist. I think these should suffice. 
There are no retry settings for: ExecuteStreamingSQL, StreamingRead and BatchWrite API calls which we don't do currently.

Deep1998:
@manitgupta please approve, will add integration tests later: [b/326374873](https://taskflow.corp.google.com/workspaces/2565865/work_item/b/326374873?edit=b%2F326374873)

manitgupta (approved):

PR #770, opened 2024-02-13
Refactor reports API and add unit tests

manitgupta:
Refactors reports generated APIs into their own reports.go file, and adds a UT for each of the APIs.

### DIFF
--- a/webv2/api/reports_test.go
+++ b/webv2/api/reports_test.go
@@ -1,0 +1,99 @@
+package api_test
+
+import (
+	"bufio"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/internal"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/internal/reports"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/webv2/api"
+	"github.com/stretchr/testify/assert"
+)
+
+type ReportMock struct{}
+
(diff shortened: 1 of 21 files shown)


webv2/api/reports_test.go
```diff
@@ -0,0 +1,99 @@
...
+	// do nothing since we don't want to test report generation here, only the API.
+}
+
+func (r *GenerateReportMock) GenerateStructuredReport(driverName string, dbName string, conv *internal.Conv, badWrites map[string]int64, printTableReports bool, printUnexpecteds bool) reports.StructuredReport {
+	return reports.StructuredReport{
+		MigrationType: "test",
+	}
+}
+
+func TestGetReportFile(t *testing.T) {
```
asthamohta, line 35: It looks like error flow is not tested, can we add tests for that?

webv2/api/rules_test.go
```diff
@@ -18,7 +18,6 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-
```
asthamohta, line 21: nit: we should remove this?

manitgupta: Didn't get you, what should we remove?

asthamohta: You've accidentally removed a line here and the whole file is now added to CL

asthamohta (approved):
LGTM

webv2/api/rules_test.go
manitgupta: Ahh gotcha thanks, will revert the change

PR #772, opened 2024-02-15
Upgrade angular to v17

manitgupta:
Upgrades Angular from v16 to v17, and resolves some open security issues along the way.


shreyakhajanchi (approved):
LGTM - based on https://update.angular.io/?v=16.0-17.0 

Just need to test the routing part is working as expected since routing API has undergone some change

PR #774, opened 2024-02-16
feat: accessors required for resource generation

asthamohta:


accessors/clients/datastream/interface.go
```diff
@@ -60,3 +64,29 @@ func (c *DatastreamClientImpl) UpdateStream(ctx context.Context, req *datastream
 	}
 
 }
+
+func (c *DatastreamClientImpl) CreateConnectionProfile(ctx context.Context, createRequest *datastreampb.CreateConnectionProfileRequest) (*operation.OperationWrapper[datastreampb.ConnectionProfile], error) {
+	op, err := c.client.CreateConnectionProfile(ctx, createRequest)
```
Deep1998, line 69: This seems to be slightly different from what we have been doing in other accessors, where the client interface has method signatures exactly equivalent to the accessors. Based on that logic, this would lie inside accessors.
However, these kind of functions would end up overloading the accessor and I think its better suited here how you have done it.

accessors/clients/operation/operation.go
Deep1998, line 1: This file and operation_test.go already seem to exist on master under operation/operation_test. Should we remove those files?

accessors/datastream/datastream_accessor_test.go
```diff
@@ -0,0 +1,156 @@
...
+		dsm					datastreamclient.DatastreamClientMock
+		connectionProfile   *datastreampb.ConnectionProfile
+		getProfileErr       error
+		expectedBucketName  string
+		expectedPrefix		string	
+		expectError 		bool
+	}{
+		{
+			name: "basic correct",
+			connectionProfile: &datastreampb.ConnectionProfile{Profile: &datastreampb.ConnectionProfile_GcsProfile{GcsProfile: &datastreampb.GcsProfile{Bucket: "bucket", RootPath: "/"}}},
```
Deep1998, line 50: Please add another test case where rootPath is /test-path, /test_path/, "" (empty string).

Deep1998: Whatever is allowed in rootPath basically, it may be possible that rootPath must always start with "/" or something like that in which case we may skip those cases.

accessors/datastream/datastream_accessor.go
```diff
@@ -0,0 +1,124 @@
...
+	gcsProfile := res.Profile.(*datastreampb.ConnectionProfile_GcsProfile).GcsProfile
+	bucketName := gcsProfile.Bucket
+	prefix := gcsProfile.RootPath + datastreamDestinationConnCfg.Prefix
+	prefix = utils.ConcatDirectoryPath(prefix, "data/")
+	return bucketName, prefix, nil
+}
+
+// Deletes a connection Profile
+func (da *DatastreamAccessorImpl) DeleteConnectionProfile(ctx context.Context, datastreamClient datastreamclient.DatastreamClient, id string, projectId string, region string) error {
+	op, err :=datastreamClient.DeleteConnectionProfile(ctx,  &datastreampb.DeleteConnectionProfileRequest{
```
Deep1998, line 63: nit: looks like formatting is disabled? ":=datastreamclient"

accessors/datastream/datastream_accessor.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// Creates new connection Profile
+func (da *DatastreamAccessorImpl) CreateConnectionProfile(ctx context.Context, datastreamClient datastreamclient.DatastreamClient, req *datastreampb.CreateConnectionProfileRequest) (*datastreampb.ConnectionProfile, error) {
+	op, err :=datastreamClient.CreateConnectionProfile(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	return op.Wait(ctx)
+}
```
Deep1998, line 84: nit: new line between functions.

accessors/datastream/datastream_accessor.go
```diff
@@ -0,0 +1,124 @@
...
+	
+	err = op.Wait(ctx)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Creates new connection Profile
+func (da *DatastreamAccessorImpl) CreateConnectionProfile(ctx context.Context, datastreamClient datastreamclient.DatastreamClient, req *datastreampb.CreateConnectionProfileRequest) (*datastreampb.ConnectionProfile, error) {
```
Deep1998, line 78: What we had decided for accessors was that we should avoid request bodies as function parameters. Instead, we create this request inside the function while the parameters just take the required strings/data etc. Like what is done in DataflowAccessorImpl.LaunchDataflowTemplate().

accessors/datastream/datastream_accessor.go
```diff
@@ -0,0 +1,124 @@
...
+func (da *DatastreamAccessorImpl) GetConnProfilesRegion(ctx context.Context, datastreamClient datastreamclient.DatastreamClient, projectId string, region string) ([]string, error) {
+	profilesIt := datastreamClient.ListConnectionProfiles(ctx, &datastreampb.ListConnectionProfilesRequest{Parent: "projects/" + projectId + "/locations/" + region})
+	var profiles []string = []string{}
+	for {
+		resp, err := profilesIt.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return profiles, err
```
Deep1998, line 95: Wondering if its a good idea to return whatever profiles we have collected till now. Most uses cases I have seen till now seem to return a nil/empty in case of error. Also keeps the output deterministic in case datastream decides to change orders.

accessors/spanner/mocks.go
```diff
@@ -18,18 +18,25 @@ import (
...
-	CreateChangeStreamMock          func(ctx context.Context, adminClient spanneradmin.AdminClient, changeStreamName, dbURI string) error
+	GetDatabaseDialectMock          	func(ctx context.Context, adminClient spanneradmin.AdminClient, dbURI string) (string, error)
+	CheckExistingDbMock             	func(ctx context.Context, adminClient spanneradmin.AdminClient, dbURI string) (bool, error)
+	CreateEmptyDatabaseMock         	func(ctx context.Context, adminClient spanneradmin.AdminClient, dbURI string) error
+	GetSpannerLeaderLocationMock    	func(ctx context.Context, instanceClient spinstanceadmin.InstanceAdminClient, instanceURI string) (string, error)
+	CheckIfChangeStreamExistsMock   	func(ctx context.Context, changeStreamName, dbURI string) (bool, error)
+	ValidateChangeStreamOptionsMock 	func(ctx context.Context, changeStreamName, dbURI string) error
+	CreateChangeStreamMock          	func(ctx context.Context, adminClient spanneradmin.AdminClient, changeStreamName, dbURI string) error
+	CreateDatabaseMock					func(ctx context.Context, adminClient spanneradmin.AdminClient, dbURI string, conv *internal.Conv, driver string, migrationType string) error
```
Deep1998, line 34: nit: even here the formatting seems to be off, do you have gofmt installed?

accessors/clients/operation/operation.go
asthamohta: My merge happened before the files were moved to a new repo, thats why I think it happened I have new files. let me move it

accessors/clients/datastream/interface.go
asthamohta: Yes, I am dealing with the operation result here as well and not just making a call to the backend, so I think this should be more suitable

accessors/datastream/datastream_accessor.go
asthamohta: In this particular case it becomes a bit complicated as is it possible to have multiple types of connection profiles(eg gcs, mysql, etc) and all of them require substantially different parameters. In either case I would need to create a struct to be passed as types may differ which raises the question for this particular case why not pass the request only?

accessors/spanner/mocks.go
asthamohta: This is a github issue, the code looks fine on my IDE and will be fine when I merge it

accessors/datastream/datastream_accessor_test.go
asthamohta: For concat, a public utils function is used which should be tested independently, adding a all empty string test case though

accessors/datastream/datastream_accessor.go
Deep1998: Fair makes sense in this case, its not a strict rule but we should do our best to adhere to that.

Deep1998 (approved):
LGTM, one minor comment

accessors/datastream/datastream_accessor_test.go
asthamohta: The root path is based on what is returned by the request itself so no special handling is required

PR #777, opened 2024-02-19
Adding Retry for UpdateStream LRO operation failures

VardhanThigle:
# Adding Retry for datastream update failure (after LRO poll)

- [X] Tests pass
   -  Have tested this PR against orchestration for 152 streams.
   -   N/A
- [ ] TODO:
   -  Unit tested the Retry Strategy.
   - Have also added accessor and tests for resource manager towards making it easier (in future) to refactor and UT the launch_stream code.
   - Requesting to allow this PR to get merged for easy handover of the load testing.


streaming/streaming.go
```diff
@@ -169,6 +175,21 @@ func dataStreamGaxRetrier() gax.Retryer {
...
+		RandomizationFactor: backoff.DefaultRandomizationFactor,
+		Multiplier:          DEFAULT_DATASTREAM_LRO_POLL_MULTIPLIER,
+		MaxInterval:         DEFAULT_DATASTREAM_LRO_POLL_MAX_DELAY,
+		MaxElapsedTime:      DEFAULT_DATASTREAM_LRO_POLL_MAX_ELAPSED_TIME,
+		Stop:                backoff.Stop,
+		Clock:               backoff.SystemClock,
+	}
+	dataStreamUpdateBackoff.Reset()
+	return dataStreamUpdateBackoff
+}
```
manitgupta, line 191: Why do we need a separate configuration for LRO exponential backoff retry and normal API retries?

manitgupta (commented):
There is a JSON file thats part of the commit, I think that needs to be removed?

streaming/streaming.go
VardhanThigle: The configuration parameters for retries when an API call errors out could be quite different than the retry when an LRO errors out and we should not tightly couple them.
For example an API could be retried after an initial delay of say 300 milliseconds and ramped up to an amount with exponential backoff, an LRO poll typically takes a few minutes to complete, so the initial retry could be in minutes instead of milliseconds.

streaming/streaming.go
```diff
@@ -590,23 +611,38 @@ func LaunchStream(ctx context.Context, sourceProfile profiles.SourceProfile, dbL
 	}
 	fmt.Println("Successfully created stream ", datastreamCfg.StreamId)
 
+	/* Note: Retrying across an LRO poll is a workaround and not a fix, use it only after checking with the API server team.
+	 * In most cases, if a long running operation leads into a retriable failure, the server would retry internally before marking the operation as failed.
+	 */
+	updateErr := backoff.Retry(func() error { return updateStream(ctx, streamInfo, projectNumberResource, datastreamCfg, dsClient) },
```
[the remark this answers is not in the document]
VardhanThigle: At the time of raising this PR, I was waiting for the initial accessor PR to be merged in (it was under review), since the 152 streams test got delayed a bit, I can do a more fuller integration with accessors now on this PR and will update the code in my next push.

streaming/streaming.go
VardhanThigle: We have most of the basic stuff needed to: 
1. Add accessor for LaunchStream
2. Rerfactor launchsetream and add ut.

I need to think through 2 things though:
1. Should we have a single accessor for laucnh stream (this involves both create and update calls)
2. Should we have 2 separate accessors
3. Should resourcemanager accessor be chained from outside or should the call to reasourcemanager accessor be made outside and only output passed to the datastream_accessor.

This is the place (making 2 API calls for 1 client and 1 API call for another client) where we probably desperately need a DI framework :(

streaming/streaming.go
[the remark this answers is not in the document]
VardhanThigle: We have most of the basic stuff needed to:

Add accessor for LaunchStream
Rerfactor launchsetream and add ut.

I need to think through 2 things though:

Should we have a single accessor for laucnh stream (this involves both create and update calls)
Should we have 2 separate accessors
Should resourcemanager accessor be chained from outside or should the call to reasourcemanager accessor be made outside and only output passed to the datastream_accessor.

This is the place (making 2 API calls for 1 client and 1 API call for another client) where we probably desperately need a DI framework :(

VardhanThigle:
> There is a JSON file thats part of the commit, I think that needs to be removed?

Thanks a ton. That was an uncleaned side effect of the load test :).
I cleaned it up from the PR.

accessors/resourcemanager/resourcemanager_accessor.go
```diff
@@ -0,0 +1,56 @@
...
+var ProjectNumberResourceCache sync.Map
+
+// ResourceManagerProjectsAccessor The ResourceManagerAccessor provides methods that internally use the resourceManagerProjectsClient.
+type ResourceManagerProjectsAccessor interface {
+	GetProjectNumberResource(ctx context.Context, rmProjectsClient resourcemanager.ProjectsClient, projectID string) string
+}
+type ResourceManagerProjectsAccessorImpl struct{}
+
+// GetProjectNumberResource returns a string that encodes the project number like `projects/12345`.
+func (rm *ResourceManagerProjectsAccessorImpl) GetProjectNumberResource(ctx context.Context, rmProjectsClient resourcemanagerclient.ResourcemanagerProjectsClient, projectID string) string {
```
manitgupta, line 38: This is still not fixed on Datastream end? Can we remove this?

manitgupta (approved):

streaming/streaming.go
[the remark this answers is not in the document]
VardhanThigle: Resolving as discssed offline. Checking this in to unblock handover of load testing. UT refactoring would be done later on.

PR #778, opened 2024-02-20
feat: json resource generation cmd

asthamohta:
The feature adds capability for customers to create source and target profiles based on json configuration passed by customers in the command line.


cmd/data.go
```diff
@@ -156,8 +162,31 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 		banner = utils.GetBanner(now, dbURI)
 	} else {
 		conv.Audit.DryRun = true
+		// If migration type is Minimal Downtime, validate if required resources can be generated
+		if !conv.UI && sourceProfile.Driver == constants.MYSQL && sourceProfile.Ty == profiles.SourceProfileTypeConfig && sourceProfile.Config.ConfigType == constants.DATAFLOW_MIGRATION {
```
manitgupta, line 166: Do we still need the `conv.UI` flag? I thought it was temporary?

conversion/data_from_database.go
```diff
@@ -221,4 +270,20 @@ func (dd *DataFromDatabaseImpl) dataFromDatabaseForBulkMigration(sourceProfile p
...
+	}
+	// Fetch the GCS path from the destination connection profile.
+	dstProf := fmt.Sprintf("projects/%s/locations/%s/connectionProfiles/%s", project, location, profileName)
+	res, err := dsClient.GetConnectionProfile(ctx, dstProf)
+	if err != nil {
+		return "", "", fmt.Errorf("could not get connection profile: %v", err)
+	}
+	gcsProfile := res.Profile.(*datastreampb.ConnectionProfile_GcsProfile).GcsProfile
+	return gcsProfile.GetBucket(), gcsProfile.GetRootPath(), nil
+}
```
manitgupta, line 289: Is this refactored or newly added code? 
Looks like something should be in `datastream_accessor`?

internal/convert.go
```diff
@@ -29,26 +29,29 @@ import (
...
+	SpDialect          string              // The dialect of the spanner database to which Spanner migration tool is writing.
+	UniquePKey         map[string][]string // Maps Spanner table name to unique column name being used as primary key (if needed).
+	Audit              Audit               `json:"-"` // Stores the audit information for the database conversion
+	Rules              []Rule              // Stores applied rules during schema conversion
+	IsSharded          bool                // Flag denoting if the migration is sharded or not
+	ConvLock           sync.RWMutex        `json:"-"` // ConvLock prevents concurrent map read/write operations. This lock will be used in all the APIs that either read or write elements to the conv object.
+	SpRegion           string              // Region for Spanner Instance
+	ResourceValidation bool                // Flag denoting if validation for resources to generated is complete
+	UI                 bool                // Flag if UI interface was used for migration. ToDo: Remove flag after resource generation is introduced to UI
 }
```
manitgupta, line 55: Has formatting changed? Do you have gofmt installed?

conversion/resource_generation.go
```diff
@@ -0,0 +1,401 @@
...
+			dstProfilesToCreate = append(dstProfilesToCreate, dstProfile)
+		}
+	}
+	return sourceProfilesToCreate, dstProfilesToCreate, err
+}
+
+// 1. For each datashard, check if source and destination connection profile exists or not
+// 2. If source connection profile doesn't exists create it or validate if creation is possible.
+// 3. If validation is false and destination connection profile doesn't exists create a corresponding gcs bucket and then a destination connection profile
+func (c *ValidateOrCreateResourcesImpl) ValidateOrCreateResourcesForShardedMigration(ctx context.Context, projectId string, instanceName string, validateOnly bool, region string, sourceProfile profiles.SourceProfile) error {
```
manitgupta, line 226: Both this method and the interface is doing a lot. Is there a challenge to break this into `validate` and `create` as different methods?

conversion/resource_generation.go
```diff
@@ -0,0 +1,401 @@
...
+
+// Method to validate if in a minimal downtime migration, required resources can be generated
+func (v *ValidateResourcesImpl) ValidateResourceGeneration(ctx context.Context, projectId string, instanceId string, sourceProfile profiles.SourceProfile, conv *internal.Conv) error {
+	spannerRegion, err := v.SpAcc.GetSpannerLeaderLocation(ctx, v.SpInstanceAdmin, "projects/"+projectId+"/instances/"+instanceId)
+	if err != nil {
+		err = fmt.Errorf("unable to fetch Spanner Region: %v", err)
+		return err
+	}
+	conv.SpRegion = spannerRegion
+	err = v.ValidateOrCreateResources.ValidateOrCreateResourcesForShardedMigration(ctx, projectId, instanceId, true, spannerRegion, sourceProfile)
```
manitgupta, line 115: This looks odd - `ValidateResource` method has a `ValidateOrCreate` inside it. Is this a naming issue or does the code require restructuring?

conversion/resource_generation.go
```diff
@@ -0,0 +1,401 @@
...
+		resourcesForCleanup = append(resourcesForCleanup, createResourceData)
+	} else {
+		fmt.Printf("Connection Profile for Datashard %v has been validated: %v\n", createResourceData.ConnectionProfile.DatashardId, createResourceData.ConnectionProfile.Id)
+	}
+
+	return common.TaskResult[*ConnectionProfileReq]{Result: createResourceData, Err: nil}
+}
+
+// If any of the resource creation fails, deletes all resources that were created
+func (r ResourceGenerationImpl) ConnectionProfileCleanUp(ctx context.Context, profiles []*ConnectionProfileReq) error {
```
manitgupta, line 175: You should name this something like `RollbackResourceCreation` or similar to demarcate what this operation is for (and not what it does)

conversion/resource_generation.go
```diff
@@ -0,0 +1,401 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Returns source and destination connection profiles to be created
+func (r ResourceGenerationImpl) GetResourcesForCreation(ctx context.Context, projectId string, sourceProfile profiles.SourceProfile, region string, validateOnly bool) ([]*ConnectionProfileReq, []*ConnectionProfileReq, error) {
```
manitgupta, line 193: Rename to something more descriptive - `GetConnectionProfilesForResources` ?

conversion/data_from_database.go
asthamohta: This is existing code refactored to use the accessor. Didn't make any other changes except use of accessor in it as it is used in a few functions in UI as well

internal/convert.go
asthamohta: used gofmt only to reformat it

conversion/resource_generation.go
asthamohta: Answering both the last and this comment here:
I have separated out things that need to only happen for validate in "ValidateResourceGeneration", but in "ValidateOrCreateResourcesForShardedMigration" the only difference is of a boolean value we send to the backend which is why if I bread them, it would be the same code with one validateOnly set as true and the other set as false

cmd/data.go
```diff
@@ -156,8 +162,31 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
...
+				return subcommands.ExitFailure
+			}
+			dsClient, err := datastreamclient.NewDatastreamClientImpl(ctx)
+			if err != nil {
+				return subcommands.ExitFailure
+			}
+			storageclient, err := storageclient.NewStorageClientImpl(ctx)
+			if err != nil {
+				return subcommands.ExitFailure
+			}
```
manitgupta, line 178: If any of these fails, what error message would be shown to the user? 
We should log the error if this would fail silently.

conversion/data_from_database.go
```diff
@@ -172,7 +220,8 @@ func (dd *DataFromDatabaseImpl) dataFromDatabaseForDataflowMigration(targetProfi
 		}
 		return common.TaskResult[*profiles.DataShard]{Result: p, Err: err}
 	}
-	_, err = common.RunParallelTasks(sourceProfile.Config.ShardConfigurationDataflow.DataShards, 20, asyncProcessShards, true)
+	r := common.RunParallelTasksImpl[*profiles.DataShard, *profiles.DataShard]{}
+	_, err = r.RunParallelTasks(sourceProfile.Config.ShardConfigurationDataflow.DataShards, 20, asyncProcessShards, true)
```
manitgupta, line 224: The parallelism should be defined as a static constant

conversion/resource_generation.go
manitgupta: I dug into this a little bit deeper, and I think the root of the problem is that the Datastream API itself is very awkward. The `CreateConnectionProfileRequest` contains the `validateOnly` flag which makes it function effectively like a `ValidateConnectionProfileRequest` object. 

This same awkwardness is percolating in our codebase, which is why we have these kind of `ValidateOrCreate` methods, since we are effectively setting a parameter in the API request.

Had they been two APIs, we would have not done so.

I think its okay for us to move ahead with existing structure given the nature of the API, but it sure would have been cleaner if we were able to separate out these flows!

asthamohta: Okay thankyou

conversion/data_from_database.go
asthamohta: So I checked, complex types can't be constant in golang. This isn't possible

manitgupta (approved):

cmd/utils.go
```diff
@@ -185,8 +190,32 @@ func migrateData(ctx context.Context, targetProfile profiles.TargetProfile, sour
...
+			return nil, err
+		}
+		storageclient, err := storageclient.NewStorageClientImpl(ctx)
+		if err != nil {
+			return nil, err
+		}
+		validateResource := conversion.NewValidateResourcesImpl(&spanneraccessor.SpannerAccessorImpl{}, spClient, &datastream_accessor.DatastreamAccessorImpl{},
+			dsClient, &storageaccessor.StorageAccessorImpl{}, storageclient)
+		err = validateResource.ValidateResourceGeneration(ctx, targetProfile.Conn.Sp.Project, targetProfile.Conn.Sp.Instance, sourceProfile, conv)
+		if err != nil {
```
Deep1998, line 211: This code piece seem duplicated thrice, maybe can be moved to a function and called instead?

conversion/conversion.go
```diff
@@ -49,7 +49,44 @@ var (
 	datastreamClient *datastream.Client
 )
 
-func getDatastreamClient(ctx context.Context) *datastream.Client {
+type ConnectionProfile struct {
```
Deep1998, line 52: nit: this feels like this should reside elsewhere, seems like we are overloading conversion.go again.

Deep1998: conversion_types.go? or maybe somewhere in profiles.go? @manitgupta WDYT

conversion/data_from_database.go
```diff
@@ -64,6 +69,49 @@ func (dd *DataFromDatabaseImpl) dataFromDatabaseForDMSMigration() (*writer.Batch
...
+		err = createResources.ValidateOrCreateResourcesForShardedMigration(ctx, targetProfile.Conn.Sp.Project, targetProfile.Conn.Sp.Instance, false, conv.SpRegion, sourceProfile)
+		if err != nil {
+			return nil, fmt.Errorf("unable to create connection profiles: %v", err)
+		}
+	}
+
+	//Set the TmpDir from the sessionState bucket which is derived from the target connection profile
+	for _, dataShard := range sourceProfile.Config.ShardConfigurationDataflow.DataShards {
+		if dataShard.TmpDir == "" {
+		bucket, rootPath, err := GetBucket(targetProfile.Conn.Sp.Project, conv.SpRegion, dataShard.DstConnectionProfile.Name)
```
Deep1998, line 107: nit/side-task: we should rename thi method to GetBucketFromDatastreamProfile or something.

internal/convert.go
```diff
@@ -29,26 +29,29 @@ import (
...
+	sampleBadRows      rowSamples          // Rows that generated errors during conversion.
+	Stats              stats               `json:"-"`
+	TimezoneOffset     string              // Timezone offset for timestamp conversion.
+	SpDialect          string              // The dialect of the spanner database to which Spanner migration tool is writing.
+	UniquePKey         map[string][]string // Maps Spanner table name to unique column name being used as primary key (if needed).
+	Audit              Audit               `json:"-"` // Stores the audit information for the database conversion
+	Rules              []Rule              // Stores applied rules during schema conversion
+	IsSharded          bool                // Flag denoting if the migration is sharded or not
+	ConvLock           sync.RWMutex        `json:"-"` // ConvLock prevents concurrent map read/write operations. This lock will be used in all the APIs that either read or write elements to the conv object.
+	SpRegion           string              // Region for Spanner Instance
```
Deep1998, line 52: nit: Leader region

internal/convert.go
```diff
@@ -29,26 +29,29 @@ import (
...
+	ResourceValidation bool                // Flag denoting if validation for resources to generated is complete
```
Deep1998, line 53: This is not being used anywhere?

internal/convert.go
```diff
@@ -29,26 +29,29 @@ import (
...
+	UI                 bool                // Flag if UI interface was used for migration. ToDo: Remove flag after resource generation is introduced to UI
```
Deep1998, line 54: Where do we set this?

profiles/source_profile.go
```diff
@@ -571,7 +571,11 @@ type DirectConnectionConfig struct {
 
 type DatastreamConnProfile struct {
 	Name     string `json:"name"`
-	Location string `json:"location"`
```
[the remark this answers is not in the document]
Deep1998: If this is just used as source profile in practice, we should be good.

conversion/resource_generation.go
```diff
@@ -0,0 +1,401 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package conversion handles initial setup for the command line tool
+// and web APIs.
+
+// TODO:(searce) Organize code in go style format to make this file more readable.
```
Deep1998, line 18: TODO not needed

conversion/resource_generation.go
```diff
@@ -0,0 +1,401 @@
...
+		if err != nil {
+			return fmt.Errorf("resource generation failed due to %s, resources created could not be cleaned up, please cleanup manually: %s", resCreationErr.Error(), err.Error())
+		} else {
+			return resCreationErr
+		}
+	} else if resCreationErr != nil {
+		return resCreationErr
+	}
+	for _, resource := range resSourceProfiles {
+		if resource.Result.Error != nil && validateOnly {
```
Deep1998, line 258: nit: validateOnly check should be outside?
```
if validateOnly{
		for _, resource := range resSourceProfiles {
				if resource.Result.Error != nil {
					// If validation failed, append to list of errors
					errorsList = append(errorsList, resource.Result.Error)
				}
		}
}
```

conversion/resource_generation.go
```diff
@@ -0,0 +1,401 @@
...
+	// If the errors occurred during validation of resource creation, return all errors
+	if len(errorsList) != 0 {
+		return multiError(errorsList)
+	}
+	// cleanup resources for cleanup if migration is successful
+	resourcesForCleanup = nil
+	return nil
+}
+
+// checks if source connection profile exists, if not, returns a request reuired to create it
```
Deep1998, line 286: required*

conversion/resource_generation.go
```diff
@@ -0,0 +1,401 @@
...
+				Region:       profile.SrcConnectionProfile.Location,
+				ValidateOnly: validateOnly},
+			Ctx: ctx,
+		}
+		return req, nil
+	}
+	return nil, nil
+}
+
+// checks if target connection profile exists, if not, returns a request reuired to create it
```
Deep1998, line 330: required*

conversion/resource_generation.go
```diff
@@ -0,0 +1,401 @@
...
+func getDstConnectionProfileForCreation(ctx context.Context, projectId string, profile *profiles.DataShard, region string, validateOnly bool, connectionProfiles map[string][]string, dsAcc datastream_accessor.DatastreamAccessor, dsClient ds.DatastreamClient) (*ConnectionProfileReq, error) {
+	dstProfileExists := false
+	var err error
+	// Destination connection profiles do not need to be validated as for their creation gcs bucket will also be created
```
Deep1998, line 334: unclear what this comment means

conversion/resource_generation.go
```diff
@@ -0,0 +1,401 @@
...
+				RootPath: "/",
+			},
+		}
+		return nil
+	}
+	return nil
+}
+
+// Clubs multiple errors into one error
+func multiError(errorMessages []error) error {
```
Deep1998, line 395: nit: rename to joinErrors/mergeErrors/concatErrors

cmd/data.go
```diff
@@ -156,8 +162,34 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 		banner = utils.GetBanner(now, dbURI)
 	} else {
 		conv.Audit.DryRun = true
+		// If migration type is Minimal Downtime, validate if required resources can be generated
+		if !conv.UI && sourceProfile.Driver == constants.MYSQL && sourceProfile.Ty == profiles.SourceProfileTypeConfig && sourceProfile.Config.ConfigType == constants.DATAFLOW_MIGRATION {
```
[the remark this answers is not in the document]
asthamohta: This is only for shraded migration which is in only mySql right now

cmd/utils.go
asthamohta: The error handling is different in each case and so are the arguments, it will end up being the same number of LOC even if if I do it in a function

internal/convert.go
asthamohta: My bad, its in the UI bit. I thought i would add the changes as part of my UI Pr, but you're right. It needs to go with this PR

profiles/source_profile.go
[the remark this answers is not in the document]
asthamohta: It is for both, but then I would need to split it, should I do so?

conversion/resource_generation.go
asthamohta: removing for other files in the same PR as well

cmd/utils.go
Deep1998: What I meant was something like `func foo() error {...}`. 
and the callers will look something like:
```
err := foo()
if err != nil {
    return nil, err // or subcommands.ExitFailure() for other implementations.
}
```

profiles/source_profile.go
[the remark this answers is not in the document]
Deep1998: If the corresponding changes in other places to handle this are'nt too big we should go ahead and do it. 

But I dont see where are these parameters being used - Host, User, Port etc? Ideally, we should'nt need these here as they can be fetched from the connection profile and stored in a separate structure.

PR #802, opened 2024-03-27
Introducing migrationProjectId and decoupling Spanner projectId

darshan-sj:

- [ ] Tests pass

### DIFF
--- a/cmd/data.go
+++ b/cmd/data.go
@@ -48,6 +48,7 @@ type DataCmd struct {
 	targetProfile   string
 	sessionJSON     string
 	filePrefix      string // TODO: move filePrefix to global flags
+	project         string
 	WriteLimit      int64
 	dryRun          bool
 	logLevel        string
@@ -84,6 +85,7 @@ func (cmd *DataCmd) SetFlags(f *flag.FlagSet) {
 	f.StringVar(&cmd.target, "target", "Spanner", "Specifies the target DB, defaults to Spanner (accepted values: `Spanner`)")
 	f.StringVar(&cmd.targetProfile, "target-profile", "", "Flag for specifying connection profile for target database e.g., \"dialect=postgresql\"")
 	f.StringVar(&cmd.filePrefix, "prefix", "", "File prefix for generated files")
+	f.StringVar(&cmd.project, "project", "", "Flag spcifying default project id for all the generated resources for the migration")
 	f.Int64Var(&cmd.WriteLimit, "write-limit", DefaultWritersLimit, "Write limit for writes to spanner")
 	f.BoolVar(&cmd.dryRun, "dry-run", false, "Flag for generating DDL and schema conversion report without creating a spanner database")
 	f.StringVar(&cmd.logLevel, "log-level", "DEBUG", "Configure the logging level for the command (INFO, DEBUG), defaults to DEBUG")
@@ -108,6 +110,13 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 	defer logger.Log.Sync()
 
 	conv := internal.MakeConv()
+	if cmd.project == "" {
+		getInfo := &utils.GetUtilInfoImpl{}
+		cmd.project, err = getInfo.GetProject()
+		if err != nil {
+			logger.Log.Error("Could not set default project id from gcloud environment", zap.Error(err))
+		}
+	}
 	// validate and parse source-profile, target-profile and source
 	sourceProfile, targetProfile, ioHelper, dbName, err := PrepareMigrationPrerequisites(cmd.sourceProfile, cmd.targetProfile, cmd.source)
 	if err != nil {
@@ -149,7 +158,7 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 	)
 	if !cmd.dryRun {
 		now := time.Now()
-		bw, err = MigrateDatabase(ctx, targetProfile, sourceProfile, dbName, &ioHelper, cmd, conv, nil)
+		bw, err = MigrateDatabase(ctx, cmd.project, targetProfile, sourceProfile, dbName, &ioHelper, cmd, conv, nil)
 		if err != nil {
 			err = fmt.Errorf("can't finish database migration for db %s: %v", dbName, err)
 			return subcommands.ExitFailure
@@ -159,14 +168,14 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 		conv.Audit.DryRun = true
 		// If migration type is Minimal Downtime, validate if required resources can be generated
 		if !conv.UI && sourceProfile.Driver == constants.MYSQL && sourceProfile.Ty == profiles.SourceProfileTypeConfig && sourceProfile.Config.ConfigType == constants.DATAFLOW_MIGRATION {
-			err := ValidateResourceGenerationHelper(ctx, targetProfile.Conn.Sp.Project, targetProfile.Conn.Sp.Instance, sourceProfile, conv)
+			err := ValidateResourceGenerationHelper(ctx, cmd.project, targetProfile.Conn.Sp.Project, targetProfile.Conn.Sp.Instance, sourceProfile, conv)
 			if err != nil {
 				return subcommands.ExitFailure
 			}
 		}
 
 		convImpl := &conversion.ConvImpl{}
-		bw, err = convImpl.DataConv(ctx, sourceProfile, targetProfile, &ioHelper, nil, conv, true, cmd.WriteLimit, &conversion.DataFromSourceImpl{})
+		bw, err = convImpl.DataConv(ctx, cmd.project, sourceProfile, targetProfile, &ioHelper, nil, conv, true, cmd.WriteLimit, &conversion.DataFromSourceImpl{})
 
 		if err != nil {
 			err = fmt.Errorf("can't finish data conversion for db %s: %v", dbName, err)
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -127,7 +127,7 @@ func PrepareMigrationPrerequisites(sourceProfileString, targetProfileString, sou
 }
 
 // MigrateData creates database and populates data in it.
-func MigrateDatabase(ctx context.Context, targetProfile profiles.TargetProfile, sourceProfile profiles.SourceProfile, dbName string, ioHelper *utils.IOStreams, cmd interface{}, conv *internal.Conv, migrationError *error) (*writer.BatchWriter, error) {
+func MigrateDatabase(ctx context.Context, migrationProjectId string, targetProfile profiles.TargetProfile, sourceProfile profiles.SourceProfile, dbName string, ioHelper *utils.IOStreams, cmd interface{}, conv *internal.Conv, migrationError *error) (*writer.BatchWriter, error) {
 	var (
 		bw  *writer.BatchWriter
 		err error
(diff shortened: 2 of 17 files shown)


cmd/utils.go
```diff
@@ -161,22 +161,22 @@ func MigrateDatabase(ctx context.Context, targetProfile profiles.TargetProfile,
...
+		return err
+	}
+	err = spA.CreateOrUpdateDatabase(ctx, adminClientImpl, dbURI, sourceProfile.Driver, conv, sourceProfile.Config.ConfigType)
+	if err != nil {
+		err = fmt.Errorf("can't create/update database: %v", err)
+		return err
+	}
+	metricsPopulation(ctx, sourceProfile.Driver, conv)
+	conv.Audit.Progress.UpdateProgress("Schema migration complete.", completionPercentage, internal.SchemaMigrationComplete)
+	return nil
```
manitgupta, line 176: Nit: Did the formatting change here? Can you revert?

conversion/conversion.go
```diff
@@ -86,7 +86,7 @@ func (ci *ConvImpl) SchemaConv(sourceProfile profiles.SourceProfile, targetProfi
 
 // DataConv performs the data conversion
 // The SourceProfile param provides the connection details to use the go SQL library.
-func (ci *ConvImpl) DataConv(ctx context.Context, sourceProfile profiles.SourceProfile, targetProfile profiles.TargetProfile, ioHelper *utils.IOStreams, client *sp.Client, conv *internal.Conv, dataOnly bool, writeLimit int64, dataFromSource DataFromSourceInterface) (*writer.BatchWriter, error) {
+func (ci *ConvImpl) DataConv(ctx context.Context, migrationProjectId string, sourceProfile profiles.SourceProfile, targetProfile profiles.TargetProfile, ioHelper *utils.IOStreams, client *sp.Client, conv *internal.Conv, dataOnly bool, writeLimit int64, dataFromSource DataFromSourceInterface) (*writer.BatchWriter, error) {
```
manitgupta, line 89: Similar comment, can this go inside `ctx`? 
Also seems like `DataConv` had ctx but SchemaConv does not (not relevant to the PR, just an observation)

manitgupta (commented):
Overall looks good. Added some comments to check if there is an opportunity to simplify by using `context` instead of adding a new parameter wherever possible.

This is not a small amount to work to figure out quickly! Thanks a lot!

common/utils/utils.go
```diff
@@ -182,7 +182,7 @@ func PreloadGCSFiles(tables []ManifestTable) ([]ManifestTable, error) {
 	return tables, nil
 }
 
-// GetProject returns the cloud project we should use for accessing Spanner.
+// GetProject returns the cloud project we should use by default.
```
shreyakhajanchi, line 185: Better to complete the statement by stating : GetProject returns the cloud project we should use by default to create resources

cmd/data.go
```diff
@@ -84,6 +85,7 @@ func (cmd *DataCmd) SetFlags(f *flag.FlagSet) {
 	f.StringVar(&cmd.target, "target", "Spanner", "Specifies the target DB, defaults to Spanner (accepted values: `Spanner`)")
 	f.StringVar(&cmd.targetProfile, "target-profile", "", "Flag for specifying connection profile for target database e.g., \"dialect=postgresql\"")
 	f.StringVar(&cmd.filePrefix, "prefix", "", "File prefix for generated files")
+	f.StringVar(&cmd.project, "project", "", "Flag spcifying default project id for all the generated resources for the migration")
```
shreyakhajanchi, line 88: Since target profile also has a field named project and we are adding one here with the note: "Flag spcifying default project id for all the generated resources for the migration" it looks a bit confusing to me. Maybe we should rename to where it is used

webv2/web.go
```diff
@@ -622,6 +623,12 @@ func migrate(w http.ResponseWriter, r *http.Request) {
 	sessionState.Conv.Audit.Progress = internal.Progress{}
 	sessionState.Conv.UI = true
 	sourceProfile, targetProfile, ioHelper, dbName, err := getSourceAndTargetProfiles(sessionState, details)
+	// TODO: Revisit the logic
+	getInfo := &utils.GetUtilInfoImpl{}
```
shreyakhajanchi, line 627: we use `sessionState.GCPProjectID` at multiple places to get connection profile, set sharded config, etc and this is the value from the UI - config. I think this also needs to be changed to  `migrationProjectId`

conversion/conversion_from_source.go
```diff
@@ -289,11 +287,12 @@ func (sads *DataFromSourceImpl) dataFromDatabase(ctx context.Context, sourceProf
 			}
 
 			monitoringResources := metrics.MonitoringMetricsResources{
-				ProjectId:            targetProfile.Conn.Sp.Project,
+				MigrationProjectId:   migrationProjectId,
 				DataflowJobId:        dfOutput.JobID,
 				DatastreamId:         streamingCfg.DatastreamCfg.StreamId,
 				JobMetadataGcsBucket: gcsBucket,
 				PubsubSubscriptionId: streamingCfg.PubsubCfg.SubscriptionId,
+				SpannerProjectId:     targetProfile.Conn.Sp.Project,
```
krishnamoorthy-r, line 295: why would monitoring resource need spanner project id? Does it try to pull metrics from cross proejct? how would auth handled?

conversion/data_from_database.go
```diff
@@ -203,7 +202,7 @@ func (dd *DataFromDatabaseImpl) dataFromDatabaseForDataflowMigration(targetProfi
 			dashboardName = strings.Split(respDash.Name, "/")[3]
 			fmt.Printf("Monitoring Dashboard for shard %v: %+v\n", p.DataShardId, dashboardName)
 		}
-		streaming.StoreGeneratedResources(conv, streamingCfg, dfOutput.JobID, dfOutput.GCloudCmd, targetProfile.Conn.Sp.Project, p.DataShardId, internal.GcsResources{BucketName: gcsBucket}, dashboardName)
+		streaming.StoreGeneratedResources(conv, streamingCfg, dfOutput.JobID, dfOutput.GCloudCmd, migrationProjectId, p.DataShardId, internal.GcsResources{BucketName: gcsBucket}, dashboardName)
```
krishnamoorthy-r, line 205: Where do we store? If it is spanner (i.e target), shouldn't this continue to be targetProfile project?

krishnamoorthy-r (commented):
Thanks for the changes, looks much bigger than what I originally thought!

cmd/data.go
```diff
@@ -108,6 +110,13 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 	defer logger.Log.Sync()
 
 	conv := internal.MakeConv()
+	if cmd.project == "" {
+		getInfo := &utils.GetUtilInfoImpl{}
+		cmd.project, err = getInfo.GetProject()
+		if err != nil {
+			logger.Log.Error("Could not set default project id from gcloud environment", zap.Error(err))
```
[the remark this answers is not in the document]
darshan-sj: Done.

cmd/data.go
```diff
@@ -149,7 +158,7 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 	)
 	if !cmd.dryRun {
 		now := time.Now()
-		bw, err = MigrateDatabase(ctx, targetProfile, sourceProfile, dbName, &ioHelper, cmd, conv, nil)
+		bw, err = MigrateDatabase(ctx, cmd.project, targetProfile, sourceProfile, dbName, &ioHelper, cmd, conv, nil)
```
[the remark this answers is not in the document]
darshan-sj: I am reluctant to use context. Variables stored in context don't have defined structure. Code maintenance will be very difficult. If we have it as an explicit parameter in methods, changes can be verified at compile time.

cmd/utils.go
darshan-sj: Done.

conversion/conversion.go
```diff
@@ -67,16 +67,16 @@ func GetDatastreamClient(ctx context.Context) *datastream.Client {
 
 type ConvInterface interface {
 	SchemaConv(sourceProfile profiles.SourceProfile, targetProfile profiles.TargetProfile, ioHelper *utils.IOStreams, schemaFromSource SchemaFromSourceInterface) (*internal.Conv, error)
```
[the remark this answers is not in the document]
darshan-sj: Changed. But interesting to see that it didn't break the build before.

conversion/conversion.go
darshan-sj: It will be very difficult to maintain if we use ctx. We won't be able to properly track and cover all the edge cases of passing and utilizing migration project id. Especially since we have so many entry points.

common/utils/utils.go
darshan-sj: Done.

cmd/data.go
darshan-sj: project field in target profile refers to Spanner's project id (since it is present in target profile). Project flag outside refers to migration project id which will be used to create other migration resources like Dataflow jobs, Datastream, Pubsub etc. Do you think anything else need to be mentioned in the flag description?

webv2/web.go
darshan-sj: Changed this. I have also changed a bunch of other UI files. Please review.

cmd/data.go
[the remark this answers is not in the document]
darshan-sj: I added a fallback to project id set in target profile.

cmd/data.go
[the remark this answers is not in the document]
darshan-sj: The cmd parameter (7th param) is received as a generic interface inside and is used to determine the type of command only. The structure of cmd object is not visible inside MigrateDatabase method.

conversion/conversion_from_source.go
darshan-sj: Monitoring resource object contains details of all the resources created for a migration. I added SpannerProjectId here for completeness. How auth and permissions needs to be handled will need to be taken up as a separate task.

conversion/data_from_database.go
darshan-sj: This method is storing the generated resources in conv object and not Spanner metadata. It doesn't need spanner project id.

cmd/data.go
```diff
@@ -114,6 +116,14 @@ func (cmd *DataCmd) Execute(ctx context.Context, f *flag.FlagSet, _ ...interface
 		err = fmt.Errorf("error while preparing prerequisites for migration: %v", err)
 		return subcommands.ExitUsageError
 	}
+	if cmd.project == "" {
+		getInfo := &utils.GetUtilInfoImpl{}
+		cmd.project, err = getInfo.GetProject()
+		if err != nil {
+			logger.Log.Error("Could not get project id from gcloud environment or --project flag. Inferring the migration project id from target profile.", zap.Error(err))
+			cmd.project = targetProfile.Conn.Sp.Project
```
manitgupta, line 124: Why are we doing this? For backward compatibility?

manitgupta: If we are proceeding ahead it should not be an ERROR log, maybe WARN. 

I think we should just fail though and not use the targetProfile value as a fallback

manitgupta (commented):
1. Please also update the Github documentation with these changes
2. Please update the PR description with what features break under what scenarios. For example - `cleanup` CLI will not work for cases when spanner project is different from migration resources project (and so on and so forth)

cmd/data.go
darshan-sj: Done.

darshan-sj:
I have updated the github documentation as well

manitgupta (approved):

PR #805, opened 2024-04-01
Dataflow version bump and project separation

aksharauke:
1. Fixes bug 330334591 by the new template versions
2. Gives ability to have Spanner in a separate project than Dataflow
3. Information about a new metric added


docs/reverse-replication/RunnigReverseReplication.md
```diff
@@ -53,6 +53,7 @@ The script takes in multiple arguments to orchestrate the pipeline. They are:
 - `sourceDbTimezoneOffset`: the timezone offset with respect to UTC for the source database.Defaults to +00:00.
 - `sourceShardsFilePath`: GCS file path for file containing shard info. Details on structure mentioned later.
 - `sourceWriterTemplateLocation` : the dataflow template location for the Source writer job.
+- `spannerProjectId`: the project id where Cloud Spanner resides, for use case when Cloud Spanner is in a different project than where Dataflow would run.
```
Deep1998, line 56: There already seems to be a `projectId`: Project id of the Spanner instance. We should only keep one

Deep1998 (approved):
One minor comment, overall LGTM

docs/reverse-replication/RunnigReverseReplication.md
aksharauke: Done, updated the description for the projectId.

PR #815, opened 2024-04-22
feat: UUID backend changes

asthamohta:
Enabling existing "Add Column Feature" as well in the PR
Tests file changes for compatibility are ~500 LOC, any test file in which corresponding changes are not made to the file it is testing can be ignored


sources/common/toddl.go
```diff
@@ -131,6 +131,10 @@ func (ss *SchemaToSpannerImpl) SchemaToSpannerDDLHelper(conv *internal.Conv, tod
 			NotNull: isNotNull,
 			Comment: "From: " + quoteIfNeeded(srcCol.Name) + " " + srcCol.Type.Print(),
 			Id:      srcColId,
+			AutoGen: ddl.AutoGenCol{
+				Name: "None",
```
bharadwaj-aditya, line 135: I see "None" being used as a string everywhere. Any reason to not use the None literal ?

sources/mysql/toddl_test.go
```diff
@@ -274,16 +274,16 @@ func TestToSpannerPostgreSQLDialectType(t *testing.T) {
...
-			"c2":  ddl.ColumnDef{Name: "b", Id: "c2", T: ddl.Type{Name: ddl.Float64}},
-			"c3":  ddl.ColumnDef{Name: "c", Id: "c3", T: ddl.Type{Name: ddl.Bool}},
-			"c4":  ddl.ColumnDef{Name: "d", Id: "c4", T: ddl.Type{Name: ddl.String, Len: int64(6)}},
-			"c5":  ddl.ColumnDef{Name: "e", Id: "c5", T: ddl.Type{Name: ddl.Numeric}},
-			"c6":  ddl.ColumnDef{Name: "f", Id: "c6", T: ddl.Type{Name: ddl.Timestamp}},
-			"c7":  ddl.ColumnDef{Name: "g", Id: "c7", T: ddl.Type{Name: ddl.JSON}},
-			"c8":  ddl.ColumnDef{Name: "h", Id: "c8", T: ddl.Type{Name: ddl.Date}},
-			"c9":  ddl.ColumnDef{Name: "i", Id: "c9", T: ddl.Type{Name: ddl.Timestamp}},
-			"c10": ddl.ColumnDef{Name: "j", Id: "c10", T: ddl.Type{Name: ddl.Bytes, Len: ddl.MaxLength}},
+			"c1":  ddl.ColumnDef{Name: "a", Id: "c1", T: ddl.Type{Name: ddl.Int64}, AutoGen: ddl.AutoGenCol{Name: "None", Type: "None"}},
```
bharadwaj-aditya, line 277: Are there tests where the autogen column is set ?

webv2/api/schema.go
```diff
@@ -1534,3 +1553,51 @@ func addShardIdToForeignKeyPerTable(isAddedAtFirst bool, table ddl.CreateTable)
...
+
+func initializeAutoGenMap() {
+	sessionState := session.GetSessionState()
+	autoGenMap = make(map[string][]types.AutoGen)
+	switch sessionState.Conv.SpDialect {
+	case constants.DIALECT_POSTGRESQL:
+		makePostgresDialectAutoGenMap()
+		return
+	case constants.DIALECT_GOOGLESQL:
+		makeGoogleSqlDialectAutoGenMap()
```
bharadwaj-aditya, line 1565: nit: can be merged with default.

schema/schema.go
```diff
@@ -119,6 +119,11 @@ type Ignored struct {
 	AutoIncrement bool
 }
 
+type AutoGenCol struct {
+	Name string
+	Type string
```
bharadwaj-aditya, line 124: Type as a variable name seems heavily overloaded here. It was not until the end of the PR could I understand what the possible values of "Type" are.

I suggest we use a more prescriptive name and add comments in the code here to clarify.

bharadwaj-aditya (changes requested):

manitgupta (commented):
Will this change the session file structure? If yes, how will we ensure backward compatibility?

sources/common/toddl.go
asthamohta: Because this will be sent to the UI and then it will get converted to an empty string if it is None Literal

schema/schema.go
asthamohta: Let me add a comment, makes sense

sources/mysql/toddl_test.go
asthamohta: uuid as a data type doesn't exist in mysql (only as a generative function) and we are only supporting mysql. Hence a uuid from mysql will never get mapped to spanner. It can only be introduced in spanner

When I add sequences, then I will add tests here

internal/convert.go
```diff
@@ -439,7 +438,7 @@ func (conv *Conv) AddShardIdColumn() {
 			colName := conv.buildColumnNameWithBase(t, ShardIdColumn)
 			columnId := GenerateColumnId()
 			ct.ColIds = append(ct.ColIds, columnId)
-			ct.ColDefs[columnId] = ddl.ColumnDef{Name: colName, Id: columnId, T: ddl.Type{Name: ddl.String, Len: 50}, NotNull: false}
+			ct.ColDefs[columnId] = ddl.ColumnDef{Name: colName, Id: columnId, T: ddl.Type{Name: ddl.String, Len: 50}, NotNull: false, AutoGen: ddl.AutoGenCol{Name: "None", Type: "None"}}
```
VardhanThigle, line 441: [nit] Could we make `AutoGen: ddl.AutoGenCol{Name: "None", Type: "None"}` as a constant?
Something like `DEFAULT_AUTOGEN_COL  ddl.AutoGenCol = ddl.AutoGenCol{Name: "None", Type: "None"}`

schema/schema.go
VardhanThigle: +1 to @Adi8885 

@asthamohta what if we make the variable name itself a bit more expressive, one thing I could think of is `GENERATION_TYPE`. But there could be other possible names too that might work here.

asthamohta:
After looking into making the changes compatible with existing session file, having a default value anything other than an empty string will break backward compatibility. Hence remove "None" and replacing it with an empty string everywhere.
@bharadwaj-aditya, @VardhanThigle, @manitgupta

bharadwaj-aditya (approved):
LGTM - as discussed

PR #818, opened 2024-05-02
docs: resource generation and cli dataflow option docs

asthamohta:


manitgupta (commented):
I think web command also has dataflow-template flag

manitgupta (approved):

PR #819, opened 2024-05-02
feat: Support for UUID UI

asthamohta:
image
image
image
image


ui/src/app/components/add-new-column/add-new-column.component.html
```diff
@@ -21,6 +21,28 @@ <h2>Column Details</h2>
             </mat-form-field>
             <br>
         </div>
+        <div *ngIf="selectedDatatype != '' && autoGenMap[selectedDatatype].length > 1 && srcDbName == 'MySQL'">
```
manitgupta, line 24: `MySQL` should not be hardcoded in the HTML.

Get this from a variable in the component implementation. It should also be list `supportedDbs` or some such with only one DB right now - MySQL

ui/src/app/components/add-new-column/add-new-column.component.html
```diff
@@ -21,6 +21,28 @@ <h2>Column Details</h2>
...
+                    <ng-container *ngFor="let autoGen of autoGenMap[selectedDatatype]">
+                        <mat-option *ngIf="autoGen.GenerationType === ''" [value]="autoGen">
+                            {{ autoGen.Name !== '' ? autoGen.Name : 'None' }}
+                        </mat-option>
+                    </ng-container>
+                    <ng-container *ngIf="processedAutoGenMap[selectedDatatype] as types">
+                        <ng-container *ngFor="let type of types | keyvalue">
+                            <mat-optgroup *ngIf="type.key !== ''" [label]="type.key">
+                                <mat-option *ngFor="let autoGen of type.value" [value]="autoGen">
+                                    {{ autoGen.Name }}
```
manitgupta, line 37: Does this list refresh/update as the user creates new sequences?

Currently this only supports for UUID so this will not arise, but please check if we need to use any observables etc. here - the end goal is that as user is creating new custom sequences, this list should be automatically updated.

ui/src/app/components/add-new-column/add-new-column.component.ts
```diff
@@ -19,9 +21,17 @@ export class AddNewColumnComponent implements OnInit {
   tableId: string = ""
   selectedNull: boolean = true
   dataTypesWithColLen: string[] = ColLength.DataTypes
+  autoGenMap : any = {}
+  selectedAutoGen: AutoGen = {
+    Name: '',
+    GenerationType: ''
+  }
+  processedAutoGenMap: GroupedAutoGens = {};
+  srcDbName: String = localStorage.getItem(StorageKeys.SourceDbName) as string
```
manitgupta, line 30: Is there another reference where we have called `localStorage` here? I have usually seen these values being set and unset in `ngOnInit()`. Please check other references in the code once.

ui/src/app/components/object-detail/object-detail.component.html
```diff
@@ -224,6 +224,36 @@ <h3 class="title">
               </td>
             </ng-container>
 
+            <ng-container matColumnDef="spAutoGen" *ngIf="srcDbName == 'MySQL'">
```
manitgupta, line 227: Same comment on MySQL hardcoding

ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -167,6 +171,11 @@ export class ObjectDetailComponent implements OnInit {
     this.localTableData = JSON.parse(JSON.stringify(this.tableData))
     this.localIndexData = JSON.parse(JSON.stringify(this.indexData))
 
+    if(this.srcDbName == SourceDbNames.MySQL && this.spColspan<7) {
+      this.spDisplayedColumns.splice(2, 0, "spAutoGen")
+      this.spColspan++
+    }
```
manitgupta, line 177: what is `this.spColspan < 7`?

ui/src/app/components/add-new-column/add-new-column.component.html
asthamohta: So this is list is created from getAutoGenMap. I will refresh it by updating the auogenMap value

ui/src/app/components/object-detail/object-detail.component.ts
asthamohta: Its so that when reredering, it doesn't keep adding the column. Modified the condition to just check for the value in the array. I forgot this was easy in js unlike golang 😅

ui/src/app/components/add-new-column/add-new-column.component.ts
asthamohta: In workspace we fetch this value in this way already. Thats the code I referenced

ui/src/app/components/object-detail/object-detail.component.html
asthamohta: Modified

manitgupta (approved):
UI changes are limited to trivial changes to existing components. We have decided existing debt needs to be covered as part of FixIT initiatives.

LGTM!

PR #820, opened 2024-05-08
feat: create sequence backend

asthamohta:


spanner/ddl/ast.go
```diff
@@ -388,13 +391,19 @@ func (agc AutoGenCol) PrintAutoGenCol() string {
 	if agc.Name == constants.UUID && agc.GenerationType == "Pre-defined" {
 		return " DEFAULT (GENERATE_UUID())"
 	}
+	if agc.GenerationType == constants.SEQUENCE {
+		return fmt.Sprintf(" DEFAULT (GET_NEXT_SEQUENCE_VALUE(SEQUENCE %s)) ", agc.Name)
```
manitgupta, line 395: How does this work? Will we allow user to write a custom `SEQUENCE`?

spanner/ddl/ast.go
```diff
@@ -565,3 +584,60 @@ func maxStringLength(s []string) int {
...
+	if seq.SkipRangeMax != "" && seq.SkipRangeMin != "" {
+		options = append(options, fmt.Sprintf(" SKIP RANGE %s %s ", seq.SkipRangeMin, seq.SkipRangeMax))
+	}
+	if seq.StartWithCounter != "" {
+		options = append(options, fmt.Sprintf(" START COUNTER WITH %s ", seq.StartWithCounter))
+	}
+
+	seqDDL := fmt.Sprintf("CREATE SEQUENCE %s", seq.Name)
+	if len(options) > 0 {
+		seqDDL += strings.Join(options, " ")
```
Deep1998, line 639: nit: we are adding triple spacing between options like this. Instead, we could remove the the trailing spaces from the Sprintfs  since Join would anyways add a space in between.

internal/convert.go
```diff
@@ -38,20 +38,22 @@ type Conv struct {
...
+	UniquePKey         map[string][]string     // Maps Spanner table name to unique column name being used as primary key (if needed).
+	Audit              Audit                   `json:"-"` // Stores the audit information for the database conversion
+	Rules              []Rule                  // Stores applied rules during schema conversion
+	IsSharded          bool                    // Flag denoting if the migration is sharded or not
+	ConvLock           sync.RWMutex            `json:"-"` // ConvLock prevents concurrent map read/write operations. This lock will be used in all the APIs that either read or write elements to the conv object.
+	SpRegion           string                  // Leader Region for Spanner Instance
+	ResourceValidation bool                    // Flag denoting if validation for resources to generated is complete
+	UI                 bool                    // Flag if UI interface was used for migration. ToDo: Remove flag after resource generation is introduced to UI
+	SpSequences        map[string]ddl.Sequence // Maps Spanner Sequences to Sequence Schema
+	SrcSequences       map[string]ddl.Sequence // Maps source-DB Sequences to Sequence schema information
```
Deep1998, line 56: Looks like this isnt being used anywhere. Is reading infoschema and generating this a todo?

Deep1998:
> Excellent PR! Love the rigor you've put in in the UTs 👍✌🏻
> 
> I don't have any major concerns, except -
> 
> Please validate backward compatibility of the session file. Can you find a way to write a test which uses an older version of the session file and the newer version (which contains sequences now) and ensure that both can be parsed and both work?
> 
> We should have a method which loads the session file which you can use to write this type of test.

Agreed, we shuold be able to add an IT here using the Spanner emulator similar to our existing ones.

spanner/ddl/ast.go
asthamohta: Screenshot 2024-05-27 at 2 29 47 PM
Screenshot 2024-05-27 at 2 29 19 PM

yes, they will be able to enter the values required to configure the sequence as shown in this UI

spanner/ddl/ast.go
asthamohta: Sure makes sense

internal/convert.go
asthamohta: I am breaking the PRs up, post the UI PR for this I will read the infoschema to check for src sequences and map them to spanner sequences

asthamohta:
@Deep1998 and @manitgupta, added unit tests reading session file into conv(new and old) to ensure an older version is also correctly read. Please check. Thanks

manitgupta (approved):

PR #827, opened 2024-05-28
Make fastExit false for orchestration and schema

manitgupta:
Switches `fastExit` to false so that queued tasks are not discarded.

### DIFF
--- a/conversion/data_from_database.go
+++ b/conversion/data_from_database.go
@@ -211,7 +211,7 @@ func (dd *DataFromDatabaseImpl) dataFromDatabaseForDataflowMigration(migrationPr
 		return common.TaskResult[*profiles.DataShard]{Result: p, Err: err}
 	}
 	r := common.RunParallelTasksImpl[*profiles.DataShard, *profiles.DataShard]{}
-	_, err = r.RunParallelTasks(sourceProfile.Config.ShardConfigurationDataflow.DataShards, 20, asyncProcessShards, true)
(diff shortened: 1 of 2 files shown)


shreyakhajanchi (approved):

conversion/data_from_database.go
```diff
@@ -211,7 +211,7 @@ func (dd *DataFromDatabaseImpl) dataFromDatabaseForDataflowMigration(migrationPr
 		return common.TaskResult[*profiles.DataShard]{Result: p, Err: err}
 	}
 	r := common.RunParallelTasksImpl[*profiles.DataShard, *profiles.DataShard]{}
-	_, err = r.RunParallelTasks(sourceProfile.Config.ShardConfigurationDataflow.DataShards, 20, asyncProcessShards, true)
+	_, err = r.RunParallelTasks(sourceProfile.Config.ShardConfigurationDataflow.DataShards, 20, asyncProcessShards, false)
```
bharadwaj-aditya, line 214: how do we log / recover the failures for folks to retry ?

manitgupta: Currently any exception from a task is logged in STDOUT. 
The logs contain the payload of the API that has failed which can be used to identify the failing shard.

The failing shard(s) will can then be manually retried depending on where the failure happened. For example, if the Dataflow job failed, only the job can be retried, but if the Datastream launch failed, SMT can be rerun with the failed shards(s) to trigger the orchestration for the failed shards again.

PR #828, opened 2024-05-29
feat: create sequence UI

asthamohta:


ui/src/app/components/object-explorer/object-explorer.component.ts
```diff
@@ -174,6 +193,14 @@ export class ObjectExplorerComponent implements OnInit {
     this.sidenav.setDisplayRuleFlag(false)
   }
 
+  openAddSequenceForm(): void {
+    this.dialog.open(AddNewSequenceComponent, {
+      width: '30vw',
```
aksharauke, line 198: Is it okay to hardcode these values? Will it work for all supported resolutions? Is there some 'base' component with some standard size that should be used instead?

ui/src/app/model/auto-gen.ts
```diff
@@ -0,0 +1,8 @@
+export default interface ICreateSequence {
+    Id?: string
+    Name: string
+    SequenceKind: string
+    SkipRangeMin?: string
```
aksharauke, line 5: Why string for the ranges and start values and not integer?

aksharauke (commented):
I assume document update will be in a different PR.
I also suggest to get a review from someone more experienced in Angular since my knowledge there is limited.

ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -1283,6 +1371,29 @@ export class ObjectDetailComponent implements OnInit {
     })
   }
 
+  dropSequence() {
+    let openDialog = this.dialog.open(DropObjectDetailDialogComponent, {
+      width: '35vw',
```
bharadwaj-aditya, line 1376: same comment about dialog boxes

ui/src/app/model/edit-table.ts
```diff
@@ -32,6 +32,14 @@ export interface IIndexData {
   spDesc: boolean | undefined | string
 }
 
+export interface ISequenceData {
+  spSeqName?: string | undefined
+  spSequenceKind?: string | undefined
+  spSkipRangeMin?: string | undefined
+  spSkipRangeMax?: string | undefined
+  spStartWithCounter?: string | undefined
```
bharadwaj-aditya, line 40: how is min different from startWithCounter ?

ui/src/app/services/conversion/conversion.service.ts
```diff
@@ -147,10 +151,47 @@ export class ConversionService {
...
+          isSpannerNode: true,
+          id: seqId,
+          parentId: '',
+          children: [],
+        }
+      }),
+    }
+
+    if (sortOrder === 'asc' || sortOrder === '') {
+      parentNode.children?.sort((a, b) => (a.name > b.name ? 1 : b.name > a.name ? -1 : 0))
```
bharadwaj-aditya, line 180: nit: can be extracted out into a function

ui/src/app/components/object-explorer/object-explorer.component.ts
asthamohta: This is based on precedence where we have previously hardcoded such values. I can replace them with constants but as for if it will be supported by all resolutions, perhaps no, but we expect people to use this application in desktop mode so for that this is is fine

ui/src/app/model/auto-gen.ts
asthamohta: Because if someone does not set it, then the default value will be 0, which is incorrect. hence the value will be parsed in backend to check if it is unset or set.

ui/src/app/app.constants.ts
```diff
@@ -24,6 +24,8 @@ export enum ObjectExplorerNodeType {
   Table = 'tableName',
   Indexes = 'indexes',
   Index = 'indexName',
+  Sequences = 'sequences',
```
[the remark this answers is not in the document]
asthamohta: This is used in showing the objects in the explorer in the left side bar. It encapsulates all the sequences below it, but even if there is no sequence the sequence heading will be shown

ui/src/app/components/add-new-sequence/add-new-sequence.component.html
```diff
@@ -0,0 +1,39 @@
...
+            <input matInput placeholder="Skip Range Min" type="text" formControlName="skipRangeMin" />
+        </mat-form-field>
+        <br>
+        <mat-form-field class="full-width" appearance="outline">
+            <mat-label>Skip Range Maximum</mat-label>
+            <input matInput placeholder="Skip Range Max" type="text" formControlName="skipRangeMax" />
+        </mat-form-field>
+        <br>
+        <mat-form-field class="full-width" appearance="outline">
+            <mat-label>Start With Counter</mat-label>
```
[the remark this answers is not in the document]
asthamohta: Yes, I will raise a separate PR for Documentation

ui/src/app/model/auto-gen.ts
aksharauke: I see, thanks! Does the component not have a method indicating is the value was set/unset? Also when someone actually sets the value 0 - they will be allowed or there is validation? I saw a regex somewhere but that just ensures numeric, which means 0 would be allowed - please correct if am wrong there.

ui/src/app/components/object-detail/object-detail.component.html
```diff
@@ -176,6 +184,13 @@ <h3 class="title">
                       ADD COLUMN
                     </button>
                   </div>
+                  <div *ngIf="!isEditMode && !currentObject.isDeleted && autoGenSupported">
```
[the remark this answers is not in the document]
asthamohta: Sorry I didn't understand, can you elaborate?

For context, only if autoGen is supported(This is true only for source MySql databases as the scope was defined to add support only for it), the add sequence button will be shown

ui/src/app/components/object-detail/object-detail.component.scss
```diff
@@ -62,6 +62,17 @@
   }
 }
 
+.table {
+  height: 100%;
+  background-color: white;
+  .mat-mdc-tab-group {
+    height: 93%;
```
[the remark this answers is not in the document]
asthamohta: I tested and tweaked with these numbers to make sure sequence table and other tables show properly

ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -439,6 +475,15 @@ export class ObjectDetailComponent implements OnInit {
     })
   }
 
+  addNewSequence() {
+    this.dialog.open(AddNewSequenceComponent, {
```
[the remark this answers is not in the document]
asthamohta: Yes, we can but so far we have always defined separate dialogue boxes so i used the same convention. Let me modify this to use a default value

ui/src/app/model/edit-table.ts
asthamohta: Users can choose to set a skip range or a start with counter or both. These are properties defined by the sequence team, hence I have just provided users the option to set them

ui/src/app/model/auto-gen.ts
asthamohta: Yes, 0 is a valid value. The method to check if it is set or unset is part of the backend code which skips this property if the string is empty

ui/src/app/components/object-detail/object-detail.component.html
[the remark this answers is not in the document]
bharadwaj-aditya: Got it

ui/src/app/components/object-detail/object-detail.component.scss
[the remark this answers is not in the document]
bharadwaj-aditya: So the concern here is that depending on screen sizes and resolutions this stuff changes.
So it is better to use a standard CSS for all the elements dependent on type, so that these changes don't go completely random on different screen sizes.

bharadwaj-aditya:
Looks much better now. PTAL at the common CSS bit

ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -173,6 +195,13 @@ export class ObjectDetailComponent implements OnInit {
 
     this.localTableData = JSON.parse(JSON.stringify(this.tableData))
     this.localIndexData = JSON.parse(JSON.stringify(this.indexData))
+    this.localSequenceData = JSON.parse(JSON.stringify(this.sequenceData))
+
+    if (this.srcDbName == SourceDbNames.MySQL && !this.spDisplayedColumns.includes("spAutoGen")) {
+      this.spDisplayedColumns.splice(2, 0, "spAutoGen");
+      this.displayedPkColumns.splice(8, 0, "spAutoGen");
+      this.spColspan++;
+    }
```
shreyakhajanchi, line 204: The if block below can be removed, looks repetitive

ui/src/app/model/auto-gen.ts
```diff
@@ -0,0 +1,8 @@
+export default interface ICreateSequence {
+    Id?: string
```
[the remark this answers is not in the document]
asthamohta: Because Id is generated in the [backend](https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/4af47476d2d61edd9b69f8bfc8c2314f1103468f/internal/helpers.go#L66). The same interface is used for both storing existing sequences and creating new ones. At time of creation sequence Id doesn't exist.

ui/src/app/components/object-detail/object-detail.component.scss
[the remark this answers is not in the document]
asthamohta: Looking back, I think I made a blunder. These values are already set, but the values are same. Should we modify them or should that be a separate effort?:
https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/master/ui/src/app/components/object-detail/object-detail.component.scss#L54-L63

ui/cypress/e2e/spec.cy.ts
```diff
@@ -82,7 +82,7 @@ describe('template spec', () => {
     // Check that workspace is rendered with correct number of tables in Object Viewer
     cy.url().should('include', '/workspace');
     cy.fixture('mysql-config').then((json) => {
-      cy.get('#table-and-index-list').find('tbody tr').should('have.length', json.tableCount + 2);
+      cy.get('#schema-object-list').find('tbody tr').should('have.length', json.tableCount + 3);
     })
```
shreyakhajanchi, line 86: Add a verification that the sequence node has 1 value in it , since its collapsed

ui/src/app/services/conversion/conversion.service.ts
```diff
@@ -161,6 +197,14 @@ export class ConversionService {
     ]
   }
 
+  sortNodeChildren(node: ISchemaObjectNode, sortOrder: string) {
```
shreyakhajanchi, line 200: Please add UTs for sortNodeChildren and getSequenceMapping too

shreyakhajanchi (approved):
Overall LGTM. left few minor comments

PR #831, opened 2024-06-03
Update schema.go 

aasthabharill:
Modified UpdateIndexes method and it's corresponding unit test to fix b/312081666 - When changing order of index columns on Spanner side, it also changes the order of columns on the MySQL side

- [x] Tests pass
  Successfully passed testing on multiple composite indices with varying number of columns and edits.


Deep1998 (approved):
LGTM

shreyakhajanchi (approved):
LGTM

PR #832, opened 2024-06-03
feat: Add support for converting auto-increment to sequence

asthamohta:
image

UI Changes:
- A new type of issue is added
- Src Schema shows Auto-Generated Column

Changes: 
- Auto-Increment in MySql is mapped to a new sequence which is created


cmd/data.go
```diff
@@ -230,7 +230,7 @@ func validateExistingDb(ctx context.Context, spDialect, dbURI string, adminClien
 	}
 	spannerConv := internal.MakeConv()
 	spannerConv.SpDialect = spDialect
-	err = utils.ReadSpannerSchema(ctx, spannerConv, client)
+	err = utils.ReadSpannerSchema(ctx, spannerConv, client, driver)
```
manitgupta, line 233: Why is driver relevant to reading the spanner schema? This interface looks off to me.

manitgupta (changes requested):
Bunch of interfaces are polluted it seems to be by adding `driver`. Can you add the reason why source database type is relevant to functions like `ReadSpannerSchema`, `ValidateExistingDb` etc.?

cmd/data.go
asthamohta: I have made changes to SchemaToSpannerDDL which is eventually called inside all of these functions required the driver. Unfortunately if we look at the tree, the last place where driver info is available around 5-6 functions above the one where I need it

manitgupta: Please evaluate if there is a way better to make these changes without having a ripple effect change on the signatures.
We can also have a call to figure out solutions if required.

sources/common/infoschema.go
```diff
@@ -33,7 +33,7 @@ type InfoSchema interface {
 	GetToDdl() ToDdl
 	GetTableName(schema string, tableName string) string
 	GetTables() ([]SchemaAndName, error)
-	GetColumns(conv *internal.Conv, table SchemaAndName, constraints map[string][]string, primaryKeys []string) (map[string]schema.Column, []string, error)
+	GetColumns(conv *internal.Conv, table SchemaAndName, constraints map[string][]string, primaryKeys []string, tableId string) (map[string]schema.Column, []string, error)
```
manitgupta, line 36: There seems to be a table `SchemaAndName`, what is the purpose of `tableId`?

sources/common/infoschema.go
```diff
@@ -191,6 +192,7 @@ func (is *InfoSchemaImpl) processTable(conv *internal.Conv, table SchemaAndName,
 		return t, fmt.Errorf("couldn't get foreign key constraints for table %s.%s: %s", table.Schema, table.Name, err)
 	}
 
+	table.Id = internal.GenerateTableId()
```
manitgupta, line 195: Where was this Id generated earlier?

asthamohta: Sorry I had to assign the value not generate it. It is generated a few line before this only in the same function

manitgupta (approved):

PR #835, opened 2024-06-12
feat:  add support to configure metadata gcs bucket

asthamohta:


ui/src/app/components/gcs-metadata-details-form/gcs-metadata-details-form.component.html
```diff
@@ -0,0 +1,27 @@
...
+      <mat-label>GCS Name</mat-label>
+      <input matInput placeholder="GCS Name" type="text" formControlName="gcsName"
+        ng-value="TargetDetails.GcsMetadataName" />
+    </mat-form-field>
+    <br>
+    <mat-form-field class="full-width" appearance="outline">
+      <mat-label>GCS Root Path</mat-label>
+      <input matInput placeholder="GCS Root Path" type="text" formControlName="gcsRootPath"
+        ng-value="TargetDetails.GcsMetadataRootPath" />
+    </mat-form-field>
```
shreyakhajanchi, line 17: Rename it to bucket name and path prefix respectively to keep in line with what appears on the pantheon UI

webv2/web.go
```diff
@@ -773,8 +773,17 @@ func getSourceAndTargetProfiles(sessionState *session.SessionState, details type
 	} else {
 		sessionState.Conv.Audit.MigrationRequestId, _ = utils.GenerateName("smt-job")
 		sessionState.Conv.Audit.MigrationRequestId = strings.Replace(sessionState.Conv.Audit.MigrationRequestId, "_", "-", -1)
-		sessionState.Bucket = strings.ToLower(sessionState.Conv.Audit.MigrationRequestId)
-		sessionState.RootPath = "/"
+		if details.TargetDetails.GcsMetadataPath.GcsBucketName != "" {
+			sessionState.Bucket = details.TargetDetails.GcsMetadataPath.GcsBucketName
+			if details.TargetDetails.GcsMetadataPath.GcsBucketRootPath != "" {
+				sessionState.RootPath = details.TargetDetails.GcsMetadataPath.GcsBucketRootPath
```
shreyakhajanchi, line 779: We might need to add / at the end if user's specified root path doesn't end with /

shreyakhajanchi (commented):
Overall looks good. Left minor comments

webv2/web.go
asthamohta: Makes sense. Now that I think about it, technically then we should always add "/" at the end of name of gcs bucket name as well right?

shreyakhajanchi: yes

ui/src/app/components/gcs-metadata-details-form/gcs-metadata-details-form.component.html
```diff
@@ -5,13 +5,13 @@ <h2>GCS Metadata Storage<mat-icon class="configure"
...
-      <mat-label>GCS Name</mat-label>
+      <mat-label>GCS Bucket Name</mat-label>
       <input matInput placeholder="GCS Name" type="text" formControlName="gcsName"
         ng-value="TargetDetails.GcsMetadataName" />
     </mat-form-field>
     <br>
     <mat-form-field class="full-width" appearance="outline">
-      <mat-label>GCS Root Path</mat-label>
+      <mat-label>Path Prefix</mat-label>
       <input matInput placeholder="GCS Root Path" type="text" formControlName="gcsRootPath"
```
shreyakhajanchi, line 15: update the placeholders too

webv2/web.go
```diff
@@ -774,11 +774,9 @@ func getSourceAndTargetProfiles(sessionState *session.SessionState, details type
 		sessionState.Conv.Audit.MigrationRequestId, _ = utils.GenerateName("smt-job")
 		sessionState.Conv.Audit.MigrationRequestId = strings.Replace(sessionState.Conv.Audit.MigrationRequestId, "_", "-", -1)
 		if details.TargetDetails.GcsMetadataPath.GcsBucketName != "" {
-			sessionState.Bucket = details.TargetDetails.GcsMetadataPath.GcsBucketName
+			sessionState.Bucket = details.TargetDetails.GcsMetadataPath.GcsBucketName + "/"
 			if details.TargetDetails.GcsMetadataPath.GcsBucketRootPath != "" {
-				sessionState.RootPath = details.TargetDetails.GcsMetadataPath.GcsBucketRootPath
-			} else {
-				sessionState.RootPath = "/"
+				sessionState.RootPath = details.TargetDetails.GcsMetadataPath.GcsBucketRootPath + "/"
```
shreyakhajanchi, line 779: its better to first check if user has not already added a `/` at the end of the path prefix

shreyakhajanchi (approved):
Please address the mentioned changes before merging

PR #836, opened 2024-06-13
Fetching Default Values for MySQL

NirnayaSindhuSuthari:
Fixes Fetching Default Values from Source Database

-Fetching Default Values from source Database.
-Stores in conv object.

-Screenshot of Default Values in json 
image


sources/mysql/infoschema.go
```diff
@@ -207,12 +207,17 @@ func (isi InfoSchemaImpl) GetColumns(conv *internal.Conv, table common.SchemaAnd
 			ignored.AutoIncrement = true
 		}
 		colId := internal.GenerateColumnId()
+		defaultVal := ""
+		if colDefault.Valid {
+			defaultVal=colDefault.String
+		}
```
Deep1998, line 213: The lint seems to be off, please configure vs code correctly to auto-handle linting.

sources/mysql/infoschema.go
```diff
@@ -207,12 +207,17 @@ func (isi InfoSchemaImpl) GetColumns(conv *internal.Conv, table common.SchemaAnd
 			ignored.AutoIncrement = true
```
Deep1998, line 207: We are still setting ignored.Default if we find a default column, we should remove that.

NirnayaSindhuSuthari: To be done later.

sources/mysql/infoschema.go
NirnayaSindhuSuthari: sorry, will correct it.

PR #843, opened 2024-06-19
fix: deprecate input file pattern

asthamohta:


manitgupta (commented):
So if the `inputFilePattern` is removed and the `gcsPubSubSubscription` is made mandatory, any customer reliant on`inputFilePattern` is already gonna break given `gcsPubSubSubscription` is mandatory now. 

This means they will have to update how they are running. If they are updating, they can remove `inputFilePattern` as well. This should mean we can just remove this parameter altogether from the template instead of marking it as deprecated.

What am I missing? @darshan-sj / @asthamohta

manitgupta (approved):

PR #847, opened 2024-06-25
verify generated files exist in ui/dist

shreyakhajanchi:


.github/scripts/verify-files.sh
```diff
@@ -0,0 +1,45 @@
...
+      MISSING_FILES=true
+    else
+      echo "File exists: $FILE_PATH"
+    fi
+  fi
+done
+
+# Exit with an error if any file was missing
+if [ "$MISSING_FILES" = true ]; then
+  echo "Some files are missing. Verification failed."
```
manitgupta, line 41: Error message is not that helpful - 

1. We should list what the error is, i.e files in index.html do not match the files in the `XYZ` directory.
2. We should list which files are missing so that the user can do something about it.

shreyakhajanchi: Line no 31 prints the files that are not found

manitgupta: Got it, thanks. Maybe just mention that file mismatch issue and something like "did you forget to run ng build?"

shreyakhajanchi: done

manitgupta (approved):

PR #854, opened 2024-07-03
MySQL foreign key action handling at the backend with warning generation

aasthabharill:
Tasks Handled in this PR:
1. Backend changes to incorporate foreign key actions in MySQL end to end including:
   a. Inferring source actions
   b. Conversion of action from source schema to spanner schema
   c. DDL generation 
2. Warning generation in Report 

Unit tests for existing functions were extended and new unit tests were added for new functions.

- [x] Tests pass


sources/common/infoschema.go
```diff
@@ -53,31 +53,33 @@ type SchemaAndName struct {
...
 
 type InfoSchemaInterface interface {
 	GenerateSrcSchema(conv *internal.Conv, infoSchema InfoSchema, numWorkers int) (int, error)
 	ProcessData(conv *internal.Conv, infoSchema InfoSchema, additionalAttributes internal.AdditionalDataAttributes)
 	SetRowStats(conv *internal.Conv, infoSchema InfoSchema)
 	processTable(conv *internal.Conv, table SchemaAndName, infoSchema InfoSchema) (schema.Table, error)
-	GetIncludedSrcTablesFromConv(conv *internal.Conv) (schemaToTablesMap map[string]internal.SchemaDetails, err error) 
+	GetIncludedSrcTablesFromConv(conv *internal.Conv) (schemaToTablesMap map[string]internal.SchemaDetails, err error)
 }
-type InfoSchemaImpl struct {}
```
Deep1998, line 69: It looks like the lint is off. Please install the golint plugin in vscode (or run int manually) so that everytime you save locally, it standardizes the formatting.

internal/mapping.go
```diff
@@ -142,6 +142,58 @@ func ToSpannerForeignKey(conv *Conv, srcFkName string) string {
...
+func ToSpannerOnDelete(conv *Conv, srcTableId string, srcDeleteRule string) string {
+	srcDeleteRule = strings.ToUpper(srcDeleteRule)
+	if srcDeleteRule == "NO ACTION" || srcDeleteRule == "CASCADE" || srcDeleteRule == "" {
+		return srcDeleteRule
+	}
+
+	if conv.SchemaIssues == nil {
+		conv.SchemaIssues = make(map[string]TableIssues)
+	}
+	conv.SchemaIssues[srcTableId] = AddTableLevelIssue(conv.SchemaIssues[srcTableId], ForeignKeyOnDelete)
```
Deep1998, line 167: This AddTableLevelIssue method seems like an overkill for what its doing. 

One can simply do: 
`conv.SchemaIssues[srcTableId].TableLevelIssues = append(conv.SchemaIssues[srcTableId].TableLevelIssues, ForeignKeyOnDelete)
`

internal/mapping.go
```diff
@@ -142,6 +142,58 @@ func ToSpannerForeignKey(conv *Conv, srcFkName string) string {
...
+func ToSpannerOnUpdate(conv *Conv, srcTableId string, srcUpdateRule string) string {
+	srcUpdateRule = strings.ToUpper(srcUpdateRule)
+	if srcUpdateRule == "NO ACTION" || srcUpdateRule == "" {
+		return srcUpdateRule
+	}
+
+	if conv.SchemaIssues == nil {
+		conv.SchemaIssues = make(map[string]TableIssues)
+	}
+	conv.SchemaIssues[srcTableId] = AddTableLevelIssue(conv.SchemaIssues[srcTableId], ForeignKeyOnUpdate)
```
Deep1998, line 193: Same here.

internal/mapping.go
```diff
@@ -142,6 +142,58 @@ func ToSpannerForeignKey(conv *Conv, srcFkName string) string {
...
+// Since only MySQL and PostgreSQL have this functionality
+// as of yet, for other sources OnDelete fields are
+// kept empty i.e. "" is mapped to ""
+//
+// For all source actions converted to a different action,
+// an issue is appended to it's TableLevelIssues to
+// generate a warning message for the user
+func ToSpannerOnDelete(conv *Conv, srcTableId string, srcDeleteRule string) string {
+	srcDeleteRule = strings.ToUpper(srcDeleteRule)
+	if srcDeleteRule == "NO ACTION" || srcDeleteRule == "CASCADE" || srcDeleteRule == "" {
```
Deep1998, line 160: Please add a todo to add Postgres, can be removed in the followup pr.

webv2/api/schema.go
```diff
@@ -362,12 +362,15 @@ func GetTableWithErrors(w http.ResponseWriter, r *http.Request) {
 	defer sessionState.Conv.ConvLock.RUnlock()
 	var tableIdName []types.TableIdAndName
 	for id, issues := range sessionState.Conv.SchemaIssues {
-		if len(issues.TableLevelIssues) != 0 {
-			t := types.TableIdAndName{
-				Id:   id,
-				Name: sessionState.Conv.SpSchema[id].Name,
+		for _, issue := range issues.TableLevelIssues {
+			if reports.IssueDB[issue].Severity == reports.Errors {
+				// if issue == internal.RowLimitExceeded {
```
Deep1998, line 367: Please remove this comment.

Deep1998 (approved):
Overall, very well written code with great test coverage across MySQL dump and PG dialect as well. 

I also see some new tests added for methods that you only modified. Thanks for going above and beyond.

LGTM with minor comments.

internal/mapping.go
manitgupta, line 160: In MYSQL `RESTRICT` and `NO ACTION` mean the same thing. 

Reference - https://stackoverflow.com/questions/39338474/whats-the-different-between-restrict-and-no-action-in-mysql

`RESTRICT` in MySQL should also get mapped to `NO ACTION` in Spanner,

internal/mapping.go
```diff
@@ -142,6 +142,58 @@ func ToSpannerForeignKey(conv *Conv, srcFkName string) string {
...
+		return srcDeleteRule
+	}
+
+	if conv.SchemaIssues == nil {
+		conv.SchemaIssues = make(map[string]TableIssues)
+	}
+	conv.SchemaIssues[srcTableId] = AddTableLevelIssue(conv.SchemaIssues[srcTableId], ForeignKeyOnDelete)
+	return "NO ACTION"
```
manitgupta, line 168: We should define FKAs as `const` somewhere `NO ACTION`, `CASCADE` etc. instead of hardcoding strings.

internal/mapping.go
aasthabharill: All actions other than CASCADE are mapped to NO ACTION

internal/mapping.go
aasthabharill: incorporated this change in latest commit

shreyakhajanchi (approved):
LGTM. Great work on adding good test coverage and quality code with comments.

PR #855, opened 2024-07-04
Advanced transformation integration for forward migration

shreyakhajanchi:
Tests run:
1. Minimal downtime migration via sharded flow
2. Minimal downtime migration via non-sharded flow

UI:


streaming/streaming.go
```diff
@@ -763,6 +766,13 @@ func LaunchDataflowJob(ctx context.Context, migrationProjectId string, targetPro
 			KmsKeyName:            dataflowCfg.KmsKeyName,
 		},
 	}
+
+	if dataflowCfg.CustomClassName != "" && dataflowCfg.CustomJarPath != "" {
+		launchParameters.Parameters["transformationJarPath"] = dataflowCfg.CustomJarPath
```
VardhanThigle, line 771: Do we want to add some level of validaion in the backend to fail fast (like gs paths is formatted correctly, if only one option is set, do we throw an error etc?

In the most ideal case we could delegate to dataflow for validation but in case we are spawnning more than one DF job for a sharded migration, it might make sense to fail fast in SMT too.

VardhanThigle (approved):

streaming/streaming.go
shreyakhajanchi: Makes sense. Added gcs file path validation in UI and also in dataflow, and is one option set in backend.

PR #861, opened 2024-07-08
test: Add golden tests to PostgreSQL

thiagotnunes:
Adds golden tests for PostgreSQL schema conversion (using pg_dump).


testing/postgres/golden_test.go
```diff
@@ -0,0 +1,89 @@
...
+func TestGoldens(t *testing.T) {
+	logger.Log = zap.NewNop()
+	entries, err := os.ReadDir(GoldenTestsDir)
+	if err != nil {
+		t.Fatalf("error when reading entries of golden tests dir %s: %s", GoldenTestsDir, err)
+	}
+
+	schemaToSpanner := common.SchemaToSpannerImpl{}
+
+	for _, entry := range entries {
```
manitgupta, line 51: This would execute everything as one test (maybe that's the intention? - Golden test "pass" or "fail"?)

I personally would prefer these tests to be part of the `golden_test.go` suite of tests, but which execute disjointly (and also in parallel?). 

The typical Golang way to achieve this is creating a test struct like so - 

```
tc := []struct {
            ........
	}{
              { // test data 1},
              { // test data 2},
        }
for _, tc := range tc {
}
```

This ensures that each test from the `GoldenTestDir` you read is executed independently, I would also recommend that you **name** the test in the `create_test.test` and then use that in the struct to name the test for easy debugging.

Reference - 

https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/1591ca5ee54ccbbe0a5083ffbf9323bc5c54c087/sources/common/utils_test.go#L142

thiagotnunes: Got it, but I would like to have a way to separate data from the test. What is the idiomatic way to do this in Golang?

I believe this makes it easier to add test cases, as one does not have to parse through the code to do that.

I don't think parallelisation would be a problem here, unless you think these tests would take too long?

manitgupta (approved):

testing/postgres/golden_test.go
thiagotnunes: Discussed offline, resolving for now

PR #863, opened 2024-07-08
Postgres and UI support for Foreign Key Actions 

aasthabharill:
1. Extending support for FK actions to Postgres along with unit tests
2. UI changes to display FK Actions on 'Foreign Key' tab
    a. Displays OnDelete and OnUpdate rule for both source and Spanner schema
    b. This is conditional - only for MySQL and PostgreSQL, all other sources will show no new columns for FK Actions

- [ ] Tests pass


ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -1120,6 +1143,8 @@ export class ObjectDetailComponent implements OnInit {
         fk.spColIds = []
         fk.spReferColumnIds = []
         fk.spReferTableId = ''
+        fk.spOnDelete = '' //check if this needs to be changed
```
Deep1998, line 1146: Remove this comment if not needed

Deep1998 (commented):
LGTM

shreyakhajanchi (approved):

PR #865, opened 2024-07-09
Foreign key actions - warnings for unsupported sources (oracle,etc) and tests for reports

aasthabharill:
- Sources other than MySQL and PostgreSQL will now display warning that Foreign Key Actions are not supported if a foreign key exists ([screenshot](https://screenshot.googleplex.com/7FWCZtKbwmWVfJp))
- Unit tests for report generation modified to test FK Actions too (for MySQL and Postgres)

- [x] Tests pass


internal/convert.go
```diff
@@ -126,6 +126,7 @@ const (
 	ForeignKeyOnDelete
 	ForeignKeyOnUpdate
 	SequenceCreated
+	ForeignKeyActionsNotSupported
```
Deep1998, line 129: nit: I think this should be ForeignKeyActionNotSupported? As in singular to be consistent with the wordings.

internal/mapping.go
```diff
@@ -149,25 +149,43 @@ func ToSpannerForeignKey(conv *Conv, srcFkName string) string {
...
 // generate a warning message for the user
+//
+// Since only MySQL and PostgreSQL have this functionality
+// as of yet, for other sources OnDelete fields are
+// kept empty i.e. "" is mapped to "" and an issue is appended
+// to generate warning message (prints only once for each table with FK Actions)
 func ToSpannerOnDelete(conv *Conv, srcTableId string, srcDeleteRule string) string {
 	srcDeleteRule = strings.ToUpper(srcDeleteRule)
-	if srcDeleteRule == constants.NO_ACTION || srcDeleteRule == constants.CASCADE || srcDeleteRule == "" {
+	if srcDeleteRule == constants.NO_ACTION || srcDeleteRule == constants.CASCADE {
```
Deep1998, line 162: nit: Coment related to a previous change, we should actually keep the constant variable name ot FK_NO_ACTION , FK_CASCADE or FOREIGN_KEY_NO_ACTION (thats upto you) but itd be nice if we can mkeep the naming specific

internal/mapping.go
```diff
@@ -149,25 +149,43 @@ func ToSpannerForeignKey(conv *Conv, srcFkName string) string {
...
 		return srcDeleteRule
 	}
-
 	if conv.SchemaIssues == nil {
 		conv.SchemaIssues = make(map[string]TableIssues)
 	}
+	tableIssues := conv.SchemaIssues[srcTableId]
+
+	if srcDeleteRule == "" {
```
Deep1998, line 170: Add a comment saying this is empty only for unsupported sources. Similarly for onUpdate method.

internal/mapping.go
```diff
@@ -149,25 +149,43 @@ func ToSpannerForeignKey(conv *Conv, srcFkName string) string {
...
+		for i := 0; i < len(tableIssues.TableLevelIssues); i++ {
+			if tableIssues.TableLevelIssues[i] == ForeignKeyActionsNotSupported {
+				flag = true
+				break
+			}
+		}
+		if !flag {
+			conv.SchemaIssues[srcTableId] = TableIssues{
+				TableLevelIssues:  append(tableIssues.TableLevelIssues, ForeignKeyActionsNotSupported),
+				ColumnLevelIssues: tableIssues.ColumnLevelIssues}
```
Deep1998, line 182: This can be made more readable using slices.Contains:
 ```

if (!slices.Contains(tableIssues.TableLevelIssues, ForeignKeyActionsNotSupported)){
conv.SchemaIssues[srcTableId] = TableIssues{
				TableLevelIssues:  append(tableIssues.TableLevelIssues, ForeignKeyActionsNotSupported),
				ColumnLevelIssues: tableIssues.ColumnLevelIssues}
}
```

File: internal/reports/report_helpers.go [hunk 5]
```diff
@@ -546,10 +556,11 @@ var IssueDB = map[internal.SchemaIssue]struct {
...
-	internal.ForeignKeyOnDelete:    {Brief: "Spanner supports only ON DELETE CASCADE/NO ACTION", Severity: warning, Category: "FOREIGN_KEY_ACTIONS"},
-	internal.ForeignKeyOnUpdate:    {Brief: "Spanner supports only ON UPDATE NO ACTION", Severity: warning, Category: "FOREIGN_KEY_ACTIONS"},
+	internal.ArrayTypeNotSupported:         {Brief: "Array datatype migration is not fully supported. Please validate data after data migration", Severity: warning, Category: "ARRAY_TYPE_NOT_SUPPORTED"},
+	internal.SequenceCreated:               {Brief: "Auto Increment has been converted to Sequence, set Skipped Range or Start with Counter to avoid duplicate value errors", Severity: warning, Category: "SEQUENCE_CREATED"},
+	internal.ForeignKeyOnDelete:            {Brief: "Spanner supports only ON DELETE CASCADE/NO ACTION", Severity: warning, Category: "FOREIGN_KEY_ACTIONS"},
+	internal.ForeignKeyOnUpdate:            {Brief: "Spanner supports only ON UPDATE NO ACTION", Severity: warning, Category: "FOREIGN_KEY_ACTIONS"},
+	internal.ForeignKeyActionsNotSupported: {Brief: "Spanner doesn't support foreign key action migration", Severity: warning, Category: "FOREIGN_KEY_ACTIONS"},
```
Deep1998, line 563: The message is misleading since Spanner does not support any migration, it should be FK Action migration not supported for source oracle/ onyl support for mysql/pg by SMT, something along these lines.

webv2/api/schema_test.go
```diff
@@ -258,7 +258,7 @@ func TestGetDDL(t *testing.T) {
 						"c2": ddl.ColumnDef{Name: "b", T: ddl.Type{Name: ddl.Int64}, NotNull: true},
 						"c3": ddl.ColumnDef{Name: "c", T: ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, NotNull: true}},
 					PrimaryKeys: []ddl.IndexKey{ddl.IndexKey{ColId: "c1", Desc: false}},
-					ForeignKeys: []ddl.Foreignkey{ddl.Foreignkey{Name: "fk1", ColIds: []string{"c1"}, ReferTableId: "t2", ReferColumnIds: []string{"c4"}, OnDelete: constants.CASCADE, OnUpdate: constants.NO_ACTION}},
+					ForeignKeys: []ddl.Foreignkey{ddl.Foreignkey{Name: "fk1", ColIds: []string{"c1"}, ReferTableId: "t2", ReferColumnIds: []string{"c4"}, OnDelete: constants.FK_CASCADE, OnUpdate: constants.FK_NO_ACTION}},
```
shreyakhajanchi, line 261: I don't think we seem to be handling the case where the foreign key from the table gets dropped, in that case the warning will also need to be dropped if its the only foreign key with FK actions

aasthabharill: Was planning to handle this with the interleaving bit as it directly calls DropForeignKey - is that okay?

aasthabharill: For the UI part, the warning are dropped from the tab and the count is also lowered, just that internally the issue list might have duplicate issues; had discussed this with deep and he said to de-prioritize it so kept it for the end

shreyakhajanchi: okay sure, sounds good

shreyakhajanchi (approved):
LGTM

Deep1998:
LGTM

PR #871, opened 2024-07-16
FK Actions: Handling interleaving & dropping issues correctly when dropping FKs 

aasthabharill:
1. Interleaving support:
    a. Propagate foreign key action to interleaved table from foreign key
    b. Ensure DDL generated includes on delete action
Major Changes: Instead of storing parentId as string in Conv, it now stores ParentTable struct containing parent id and on delete action. 

2. Drop issues when foreign key is dropped
    a. Foreign key actions added TableLevelIssues to SchemaIssues of Conv that need to be dropped (both internally and from UI and reports) when foreign key is dropped, while ensuring warnings raised by other FKs remain intact

3. Unit test coverage added to all code changes (Primarily testing fk->interleaved action migration and SchemaIssues)


webv2/table/review_table_schema_test.go
```diff
@@ -1715,7 +1715,7 @@ func TestReviewTableSchema(t *testing.T) {
 							"c2": {Name: "b", Id: "c2", T: ddl.Type{Name: ddl.String, Len: 6}, AutoGen: ddl.AutoGenCol{Name: "UUID", GenerationType: "Pre-defined"}},
 						},
 						PrimaryKeys: []ddl.IndexKey{{ColId: "c1"}},
-						ParentId:    "t2",
+						ParentTable: ddl.InterleavedParent{Id: "t2", OnDelete: constants.FK_CASCADE},
```
shreyakhajanchi, line 1718: Add cases where source action is unsupported in spanner

webv2/table/update_table_schema_test.go
```diff
@@ -622,7 +622,7 @@ func TestUpdateTableSchema(t *testing.T) {
 							"c2": {Name: "b", Id: "c2", T: ddl.Type{Name: ddl.String, Len: 6}},
 						},
 						PrimaryKeys: []ddl.IndexKey{{ColId: "c1"}},
-						ParentId:    "t2",
+						ParentTable: ddl.InterleavedParent{Id: "t2", OnDelete: constants.FK_CASCADE},
```
shreyakhajanchi, line 625: Add cases where source action is unsupported in spanner

common/utils/utils.go
```diff
@@ -521,11 +521,12 @@ func ReadSpannerSchema(ctx context.Context, conv *internal.Conv, client *sp.Clie
 		conv.Unexpected(fmt.Sprintf("error trying to fetch interleave table info from schema: %v", err))
 	}
 	// Assign parents if any.
+	// TODO: Assign ParentTable.OnDelete too
```
shreyakhajanchi, line 524: why have we marked this as todo and not added it here? Same comment for the other TODO

spanner/ddl/ast.go
```diff
@@ -282,6 +282,14 @@ type Foreignkey struct {
 	OnUpdate       string
 }
 
+// InterleavedParent encodes the following DDL definition:
+//
+//	INTERLEAVE IN PARENT parent_name ON DELETE delete_rule
```
shreyakhajanchi, line 287: Indentation of the comment is off

webv2/utilities/utilities.go
```diff
@@ -236,15 +246,38 @@ func RemovePk(slice []ddl.IndexKey, s int) []ddl.IndexKey {
...
 	for i, fk := range slice {
 		if fk.Id == fkId {
 			pos = i
+
+			// Remove foreign key action conversion warnings for dropped FK
+			//
+			// Issues ForeignKeyOnDelete and ForeignKeyOnUpdate might be added to SchemaIssues[tableId] multiple times
+			// due to existance of multiple FKs. We only delete one instance of these issues to maintain
+			// correct warnings for other FKs
+			srcFk, err := internal.GetSrcFkFromId(srcSchema.ForeignKeys, fk.Id)
```
shreyakhajanchi, line 263: Are we adding multiple warnings with same message and no identifier in case of multiple unsupported FK actions?

webv2/api/schema.go
```diff
@@ -1028,11 +1027,15 @@ func parentTableHelper(tableId string, update bool) *types.TableInterleaveStatus
...
+			if update && sp.ParentTable.Id == "" && colIdNotInOrder == "" {
 				usedNames := sessionState.Conv.UsedNames
 				delete(usedNames, strings.ToLower(sp.ForeignKeys[i].Name))
-				sp.ParentId = refTableId
-				sp.ForeignKeys = utilities.RemoveFk(sp.ForeignKeys, sp.ForeignKeys[i].Id)
+				sp.ParentTable.Id = refTableId
+				sp.ParentTable.OnDelete = onDelete
+				sp.ForeignKeys, err = utilities.RemoveFk(sp.ForeignKeys, sp.ForeignKeys[i].Id, sessionState.Conv.SrcSchema[tableId], tableId)
+				if err != nil {
+					continue
```
Deep1998, line 1037: Do we want to silently ignore this error?

Deep1998:
Overall LGTM, please resolve Shreya's comments before merging. Thanks for the changes, this was a rather tricky scenario handling interleaved tables and the issue management.

spanner/ddl/ast.go
aasthabharill: idk why but comments above earlier functions followed this indentation so I did it too, lmk if it still needs to be changed

webv2/utilities/utilities.go
aasthabharill: The warning body contains all the details of the foreign key and are not repeated, but internally the issue is added without this information so there's no way to differentiate

spanner/ddl/ast_test.go
```diff
@@ -187,6 +200,17 @@ func TestPrintCreateTable(t *testing.T) {
 			"interleaved",
 			false,
 			t2,
+			"CREATE TABLE mytable (\n" +
+				"	col1 INT64 NOT NULL ,\n" +
+				"	col2 STRING(MAX),\n" +
+				"	col3 BYTES(42),\n" +
+				") PRIMARY KEY (col1 DESC),\n" +
+				"INTERLEAVE IN PARENT  ON DELETE CASCADE",
```
Deep1998, line 208: Hmm the empty parent table name looks odd, this double spacing looks like an error at first.

Can we genearte a dummy spSchema so that the parent table name is also populated correctly?

Deep1998 (approved):

PR #875, opened 2024-07-19
Documentation for custom transformation

shreyakhajanchi:
https://shreyakhajanchi.github.io/harbourbridge/


docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
+---
+layout: default
+title: Custom Transformation
+parent: Minimal downtime migrations
+nav_order: 4
+---
+
+# Minimal downtime migrations for MySQL
```
manitgupta, line 8: Does the heading need to change?

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+{: .no_toc }
+
+For cases where a user wants to handle a custom transformation logic, they need to specify the following parameters in the [Datastream To Spanner](https://github.com/GoogleCloudPlatform/DataflowTemplates/tree/main/v2/datastream-to-spanner) template - a GCS path that points to a custom jar, fully classified custom class name of the class containing custom transformation logic and custom parameters which might be used by the jar to invoke custom logic to perform transformation.
```
manitgupta, line 11: Maybe we can make this a list and call them pre-reqs to do transformations?

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+
+Steps to perfrom customization:
+1. Implement custom transformation logic for forward migration in the [toSpannerRow](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-custom-shard/src/main/java/com/custom/CustomTransformationFetcher.java#L42) method of the **CustomTransformationFetcher.java**. Details of the MigrationTransformationRequest class can be found [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-migrations-sdk/src/main/java/com/google/cloud/teleport/v2/spanner/utils/MigrationTransformationRequest.java).
+2. Build the [JAR](https://github.com/GoogleCloudPlatform/DataflowTemplates/tree/main/v2/spanner-custom-shard) and upload the jar to GCS
```
manitgupta, line 15: Instructions to build the JAR?

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+3. Invoke the datastream-to-spanner template by passing the custom jar path and custom class path.
```
manitgupta, line 16: Sample command? This is not very clear, do you mean via SMT?

docs/reverse-replication/ReverseReplicationUserGuide.md
```diff
@@ -378,10 +381,53 @@ In order to make it easier for users to customize the shard routing logic, the [
...
+2. [MigrationTransformationResponse](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-migrations-sdk/src/main/java/com/google/cloud/teleport/v2/spanner/utils/MigrationTransformationResponse.java) contains the following information - 
+    - responseRow - It is a map where key is the source column name and value is source column value.
+    - isEventFiltered - If set to true, event will be skipped and not written to source.
+3. Values in the response row should be exactly in the format compatible with source schema, which means users also need to enclose string values in **single quotes** as they would normally do in an INSERT statement.
+4. Please throw **InvalidTransformationException** in case of any error while processing a particular event in custom jar.
+
+Here is a table that details the source data type for **MySQL**, its corresponding request row object type, spanner datatype and the expected response format:
+|Spanner datatype | Source datatype         | Request object type                       | Response format                                                                        |
```
manitgupta, line 409: Does this exist somewhere else? Can we link?

docs/minimal/transformations.md
manitgupta, line 8: Seems like this needs to be broken down into two parts - 

1. How to Implement a custom transform. (Implement, Unit test?)
3. How to use the custom transform in a pipeline (forward or reverse).

Currently does not read well IMO.

Deep1998: We should also put the following contents in a separate page and make it agnostic of minimal downtime migration (because bulk docs will refer to these) :
- Building a JAR
- Parameter details
- Java Object type mappings
Anything else that is not minimal downtime specific

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+nav_order: 4
+---
+
+# Minimal downtime migrations for MySQL
+{: .no_toc }
+
+For cases where a user wants to handle a custom transformation logic, they need to specify the following parameters in the [Datastream To Spanner](https://github.com/GoogleCloudPlatform/DataflowTemplates/tree/main/v2/datastream-to-spanner) template - a GCS path that points to a custom jar, fully classified custom class name of the class containing custom transformation logic and custom parameters which might be used by the jar to invoke custom logic to perform transformation.
+
+Steps to perfrom customization:
+1. Implement custom transformation logic for forward migration in the [toSpannerRow](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-custom-shard/src/main/java/com/custom/CustomTransformationFetcher.java#L42) method of the **CustomTransformationFetcher.java**. Details of the MigrationTransformationRequest class can be found [here](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-migrations-sdk/src/main/java/com/google/cloud/teleport/v2/spanner/utils/MigrationTransformationRequest.java).
```
Deep1998, line 14: Explicitly call out implementing toSourceRow() may be skipped if not using reverse replication?

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+3. Invoke the datastream-to-spanner template by passing the custom jar path and custom class path.
+4. If any custom parameters are needed in the custom transformation logic, they can be passed via the *customParameters* input to the template. These parameters will be passed to the *init* method of the custom class. The *init* method is invoked once per worker setup.
+
+Implementation details for custom transformation:
+1. [MigrationTransformationRequest](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-migrations-sdk/src/main/java/com/google/cloud/teleport/v2/spanner/utils/MigrationTransformationRequest.java) contains the following information - 
+    - tableName - Name of the source table to which the event belongs to.
+    - shardId - Logical shard id of the record.
+    - eventType - The event type can either be INSERT, UPDATE-INSERT, UPDATE, UPDATE_DELETE or DELETE. Please refer to the [datastream documentation](https://cloud.google.com/datastream/docs/events-and-streams) for more details.
+    - requestRow - It is a map where key is the source column name and value is source column value.
```
Deep1998, line 24: Map of type `Map<java.lang.String, java.lang.Object>` where ...

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+2. [MigrationTransformationResponse](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-migrations-sdk/src/main/java/com/google/cloud/teleport/v2/spanner/utils/MigrationTransformationResponse.java) contains the following information - 
+    - responseRow - It is a map where key is the spanner column name and value is spanner column value.
```
Deep1998, line 26: We should Mention explicitly that the the toSpannerRow method expects this struct to be returned.

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+    - isEventFiltered - If set to true, event will be skipped and not written to spanner.
+3. Values in the response row should be of same datatype as the spanner schema.
```
Deep1998, line 28: This point should be modified to something like:

Values in the response row must be compatible with their corresponding Spanner column types. The following table outlines the recommended Java types for each Spanner data type to ensure successful data insertion:
<INSERT response table>

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+4. Please throw **InvalidTransformationException** in case of any error while processing a particular event in custom jar.
+
+Here is a table that details the source data type for **MySQL**, its corresponding request row object type, spanner datatype and the expected response format:
+| Source datatype         | Request object type                 | Spanner datatype      | Response format                                                                        |
+|-------------------------|-------------------------------------|-----------------------|----------------------------------------------------------------------------------------|
```
Deep1998, line 33: We should break this into two tables:
-  mapping source to request object type
-  mapping spanner to response object type

primarily because a source type can map to different spanner types.
And also, this way we can put this first table right with the parameter definition for MigrationTransformationRequest.requestRow and response row respectively.

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+| TEXT                    | String                              | STRING     | String                         |
+| ENUM                    | String                              | STRING     | String                               |
+| BLOB                    | String (hex encoded)             | BYTES     | Binary String                                                                          |
+| BINARY                  | String (hex encoded)             | BYTES     | Binary String                                                                          |
+| BIT                     | Long             | BYTES     | Long                                                                          |
+| DATE                    | String (Format: yyyy-MM-dd))        | DATE     | String( Format: yyyy-MM-dd)            |
+| DATETIME                | String (ex: 2024-01-01T12:34:56Z)   | TIMESTAMP     | String                                                                                 |
+| TIMESTAMP               | String (ex: 2024-01-01T12:34:56Z)   | TIMESTAMP     | String                                                                                 |
+
```
Deep1998, line 51: It seems like we are missing certain types for mysql:

This is a fuller list here:
https://cloud.google.com/datastream/docs/unified-types#map-mysql

This doc still misses unsupported spatial types:
"spatial_linestring",
            "spatial_multilinestring",
            "spatial_multipoint",
            "spatial_multipolygon",
            "spatial_point",
            "spatial_polygon"

For these we can just say expect null since not supported.

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+Here is a table that details the source data type for **MySQL**, its corresponding request row object type, spanner datatype and the expected response format:
+| Source datatype         | Request object type                 | Spanner datatype      | Response format                                                                        |
+|-------------------------|-------------------------------------|-----------------------|----------------------------------------------------------------------------------------|
+| TINYINT                 | Long                              | INT64     | Long                                                                                 |
+| INT                     | Long                              | INT64     | Long                                                                                 |
+| BIGINT                  | Long                              | INT64    | Long                                                                                 |
+| TIME                    | Long ([time-micros](https://avro.apache.org/docs/current/specification/_print/#time-microsecond-precision) ex: 45296000000 for 12:34:56) | STRING      | Long|
```
Deep1998, line 37: Note: b/349257952 tracks this which would cause this mapping to change, for both source and destination.

docs/minimal/transformations.md
Deep1998, line 1: We shuold also document somewhere that not all column values need to be returned, only the ones returned will be updated. The rest will be migrated as normal. And the user can also add newer columns not there in source (but there on spanner).

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+| DOUBLE                  | Double                          | FLOAT64     | Double                                                                                 |
+| DECIMAL                 | String                              | NUMERIC      | String                                                                                 |
+| BOOLEAN                 | Long                | BOOLEAN     | Long                                                                                 |
+| TEXT                    | String                              | STRING     | String                         |
+| ENUM                    | String                              | STRING     | String                               |
+| BLOB                    | String (hex encoded)             | BYTES     | Binary String                                                                          |
+| BINARY                  | String (hex encoded)             | BYTES     | Binary String                                                                          |
```
Deep1998, line 46: ~Binary string~ Hex Encoded String?

docs/minimal/transformations.md
```diff
@@ -0,0 +1,53 @@
...
+| BIT                     | Long             | BYTES     | Long                                                                          |
+| DATE                    | String (Format: yyyy-MM-dd))        | DATE     | String( Format: yyyy-MM-dd)            |
+| DATETIME                | String (ex: 2024-01-01T12:34:56Z)   | TIMESTAMP     | String                                                                                 |
+| TIMESTAMP               | String (ex: 2024-01-01T12:34:56Z)   | TIMESTAMP     | String                                                                                 |
```
Deep1998, line 50: Add example for response type as well, mention UTC format.

docs/minimal/transformations.md
shreyakhajanchi: Created a different page and changed the heading

docs/minimal/transformations.md
shreyakhajanchi: Added steps to implement custom transformation

docs/minimal/transformations.md
shreyakhajanchi: Broke it down into multiple sections and moved it to a common place

docs/minimal/transformations.md
shreyakhajanchi: Grouped into request and response, so shouldn't be required now IMHO

docs/minimal/transformations.md
shreyakhajanchi: Makes sense, done

docs/minimal/transformations.md
shreyakhajanchi: Added

docs/minimal/transformations.md
shreyakhajanchi: Will update the documentation once the bug is fixed

docs/transformations/CustomTransformation.md
```diff
@@ -0,0 +1,212 @@
...
+#### Request
+[MigrationTransformationRequest](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-migrations-sdk/src/main/java/com/google/cloud/teleport/v2/spanner/utils/MigrationTransformationRequest.java) contains the following information - 
+- tableName - Name of the source table to which the event belongs to.
+- shardId - Logical shard id of the record.
+- eventType - The event type can either be INSERT, UPDATE-INSERT, UPDATE, UPDATE_DELETE or DELETE. Please refer to the [datastream documentation](https://cloud.google.com/datastream/docs/events-and-streams) for more details.
+- requestRow - It is a map of type `Map<java.lang.String, java.lang.Object>` where key is the source column name and value is source column value.
+
+The following table outlines the Java object type that will be sent in the request to the `toSpannerRow` for each **MySQL** datatype:
+
+| MYSQL datatype | Java object type    |
```
Deep1998, line 49: Thanks, this looks much better.

nit: for datetime and timestamp types, instead of eg, maybe we should say format to indicate utc is always the format.

docs/transformations/CustomTransformation.md
```diff
@@ -0,0 +1,212 @@
...
+    public MigrationTransformationResponse toSpannerRow(MigrationTransformationRequest request throws InvalidTransformationException {
+    return new MigrationTransformationResponse(null, false);
+    }
+    ```
+3. If any custom parameters are needed in the custom transformation logic, they can be passed to the *init* method of the custom class. The *init* method is invoked once per worker setup.
+3. Please test the modified code by writing unit and cross functional tests in [CustomTransformationFetcherTest.java](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/main/v2/spanner-custom-shard/src/test/java/com/custom/CustomTransformationFetcherTest.java) 
+4. Build the `spanner-custom-shard` module by running the below commands:
+    ```
+    cd v2/spanner-custom-shard
+    mvn install
```
Deep1998, line 184: will the user need to do an mvn install on the parent directory for htis to work? I remember one had to do this one-time mvn install (Even if it fails its okay as long as a certain module gets built) on parent before building a template. Not sure if it is a requirement on this as well.

docs/troubleshoot/minimal.md
```diff
@@ -50,29 +50,38 @@ Migration progress can be tracked by monitoring the Dataflow job and following c
...
+| Metric Name                                   | Description                                                                                                                      |
+|-----------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------|
+| Successful events                             | Total number of events successfully processed and applied to Spanner database                                                    |
+| Retryable errors                              | The count of events that were errored out but will be retried                                                                    |
+| Total permanent errors                        | The number of events that are errored out with non-retriable errors in addition to the number of errors after exhausting retries |
+| Conversion errors                             | Number of events that could not be converted to Spanner.This is a permanent error category.                                      |
```
Deep1998, line 58: nit: Spanner. This - Add space

Deep1998 (commented):
LGTM with minor comments

docs/transformations/CustomTransformation.md
shreyakhajanchi: I have tested this on a fresh cloned repo, it works. It doesn't have dependency on parent directories that use to cause this issue in the past

Deep1998 (approved):

PR #877, opened 2024-07-19
Default Values Draft Struct Changes

NirnayaSindhuSuthari:
-Added DefaultValue field of DefaultValue  type in the column struct.
-DefaultValue struct has IsPresent and Value fields.


NirnayaSindhuSuthari:
-This is a draft kindly ignore it.

PR #881, opened 2024-07-23
feature: translate numeric primary keys to text in PostgreSQL

thiagotnunes:
PostgreSQL dialect does not support numeric primary keys. We instead translate the numeric components of the primary key to text and warn users.


webv2/utilities/get_type.go
```diff
@@ -33,22 +33,23 @@ func GetType(conv *internal.Conv, newType, tableId, colId string) (ddl.CreateTab
 
 	sp := conv.SpSchema[tableId]
 	srcCol := conv.SrcSchema[tableId].ColDefs[colId]
+	isPk := common.IsPrimaryKey(colId, conv.SrcSchema[tableId])
```
VardhanThigle, line 36: Would it make sense for this check to be decided by the spanner schema instead ? Not sure if it makes a difference in the current state SMT, but just in case we support any kind of transformations on primary keys it might change.

VardhanThigle (approved):

webv2/utilities/get_type.go
thiagotnunes: This check is done before we have the Spanner Schema (it is used to create it).

thiagotnunes:
Tested with live migration in SMT. Job succeeded.

PR #913, opened 2024-10-07
added flag to support sharded migrations via cli

bharadwaj-aditya:
Added support to perform sharded migrations from SMT CLI.
This defaults to not adding the migration_shard_id in the primary key.

Options to add in the primary key will be taken on as follow up work.

- [x] Tests pass


conversion/conversion_from_source.go
```diff
@@ -59,18 +59,32 @@ func (sads *SchemaFromSourceImpl) schemaFromDatabase(migrationProjectId string,
 	var infoSchema common.InfoSchema
 	var err error
 	isSharded := false
+	fmt.Printf("fetching schema from database %+v\n", sourceProfile.Config.ShardConfigurationBulk)
```
manitgupta, line 62: This might print sensitive details to the log? Please check print statements added (in other parts of PR as well)

docs/cli/flags.md
```diff
@@ -48,6 +48,8 @@ such as `avro` etc.
 Please note that streaming migration is only supported for MySQL, Oracle and PostgreSQL databases currently.
 Here is an example of a [streamingCfg JSON](./config-json.md#streamingcfg-for-non-sharded-minimal-downtime-migrations) and [how to use it in the CLI](./schema-and-data.md#examples).
 
+* **`shardingConfig`**: Specifies the sharding config file.
```
manitgupta, line 51: Let's add an example sharding config here.

profiles/source_profile.go
```diff
@@ -805,8 +806,25 @@ func NewSourceProfile(s string, source string, n NewSourceProfileInterface) (Sou
...
+				config.ShardIdConfig = shardIdConfig
+			case constants.SHARD_ID_NONE:
+				config.ShardIdConfig = shardIdConfig
+			default:
+				return SourceProfile{}, fmt.Errorf("invalid shard id config passed")
+			}
+		} else {
+			//Default to setting shard id config to none
+			config.ShardIdConfig = constants.SHARD_ID_NONE
+		}
```
manitgupta, line 826: `n.NewSourceProfileConfig` seems to be a misnomer now since it is supposed to return a `SourceProfileConfig` but the `SourceProfileConfig` is now actually a combination of `shardIdConfig` and `ShardConfiguration*` (bulk, dms or dataflow). The `shardIdConfig` only gets added in later..

We should ideally rename `NewSourceProfileConfig` to something like `parseShardConfigurationFile` or shift this logic inside it so that it still returns the completed `SourceProfileConfig` implementation.

manitgupta:
I am actually a bit confused on what this PR is trying to do. I think I initially misunderstood this as adding the ability to modify primary key, but it seems like it just allows user to do sharded migrations via CLI by passing in the cfg?

If so, this is already supported today via the `config=` flag. 

https://screenshot.googleplex.com/5cD3UUPDhhX5nJU

bharadwaj-aditya:
> I am actually a bit confused on what this PR is trying to do. I think I initially misunderstood this as adding the ability to modify primary key, but it seems like it just allows user to do sharded migrations via CLI by passing in the cfg?
> 
> If so, this is already supported today via the `config=` flag.
> 
> https://screenshot.googleplex.com/5cD3UUPDhhX5nJU

It started with that idea, and pulled back to break into 2 PR's as the larger PR was getting too big owing to a bunch of refactoring. 

But i agree, this PR doesnt seem super useful without the overall flow, so I'll probably close this and just raise the larger PR directly.

PR #936, opened 2024-11-18
feat: Refactor SpannerAccessor and add new APIs 

manitgupta:
Refactors SpannerAccessor to simplify the APIs. Instead of clients being passed into every API, the clients have been moved into the method receiver. This avoids pollution of the API signature and abstracts out  the details of the the construction of the SpannerAccessor from the API.


cmd/utils.go
```diff
@@ -161,12 +161,12 @@ func MigrateDatabase(ctx context.Context, migrationProjectId string, targetProfi
...
 func migrateSchema(ctx context.Context, targetProfile profiles.TargetProfile, sourceProfile profiles.SourceProfile,
 	ioHelper *utils.IOStreams, conv *internal.Conv, dbURI string, adminClient *database.DatabaseAdminClient) error {
-		spA := spanneraccessor.SpannerAccessorImpl{}
 		adminClientImpl, err := spanneradmin.NewAdminClientImpl(ctx)
 		if err != nil {
 			return err
 		}
-		err = spA.CreateOrUpdateDatabase(ctx, adminClientImpl, dbURI, sourceProfile.Driver, conv, sourceProfile.Config.ConfigType)
+		spA := spanneraccessor.SpannerAccessorImpl{AdminClient: adminClientImpl}
+		err = spA.CreateOrUpdateDatabase(ctx, dbURI, sourceProfile.Driver, conv, sourceProfile.Config.ConfigType)
```
asthamohta, line 169: I suggest that we add a createNewSpannerAccessor implementation which returns and accessor with all three values, adminClient, instanceClient and spannerClient set and call that

manitgupta: Different portions of the code do not need to initialize all the accessors. Most of the code references only use the adminClient, with instanceClient being used sparingly...
`createNewSpannerAccessor` method will result in a bunch of `null` being passed in various portions of the code to call it; feels more clunky to me..

asthamohta: Having the method doesn't necessitate its usage. Since spanner accessor is passed as an argument in many places and hence its used it becomes clunky in my opinion. Just a thought.

asthamohta (approved):

cmd/utils.go
bharadwaj-aditya: minor optimization - it seems like the adminClient is not being used in these place - and in other parts of the code. May make sense to change the method signature to just access a context and return the Spanner Accessor.

bharadwaj-aditya (commented):
Overall looks good. The simplification of the code was much needed in some places.

Couple of minor comments.

cmd/utils.go
manitgupta: Thanks! Had a discussion with @asthamohta offline. Will incorporate this - @bharadwaj-aditya we have followed a convention in other parts of the code that I will use here now.

accessors/clients/spanner/client/interface.go
```diff
@@ -0,0 +1,61 @@
+package spannerclient
```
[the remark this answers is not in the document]
manitgupta: We have looks of files called `interface.go`. This a convention followed across many packages in SMT currently. Will keep it the same for now, but add a backlog bug for renaming (will rename others together as well)

PR #959, opened 2024-12-15
Upgrading go version to 1.21

darshan-sj:
Upgraded go version to 1.21

**Tested:**
1. go get ./...
2. go mod tidy
3. built and ran a sharded migration from mysql to spanner and cleaned up the resources.
4. built and ran a non sharded migration from mysql to spanner and cleaned up the resources.
5. No errors in the SMT log in all the above steps.


asthamohta (approved):

shreyakhajanchi (changes requested):
We should verify that the version changes are compatible with the kokoro script in SMT GoB repository, ideally it would require to update the version in GoB release script first. 
cc: @manitgupta

PR #963, opened 2024-12-17
feat: default value for mysql source 

asthamohta:
Displaying default values for mysql source
image


sources/common/infoschema.go
```diff
@@ -253,3 +254,14 @@ func (is *InfoSchemaImpl) GetIncludedSrcTablesFromConv(conv *internal.Conv) (sch
 	}
 	return schemaToTablesMap, nil
 }
+
+// SanitizeDefaultValue removes extra characters added to Default Value in information schema in MySQL.
+func SanitizeDefaultValue(defaultValue string, ty string, generated bool) string {
+	defaultValue = strings.ReplaceAll(defaultValue, "_utf8mb4", "")
+	defaultValue = strings.ReplaceAll(defaultValue, "\\\\", "\\")
+	defaultValue = strings.ReplaceAll(defaultValue, "\\'", "'")
+	if !generated && (ty == "char" || ty == "varchar" || ty == "text" || ty=="STRING") && !strings.HasPrefix(defaultValue, "'") && !strings.HasSuffix(defaultValue, "'") {
```
shreyakhajanchi, line 263: Why are we checking for `STRING` if these are MySQL types, also I think the list isn't complete: https://www.w3schools.com/mysql/mysql_datatypes.asp

shreyakhajanchi: It would be better to declare a list of string types and check if ty is present in the list and then do the sanitization

ui/src/app/services/conversion/conversion.service.ts
```diff
@@ -331,6 +331,7 @@ export class ConversionService {
           Name: '',
           GenerationType: ''
         },
+        srcDefaultValue: data.SrcSchema[tableId].ColDefs[colId].DefaultValue.Value.Query,
```
shreyakhajanchi, line 334: Will this not throw an error if data.SrcSchema[tableId].ColDefs[colId].DefaultValue is null or undefined?

shreyakhajanchi (approved):
`index.html` has different main.js version compared to the one getting checked in , please ensure all the file versions match in `dist/` before merging

PR #965, opened 2024-12-17
feat: APIs for Backend Changes for Default Values

asthamohta:
The following PR contains APIs which will be used for Default Value migrations. For completeness on usage of these API, refer to the following PR: https://github.com/GoogleCloudPlatform/spanner-migration-tool/pull/952


sources/common/infoschema.go
```diff
@@ -253,3 +254,14 @@ func (is *InfoSchemaImpl) GetIncludedSrcTablesFromConv(conv *internal.Conv) (sch
 	}
 	return schemaToTablesMap, nil
 }
+
+// SanitizeDefaultValue removes extra characters added to Default Value in information schema in MySQL.
```
asthamohta, line 258: This code is added in another PR for review and merging, the testing for it is added there. Hence its missing here:
https://github.com/GoogleCloudPlatform/spanner-migration-tool/pull/963

webv2/table/utilities.go
```diff
@@ -335,3 +337,37 @@ func getSequenceId(sequenceName string, spSeq map[string]ddl.Sequence) string {
 	}
 	return ""
 }
+
+func UpdateDefaultValue(dv ddl.DefaultValue, tableId, colId string, conv *internal.Conv) {
```
asthamohta, line 341: This code will be tested through the Update API and tests will be added there

expressions_api/expression_verify.go
```diff
@@ -129,3 +166,67 @@ func (ev *ExpressionVerificationAccessorImpl) removeExpressions(inputConv *inter
...
+}
+
+func (ddlv *DDLVerifierImpl) GetSpannerExpressionDetails(conv *internal.Conv, tableIds []string) []internal.ExpressionDetail {
+	expressionDetails := []internal.ExpressionDetail{}
+	// Collect default values for verification
+	for _, tableId := range tableIds {
+		spTable := conv.SpSchema[tableId]
+		for _, spColId := range spTable.ColIds {
+			spCol := spTable.ColDefs[spColId]
+			if spCol.DefaultValue.IsPresent {
```
darshan-sj, line 213: Where will the spCol.DefaultValue.IsPresent be set to true?

expressions_api/expression_verify.go
```diff
@@ -18,22 +19,49 @@ const THREAD_POOL = 500
 type ExpressionVerificationAccessor interface {
 	//Batch API which parallelizes expression verification calls
 	VerifyExpressions(ctx context.Context, verifyExpressionsInput internal.VerifyExpressionsInput) internal.VerifyExpressionsOutput
+	RefreshSpannerClient(ctx context.Context, project string, instance string) error
```
darshan-sj, line 22: I'm assuming the changes to call the VerifyExpressions will be in next PRs. But, I thought RefreshSpannerClient will happen inside `VerifyExpressions` method and the user need not call RefreshSpannerClient before calling VerifyExpressions. 

Either design is fine, I just want to make sure this is intentionally done this way.

darshan-sj (commented):
Left few comments.
Good work covering most of the methods with Unit tests!

webv2/table/utilities.go
asthamohta: There are a couple of differences:
1. This API is only for default values but the `spannerSchemaApplyExpressions` will service for all expression related values such as generated columns, etc.
2. This is called as part of the Edit column functionality and updates the value without verification where as the `spannerSchemaApplyExpressions` also handles any errors when verification api is called
3. This handles update, add, delete for default values which that is just to applies the first round of default values

conversion/conversion_from_source.go
```diff
@@ -159,6 +162,9 @@ func (sads *DataFromSourceImpl) dataFromCSV(ctx context.Context, sourceProfile p
 		return nil, fmt.Errorf("dbName is mandatory in target-profile for csv source")
 	}
 	conv.SpDialect = targetProfile.Conn.Sp.Dialect
+	conv.SpProjectId = targetProfile.Conn.Sp.Project
```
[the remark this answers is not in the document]
asthamohta: I haven't added this anywhere no required yet. Currently project and instance is only needed for schema migration not data migration. I will let it be added when it is required

expressions_api/expression_verify.go
asthamohta: Yes this was intentional as this ends up only being called during ProcessSchema(when the client is used again), I didn't want to increase redundancy of calling this is session creation is expensive

expressions_api/expression_verify.go
```diff
@@ -79,6 +107,15 @@ func (ev *ExpressionVerificationAccessorImpl) VerifyExpressions(ctx context.Cont
 	return verifyExpressionsOutput
 }
 
+func (ev *ExpressionVerificationAccessorImpl) RefreshSpannerClient(ctx context.Context, project string, instance string) error {
+	spannerClient, err := spannerclient.NewSpannerClientImpl(ctx, fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, "smt-staging-db"))
```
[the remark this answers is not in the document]
asthamohta: Manit has already merged the PR for that. This work was done by him

expressions_api/expression_verify.go
asthamohta: Since this is only the backend API's, actually setting this will happen in the full changes. You see those here:
https://github.com/GoogleCloudPlatform/spanner-migration-tool/pull/960/files#diff-d701115d87a06e8b04b128cb7bbb7500236bec7d6eb7a06941fca057f744d451R222-R233

sources/common/toddl.go
```diff
@@ -374,3 +374,31 @@ func CvtIndexHelper(conv *internal.Conv, tableId string, srcIndex schema.Index,
...
+func spannerSchemaApplyExpressions(conv *internal.Conv, expressions internal.VerifyExpressionsOutput) {
+	for _, expression := range expressions.ExpressionVerificationOutputList {
+		switch expression.ExpressionDetail.Type {
+		case "DEFAULT":
+			{
+				tableId := expression.ExpressionDetail.Metadata["TableId"]
+				columnId := expression.ExpressionDetail.Metadata["ColId"]
+
+				if expression.Result {
+					col := conv.SpSchema[tableId].ColDefs[columnId]
```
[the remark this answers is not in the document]
asthamohta: Yes, this is called when the initial migration is done, hence all columns and tables are migrated

sources/common/toddl.go
```diff
@@ -374,3 +374,31 @@ func CvtIndexHelper(conv *internal.Conv, tableId string, srcIndex schema.Index,
...
+					col.DefaultValue = ddl.DefaultValue{
+						IsPresent: true,
+						Value: ddl.Expression{
+							ExpressionId: expression.ExpressionDetail.ExpressionId,
+							Query:        expression.ExpressionDetail.Expression,
```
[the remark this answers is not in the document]
asthamohta: It refers to the Expression, hence I used this name

webv2/table/utilities.go
darshan-sj: Okay 👍 
Can you add a comment above both these comments? A small comment distinguishing them would be enough.

darshan-sj (approved):
LGTM

sources/common/toddl.go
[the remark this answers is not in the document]
darshan-sj: Can you point me to the caller of this method? I dont think we need to make the assumption that this will be always called during the initial migration.

We can check if the tableId and columnId is present in spSchema before trying to read the `col` object here. Then we can call this utility method in other flows as well (for example during expr validation on click of prepare migration)

sources/common/toddl.go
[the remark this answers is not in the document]
darshan-sj: Query doesn't sound right to me. Maybe use `Statement` ?

darshan-sj (approved):

sources/common/toddl.go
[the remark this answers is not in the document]
asthamohta: Sure, let me update this

sources/common/toddl.go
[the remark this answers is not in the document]
asthamohta: https://github.com/GoogleCloudPlatform/spanner-migration-tool/pull/960/files#diff-d26a50b05c1e5c3bf6a8442acd55441b32da1dec3ae486708c0ae2906da91a6fR79-R86

webv2/table/utilities.go
asthamohta: Sure!

expressions_api/expression_verify.go
```diff
@@ -18,22 +19,49 @@ const THREAD_POOL = 500
...
-		return nil, err
+	var spannerAccessor *spanneraccessor.SpannerAccessorImpl
+	var err error
+	if project != "" && instance != "" {
+		spannerAccessor, err = spanneraccessor.NewSpannerAccessorClientImplWithSpannerClient(ctx, fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, "smt-staging-db"))
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		spannerAccessor, err = spanneraccessor.NewSpannerAccessorClientImpl(ctx)
```
Deep1998, line 38: Dont have full context but Why are we multiplexing this here? Seems like this project id based if else logic should reside inside NewSpannerAccessorClientImpl() that creates accessor without spanner client if fields are empty.

expressions_api/expression_verify.go
```diff
@@ -18,22 +19,49 @@ const THREAD_POOL = 500
...
+		if err != nil {
+			return nil, err
+		}
 	}
 	return &ExpressionVerificationAccessorImpl{
 		SpannerAccessor: spannerAccessor,
 	}, nil
 }
 
+type DDLVerifier interface {
```
Deep1998, line 48: nit: in general its good practice to document behaviour for methods in interface.

expressions_api/expression_verify.go
Deep1998, line 111: "smt-staging-db" should be a constant string across the codebase.

expressions_api/expression_verify.go
asthamohta: Because for certain scenarios we will not have the project and instance id and we will refresh the spanner accessor later

expressions_api/expression_verify.go
asthamohta: Makes sense

PR #968, opened 2024-12-19
fix: github issue fix

asthamohta:


.github/workflows/integration-tests-against-emulator.yaml
```diff
@@ -22,7 +22,7 @@ on:
 name: integration-tests-against-emulator
 jobs:
   integration-tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
VardhanThigle, line 25: Downgrading ubuntu might not be the best fix unless we are sure we cant work with latest.
A downgraded ubuntu would make other vulnerabilty fixes hard to apply sooner or later.

VardhanThigle: #969 raises for the oncall issue has all the dependencies now moved to ubuntu-24. So let's consider taking that in.

VardhanThigle: Thanks a ton!

PR #1003, opened 2025-02-04
fix: Query modification and Refresh Client Issue

taherkl:
1. Updated the query to get constraint details.  Issue fixed when the same constraint names are used across different databases.
2. Additional error handling when SchemaToSpannerDDL fails.
3. Added RefreshSpannerClient call for VerifyCheckConstraintExpression API call

### DIFF
--- a/sources/mysql/infoschema.go
+++ b/sources/mysql/infoschema.go
@@ -292,7 +292,7 @@ func (isi InfoSchemaImpl) getConstraintsDQL() (string, error) {
             AND t.TABLE_NAME = k.TABLE_NAME
             LEFT JOIN INFORMATION_SCHEMA.CHECK_CONSTRAINTS AS c
             ON t.CONSTRAINT_NAME = c.CONSTRAINT_NAME
-			AND t.CONSTRAINT_SCHEMA = k.CONSTRAINT_SCHEMA
+			AND t.TABLE_SCHEMA = c.CONSTRAINT_SCHEMA
             WHERE t.TABLE_SCHEMA = ? 
             AND t.TABLE_NAME = ?;`, nil
 	}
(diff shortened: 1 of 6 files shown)


sources/mysql/infoschema.go
```diff
@@ -292,6 +292,7 @@ func (isi InfoSchemaImpl) getConstraintsDQL() (string, error) {
             AND t.TABLE_NAME = k.TABLE_NAME
             LEFT JOIN INFORMATION_SCHEMA.CHECK_CONSTRAINTS AS c
             ON t.CONSTRAINT_NAME = c.CONSTRAINT_NAME
+			AND t.CONSTRAINT_SCHEMA = k.CONSTRAINT_SCHEMA
```
bharadwaj-aditya, line 295: please also add a join clause on table name. It is possible to have the same name of constraint in different tables in a given schema.

bharadwaj-aditya: Correction - it seems this is a difference between MariaDB and MySQL. Mysql allows only one constraint per schema, while Maria requires only that they are unique per table.

But having this additional check should not lead to any harm.

taherkl: @bharadwaj-aditya Incorrect condition was pushed. The correct condition we added to the join is as shown below. Since the constraint name can be same but different expression across different schema's, this led to the issue. The information_schema`.`CHECK_CONSTRAINTS` table does not contain table name so we cannot join with it. 

Previous - ON t.CONSTRAINT_NAME = c.CONSTRAINT_NAME
Modified - ON t.CONSTRAINT_NAME = c.CONSTRAINT_NAME AND t.TABLE_SCHEMA = c.CONSTRAINT_SCHEMA

asthamohta (approved):

PR #1009, opened 2025-02-06
fix(spanner): View assessment does not show detailed report and the download report functionality was not working

pratickchokhani:
- ReportGenerator was null as the respective interface was not assigned.
- If any report has no issues, then all the issues was being reset.


bharadwaj-aditya (approved):
I don't quite understand the root of issue and how this fixed it. But if you have tested and it is working, then it should be ok.

PR #1011, opened 2025-02-06
Use `spanner_migration_tool_output/config` as config file path

VardhanThigle:

- [x] Tests pass
- [ ] Appropriate changes to README are included in PR - NA

### DIFF
--- a/webv2/config/config_service.go
+++ b/webv2/config/config_service.go
@@ -20,12 +20,40 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path/filepath"
+	"sync"
 )
 
-var configFilePath string = "./webv2/config.json"
+var configFilePath string
+var configFilePathMutex sync.Mutex
 
+func getConfigFilePath() (string, error) {
+	dirPath := "spanner_migration_tool_output/config"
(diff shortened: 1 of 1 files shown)


webv2/config/config_service.go
```diff
@@ -20,12 +20,40 @@ import (
...
+	"path/filepath"
+	"sync"
 )
 
-var configFilePath string = "./webv2/config.json"
+var configFilePath string
+var configFilePathMutex sync.Mutex
 
+func getConfigFilePath() (string, error) {
+	dirPath := "spanner_migration_tool_output/config"
```
bharadwaj-aditya, line 31: nit: any reason not to call this just output ?

bharadwaj-aditya (approved):

webv2/config/config_service.go
VardhanThigle: This path is also used for storing the session state. [ref](https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/aa3d540e59aeaa7773aa254fed0f13270e91daed/webv2/session/local_session_store.go#L23).

shreyakhajanchi (approved):

PR #1019, opened 2025-02-18
fix: back tick foreign key column name

asthamohta:


manitgupta (commented):
Have two questions - 

1. Can you validate if the PG dialect reserved keyword handling is also correct?
2. Can you check for column name handling in indexes, constraints, and default as well?

asthamohta:
@manitgupta checked for index and cc. it works fine

shreyakhajanchi (commented):
I think we have missed adding quotes in sequences, auto gen and default values as well.

spanner/ddl/ast.go
```diff
@@ -537,8 +537,8 @@ func isStoredColumnKeyPartOfPrimaryKey(ct CreateTable, colId string) bool {
 func (k Foreignkey) PrintForeignKeyAlterTable(spannerSchema Schema, c Config, tableId string) string {
 	var cols, referCols []string
 	for i, col := range k.ColIds {
```
shreyakhajanchi, line 539: Please add quotes to [check constraints](https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/525dd1ab35c1bb55cd8d73cc87ea3d82a640fdf0/spanner/ddl/ast.go#L560) as well

asthamohta: Check constraints already adds the backticks.

shreyakhajanchi (approved):

PR #1020, opened 2025-02-19
feat: Allow ddl download without comments in UI

asthamohta:


ui/src/app/services/fetch/fetch.service.ts
```diff
@@ -131,6 +131,10 @@ export class FetchService {
     return this.http.get<string>(`${this.url}/downloadDDL`)
   }
 
+  getSpannerDDLWoComments(){
+    return this.http.get<string>(`${this.url}/downloadDDLWoComments`)
+  }
+
```
manitgupta, line 137: I understand that you are following the code convention here, but it looks weird that we decided to create 5 APIs for downloading different artifacts. This should have been a single parameterized API IMO. 
Can you create a backlog bug to refactor this?

manitgupta (approved):

PR #1038, opened 2025-03-04
added db impact implementation

bharadwaj-aditya:
Added db impact to assessment

### DIFF
--- a/assessment/report_generator.go
+++ b/assessment/report_generator.go
@@ -81,7 +81,6 @@ func generateSchemaReport(assessmentOutput utils.AssessmentOutput) [][]string {
 		row = append(row, schemaRow.targetName)
 		row = append(row, schemaRow.targetDefinition)
 		row = append(row, schemaRow.dbChangeEffort)
-		row = append(row, schemaRow.dbChangeType)
 		row = append(row, schemaRow.dbChanges)
 		row = append(row, schemaRow.dbImpact)
 		row = append(row, schemaRow.codeChangeEffort)
@@ -103,7 +102,6 @@ func getHeaders() []string {
 		"Target Name",
 		"Target Definition",
 		//DB
-		"DB Change Type",
 		"DB Change Effort",
 		"DB Changes",
 		"DB Impact",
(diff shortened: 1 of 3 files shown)


assessment/report_generator.go
```diff
@@ -189,3 +196,106 @@ func sourceColumnDefinitionToString(columnDefinition utils.SrcColumnDetails) str
...
+// TODO move calculation logic to assessment engine
+func calculateTableDbChangesAndImpact(srcTable utils.TableDetails, spTable utils.TableDetails) (string, string) {
+	changes := []string{}
+	impact := []string{}
+	if !strings.Contains(srcTable.Charset, "utf8") { // TODO add charset level comparisons - per source
+		changes = append(changes, "charset")
+		impact = append(impact, "storage increase")
+	}
+	if len(changes) == 0 {
+		changes = append(changes, "None")
```
shreyakhajanchi, line 209: if len(impact) == 0 {
		impact = append(impact, "None")
	}
	
This also needs to be added

assessment/report_generator.go
```diff
@@ -189,3 +196,106 @@ func sourceColumnDefinitionToString(columnDefinition utils.SrcColumnDetails) str
...
+	case "bit":
+		return srcCol.Mods[0] / 8
+	case "int":
+		return 4
+	case "integer":
+		return 4
+	case "float":
+		return 4 // Add precision pspecific handling
+	case "text":
+		return 2 ^ 16
```
shreyakhajanchi, line 256: text column takes in size right?

assessment/report_generator.go
```diff
@@ -189,3 +196,106 @@ func sourceColumnDefinitionToString(columnDefinition utils.SrcColumnDetails) str
...
+}
+
+// TODO - move to source specific interfaces - also account for charsets
+func getSrcColSizeBytesMySQL(srcCol utils.SrcColumnDetails) int64 {
+	switch strings.ToLower(srcCol.Datatype) {
+	case "date":
+		return 4
+	case "timestamp":
+		return 4
+	case "bit":
```
shreyakhajanchi, line 247: Following [this](https://www.mysqldatatypes.com/), maybe we should have : (srcCol.Mods[0] + 7)/ 8

assessment/report_generator.go
```diff
@@ -189,3 +196,106 @@ func sourceColumnDefinitionToString(columnDefinition utils.SrcColumnDetails) str
...
+		return 4
+	}
+}
+
+// TODO - move to source specific interfaces - also account for charsets
+func getSpColSizeBytes(spCol utils.SpColumnDetails) int64 {
+	var size int64
+	switch strings.ToUpper(spCol.Datatype) {
+	case "ARRAY":
+		size = spCol.Len
```
shreyakhajanchi, line 272: In case of array datatype , datatype would be string , int, etc and IsArray would be set as true

assessment/report_generator.go
bharadwaj-aditya: added

assessment/report_generator.go
bharadwaj-aditya: changed.

assessment/report_generator.go
bharadwaj-aditya: the current flow works if 1 byte records are used. this logic will be a lot more involved based on underlying type. added a todo here to take up in a follow up PR

shreyakhajanchi (approved):

PR #1039, opened 2025-03-05
fix(spanner): View assessment does not show detailed report and the download report functionality was not working

pratickchokhani:


bharadwaj-aditya (approved):
LGTM

PR #1050, opened 2025-03-11
[DRAFT] Initial commit for data import interface

rohitwali:
This PR proposes an initial interface for data import with basic integ and dummy unit test. 
The dummy unit test helps follow TDD as we iterate on to build concrete functionality


cmd/import_data.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"context"
+	"flag"
+	"fmt"
+
+	"github.com/google/subcommands"
+)
+
+type ImportDataCmd struct {
```
VardhanThigle, line 25: [Super-nit] Could we add a comment for future contributors that any changes in options need to be ported back to the gcloud import surface?

VardhanThigle (approved):

cmd/import_data.go
rohitwali: valid point. I thought about it but was not sure if its reasonable to leak details about gcloud on open source project.
is it already happening in other places in the project?

alternately, I can add it in gcloud, the other way round..LMK

PR #1052, opened 2025-03-12
Add dockerfile and terraform scripts

Deep1998:
Add docker file and terraform scripts. Note that htese are not final and need to be updated once the endpoint client library is made open source.

### DIFF
--- a/sources/cassandra/main.tf
+++ b/sources/cassandra/main.tf
@@ -1,0 +1,86 @@
+resource "google_compute_firewall" "allow_ports" {
+  name    = "allow-zdm-ports"
+  network = var.common_params.network
+
+  allow {
+    protocol = "tcp"
+    ports    = ["14002", "9042"]
+  }
+
+  source_ranges = ["0.0.0.0/0"]
+  target_tags   = ["zdm"]
+}
+
+resource "google_compute_instance" "cos_instances" {
+  count        = var.common_params.instance_count
+  name         = "${var.common_params.instance_name_prefix}-${count.index}"
+  machine_type = var.common_params.machine_type
+  zone         = var.common_params.zone
+
+  boot_disk {
+    initialize_params {
+      image = "projects/cos-cloud/global/images/cos-105-17412-535-55"
+      size  = 32
+      type  = "pd-balanced"
+    }
+  }
+
+  network_interface {
+    network = var.common_params.network
+    access_config {}
+  }
+
+  tags = ["zdm"]
+
+  metadata = {
+    enable-oslogin = "TRUE"
+  }
+}
+
+resource "local_file" "zdm_config" {
+  count    = var.common_params.instance_count
+  filename = "zdm-config-${count.index}.yaml"
+  content = yamlencode({
+    origin_contact_points    = var.zdm_config.origin_contact_points
+    target_contact_points    = var.zdm_config.target_contact_points
+    origin_username          = var.zdm_config.origin_username
+    origin_password          = var.zdm_config.origin_password
+    target_username          = var.zdm_config.target_username
+    target_password          = var.zdm_config.target_password
+    proxy_listen_address     = var.zdm_config.proxy_listen_address
+    proxy_topology_addresses = join(",", google_compute_instance.cos_instances[*].network_interface[0].network_ip)
+    proxy_topology_index     = count.index
+  })
+  depends_on = [google_compute_instance.cos_instances]
+}
+
+resource "null_resource" "setup_instances" {
+  count = var.common_params.instance_count
+
+  provisioner "local-exec" {
+    command = <<-EOT
+      for i in {1..3}; do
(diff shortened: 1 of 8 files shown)


sources/cassandra/main.tf
```diff
@@ -0,0 +1,102 @@
...
+  count        = var.common_params.instance_count
+  name         = "${var.common_params.instance_name_prefix}-${count.index}"
+  machine_type = var.common_params.machine_type
+  zone         = var.common_params.zone
+
+  boot_disk {
+    initialize_params {
+      image = "projects/cos-cloud/global/images/cos-105-17412-535-55"
+      size  = 32
+      type  = "pd-balanced"
```
VardhanThigle, line 24: If we are not using local storage for anything performance critical, would `pd-standard` give the customer better price point?

sources/cassandra/main.tf
```diff
@@ -0,0 +1,102 @@
...
+
+  network_interface {
+    network = var.common_params.network
+    access_config {}
+  }
+
+  tags = ["zdm"]
+
+  metadata = {
+    enable-oslogin = "TRUE"
```
VardhanThigle, line 36: Do we need `osLogin` to be forced as enable? If it's only for ease of  debugging, could we please take this in as a variable?

There could be also be configured by the customer's policy at project or org level
https://cloud.google.com/compute/docs/oslogin/manage-oslogin-in-an-org

The VM IIRC, overrides the project or org config and may be that's not the security posture the customer wants.

sources/cassandra/terraform.tfvars
```diff
@@ -0,0 +1,17 @@
...
+  spanner_project_id   = "test-project"
+  spanner_instance_id  = "test-instance"
+  spanner_database_id  = "test-db"
+  region               = "us-central1"
+  zone                 = "us-central1-a"
+  instance_count       = "10"
+  instance_name_prefix = "node-10-zdm"
+  network              = "default"
+  machine_type         = "n2-standard-8"
+  service_account_key  = "path/to/local/key/file"
```
VardhanThigle, line 11: Sorry I might be missing the entire context, but why do we need `service_account_key`?
Can we avoid it?

Many times production projects (including ones in `google.com`) have service account key creation disabled
This is as per Google's recommendations for Customer's security posture
https://cloud.google.com/iam/docs/best-practices-for-managing-service-account-keys

So it's quite possible that we might encouter a Cx who can not generate a key from an SA.

sources/cassandra/main.tf
```diff
@@ -0,0 +1,102 @@
...
+  })
+  depends_on = [google_compute_instance.cos_instances]
+}
+
+resource "null_resource" "setup_instances" {
+  count = var.common_params.instance_count
+
+  provisioner "local-exec" {
+    command = <<-EOT
+      for i in {1..3}; do
```
VardhanThigle, line 71: Is 3 number of nodes?
If yes, can we make it configurable.

VardhanThigle (approved):
Over all looks good.
If any of the comments needs more work in terms of testing the whole flow once again, feel free to resolve, and add a TODO bug for later.

sources/cassandra/main.tf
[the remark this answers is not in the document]
Deep1998: Because this does an scp which can have transient network issues, so 3 times is for retrial.

Deep1998: This is a retry loop. I faced network timeouts with the scp command when running orchestrating large number of proxies, so this retries scp thrice before failing. Ideally we would want it configurable but did not want to overload params for the user unless we feel we need more.

sources/cassandra/terraform.tfvars
Deep1998: The service account key is to allow the sidecar allow access to Spanner. This seems consistent with the official documentation of how the cassandra-to-spanner-proxy recommends running it.

sources/cassandra/main.tf
Deep1998: Agree that production setups should have this disabled. But I do not expect customers to use this directly for production (just like our other terraform scripts) and this is primarily for reference and basic testing.

Adding a flag for this niche detail is a little out of place, as there are many other things customers would like to tweak in their GCE vm config, firewall, networking settings etc itself. This is just a reference on how to get the proxy stuff orchestrated.

VardhanThigle:
```
Adding a flag for this niche detail is a little out of place
```
Agreed, but my slight worry was that:
If we are overriding it enable at VM level, we might end up contradicting the policy that the customer's place has setup at a project or org level.
Should we just let it inherit the project instead?

If this is primarily for reference - 
  1. please let us call it out as such in the documentation
  2. if it's primarily for reference I might lean on not enabling OS login here and letting it inheirt from the project properties. We can always advise the customer to change the script if they want to override this in the docs.

sources/cassandra/terraform.tfvars
Deep1998: Added a todo to find more secure ways of doing this

PR #1060, opened 2025-03-18
Update WaitGroup handling

manitgupta:
Fixed flaky test: b/404432832

### Investigation

This is what was happening initially, this was the error - 

`panic: sync: WaitGroup misuse: Add called concurrently with Wait`

I solved this by protecting the Add() with the mutex that is already passed to the processAsync method. When I did that, the error above went away, instead I got this error - 

`panic: sync: WaitGroup is reused before previous Wait has returned [recovered]`

From my research, the above happens when the WaitGroup is accessed while a previous Wait() is in progress.

This is when I made the fix in this PR.

My hypothesis is that whenever the fastExit is triggered, because we would call two wg.Wait(), one in the fastExit code, and the other in the defer - it can result in this concurrency issue, where the wg.Wait() from the fastExit has not returned and the wg.Wait() from the defer is triggered. 
Because fastExit code would get triggered in cases when an individual goroutine _happens_ to return an error, it would also explain the flakiness. 

I add also considered adding all the routines together as you are recommending wg.Add(n), but I found this blog which seems to suggest that doing that is error-prone. 

https://victoriametrics.com/blog/go-sync-waitgroup/#:~:text=%E2%80%9CWhy%20is%20wg.Add(n)%20considered%20error%2Dprone%3F%E2%80%9D

### Change

Currently, `wg.Wait()` is executed inside a `defer` block, which means it only executes after the RunParallelTasks function completes, regardless of how the function itself exits. In the fastExit scenario, we are explicitly calling `wg.Wait()` within the function. Later, when the function exits the `defer` block will also be triggered. This can theoretically lead to the "WaitGroup is reused before previous Wait has returned" error.
The `wg.Wait()` is now called after the loop that reads from `outputChannel`. This ensures that all workers have finished before the function returns. The deferred `wg.Wait()` has been removed.

### Testing 

1. Ran the `integration test` suite multiple items on this PR. This issue did not appear again.
2. Ran a large scale schema migration with over 1400 tables and ~1000 default and check constraints which totaled a completion of 4387 internal goroutines during processing. This test was run 2 times. No error was observed.

### DIFF
--- a/common/task/task.go
+++ b/common/task/task.go
@@ -48,12 +48,12 @@ type TaskResult[O any] struct {
 // 2. Defer execution to wg.Wait(). This ensures that all tasks added to the waitGroup will be processed.
 // 3. Add len(input) tasks to the thread. This ensures that the waitGroup will track the completion of the tasks.
 // 4. We then begin the execution by executing the processAync in goroutines (equal to the numWorkers)
-// 5. When the processAsync processes a task, there are two possible outcomes - it will either pass or fail.
+// 5. When processAsync processes a task, there are two possible outcomes - it will either pass or fail.
 // 6. No matter the result, once the processing is complete, the task is acknowledged to be completed using wg.Done()
 // 7a. [If no fastExit] No matter if it passes or fails, the results are added to the TaskResult[0] array.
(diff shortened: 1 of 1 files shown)


manitgupta:
@bharadwaj-aditya / @VardhanThigle Please review.

PR #1063, opened 2025-03-21
Assessment report update

bharadwaj-aditya:
Updated assessment report with required code fields.

### DIFF
--- a/assessment/collectors/parser/response_parser.go
+++ b/assessment/collectors/parser/response_parser.go
@@ -19,6 +19,7 @@ package assessment
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	. "github.com/GoogleCloudPlatform/spanner-migration-tool/assessment/utils"
 	"github.com/GoogleCloudPlatform/spanner-migration-tool/logger"
@@ -60,7 +61,7 @@ func parseAnyToString(anyType any) string {
 	return fmt.Sprintf("%v", anyType)
 }
 
-func ParseSchemaImpact(schemaImpactResponse map[string]any, filePath string) (*Snippet, error) {
+func ParseSchemaImpact(schemaImpactResponse map[string]any, projectPath, filePath string) (*Snippet, error) {
 	logger.Log.Debug("schemaImpactResponse:", zap.Any("sec: ", schemaImpactResponse))
 	return &Snippet{
 		SchemaChange:          parseAnyToString(schemaImpactResponse["schema_change"]),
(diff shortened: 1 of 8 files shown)


assessment/collectors/parser/response_parser.go
```diff
@@ -60,36 +62,56 @@ func parseAnyToString(anyType any) string {
...
+}
+
+func ParseSchemaImpact(schemaImpactResponse map[string]any, projectPath, filePath string) (*Snippet, error) {
 	logger.Log.Debug("schemaImpactResponse:", zap.Any("sec: ", schemaImpactResponse))
 	return &Snippet{
 		SchemaChange:          parseAnyToString(schemaImpactResponse["schema_change"]),
 		TableName:             parseAnyToString(schemaImpactResponse["table"]),
 		ColumnName:            parseAnyToString(schemaImpactResponse["column"]),
-		NumberOfAffectedLines: parseAnyToString(schemaImpactResponse["number_of_affected_lines"]),
+		NumberOfAffectedLines: parseAnyToInteger(schemaImpactResponse["number_of_affected_lines"]),
```
pratickchokhani, line 81: If the string is not parsable to an Int, shouldn't we still make it part of the report as it may have some parsing issues?

Else, can we show '-' or empty string?

pratickchokhani (approved):

assessment/collectors/parser/response_parser.go
gauravpurohit06: (+1) for keeping it as String... LLM may return non numeric data for this field in which case `parseAnyToInteger` will get failed..

assessment/report_generator.go
```diff
@@ -56,16 +68,16 @@ func dumpCsvReport(fileName string, records [][]string) {
 	defer f.Close()
 
 	w := csv.NewWriter(f)
-	w.Comma = '|'
+	w.Comma = '\t'
```
gauravpurohit06, line 71: Code Snippets in the LLM response often have `\t`, `\n`.. I think using this as a delimiter in csv may mess up the formatting of data.

gauravpurohit06 (changes requested):

assessment/report_generator.go
bharadwaj-aditya: There is a sanitize method called before this which strips out all the \t and \n.
It may mess up the readability of the code, but it won't cause errors.

assessment/collectors/parser/response_parser.go
bharadwaj-aditya: the parseAnyToInteger handles this and returns 0 in case it can not be parsed.

PR #1064, opened 2025-03-21
moved snippets into app code assessment and corrected default null ha…

bharadwaj-aditya:
1. moved app assessment to correct struct.
2. corrected handling of null in default values

### DIFF
--- a/assessment/collectors/app_migration_analysis.go
+++ b/assessment/collectors/app_migration_analysis.go
@@ -476,17 +476,27 @@ func (m *MigrationSummarizer) AnalyzeProject(ctx context.Context) (*CodeAssessme
 
 	runParallel := &task.RunParallelTasksImpl[*AnalyzeFileInput, *AnalyzeFileResponse]{}
 	fileIndex := 0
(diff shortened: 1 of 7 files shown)


assessment/collectors/app_migration_analysis.go
```diff
@@ -524,9 +534,38 @@ func (m *MigrationSummarizer) AnalyzeProject(ctx context.Context) (*CodeAssessme
...
 		}
 	}
+	codeAssessment.Language = language
+	codeAssessment.Framework = framework
+	codeAssessment.TotalLoc = totalLoc
+	codeAssessment.TotalFiles = fileIndex
 	return codeAssessment, nil
 }
 
+func getFramework(fileContent string) string {
```
gauravpurohit06, line 544: This method is specific to Go ecosystem, we should also update it as part of refactoring to support multiple languages and frameworks.

gauravpurohit06 (approved):
LGTM, there are few things which we need to refactor as it's specific to Go ecosystem.

PR #1068, opened 2025-03-24
Data import: Basic functionality. Works with existing schema

rohitwali:
This PR implements a basic data load from CSV, assumes target table already exists on the Spanner schema.
Subsequent PRs are expected to remove that assumption.


cmd/import_data.go
```diff
@@ -47,14 +61,143 @@ func (cmd *ImportDataCmd) SetFlags(set *flag.FlagSet) {
...
+		importCsv(ctx, cmd, &csv.CsvImpl{})
+	default:
+		fmt.Printf("format %s not supported yet", cmd.sourceFormat)
+	}
+
 	return 0
 }
 
+func importCsv(ctx context.Context, cmd *ImportDataCmd, csv csv.CsvInterface) subcommands.ExitStatus {
+	// TODO: start with single table imports
```
manitgupta, line 77: I think we should set the structure correctly and then iteratively implement the functionality. This involves defining the method signatures, file structure etc. upfront so that they don't change much (some changes are fine) with successive PRs.

Some suggestions - 

1. `import_data.go` could only be responsible for parsing the incoming command and then calling the appropriate internal libraries. 
2. The CSV import functionality (`importCsv` method) should be implemented in its own file. This will allow for easier unit testing of this method as well.

This will keep the imports of this file clean and delegate functionality as required, and allow for easier testing.

cmd/import_data.go
```diff
@@ -47,14 +61,143 @@ func (cmd *ImportDataCmd) SetFlags(set *flag.FlagSet) {
...
+		err = fmt.Errorf("can't create admin client: %v", parse.AnalyzeError(err, dbURI))
+		return nil, nil, err
+	}
+	client, err := utils.GetClient(ctx, dbURI)
+	if err != nil {
+		err = fmt.Errorf("can't create client for db %s: %v", dbURI, err)
+		return adminClient, nil, err
+	}
+	return adminClient, client, nil
+}
```
manitgupta, line 197: You should use accessors to implement Spanner associated functionality - 

https://github.com/GoogleCloudPlatform/spanner-migration-tool/blob/master/accessors/spanner/spanner_accessor.go

Add a method to the accessor if one suitable for your use-case does not exist. This will keep consistency across the code base and make unit testing much more easier.

cmd/import_data.go
```diff
@@ -47,14 +61,143 @@ func (cmd *ImportDataCmd) SetFlags(set *flag.FlagSet) {
...
+	conv.SetDataMode()
+	conv.SetDataSink(
+		func(table string, cols []string, vals []interface{}) {
+			batchWriter.AddRow(table, cols, vals)
+		})
+	conv.DataFlush = func() {
+		batchWriter.Flush()
+	}
+	return batchWriter
+}
```
manitgupta, line 180: Seems like this should be part of the constructor/initialization method of the `BatchWriter`? We should be able to pass configurations (including `func()`) on how we want the BatchWriter to be set up and then just use it I think.

cmd/import_data.go
```diff
@@ -47,14 +61,143 @@ func (cmd *ImportDataCmd) SetFlags(set *flag.FlagSet) {
 }
 
 func (cmd *ImportDataCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...interface{}) subcommands.ExitStatus {
```
manitgupta, line 63: Does the unit test need to be updated? I think because the return status of `importCsv` is not consumed, the current UT always succeeds since the command always returns 0.

sources/csv/data.go
```diff
@@ -222,32 +216,12 @@ func (c *CsvImpl) ProcessCSV(conv *internal.Conv, tables []utils.ManifestTable,
...
 			}
+			colDefs := conv.SpSchema[tableId].ColDefs
 
-			srcCols, err := r.Read()
-			if err == io.EOF {
-				conv.Unexpected(fmt.Sprintf("error processing table %s: file %s is empty.", table.Table_name, filePath))
-				continue
-			}
+			err = c.ProcessSingleCSV(conv, table.Table_name, colNames, colDefs,
+				filePath, nullStr, delimiter)
```
manitgupta, line 222: `colDefs` variable is set using the `conv` object, but then later in the `ProcessSingleCSV` method, both `colDefs` and `conv` are passed - this seems redundant. `conv` already contains `colDefs` implicitly and there is no need to pass it as a separate parameter down the stack.

This has caused a ripple effect signature modification of other downstream methods (`processDataRow`, `convertData`) which can be avoided.

darshan-sj (approved):
Looks good

cmd/import_data.go
rohitwali: I acknowledge that the PR is a bit light on UTs. The intent is to keep changes light for easier and faster reivew. there will be a separate PR for UT

cmd/import_data.go
```diff
@@ -47,14 +61,143 @@ func (cmd *ImportDataCmd) SetFlags(set *flag.FlagSet) {
...
-	//TODO implement me
-	fmt.Printf("executing dummy import data\n")
 	fmt.Printf("instanceId %s, dbName %s, schemaUri %s\n", cmd.instanceId, cmd.dbName, cmd.schemaUri)
```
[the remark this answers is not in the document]
rohitwali: ack

cmd/import_data.go
```diff
@@ -47,14 +61,143 @@ func (cmd *ImportDataCmd) SetFlags(set *flag.FlagSet) {
...
+
+	switch cmd.sourceFormat {
+	case constants.CSV:
+		importCsv(ctx, cmd, &csv.CsvImpl{})
```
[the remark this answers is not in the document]
rohitwali: 1. I am taking inspiration from other places in the code and seems like a pattern we are following. Let me know in case I  may be missing something
2. the return code interface needs to be revisited so was being lazy. have fixed this now. thanks for pointing

cmd/import_data.go
```diff
@@ -47,14 +61,143 @@ func (cmd *ImportDataCmd) SetFlags(set *flag.FlagSet) {
...
+	default:
+		fmt.Printf("format %s not supported yet", cmd.sourceFormat)
+	}
+
 	return 0
```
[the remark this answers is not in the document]
rohitwali: fixed

cmd/import_data.go
rohitwali: wasnt aware. thanks for the pointer. will update

cmd/import_data.go
```diff
@@ -47,14 +61,143 @@ func (cmd *ImportDataCmd) SetFlags(set *flag.FlagSet) {
...
+		batchWriter.Flush()
+	}
+	return batchWriter
+}
+
+func GetSpannerClient(ctx context.Context, project string, instance string,
+	dbName string) (*database.DatabaseAdminClient, *sp.Client, error) {
+
+	dbURI := fmt.Sprintf("projects/%s/instances/%s/databases/%s", project, instance, dbName)
+	adminClient, err := utils.NewDatabaseAdminClient(ctx)
```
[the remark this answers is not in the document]
rohitwali: yes, it doesnt :)

cmd/import_data.go
rohitwali: updated. PTAL

cmd/import_data.go
rohitwali: I am not sure about constructor as that will require leaking details to source caller. I fell these are better placed with source handlers that have visibility into row sizes, etc to tweak around mutation configs, etc

cmd/import_data.go
rohitwali: done. PTAL

cmd/import_data.go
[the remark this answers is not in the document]
rohitwali: updated. PTAL

cmd/import_data.go
rohitwali: anyway PTAL let the newer revision and LMK

test_data/basic_csv_schema.csv
```diff
@@ -1,2 +1,2 @@
-c3,INT64
-c4,STRING
\ No newline at end of file
+c3,INT64,1,1
+c4,STRING(250),1,0
```
manitgupta, line 2: I'm a bit confused by this, this is a `csv based schema` ? Can you elaborate what that means are what do the new numbers `1,1 and 1,0` added mean?

manitgupta:
You should consider rebasing so that codecov can run now since the PR is in a much better shape now.

cmd/import_data.go
```diff
@@ -47,14 +54,79 @@ func (cmd *ImportDataCmd) SetFlags(set *flag.FlagSet) {
...
+	}
+
+	startTime := time.Now()
+	csvSchema := import_data.CsvSchemaImpl{}
+	csvSchema.ProjectId = cmd.project
+	csvSchema.InstanceId = cmd.instanceId
+	csvSchema.TableName = cmd.tableName
+	csvSchema.DbName = cmd.dbName
+	csvSchema.SchemaUri = cmd.schemaUri
+	csvSchema.CsvFieldDelimiter = cmd.csvFieldDelimiter
```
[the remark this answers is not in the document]
rohitwali: ack. will do

cmd/import_data.go
```diff
@@ -47,14 +54,79 @@ func (cmd *ImportDataCmd) SetFlags(set *flag.FlagSet) {
...
+	dbURI := fmt.Sprintf("projects/%s/instances/%s/databases/%s", cmd.project, cmd.instanceId, cmd.dbName)
+	infoSchema, err := spanner.NewInfoSchemaImplWithSpannerClient(ctx, dbURI, constants.DIALECT_GOOGLESQL)
+	if err != nil {
+		logger.Log.Error(fmt.Sprintf("Unable to read Spanner schema %v", err))
+		return subcommands.ExitFailure
+	}
+
+	switch cmd.sourceFormat {
+	case constants.CSV:
+		err := cmd.handleCsv(ctx, infoSchema)
```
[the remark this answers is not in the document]
rohitwali: may be, once we have a use case. but, at the moment, its a bit futuristic to me.

sources/spanner/infoschema.go
```diff
@@ -128,6 +145,39 @@ func (isi InfoSchemaImpl) GetTables() ([]common.SchemaAndName, error) {
 	return tables, nil
 }
 
+func (sp *InfoSchemaImpl) PopulateSpannerSchema(ctx context.Context, conv *internal.Conv) error {
```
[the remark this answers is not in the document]
rohitwali: probably... probably not! the current PR is already getting out of manageable limits. thanks for your patience to got through it though :)
I will prefer iterating in a subsequent PR, if you are ok

test_data/basic_csv_schema.csv
rohitwali: If you look at ColumnDefinition struct in the csv_schema, that should give details.
But for TLDR,  the current format is
column name, spanner type, is null, primary key order (if part of primary key)

csv seems simple and sufficient for now. but open to suggestions

manitgupta (approved):
LGTM! CodeCov has not run for some reason though...

PR #1074, opened 2025-04-03
Adjust code Codecov

VardhanThigle:
Temporarily reduce codecov threshold.
Currently the project has dropped to 39%. We need to leave some threshold untill UTs are added back again. 
Moving it to 32%.


bharadwaj-aditya (approved):
LGTM - Temporarily dropping this until we get the code coverage back up.

PR #1075, opened 2025-04-03
Adding package.lock to pin dependencies. Also add a Workflow for checking package.lock

VardhanThigle:
Fixes Vulnerability - package-lock.json not found.
Also adds a workflow to detect:
1. if the package-lock.json is not modified along with package.json (or is deleted) 
2. In situations where the workflow needs to be overridden  (ideally there shouldn't be a case but we need to have a way to avoid an impossible impasse ) , it also allows for an override. The override is to add these 3 words concatenated with a `-` (hyphen) `NO`, `LOCKFILE` and `CHECK` in the PR description.  The new work flow might re-run each time a PR description is edited but it completed quite quickly.

- [ ] Tests pass
   - [x]  https://screenshot.googleplex.com/BGGLBSrEmsSHT7z - For the case where file is deleted.
   - [x] https://screenshot.googleplex.com/4atbdbT4fUYTTPP - For testing over-ride 
   - [ ] TODO: After checkin, test with file not changed across base commit.
- [ ] TODO
   - [ ] After Github, GOB changes for the same. 

### DIFF
--- a/.github/workflows/check-package-lock.yaml
+++ b/.github/workflows/check-package-lock.yaml
@@ -1,0 +1,59 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


manitgupta (approved):
Thanks, this is handled pretty neatly!

PR #1078, opened 2025-04-03
feat: Adding support for JDBC Projects in Application Code Assessment

gauravpurohit06:
1. Adding RAG dataset for JDBC Projects.
2. Moving Prompt Template to Files and removing from Code.
3. Making Prompts Generic.


assessment/collectors/app_migration_analysis.go
```diff
@@ -92,10 +104,24 @@ func NewMigrationSummarizer(ctx context.Context, googleGenerativeAIAPIKey *strin
...
-	conceptExampleDB, err := assessment.NewMysqlConceptDb(projectID, location)
+	conceptExampleDB, err := assessment.NewMysqlConceptDb(projectID, location, language)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load code example DB: %w", err)
 	}
+
+	var sourceFramework, targetFramework string
+
+	switch language {
+	case "go":
```
bharadwaj-aditya, line 115: The frameworks should be detected. If that is not possible now please add a TODO here.

bharadwaj-aditya (approved):
Looks much cleaner. LGTM apart from 1 comment around framework. Please address that and merge the PR.

assessment/collectors/app_migration_analysis.go
gauravpurohit06: yes we can detect but the cleaner approach would be take it as input from user as part of `AssessmentConfig`

PR #1089, opened 2025-04-10
feat(spanner): Add support for MYSQLDUMP in import command:

pratickchokhani:
1. This enables importing data and schema from mysql dump.
2. It creates a database if not already created.
3. Unit test will be added in a follow up PR.


cmd/import_data.go
```diff
@@ -111,8 +119,79 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
 
 }
 
+func (cmd *ImportDataCmd) handleDump(ctx context.Context, dbUri, dialect string, spannerAccessor *spanneraccessor.SpannerAccessorImpl) error {
+	// TODO: handle POSTGRESQL
+	driver := constants.MYSQLDUMP
+
+	// TODO: handle GCS
+	dumpReader, err := os.Open(cmd.sourceUri)
```
rohitwali, line 127: may want to move this to import_from_dump class

rohitwali: if the intent is validation, you may want to use exists API instead

cmd/import_data.go
```diff
@@ -111,8 +119,79 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
...
+	importDump := &import_data.ImportFromDumpImpl{
+		ProjectId:  cmd.project,
+		InstanceId: cmd.instanceId,
+		DbName:     cmd.dbName,
+		DumpUri:    cmd.sourceUri,
+		DumpReader: dumpReader,
+		Driver:     driver,
+	}
+
+	expressionVerificationAccessor := &expressions_api.ExpressionVerificationAccessorImpl{
```
rohitwali, line 145: can this be pushed down?

cmd/import_data.go
```diff
@@ -111,8 +119,79 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
...
+	}
+
+	defer dumpReader.Close()
+
+	importDump := &import_data.ImportFromDumpImpl{
+		ProjectId:  cmd.project,
+		InstanceId: cmd.instanceId,
+		DbName:     cmd.dbName,
+		DumpUri:    cmd.sourceUri,
+		DumpReader: dumpReader,
```
rohitwali, line 141: do you need this given you are passing the URI?

cmd/import_data.go
```diff
@@ -111,8 +119,79 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
...
+		Driver:     driver,
```
rohitwali, line 142: unclear. Can you please disambiguate?

cmd/import_data.go
```diff
@@ -111,8 +119,79 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
...
+		return fmt.Errorf(fmt.Sprintf("can't create schema: %v\n", err))
+	}
+
+	dumpReader, err = resetReader(dumpReader, cmd.sourceUri)
+
+	if err != nil {
+		return fmt.Errorf(fmt.Sprintf("can't reset reader: %v\n", err))
+	}
+
+	err = spannerAccessor.CreateOrUpdateDatabase(ctx, dbUri, driver, conv, driver)
```
rohitwali, line 165: database is created after the schema?

import_data/import_from_dump.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+	spannerclient "github.com/GoogleCloudPlatform/spanner-migration-tool/accessors/clients/spanner/client"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/conversion"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/internal"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/logger"
+	"go.uber.org/zap"
+	"io"
+)
+
+type ImportFromDump interface {
```
rohitwali, line 14: would it make sense to split this into mysqldump and pg_dump?

import_data/import_from_dump.go
```diff
@@ -0,0 +1,57 @@
...
+	err := processDump.ProcessDump(source.Driver, conv, r)
+	if err != nil {
+		logger.Log.Error("Failed to parse the dump file:", zap.Error(err))
+		return nil, fmt.Errorf("failed to parse the dump file")
+	}
+	return conv, nil
+}
+
+func (source *ImportFromDumpImpl) ImportData(conv *internal.Conv, processDump conversion.ProcessDumpByDialectInterface, client spannerclient.SpannerClient) error {
+	logger.Log.Info(fmt.Sprintf("Importing %d rows.", conv.Rows()))
```
rohitwali, line 46: does this need to be info level?

import_data/import_from_dump.go
```diff
@@ -0,0 +1,57 @@
...
+type ImportFromDumpImpl struct {
+	ProjectId  string
+	InstanceId string
+	DbName     string
+	DumpUri    string
+	DumpReader io.Reader
+	Driver     string
+}
+
+func (source *ImportFromDumpImpl) CreateSchema(dialect string, processDump conversion.ProcessDumpByDialectInterface) (*internal.Conv, error) {
```
rohitwali, line 28: can we add some unit tests for this class?

rohitwali (commented):
>It creates a database if not already created.
We should assume database will be created outside of SMT. table creation can be handled in the PR.

Please also add a section on testing done.

cmd/import_data.go
pratickchokhani: Initiating the interface here helps in using Mocks in ImportFromDump

cmd/import_data.go
pratickchokhani: Not intended for validation. There are following benefits of initialising the reader here:
1. If file is seek able, then the same reader will be used again.
2. For GCS files, same storage client can be used. (It's part of subsequent PR)
3. Makes ImportFromDump unaware of the file format.

cmd/import_data.go
pratickchokhani: See above.

cmd/import_data.go
pratickchokhani: In CreateSchema, the source schema is processed and converted to Spanner Schema.

This function creates Spanner database.

import_data/import_from_dump.go
pratickchokhani: The will be a help full metrics for customers.

import_data/import_from_dump.go
pratickchokhani: Unit test will be added in subsequent PR. Adding them here was making the PR very long.

cmd/import_data.go
pratickchokhani: Source data type: mysqldump or pg_dump.

pratickchokhani:
> > It creates a database if not already created.
> > We should assume database will be created outside of SMT. table creation can be handled in the PR.
> 
> Please also add a section on testing done.

Unit tests will be added in subsequent PR.

cmd/import_data.go
```diff
@@ -111,8 +119,79 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
...
+
+	endTime1 := time.Now()
```
bharadwaj-aditya, line 167: please name this a little better.

cmd/import_data.go
```diff
@@ -111,8 +119,79 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
...
+	elapsedTime := endTime1.Sub(startTime)
+	fmt.Println("Schema creation took ", elapsedTime.Seconds(), "  secs")
```
bharadwaj-aditya, line 169: please use loggers everywhere.

import_data/import_from_dump.go
```diff
@@ -0,0 +1,57 @@
+package import_data
```
bharadwaj-aditya, line 1: Seems like this does both schema and data.  please name the package as import instead

cmd/import_data.go
```diff
@@ -18,13 +18,16 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/conversion"
```
bharadwaj-aditya, line 21: this command does schema and data. Can you use a more indicative name ?

cmd/import_data.go
```diff
@@ -111,8 +119,79 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
...
+		DumpUri:    cmd.sourceUri,
+		DumpReader: dumpReader,
+		Driver:     driver,
+	}
+
+	expressionVerificationAccessor := &expressions_api.ExpressionVerificationAccessorImpl{
+		SpannerAccessor: spannerAccessor,
+	}
+
+	processDump := &conversion.ProcessDumpByDialectImpl{
```
bharadwaj-aditya, line 149: is the test coverage for this method sufficient ?

bharadwaj-aditya (dismissed):

bharadwaj-aditya (commented):
Please update the supports values in the flag description

cmd/import_data.go
```diff
@@ -57,34 +60,39 @@ func (cmd *ImportDataCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...
...
 	case constants.CSV:
-		err := cmd.handleCsv(ctx, infoSchema)
+		err := cmd.handleCsv(ctx, dbURI, sp)
 		if err != nil {
 			logger.Log.Error(fmt.Sprintf("Unable to handle Csv %v", err))
 			return subcommands.ExitFailure
 		}
 		return subcommands.ExitSuccess
+	case constants.MYSQLDUMP:
+		err := cmd.handleDump(ctx, dbURI, constants.DIALECT_GOOGLESQL, sp)
```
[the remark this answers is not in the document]
pratickchokhani: Done.

cmd/import_data.go
```diff
@@ -57,34 +60,39 @@ func (cmd *ImportDataCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...
...
 }
 
-func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.InfoSchemaImpl) error {
+func (cmd *ImportDataCmd) handleCsv(ctx context.Context, dbURI string, sp *spanneraccessor.SpannerAccessorImpl) error {
 	//TODO: handle POSTGRESQL
 	dialect := constants.DIALECT_GOOGLESQL
-
-	dbURI := fmt.Sprintf("projects/%s/instances/%s/databases/%s", cmd.project, cmd.instanceId, cmd.dbName)
-	sp, err := spanneraccessor.NewSpannerAccessorClientImplWithSpannerClient(ctx, dbURI)
+	infoSchema, err := spanner.NewInfoSchemaImplWithSpannerClient(ctx, dbURI, constants.DIALECT_GOOGLESQL)
```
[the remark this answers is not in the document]
pratickchokhani: Done.

cmd/import_data.go
```diff
@@ -111,8 +119,79 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
 
 }
 
+func (cmd *ImportDataCmd) handleDump(ctx context.Context, dbUri, dialect string, spannerAccessor *spanneraccessor.SpannerAccessorImpl) error {
```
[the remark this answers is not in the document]
pratickchokhani: This method is common for both Mysql and PG. Have renamed it to reflect that.

cmd/import_data.go
pratickchokhani: Done.

cmd/import_data.go
pratickchokhani: Done.

cmd/import_data.go
```diff
@@ -111,8 +119,79 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
...
+	err = importDump.ImportData(conv, processDump, spannerAccessor.SpannerClient)
+
+	endTime2 := time.Now()
+	elapsedTime = endTime2.Sub(endTime1)
+	fmt.Println("Data import took ", elapsedTime.Seconds(), "  secs")
+	return err
+
+}
+
+func resetReader(dumpReader *os.File, fileUri string) (*os.File, error) {
```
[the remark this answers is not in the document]
pratickchokhani: Have moved it. In schema_and_data, we copy the file to TMP Dir for making it seekable. Hence, there is not equivalent function.

import_data/import_from_dump.go
pratickchokhani: Done.

cmd/import_data.go
pratickchokhani: Done.

cmd/import_data.go
pratickchokhani: This interface expose only 1 function. The job of the function is to call the respective API for the specific source format. 

UT for this will be added in the subsequent PR.

pratickchokhani:
> Please update the supports values in the flag description

Done. 2 flags (dialect and dry-run) is not added as there support is to be implemented.

cmd/import_data.go
rohitwali: I would argue the other way round. :)

This class has no need to do file handling. Where as import_from_dump needs to do this anyways

cmd/import_data.go
rohitwali: we should assume database is already created and validate that

import_data/import_from_dump.go
rohitwali: they can wc -l on the file to find this?

import_data/import_from_dump.go
rohitwali: We should add a few to ensure the interfaces introduced in the PR are conjucive to unit testing

pratickchokhani: Done.

import_data/import_from_dump.go
pratickchokhani: As discussed, this is resolved.

cmd/import_data.go
pratickchokhani: I have added a TODO for this. There is a bug within DDL verification. This need to be fixed before changing this. Will be doing that next.

cmd/import_data.go
pratickchokhani: Have updated this as per discussion.

cmd/import_data.go
pratickchokhani: Have pushed it.

cmd/import.go
```diff
@@ -18,13 +18,13 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/import_file"
```
rohitwali, line 21: can we call this just import?

cmd/import_test.go
```diff
@@ -0,0 +1,164 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.*/
+
+package cmd
+
+import (
```
rohitwali, line 17: is this just rename? or needs review?

import_file/import_from_dump.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/sources/mysql"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/sources/postgres"
+	"go.uber.org/zap"
+	"os"
+)
+
+type ImportFromDump interface {
+	CreateSchema(dialect string) (*internal.Conv, error)
+	ImportData(conv *internal.Conv) error
+	Finalize() error
```
rohitwali, line 21: any reason this is in interface? seems to be leaking impl details

import_file/import_from_dump.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+type ImportFromDumpImpl struct {
+	ProjectId       string
+	InstanceId      string
+	DatabaseName    string
+	DumpUri         string
+	dumpReader      *os.File
```
rohitwali, line 29: lets use either URI or File.

cmd/import_test.go
pratickchokhani: TestImportDataCmd_handleDump test has been added.

import_file/import_from_dump.go
pratickchokhani: The file opened and assigned to dumpReader needs to be closed post processing. Also, the same handler is used for both CreateSchema and ImportData. 

Exposing Finalize here will be similar to exposing Close functions.

accessors/spanner/mocks.go
```diff
@@ -38,6 +39,10 @@ type SpannerAccessorMock struct {
 	DropDatabaseMock                func(ctx context.Context, dbURI string) error
 	ValidateDMLMock                 func(ctx context.Context, query string) (bool, error)
 	TableExistsMock                 func(ctx context.Context, tableName string) (bool, error)
+	GetDatabaseNameMock             func() string
+	RefreshMock                     func(ctx context.Context, dbURI string)
+	SetSpannerClientMock            func(spannerClient spannerclient.SpannerClient)
```
manitgupta, line 44: This seems weird. You shouldn't be needing to do this. Are you not able to set the `spannerClient` as part of the constructing the `spannerAccessor`? 

I think we should be constructing immutable instances of `spannerAccessor`. If you need a different spannerClient for any reason, create a different spannerAccessor.

cmd/import.go
```diff
@@ -57,39 +57,44 @@ func (cmd *ImportDataCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...
 	logger.Log.Debug(fmt.Sprintf("instanceId %s, dbName %s, schemaUri %s\n", cmd.instanceId, cmd.databaseName, cmd.schemaUri))
 
 	dbURI := fmt.Sprintf("projects/%s/instances/%s/databases/%s", cmd.project, cmd.instanceId, cmd.databaseName)
-	infoSchema, err := spanner.NewInfoSchemaImplWithSpannerClient(ctx, dbURI, constants.DIALECT_GOOGLESQL)
+	sp, err := spanneraccessor.NewSpannerAccessorClientImplWithSpannerClient(ctx, dbURI)
 	if err != nil {
-		logger.Log.Error(fmt.Sprintf("Unable to read Spanner schema %v", err))
+		logger.Log.Error(fmt.Sprintf("Unable to instantiate spanner client %v", err))
```
manitgupta, line 62: Earlier error was more readable for the user. This one is an internal detail of what failed. I will suggest reverting to the older error. 

Same comment for the other place as well.

cmd/import.go
```diff
@@ -57,39 +57,44 @@ func (cmd *ImportDataCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...
...
+		err := cmd.handleCsv(ctx, dbURI, dialect, sp)
 		if err != nil {
 			logger.Log.Error(fmt.Sprintf("Unable to handle Csv %v", err))
 			return subcommands.ExitFailure
 		}
 		return subcommands.ExitSuccess
+	case constants.MYSQLDUMP:
+		err := cmd.handleDatabaseDumpFile(ctx, dbURI, constants.MYSQLDUMP, constants.DIALECT_GOOGLESQL, sp)
+		if err != nil {
+			logger.Log.Error(fmt.Sprintf("Unable to handle MYSQL Dump %v", err))
```
manitgupta, line 79: Nit: Can we make this error message more actionable? For example, say reach out to Google or something similar.

cmd/import.go
```diff
@@ -111,6 +116,41 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
 
 }
 
+func (cmd *ImportDataCmd) handleDatabaseDumpFile(ctx context.Context, dbUri, sourceFormat string, dialect string, spannerAccessor spanneraccessor.SpannerAccessor) error {
+
+	importDump, err := import_file.NewImportFromDump(cmd.project, cmd.instanceId, cmd.databaseName, cmd.sourceUri, sourceFormat, spannerAccessor)
+	if err != nil {
+		return err
```
manitgupta, line 123: Nit: Return similar `fmt.errorf` error as done below?

cmd/import.go
```diff
@@ -57,39 +57,44 @@ func (cmd *ImportDataCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...
...
+		//TODO: handle POSTGRESQL
+		dialect := constants.DIALECT_GOOGLESQL
+		err := cmd.handleCsv(ctx, dbURI, dialect, sp)
 		if err != nil {
 			logger.Log.Error(fmt.Sprintf("Unable to handle Csv %v", err))
 			return subcommands.ExitFailure
 		}
 		return subcommands.ExitSuccess
+	case constants.MYSQLDUMP:
+		err := cmd.handleDatabaseDumpFile(ctx, dbURI, constants.MYSQLDUMP, constants.DIALECT_GOOGLESQL, sp)
```
manitgupta, line 77: `sp` does not need to be part of the interface for `handleCsv` or the `handleDatabaseDumpFile` methods. You can make it part of the method receiver. This will ensure that the `handleCsv` is testable and the method receiver pointer contains a reference to all the dependencies it needs to run (and mock in the UTs). 

As a general rule, interface of the methods should only get parameters that are closely tied to the business logic being executed, for e.g., in this case, it would be user inputs.


**Note 1**: I know that this will appear a little off to implement since the `ImportDataCmd` structure is being used to declare the inputs to the CLI command, and you will have to add `SpannerAccessor` there now. However, this is just how the existing code is structured (and is suboptimal). If you see, we have done the exact same thing I am talking about above in classes like `spannerAccessor` as well, and is the golang recommendation. Also, `SetFlags()` is also effectively the function to care about here anyway, so it shouldn't matter. 

**Note 2:** This also goes back to my original comment earlier where I had advised that the class which declares the CLI surface should just act as a command parser and delegate ALL of the business logic to a downstream struct. Because we are not doing that (and I am not necessarily asking you to do so), my `Note 1` is the next best trade-off to make. Atleast we get to keep the method signatures clean, if not the file responsibility and structure..

cmd/import.go
```diff
@@ -111,6 +116,41 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
...
+
+	schemaStartTime := time.Now()
+	conv, err := importDump.CreateSchema(dialect)
+	if err != nil {
+		return fmt.Errorf(fmt.Sprintf("can't create schema: %v\n", err))
+	}
+
+	// TODO: Only update database
+	err = spannerAccessor.CreateOrUpdateDatabase(ctx, dbUri, sourceFormat, conv, sourceFormat)
+	spannerAccessor.Refresh(ctx, dbUri)
```
manitgupta, line 136: Is there a use-case for this? `Refresh` was added due to a vey specific limitation with expression verification and I'm not sure if it applies here.

cmd/import.go
```diff
@@ -111,6 +116,41 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
...
+	logger.Log.Info(fmt.Sprintf("Schema creation took %f secs", elapsedTime.Seconds()))
+	if err != nil {
+		return err
+	}
+
+	err = importDump.ImportData(conv)
+
+	dataEndTime := time.Now()
+	elapsedTime = dataEndTime.Sub(schemaEndTime)
+	logger.Log.Info(fmt.Sprintf("Data import took %f secs", elapsedTime.Seconds()))
```
manitgupta, line 149: Trying to recall, is there some sort of a progress bar that appears on the screen when import is in progress? This will be important from a UX perspective.

Same comment applies to schema as well - Does schema creation take long? I feel like that there should be some notion of progress (like a timer or something) that should denote long-running operations so that the user knows that they are not stuck.

import_file/import_from_dump.go
```diff
@@ -0,0 +1,128 @@
...
+	// TODO: handle GCS
+	dumpReader, err := os.Open(dumpUri)
+	if err != nil {
+		if err != nil {
+			return nil, fmt.Errorf(fmt.Sprintf("can't read dump file: %s due to: %v", dumpUri, err))
+		}
+	}
+
+	schemaToSpanner := &common.SchemaToSpannerImpl{
+		ExpressionVerificationAccessor: &expressions_api.ExpressionVerificationAccessorImpl{SpannerAccessor: spannerAccessor},
```
manitgupta, line 56: Will the`mysqldump` flow call the expression verification API? If yes, this substantially increase the time required for schema conversion?

import_file/utils.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	"fmt"
+	spannerclient "github.com/GoogleCloudPlatform/spanner-migration-tool/accessors/clients/spanner/client"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/internal"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/logger"
+	"github.com/GoogleCloudPlatform/spanner-migration-tool/spanner/writer"
+	"os"
+	"sync/atomic"
+)
+
```
manitgupta, line 14: Can we avoid creating a `utils.go` file? In my experience, it becomes a dumping ground for all sorts of random, unrelated methods? 

For example, can we definge the `getBatchWriterWithConfig` method in the `writer.BatchWriter` struct itself? Do you see a risk?

manitgupta (commented):
Great effort on adding coverage and UTs! I do see some lines that are uncovered - 

1. https://screenshot.googleplex.com/9Mh6aWgaPQ54C27
2. https://screenshot.googleplex.com/9cs8CBBXa8Krby8

Will it be possible to cover these? Please let me know. 

Added some general comments - One a bigger comment on structure and responsibility. Please take a look.

accessors/spanner/mocks.go
pratickchokhani: This is required by ExpressionVerificationAccessorImpl. Without this, we will not be able to use the interface within ExpressionVerificationAccessorImpl.


https://screenshot.googleplex.com/7THJ3K5n45d7fn5

cmd/import.go
pratickchokhani: If spanner database is created in CreateOrUpdateDatabase, then. the SpannerClient within SpannerAccessor becomes stale. 

Hence, refresh is required.

cmd/import.go
pratickchokhani: From UX perspective in:
CLI: there are logs which depict the progress.

WEB Ui: this will be thought through when UI integration is done.

import_file/utils.go
pratickchokhani: Have moved it to writer.BatchWriter

cmd/import.go
pratickchokhani: Done.

cmd/import.go
pratickchokhani: Done.

pratickchokhani:
> Great effort on adding coverage and UTs! I do see some lines that are uncovered -
> 
> 1. https://screenshot.googleplex.com/9Mh6aWgaPQ54C27
> 2. https://screenshot.googleplex.com/9cs8CBBXa8Krby8
> 
> Will it be possible to cover these? Please let me know.
> 
> Added some general comments - One a bigger comment on structure and responsibility. Please take a look.

For the first section, have added UTs. For second, will be working as a separate PR.

cmd/import.go
manitgupta: Can you show me a screenshot of what the logs look like? Yes, web UI is not a concern right now. Only asking for UX of the CLI.

import_file/import_from_dump.go
manitgupta: Using expression verification API may not be viable for the kind of UX we want to provide for import flows. 

For context, this API makes calls to an actual spanner instance to verify the expressions that are part of the MySQL schema. For a small instance, something like a 100 processing unit instance (the free ones) this can take prohibitively long.

Since the goal of the import CLI is lightweight experimentation, I expect customers to be using such instances more so than migration, where a large instance is expected (and hence the above trade-off was acceptable).

FYI: @rohitwali

manitgupta:
> For second, will be working as a separate PR. 

Okay. Curious to know, is there a reason to push this out? Reason I'm asking is that next time it will not be in patch coverage and becomes best-effort. If this is about size of the PR, I am happy reviewing a slightly larger PR if it means we can improve the coverage in this PR itself.

cmd/import.go
pratickchokhani: https://screenshot.googleplex.com/9hKYBhMjZBzdqdZ

import_file/import_from_dump.go
pratickchokhani: The final goal is to remove the verifier but I am not making this part of this PR as it will require changes to SchemaToSpannerImpl, This will be done is a following PR.

cmd/import.go
pratickchokhani: As discussed over chat, we have created a follow up bug on this.

import_file/import_from_dump.go
pratickchokhani: I have updated the PR to include this.

pratickchokhani:
> > For second, will be working as a separate PR.
> 
> Okay. Curious to know, is there a reason to push this out? Reason I'm asking is that next time it will not be in patch coverage and becomes best-effort. If this is about size of the PR, I am happy reviewing a slightly larger PR if it means we can improve the coverage in this PR itself.

This was planned to reduce the size of this PR. I have added the respective unit tests to resolve this.
https://screenshot.googleplex.com/8cKXhLxXv5wjb5E
https://screenshot.googleplex.com/KNuv42hnRnpKJtC

manitgupta (approved):
Thank you for the rigor on the UTs and the openness to incorporate feedback. Appreciate it! Thanks Pratick! :)

PR #1113, opened 2025-05-03
switching schema format from csv to json

rohitwali:
1. switching schema format from csv to json
2. Increasing test coverage form import package to >80%


cmd/import_data.go
```diff
@@ -77,10 +100,118 @@ func (cmd *ImportDataCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...
...
+	}
+
+	if input.sourceFormat == constants.CSV && len(input.schemaUri) == 0 {
+		return fmt.Errorf("Please specify schemaUri using the --schema-uri parameter. Received  schemaUri: %v", input.sourceFormat)
+	}
+
+	return err
+}
+
+func isSpannerAccessible(ctx context.Context, projectID, instanceId, databaseName string) bool {
```
pratickchokhani, line 150: Instead of implementing a separate client creation, using NewSpannerAccessorClientImplWithSpannerClient will ensure that the client is initialised once.

It will also prevent code duplication.

cmd/import_data.go
```diff
@@ -102,7 +233,7 @@ func (cmd *ImportDataCmd) handleCsv(ctx context.Context, infoSchema *spanner.Inf
 
 	csvData := import_data.CsvDataImpl{ProjectId: cmd.project, InstanceId: cmd.instanceId,
 		TableName: cmd.tableName, DbName: cmd.databaseName, SourceUri: cmd.sourceUri, CsvFieldDelimiter: cmd.csvFieldDelimiter}
-	err = csvData.ImportData(ctx, infoSchema, dialect)
+	err = csvData.ImportData(ctx, infoSchema, dialect, internal.MakeConv(), &common.InfoSchemaImpl{}, &csv.CsvImpl{})
```
pratickchokhani, line 236: I feel Conv object creation should happen 1 level below.

import_data/csv_data.go
```diff
@@ -63,8 +61,7 @@ func (source *CsvDataImpl) ImportData(ctx context.Context, infoSchema *spanner.I
 	return err
 }
 
-func getConvObject(projectId, instanceId, dialect string) *internal.Conv {
-	conv := internal.MakeConv()
```
pratickchokhani, line 67: Why are we removing creation of Conv object here?

cmd/import_data.go
rohitwali: fair point. will update

cmd/import_data.go
rohitwali: this helps me interject the object from testing. otherwise it become hard to simulate alternate behaviour in tests

import_data/csv_data.go
rohitwali: I am passing this as a parameter. this help with unit tests

import_data/csv_data.go
```diff
@@ -29,15 +30,13 @@ type CsvDataImpl struct {
 	CsvFieldDelimiter string
 }
 
-func (source *CsvDataImpl) ImportData(ctx context.Context, infoSchema *spanner.InfoSchemaImpl, dialect string) error {
-	// TODO: start with single table imports
+func (source *CsvDataImpl) ImportData(ctx context.Context, spannerInfoSchema *spanner.InfoSchemaImpl, dialect string, conv *internal.Conv, commonInfoSchema common.InfoSchemaInterface, csv csv.CsvInterface) error {
```
pratickchokhani, line 33: common.InfoSchema interface should be used instead of spanner.InfoSchemaImpl 

Using struct will prevent mocking.

import_data/csv_schema.go
```diff
@@ -47,7 +45,7 @@ type PrimaryKey struct {
 func (source *CsvSchemaImpl) CreateSchema(ctx context.Context, dialect string, sp *spanneraccessor.SpannerAccessorImpl) error {
```
pratickchokhani, line 45: spanneraccessor.SpannerAccessor interface should be used.

import_data/csv_data.go
rohitwali: We are using Populate schema, which is not available in common infoschema. Adding it in common infoschema interface will force other sources to implement as well.

In this case mocking is not really a blocker and we are relying on stubs and deeper level mocks on SpannerClient

cmd/import_data.go
pratickchokhani: Can we reuse the client at line number 204?

pratickchokhani (approved):

PR #1120, opened 2025-05-09
chore(spanner): Minor refactoring:

pratickchokhani:
- Used SpannerAccessor interface instead of SpannerAccessorImpl.
- File validation to be happening on creation of FileReader.


cmd/import.go
```diff
@@ -103,10 +101,6 @@ func (cmd *ImportDataCmd) Execute(ctx context.Context, f *flag.FlagSet, args ...
 }
 
 func validateInputRemote(ctx context.Context, input *ImportDataCmd) error {
-	if !isSpannerAccessible(ctx, input.project, input.instanceId, input.databaseName) {
```
rohitwali, line 106: Can we add the check back? Its better to be explicit about validation in the code and it reads well

pratickchokhani: Done. I have created a separate function. This will enable to pass the SpannerAccessor object.

rohitwali (approved):

PR #1123, opened 2025-05-13
chore(spanner): Added support of PG_DUMP source format and Postgresql dialect in IMPORT command.

pratickchokhani:


cmd/import.go
```diff
@@ -130,12 +130,27 @@ func createDatabase(ctx context.Context, dbURI, dialect string, spannerAccessor
...
 	spannerAccessor, err := import_file.NewSpannerAccessor(ctx, dbURI)
 	if err != nil {
 		logger.Log.Error(fmt.Sprintf("Unable to instantiate spanner client %v", err))
 		return nil, fmt.Errorf("unable to instantiate spanner client %v", err)
 	}
+
+	skipDialectValidation := os.Getenv("IMPORT_CMD_SKIP_DIALECT_VALIDATION")
+
+	// Only used for Emulator integration testing.
+	// TODO: Remove once Cl for fix within Emulator is release.
```
rohitwali: 1. pl link the tracking bug.
2. we recently decided to push database creation to SMT from gcloud. can you pl rebase with latest? would like to see what happens if the database doesnt exist?

cmd/import.go
```diff
@@ -166,7 +181,7 @@ func getDialectWithDefaults(dialect string) string {
 	case constants.DIALECT_POSTGRESQL:
 		return dialect
 	default:
-		logger.Log.Warn(fmt.Sprintf("Dialect passed is %s . Defaulting to %s", dialect, constants.DIALECT_GOOGLESQL))
```
rohitwali: this really should be warn no? do you see execution flow reaching here normally?

cmd/import.go
pratickchokhani: 1. Done.
2. Have validated the database creation flow.

cmd/import.go
pratickchokhani: Dialect field is optional. Hence, for GOOGLESQL, we expect the dialect field to be empty.

rohitwali: arent we defaulting in line 66?

cmd/import.go
rohitwali: 2. wont the current flow fail at line 147 if the database doesnt exist?

pratickchokhani: 2. Have fixed the issue.

cmd/import.go
pratickchokhani: Make sense. Have reverted the change.

cmd/import.go
```diff
@@ -130,12 +130,32 @@ func createDatabase(ctx context.Context, dbURI, dialect string, spannerAccessor
...
+
+		dialect, err := spannerAccessor.GetDatabaseDialect(ctx, dbURI)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get database dialect %v", err)
+		}
+
+		if dialect != targetDialect {
+			return nil, fmt.Errorf("database dialect is different for target dialect. Provided dialect: %s, Database dialect: %s", targetDialect, dialect)
+		}
+	}
```
rohitwali: would it make sense to move this to create database method?

pratickchokhani: I wanted to discuss this with you. I feel create database should not be called until schema is validated. 2 reasons for this:
1. If there is an error, then the database will be created unnecessarily.
2. While creating the database, we send the DDL statements in the same RPC. This will be more optimised.

rohitwali: Whenever/however we create the database, this check still needs to happen after that

pratickchokhani: Got it. Have updated the code as discussed.

rohitwali (approved):

PR #1136, opened 2025-06-04
Bug fix in Mysql check constraint

pratickchokhani:
Bug fix in Mysql check constraint and unit tests added.


internal/helpers.go
```diff
@@ -147,6 +147,15 @@ func GetSrcFkFromId(fks []schema.ForeignKey, fkId string) (schema.ForeignKey, er
 	return schema.ForeignKey{}, fmt.Errorf("foreign key not found")
 }
 
+func GetCheckConstraintFromId(checkConstraints []ddl.CheckConstraint, checkConstraintName string) (ddl.CheckConstraint, error) {
```
shreyakhajanchi: since this method is used only in test assertions , it might be better to move it to `test_util.go`

internal/test_util.go
```diff
@@ -32,9 +32,20 @@ func AssertSpSchema(conv *Conv, t *testing.T, expectedSchema, actualSchema map[s
...
 		assertSpFk(conv, t, tableId, expectedTable.ForeignKeys, actualSchema[tableId].ForeignKeys)
 		assertSpIndexes(conv, t, tableId, expectedTable.Indexes, actualSchema[tableId].Indexes)
+		assertCheckConstraints(conv, t, tableId, expectedTable.CheckConstraints, actualSchema[tableId].CheckConstraints)
 	}
 }
 
+func assertCheckConstraints(conv *Conv, t *testing.T, tableId string, expectedCheckConstraints, actualCheckConstraints []ddl.CheckConstraint) {
+	assert.Equal(t, len(expectedCheckConstraints), len(actualCheckConstraints))
+	for _, expectedCheckConstraint := range expectedCheckConstraints {
+		actualCheckConstraint, err := GetCheckConstraintFromId(actualCheckConstraints, expectedCheckConstraint.Name)
```
shreyakhajanchi: maybe rename the method to `GetCheckConstraintFromName` since we are passing name here

shreyakhajanchi (commented):
Emulator test seems to be failing at: `testing/import_cmd/integration_test.go:151`

internal/helpers.go
pratickchokhani: Done.

internal/test_util.go
pratickchokhani: Done.

shreyakhajanchi (approved):

PR #1145, opened 2025-06-23
Add locks to avoid race condition during schema discovery

Deep1998:
b/425603969


manitgupta (approved):
Please create a backlog bug to refactor schema migration code flow to not use shared objects at all. The code in its present state is very brittle and prone to crashes.

PR #1146, opened 2025-06-25
chore(spanner): Added self hosted runner to integration tests.

pratickchokhani:


.github/workflows/integration-tests-against-emulator.yaml
```diff
@@ -183,12 +164,14 @@ jobs:
       - name: Cypress run
         run: |
           cd ui
+          npm install cypress --save-dev
           npx cypress run
 
       - uses: actions/setup-go@v2
         with:
-          go-version: "1.24.3"
+          go-version: "1.23.0"
```
darshan-sj, line 172: Why are we decreasing go version?

darshan-sj (approved):

.github/workflows/integration-tests-against-emulator.yaml
pratickchokhani: Changing this to match with the the Build version set in go.mod.

PR #1155, opened 2025-07-03
Added Function testing for PG DUMP in Import command

pratickchokhani:


testing/import_cmd/integration_test.go
```diff
@@ -74,60 +83,93 @@ func onlyRunForEmulatorTest(t *testing.T) {
...
+			"-source-uri=%s -database-dialect=%s",
+		testData.sourceFormat, projectID, instanceID, testData.dbName, dumpFilePath, dialect)
+	fmt.Printf("Executing: %s\n", args)
+	err := common.RunCommand(args, projectID)
+	assert.NoError(t, err)
 
-			// TODO validation to be added.
-		})
-	}
+	// TODO validation to be added.
```
shreyakhajanchi, line 172: why are we just importing the db without any validations?

pratickchokhani: For validations, we have unit tests. This is just to ensure that there are no errors while importing data.

PR #1169, opened 2025-07-30
fix for handling special characters in connection fields

shreyakhajanchi:
Please note that this doesn't solve for `"` in any of the fields. We will have to imply another solution like using json in such a case.

### DIFF
--- a/webv2/web.go
+++ b/webv2/web.go
@@ -33,10 +33,10 @@ import (
 	"time"
 
 	instance "cloud.google.com/go/spanner/admin/instance/apiv1"
+	cassandraaccessor "github.com/GoogleCloudPlatform/spanner-migration-tool/accessors/cassandra"
 	cc "github.com/GoogleCloudPlatform/spanner-migration-tool/accessors/clients/cassandra"
 	storageclient "github.com/GoogleCloudPlatform/spanner-migration-tool/accessors/clients/storage"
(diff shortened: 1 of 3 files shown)


webv2/web.go
```diff
@@ -830,15 +830,15 @@ func getSourceAndTargetProfiles(sessionState *session.SessionState, details type
...
+		sourceProfileString = fmt.Sprintf("\"host=%v\",\"port=%v\",\"user=%v\",\"password=%v\",\"keyspace=%v\",\"datacenter=%v\"",
 			sourceDBConnectionDetails.Host,
 			sourceDBConnectionDetails.Port,
 			sourceDBConnectionDetails.User,
 			sourceDBConnectionDetails.Password,
 			sessionState.DbName,
 			sourceDBConnectionDetails.DataCenter)
 	} else {
-		sourceProfileString = fmt.Sprintf("host=%v,port=%v,user=%v,password=%v,dbName=%v",
+		sourceProfileString = fmt.Sprintf("\"host=%v\",\"port=%v\",\"user=%v\",\"password=%v\",\"dbName=%v\"",
```
darshan-sj, line 841: We are adding quotes here. But the parsing logic in common.go doesn't remove these quotes. Does this code go through the logic written in above in `profiles/common.go` ? Have you tested this?

shreyakhajanchi: Yes, csvreader takes care of it. I have tested it, will test once more

darshan-sj (approved):

PR #1170, opened 2025-07-30
Fixing primary key sort order for source

shreyakhajanchi:
image


ui/src/app/components/object-detail/object-detail.component.ts
```diff
@@ -1141,6 +1144,35 @@ export class ObjectDetailComponent implements OnInit {
     }
   }
 
+  setSrcPkOrder() {
+    if (
+        this.currentObject &&
+        this.conv.SrcSchema[this.currentObject!.id]?.PrimaryKeys.length == this.pkData.length
```
bharadwaj-aditya: why is this logic based on length ?

PR #1174, opened 2025-08-07
Overrides support in smt

Deep1998:


internal/overrides.go
```diff
@@ -0,0 +1,50 @@
...
+package internal
+
+// OverridesFile represents the overrides file format containing schema mapping information
+type OverridesFile struct {
+	RenamedTables  map[string]string            `json:"renamedTables"`
+	RenamedColumns map[string]map[string]string `json:"renamedColumns"`
+}
+
+// ExtractOverridesFromConv extracts renamed tables and columns from the conv object
+func ExtractOverridesFromConv(conv *Conv) *OverridesFile {
```
bharadwaj-aditya, line 24: We should also be supporting dropped columns. Is that in a different PR ?

Deep1998: Dropped/extra columns are not supported in override files. Overrides only contain renames, dropped extra columns are handled in the templates using the ddl objects and custom jars.

PR #1177, opened 2025-08-12
Added randomization to temp Db.

pratickchokhani:
Multiple integration tests executes in parallel trying to create the TEMP Database. Hence have the same database name can cause conflict.
Fixes: b/437803226

### DIFF
--- a/expressions_api/expression_verify.go
+++ b/expressions_api/expression_verify.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	mtrand "math/rand"
 	"sync"
+	"time"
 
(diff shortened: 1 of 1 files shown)


manitgupta (commented):
1. Why does this error occur in the first place? Will creating a random temp DB lead to leaking resources?

2. I think the API would clean up the temp DB before starting up. This used to work because the temp DB name was constant. Will that flow still work?

pratickchokhani:
> 1. Why does this error occur in the first place? Will creating a random temp DB lead to leaking resources?
> 2. I think the API would clean up the temp DB before starting up. This used to work because the temp DB name was constant. Will that flow still work?

1. On successful execution, the db gets deleted.
2. It will still clean it up as for the DB name is stored in SpannerClient.

PR #1182, opened 2025-08-13
[Assessments] Changing IsNull to NotNull

aasthabharill:
Fixes b/438400729
To avoid cases where IsNull is not initialized and picks up "false" as the default value resulting in unnecessary addition of "NOT NULL" constraint


bharadwaj-aditya (approved):
LGTM

PR #1187, opened 2025-08-20
added NUMERIC as an option as target datatype for BIGINT

sm745052:
Fixes b/368291549
Added option to convert BIGINT to NUMERIC. Will be helpful if one has BIGINT UNSIGNED column in source with values that donot fit in spanner INT64. 

- All Tests pass

- UI shows Numeric as an option
Screenshot 2025-08-20 at 12 21 27 PM

- works as expected from CLI as well. (tested with schema-migration dry run)

### DIFF
--- a/sources/mysql/toddl.go
+++ b/sources/mysql/toddl.go
@@ -137,6 +137,8 @@ func toSpannerTypeInternal(srcType schema.Type, spType string) (ddl.Type, []inte
 		switch spType {
 		case ddl.String:
 			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.Widened}
+		case ddl.Numeric:
+			return ddl.Type{Name: ddl.Numeric}, []internal.SchemaIssue{internal.Widened}
 		default:
 			return ddl.Type{Name: ddl.Int64}, nil
 		}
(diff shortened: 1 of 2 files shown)


sources/mysql/toddl.go
```diff
@@ -137,6 +137,8 @@ func toSpannerTypeInternal(srcType schema.Type, spType string) (ddl.Type, []inte
 		switch spType {
```
VardhanThigle, line 137: We might want to do this only for `bigint unsigned`

sm745052: since we are only adding it as an option, having it for BIGINT too might not have adverse effect.

Also, in the current state, we donot identify if BIGINT it is signed or unsigned. Identification of that will be a different PR.

VardhanThigle: We are not adding a warning for `BigInt Unsigned` to `Int64` conversion in this PR.

If you are planning to add it in a subsequent PR, whenever that warning is added please ensure that it is added only to `BigInt Unsigned`.

Please resolve the comment thread once you create a tracking bug for the same.

sm745052: added a tracking bug for the same.

b/439992639

PR #1189, opened 2025-08-20
Report generator for query assessment

shreyakhajanchi:


assessment/assessment_engine.go
```diff
@@ -167,6 +171,21 @@ func performQueryAssessment(ctx context.Context, collectors assessmentCollectors
 	return translationResult, nil
 }
 
+func fetchSpannerTableNames(conv *internal.Conv, tableNames []string) []string {
+	spannerTableNames := make([]string, 0, len(tableNames))
+	for _, tableName := range tableNames {
+		tableId, err := internal.GetTableIdFromSrcName(conv.SrcSchema, tableName)
+		if err != nil {
+			logger.Log.Warn("error getting table id from source name", zap.String("tableName", tableName), zap.Error(err))
+			spannerTableNames = append(spannerTableNames, internal.GetSpannerValidName(conv, tableName))
```
bharadwaj-aditya: why is the spanner table name added if it could not be fetched ?

assessment/assessment_engine.go
```diff
@@ -167,6 +171,21 @@ func performQueryAssessment(ctx context.Context, collectors assessmentCollectors
...
+			continue
+		}
+		spannerTableName, _ := internal.GetSpannerTable(conv, tableId)
```
bharadwaj-aditya: what happens in scenarios where the spanner table does not exist ?

assessment/report_generator.go
```diff
@@ -807,3 +829,145 @@ func populateSequenceInfo(sequenceAssessmentOutput map[string]ddl.Sequence, tabl
...
+	h := sha256.New()
+	h.Write([]byte(normalized))
+	return "q" + hex.EncodeToString(h.Sum(nil))[:8]
+}
+
+func codeChangeEffort(complexity string) string {
+	switch strings.ToLower(complexity) {
+	case "simple":
+		return "Low"
+	case "moderate", "medium":
```
bharadwaj-aditya, line 944: This block seems unnecessary. We should return standard values while generating this value.

Can be left as is for now, can take this up as a follow up fix.

assessment/report_generator.go
```diff
@@ -807,3 +829,145 @@ func populateSequenceInfo(sequenceAssessmentOutput map[string]ddl.Sequence, tabl
...
+		return "Medium"
+	case "complex":
+		return "High"
+	default:
+		return ""
+	}
+}
+
+func getQueryType(normalized string) string {
+	s := strings.TrimSpace(strings.ToUpper(normalized))
```
bharadwaj-aditya: this probably belongs in the assessment engine. Every output would require this information.

assessment/utils/assessment_output.go
```diff
@@ -193,19 +193,19 @@ type PerformanceAssessmentOutput struct {
...
-	SnippetId               string             `json:"snippet_id,omitempty"`
-	NumberOfQueryOccurances int                `json:"number_of_query_occurances,omitempty"`
-	TablesAffected          []string           `json:"tables_affected"`
+	OriginalQuery           string   `json:"old_query"`
+	NormalizedQuery         string   `json:"normalized_query"`
+	SpannerQuery            string   `json:"new_query"`
+	Explanation             string   `json:"explanation"`
+	Complexity              string   `json:"complexity"`
+	TranslationError        string   `json:"translation_error,omitempty"`
+	Source                  string   // "app_code" or "performance_schema" or "app_code,performance_schema"
```
bharadwaj-aditya: rename this field to assessment source. To disambiguate from the db source.

assessment/utils/constants.go
```diff
@@ -22,3 +22,200 @@ const (
 	GEMINI_PRO_MODEL           string = "gemini-1.5-pro-002"
 	GEMINI_FLASH_MODEL         string = "gemini-2.0-flash-001"
 )
+
+// SupportedFunctions is a map of supported Spanner GoogleSQL functions.
+var SupportedFunctions = map[string]bool{
```
bharadwaj-aditya, line 27: is there a library where we can get these ? Any updates might make this stale.

assessment/assessment_engine.go
shreyakhajanchi: it generates a valid spanner table name

assessment/report_generator.go
shreyakhajanchi: noted

assessment/utils/constants.go
shreyakhajanchi: I couldn't find a library for this , one thing we can do is create some parser to go through the documentation for fetching this list

assessment/assessment_engine.go
shreyakhajanchi: In cases where there are queries that belong to a database different from the one specified in the source profile, we would want an equivalent spanner name , therefore added this logic. What is your suggestion in this case?

bharadwaj-aditya: We should mark this query as not supported - as it is a cross database join

assessment/assessment_engine.go
bharadwaj-aditya: If the table does not exist, how will a name be generated ?

PR #1194, opened 2025-08-25
Bigint unsigned handling

sm745052:
Fixes b/439992639

> Added identification of `bigint unsigned` as a different datatype from `bigint` (both from infoschema and dump file)for different handling
> Added an generalized warning for possible overflow
> raises possible overflow warning when converting `bigint unsigned` to `int64`

- all tests pass
- changes work in UI
Screenshot 2025-08-25 at 9 59 38 AM

Screenshot 2025-08-25 at 9 56 55 AM

### DIFF
--- a/sources/mysql/infoschema_test.go
+++ b/sources/mysql/infoschema_test.go
@@ -229,6 +229,7 @@ func TestProcessSchemaMYSQL(t *testing.T) {
 				{"tz", "timestamp", "timestamp", "YES", nil, nil, nil, nil, nil},
 				{"vc", "varchar", "varchar", "YES", nil, nil, nil, nil, nil},
 				{"vc6", "varchar", "varchar(6)", "YES", nil, 6, nil, nil, nil},
+				{"bu", "bigint", "bigint(20) unsigned", "YES", nil, nil, 20, 0, nil},
 			},
 		},
 		// db call to fetch index happens after fetching of column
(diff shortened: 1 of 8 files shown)


sources/mysql/infoschema_test.go
```diff
@@ -229,6 +229,7 @@ func TestProcessSchemaMYSQL(t *testing.T) {
 				{"tz", "timestamp", "timestamp", "YES", nil, nil, nil, nil, nil},
 				{"vc", "varchar", "varchar", "YES", nil, nil, nil, nil, nil},
 				{"vc6", "varchar", "varchar(6)", "YES", nil, 6, nil, nil, nil},
+				{"bu", "bigint", "bigint(20) unsigned", "YES", nil, nil, 20, 0, nil},
```
manitgupta, line 232: Can you confirm if a UT exists for the `signed bigint` as well? Want to ensure that code path works as expected.

manitgupta (commented):
Code LGTM!

There was a conversation around having follow up AIs where we should ask the user to explicitly acknowledge the possible overflow before allowing them to proceed. 

Could you please ensure that those backlog bugs are created?

We might need to look at how to prioritize the new backlog items that @sm745052 will create. Without that, its possible that the overflow is missed by the user without realizing it.

CC: @rohitwali @bharadwaj-aditya

sources/mysql/toddl.go
```diff
@@ -133,13 +133,16 @@ func toSpannerTypeInternal(srcType schema.Type, spType string) (ddl.Type, []inte
 			// capabilities between MySQL and Spanner NUMERIC.
 			return ddl.Type{Name: ddl.Numeric}, nil
 		}
-	case "bigint":
+ 	case "bigint", "bigint unsigned":
```
VardhanThigle: [nit] [style guide] Please create a separate case (even if that might mean repeating the code a bit) and avoid adding `if` inside the switch case especially and if over the argument of the switch case.

sources/mysql/mysqldump.go
```diff
@@ -618,6 +618,11 @@ func getTypeModsAndID(conv *internal.Conv, columnType string) (string, []int64)
 	if strings.Contains(id, " ") {
 		id = strings.TrimSuffix(columnType, " BINARY")
 	}
+	// Bigint unsigned comes as bigint(20) UNSIGNED in columnType and is treated as bigint in id.
+	// An extra check is added to respect to the unsigned nature. 
+	if id == "bigint" && strings.Contains(strings.ToUpper(columnType), "UNSIGNED") {
+		id = strings.Join([]string{id, "unsigned"}, " ")
```
VardhanThigle: [nit][readability] Can we just return the string `"bigint unsigned"`? The `if` here is mandating the `bigInt` part so the join might be redundant.

If you would prefer a more thorough normalization of the mySql types, you could instead refer to [bulk code](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/5a65186f293ef9dad9d451bf968da0fab88a4a1a/v2/sourcedb-to-spanner/src/main/java/com/google/cloud/teleport/v2/source/reader/io/jdbc/dialectadapter/mysql/MysqlDialectAdapter.java#L520) but I guess that will need lot more changes in this part of the codebase than the context of this PR and could be deferred.

sources/mysql/infoschema_test.go
sm745052: yes a similar test for `signed bigint` (or just `bigint`) does exist (column_name `bs`).

sources/mysql/mysqldump.go
sm745052: thats right. made it that way.

sources/mysql/toddl.go
sm745052: noted

PR #1200, opened 2025-09-03
Capping size for (`""`|`"medium"`|`"tiny"`|`"long"`)Blob to BYTES mappings reflect mySql Storage limits.

VardhanThigle:
Fixes b/442234786

> When mapping blob family of types to spanner Bytes set the default max based on the Max storage supported in MySql

- [X] Tests pass
- [x] Appropriate changes to Documentation are included in PR


docs/data-types/mysql.md
```diff
@@ -24,28 +24,35 @@ There are also nuances to handling certain specific data types. These are captur
...
-| `VARCHAR`                                         | `STRING(MAX)`    |                                                          |
-| `VARCHAR(N)`                                      | `STRING(N)`      | differences in treatment of fixed-length character types |
+|                  **MySQL Type**                   |  **Spanner Type**   | **Notes**                                                |
+|:-------------------------------------------------:|:-------------------:|:--------------------------------------------------------:|
+|        `BOOL`, `BOOLEAN`,<br/>`TINYINT(1)`        |       `BOOL`        |                                                          |
+|                     `BIGINT`                      |       `INT64`       |                                                          |
+|               `BINARY`, `VARBINARY`               |    `BYTES(MAX)`     |                                                          |
+|                      `BLOB`                       |   `BYTES(65,535)`   |                                                          |
```
darshan-sj: nit: Whenever I see commas inside parentheses it looks like different arguments - Like `BYTES(65,535)` looks like BYTES has two variables with values 65 and 535. Should we omit the commas here? WDYT?

darshan-sj (approved):
LGTM

docs/data-types/mysql.md
VardhanThigle: Thanks, makes sense. Removed the comma.

sm745052 (commented):
LGTM

PR #1229, opened 2025-11-11
Allow specifying the Spanner DB default timezone as part of schema migration

nmemond:
This PR adds the ability to specify the default timezone for the target Spanner database as part of schema migration (applies to both the `schema` and `schema-and-data` commands as well as the web UI migration flow).

This is accomplished via a new (optional) flag in the `targetProfile` command-line parameter, `defaultTimeZone`. This new flag accepts an IANA timezone identifier, and is validated using Go's `time.LoadLocation`.

On the Web UI side, a new dropdown field has been added to the target details form used when preparing the migration. The dropdown is pre-populated with the list of IANA timezones (from `Intl.supportedValuesOf('timeZone')`), and supports a basic search functionality:
image
image

If the database being migrated to already contains tables, a warning is logged indicating that Spanner does not allow updating the default timezone for a non-empty database that already includes tables, and the `defaultTimezone` value is ignored.

Note: I still need to create/update and run tests for the other three migration pillars to ensure that timestamp and datetime values are still migrated correctly.

- [X] Tests pass (Note: the UI tests are still flaky and fail intermittently...)

### DIFF
--- a/testing/mysql/integration_test.go
+++ b/testing/mysql/integration_test.go
@@ -287,7 +287,7 @@ func TestIntegration_MySQLDUMP_ReservedKeyword(t *testing.T) {
 	dumpFilePath := "../../test_data/mysql_dump_reserved_keyword.sql"
 	filePrefix := filepath.Join(tmpdir, dbName)
 	dbURI := fmt.Sprintf("projects/%s/instances/%s/databases/%s", projectID, instanceID, dbName)
-	runSchemaAndDataSubcommand(t, dbName, dbURI, filePrefix, dumpFilePath)
+	runSchemaAndDataSubcommand(t, dbName, dbURI, filePrefix, dumpFilePath, "")
 
 	defer dropDatabase(t, dbURI)
 }


profiles/target_profile.go
```diff
@@ -146,6 +147,9 @@ func NewTargetProfile(s string) (TargetProfile, error) {
 	if dialect, ok := params["dialect"]; ok {
 		sp.Dialect = strings.ToLower(dialect)
 	}
+	if defaultTimezone, ok := params["defaultTimezone"]; ok {
+		sp.DefaultTimezone = defaultTimezone
+	}
```
nmemond, line 152: This is where the new flag is parsed out of the target profile string.

profiles/target_profile.go
```diff
@@ -157,6 +161,13 @@ func NewTargetProfile(s string) (TargetProfile, error) {
 		return TargetProfile{}, fmt.Errorf("found empty string for instance. please specify instance (spanner instance) in the target-profile")
 	}
 
+	if sp.DefaultTimezone != "" {
+		_, err := time.LoadLocation(sp.DefaultTimezone)
+		if err != nil {
+			return TargetProfile{}, err
+		}
+	}
```
nmemond, line 169: Validation for the new flag: checks to ensure that the specified timezone is actually an IANA timezone. If not, an error is returned.

spanner/ddl/ast.go
```diff
@@ -772,3 +781,44 @@ func (seq Sequence) PGPrintSequence(c Config) string {
 
 	return seqDDL
 }
+
+type DatabaseOptions struct {
+	DbName string
+	DefaultTimezone string
+}
```
nmemond, line 814: I elected to add this new structure instead of just the default timezone to make adding any other database options to the schema migration easier in the future.

spanner/ddl/ast.go
```diff
@@ -772,3 +781,44 @@ func (seq Sequence) PGPrintSequence(c Config) string {
...
+	}
+
+	var options []string
+	if dbOptions.DefaultTimezone != "" {
+		options = append(options, fmt.Sprintf("default_time_zone = '%s'", dbOptions.DefaultTimezone))
+	}
+
+	dbOptionsDdl := ""
+	if len(options) > 0 {
+		dbOptionsDdl = fmt.Sprintf("ALTER DATABASE %s SET OPTIONS (%s)", dbName, strings.Join(options, ", "))
```
nmemond, line 829: The spanner dialect allows for setting multiple options in a single DDL statement. Of course if no options are specified, we should not print the statement at all.

spanner/ddl/ast.go
```diff
@@ -772,3 +781,44 @@ func (seq Sequence) PGPrintSequence(c Config) string {
...
+	}
+
+	var options []string
+	if dbOptions.DefaultTimezone != "" {
+		options = append(options, fmt.Sprintf("spanner.default_time_zone = '%s'", dbOptions.DefaultTimezone))
+	}
+
+	var dbOptionsDdls []string
+	for _, option := range options {
+		dbOptionsDdls = append(dbOptionsDdls, fmt.Sprintf("ALTER DATABASE %s SET %s", dbName, option))
```
nmemond, line 847: On the other hand, the postgres dialect only allows a single option to be set in a single DDL statement, so setting additional options would need to be done in separate statements. Note the return type of this function is an array of strings, as opposed to a single string like on the spanner dialect side.

spanner/ddl/ast.go
```diff
@@ -658,9 +658,18 @@ func GetSortedTableIdsBySpName(s Schema) []string {
...
 	var ddl []string
 
+	if c.SpDialect == constants.DIALECT_POSTGRESQL {
+		ddl = append(ddl, dbOptions.PGPrintDatabaseOptions()...)
+	} else {
+		dbOptionsDdl := dbOptions.PrintDatabaseOptions()
+		if (dbOptionsDdl != "") {
+			ddl = append(ddl, dbOptionsDdl)
+		}
+	}
```
nmemond, line 697: I elected to keep the DDL for setting the database options here with the other DDL statements, even if it isn't strictly a "migration" from the source DB (in that the user is specifying it as opposed to it being sourced from the source DB).

This does mean that the signature of the `GetDDL` method has changed slightly.

accessors/spanner/spanner_accessor.go
```diff
@@ -378,6 +378,10 @@ func (sp *SpannerAccessorImpl) CreateOrUpdateDatabase(ctx context.Context, dbURI
 	if err != nil {
 		return err
 	}
+	if conv.DatabaseOptions.DefaultTimezone != "" && len(tablesExistingOnSpanner) > 0 {
+		logger.Log.Warn("Spanner database already contains tables, can not set default time zone of '" + conv.DatabaseOptions.DefaultTimezone + "'. See https://docs.cloud.google.com/spanner/docs/set-default-time-zone#limitations.")
+		conv.DatabaseOptions.DefaultTimezone = ""
+	}
```
nmemond, line 384: This is the check for existing tables: we can't update the default timezone if the target DB already contains tables, so we have to ignore the value passed in by the user.

internal/convert.go
```diff
@@ -58,6 +58,7 @@ type Conv struct {
 	SpProjectId        string                  // Spanner Project Id
 	SpInstanceId       string                  // Spanner Instance Id
 	Source             string                  // Source Database type being migrated
+	DatabaseOptions    ddl.DatabaseOptions
```
nmemond, line 61: To avoid having to pass the target profile through a whole bunch of methods, I elected to add the new `DatabaseOptions` model to the `Conv` struct. This does align with the other sources of DDL statements, so it felt the most appropriate.

conversion/conversion.go
```diff
@@ -76,15 +77,21 @@ type ConvImpl struct{}
...
 	case constants.PGDUMP, constants.MYSQLDUMP:
 		expressionVerificationAccessor, _ := expressions_api.NewExpressionVerificationAccessorImpl(context.Background(), targetProfile.Conn.Sp.Project, targetProfile.Conn.Sp.Instance)
-		return schemaFromSource.SchemaFromDump(targetProfile.Conn.Sp.Project, targetProfile.Conn.Sp.Instance, sourceProfile.Driver, targetProfile.Conn.Sp.Dialect, ioHelper, &ProcessDumpByDialectImpl{ExpressionVerificationAccessor: expressionVerificationAccessor})
+		conv, err = schemaFromSource.SchemaFromDump(targetProfile.Conn.Sp.Project, targetProfile.Conn.Sp.Instance, sourceProfile.Driver, targetProfile.Conn.Sp.Dialect, ioHelper, &ProcessDumpByDialectImpl{ExpressionVerificationAccessor: expressionVerificationAccessor})
 	default:
 		return nil, fmt.Errorf("schema conversion for driver %s not supported", sourceProfile.Driver)
 	}
+	conv.DatabaseOptions = ddl.DatabaseOptions{
+		DefaultTimezone: targetProfile.Conn.Sp.DefaultTimezone,
+	}
```
nmemond, line 93: This seemed like the best place to set the database options in the `conv` object based on what was passed in to the target profile.

cmd/utils.go
```diff
@@ -162,6 +163,11 @@ func MigrateDatabase(ctx context.Context, migrationProjectId string, targetProfi
 	}
 	defer adminClient.Close()
 	defer client.Close()
+	// Before this point, the actual DB name to use isn't finalized...
+	conv.DatabaseOptions = ddl.DatabaseOptions{
+		DbName: targetProfile.Conn.Sp.Dbname,
```
nmemond, line 168: Before we get here, the `targetProfile.Conn.Sp.Dbname` field is potentially empty (in the case where the user didn't specify their own DB name, and so a generated name is used). At this point, the generated name (if used) has been set in this field.

ui/src/app/components/prepare-migration/prepare-migration.component.html
```diff
@@ -343,6 +343,10 @@ <h3>Source database details:</h3>
                 <span><b>Spanner Instance: </b>{{ instance }} (Nodes: {{nodeCount}}, Processing Units:
                     {{processingUnits}})</span>
             </p>
+            <p class="point">
+                <span class="bullet">5</span>
+                <span><b>Default Timezone: </b>{{ targetDetails.DefaultTimezone || 'None' }}</span>
+            </p>
```
nmemond, line 349: This displays the specified timezone (or None if not specified) in the target details summary.
image
image

ui/src/app/components/target-details-form/target-details-form.component.html
```diff
@@ -17,6 +17,16 @@ <h2>Spanner Database Details<mat-icon class="configure"
...
     <br>
+    <ng-select appearance="outline"
+               [items]="timezones"
+               [searchFn]="customSearchFn"
+               placeholder="Default Timezone"
+               [clearable]="true"
+               formControlName="defaultTimezone"
+               bindLabel="timezone"
+               bindValue="timezone">
+    </ng-select>
```
nmemond, line 28: This is the new dropdown for default timezone selection (see screenshots in the description).

ui/src/app/components/target-details-form/target-details-form.component.scss
```diff
@@ -7,3 +7,16 @@
...
+
+::ng-deep {
+  .ng-select[appearance=outline] {
+    &.ng-select-single .ng-select-container {
+      .ng-value-container .ng-input {
+        position: absolute;
+        bottom: auto;
+        left: auto;
+        padding: 0 .25em;
+      }
```
nmemond, line 19: There was some weird alignment with the text input in the select textbox, so I ended up having to do this to get it to align properly.

ui/tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "importHelpers": true,
     "target": "es2017",
     "module": "es2020",
-    "lib": ["es2020", "dom"]
+    "lib": ["es2022", "dom"]
```
nmemond, line 21: This is to allow the call to `Intl.supportedValuesOf('timeZone')`.

ui/src/app/components/prepare-migration/prepare-migration.component.html
VardhanThigle: Would it be possible for Spanner migration tool mirror the default time of of Spanner if the user has not chosen or explicitly set a default value?

nmemond: Yep, that can be done.

My only concern would be if the default timezone ever changes on the Spanner side, we'd likely want to change it here as well otherwise it could be a bit confusing/misleading, but there would likely still be a period of time where there's a misalignment between Spanner and this tool.

Let me know your preference, I can make the change if needed.

PR #1233, opened 2025-11-19
Implement migration of SERIAL columns from postgres sources, mapping to IDENTITY columns in Spanner

nmemond:
This adds support for migrating SERIAL columns from postgres sources to IDENTITY columns in Spanner.

Note that the SERIAL types in postgres are essentially just an alias, and once the table is created, querying the data types of SERIAL columns does not actually return SERIAL (it returns whatever the corresponding integer data type is). As such, identifying SERIAL columns is more involved than identifying AUTO_INCREMENT columns in MySQL.

SERIAL columns have the following three properties:
- They have an integer data type (integer, bigint, smallint)
- They have a DEFAULT value generated by a sequence using `nextval()`
- They own the sequence mentioned above

At the schema-level, we check `pg_attribute` and `pg_attrdef` to find all columns that meet the above criteria for a given table. Note that the function `pg_get_serial_sequence()` alone is not sufficient, since it doesn't check that the column actually uses the sequence to generate its values (see the description of the method [here](https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-INFO-CATALOG)).

At the dump-level, we do have a basic check for the SERIAL types just in case, but dumps generated by pg_dump don't actually use the SERIAL types, they instead have individual statements for creating the sequence, assigning the column as its owner, and altering the table to the column's default value. As such, I've added the concept of columns owning sequences to the `ddl.Sequence` struct, and also have updated the postgres dump processor to keep track of sequences as well as sequence-based AutoGen for columns. This gives us enough information to check whether a particular column is a SERIAL column. Note also that since the statements can show up in roughly any order, we check for SERIAL after processing every relevant statement.

The web UI has also been updated to show auto-gen components for Postgres in addition to MySQL, but restricted to only IDENTITY columns (no UUID or sequence-based generation).

- [X] Tests pass (note, there are still transient UI test failures...)

### DIFF
--- a/test_data/pg_dump.test.out
+++ b/test_data/pg_dump.test.out
@@ -92,7 +92,6 @@ ALTER TABLE public.test3 OWNER TO postgres;
 --
 
 CREATE SEQUENCE public.test_id_seq
-    AS integer
     START WITH 1
     INCREMENT BY 1
     NO MINVALUE


sources/postgres/infoschema.go
```diff
@@ -318,21 +320,48 @@ func (isi InfoSchemaImpl) GetColumns(conv *internal.Conv, table common.SchemaAnd
...
+			AutoGen: toAutoGen(isSerialColumn),
 		}
 		colDefs[colId] = c
 		colIds = append(colIds, colId)
 	}
 	return colDefs, colIds, nil
 }
 
+func (isi InfoSchemaImpl) getSerialColumns(conv *internal.Conv, table common.SchemaAndName) []string {
+	serialColsQuery := `SELECT a.attname FROM pg_attribute a
```
pratickchokhani, line 341: Can we have integration tests for these flows?

nmemond:
Also, the failing UI tests appear to be spurious, they do pass locally:
image
Likely just need to retry them.

sources/postgres/pgdump.go
```diff
@@ -735,6 +892,15 @@ func updateSchema(conv *internal.Conv, tableId string, cs []constraint, stmtType
 			ct := conv.SrcSchema[tableId]
 			ct.Indexes = append(ct.Indexes, schema.Index{Name: c.name, Unique: true, Keys: toSchemaKeys(conv, tableId, c.cols, colNameIdMap)})
 			conv.SrcSchema[tableId] = ct
+		case pg_query.ConstrType_CONSTR_DEFAULT:
```
pratickchokhani, line 895: Please add an integration test for this.

sources/postgres/infoschema.go
nmemond: @pratickchokhani done, I've updated the existing postgres integration tests to check that the serial column was migrated as expected.

sources/postgres/pgdump.go
nmemond: @pratickchokhani done, I've updated the existing postgres integration tests to check that the serial column was migrated as expected.

PR #1236, opened 2025-11-20
Add CLI flag for specifying default SKIP RANGE and START COUNTER WITH values for IDENTITY columns

nmemond:
This PR implements optional CLI flags in the `targetProfile` parameter for specifying default SKIP RANGE and START COUNTER WITH values, as requested in [this comment](https://github.com/GoogleCloudPlatform/spanner-migration-tool/pull/1227#discussion_r2532779636).

I elected to add this as part of the `targetProfile` parameter instead of adding a new parameter in order to reduce the surface area of changes. I'm open to adding a new parameter instead if that's preferred.

- [X] Tests pass

### DIFF
--- a/profiles/target_profile.go
+++ b/profiles/target_profile.go
@@ -205,6 +205,8 @@ func extractDefaultIdentityOptions(params map[string]string) (DefaultIdentityOpt
 	return defaultIdentityOptions, nil
 }
 
+// This attempts to parse the skip range assuming the following format: "<min>-<max>" where both <min> and <max> are
+// positive integers and <min> is less than <max>.
 func parseDefaultSkipRange(defaultSkipRangeStr string) (string, string, error) {
 	parts := strings.Split(defaultSkipRangeStr, "-")
 	if len(parts) < 2 {


conversion/conversion.go
```diff
@@ -81,7 +81,7 @@ func (ci *ConvImpl) SchemaConv(migrationProjectId string, sourceProfile profiles
 		return schemaFromSource.schemaFromDatabase(migrationProjectId, sourceProfile, targetProfile, &GetInfoImpl{}, &common.ProcessSchemaImpl{})
 	case constants.PGDUMP, constants.MYSQLDUMP:
 		expressionVerificationAccessor, _ := expressions_api.NewExpressionVerificationAccessorImpl(context.Background(), targetProfile.Conn.Sp.Project, targetProfile.Conn.Sp.Instance)
-		return schemaFromSource.SchemaFromDump(targetProfile.Conn.Sp.Project, targetProfile.Conn.Sp.Instance, sourceProfile.Driver, targetProfile.Conn.Sp.Dialect, ioHelper, &ProcessDumpByDialectImpl{ExpressionVerificationAccessor: expressionVerificationAccessor})
+		return schemaFromSource.SchemaFromDump(targetProfile.Conn.Sp.Project, targetProfile.Conn.Sp.Instance, sourceProfile.Driver, targetProfile.Conn.Sp.Dialect, ioHelper, &ProcessDumpByDialectImpl{ExpressionVerificationAccessor: expressionVerificationAccessor}, targetProfile.DefaultIdentityOptions)
```
nmemond: I did consider refactoring this to just pass in `targetProfile` itself, but turns out this is also used from `webv2/api/schema.go` which doesn't have use the `targetProfile` object.

conversion/conversion_from_source.go
```diff
@@ -60,6 +61,11 @@ func (sads *SchemaFromSourceImpl) schemaFromDatabase(migrationProjectId string,
 	conv.SpProjectId = targetProfile.Conn.Sp.Project
 	conv.SpInstanceId = targetProfile.Conn.Sp.Instance
 	conv.Source = sourceProfile.Driver
+	conv.DefaultIdentityOptions = ddl.IdentityOptions{
+		SkipRangeMin: targetProfile.DefaultIdentityOptions.SkipRangeMin,
+		SkipRangeMax: targetProfile.DefaultIdentityOptions.SkipRangeMax,
+		StartCounterWith: targetProfile.DefaultIdentityOptions.StartCounterWith,
+	}
```
nmemond, line 68: This is where we set the default identity options for schema migrations from a direct database connection.

conversion/conversion_from_source.go
```diff
@@ -125,6 +131,11 @@ func (sads *SchemaFromSourceImpl) SchemaFromDump(SpProjectId string, SpInstanceI
 	conv.Source = driver
 	conv.SpProjectId = SpProjectId
 	conv.SpInstanceId = SpInstanceId
+	conv.DefaultIdentityOptions = ddl.IdentityOptions{
+		SkipRangeMin: defaultIdentityOptions.SkipRangeMin,
+		SkipRangeMax: defaultIdentityOptions.SkipRangeMax,
+		StartCounterWith: defaultIdentityOptions.StartCounterWith,
+	}
```
nmemond, line 138: This is where we set the default identity options for schema migrations from a database dump file.

profiles/target_profile.go
```diff
@@ -152,11 +160,60 @@ func NewTargetProfile(s string) (TargetProfile, error) {
 		return TargetProfile{}, fmt.Errorf("dialect not supported %v", sp.Dialect)
 	}
 
+	defaultIdentityOptions, err := extractDefaultIdentityOptions(params)
```
nmemond, line 167: This is where we parse the default identity options from the CLI argument. This method includes validation of the values.

profiles/target_profile.go
```diff
@@ -169,5 +182,49 @@ func NewTargetProfile(s string) (TargetProfile, error) {
...
+		startCounterWithInt, err := strconv.Atoi(defaultStartCounterWith)
+		if err != nil || startCounterWithInt <= 0 {
+			return DefaultIdentityOptions{}, fmt.Errorf("Invalid value for defaultIdentityStartCounterWith: %s, expected a positive integer.", defaultStartCounterWith)
+		}
+		defaultIdentityOptions.StartCounterWith = defaultStartCounterWith
+	}
+	return defaultIdentityOptions, nil
+}
+
+func parseDefaultSkipRange(defaultSkipRangeStr string) (string, string, error) {
```
pratickchokhani, line 210: Add comment to highlight the expected string format.

profiles/target_profile.go
```diff
@@ -169,5 +182,49 @@ func NewTargetProfile(s string) (TargetProfile, error) {
 	}
 
 	conn := TargetProfileConnection{Ty: TargetProfileConnectionTypeSpanner, Sp: sp}
-	return TargetProfile{Ty: TargetProfileTypeConnection, Conn: conn}, nil
+	return TargetProfile{Ty: TargetProfileTypeConnection, Conn: conn, DefaultIdentityOptions: defaultIdentityOptions}, nil
+}
+
+func extractDefaultIdentityOptions(params map[string]string) (DefaultIdentityOptions, error) {
+	defaultIdentityOptions := DefaultIdentityOptions{}
+	if defaultSkipRangeStr, ok := params["defaultIdentitySkipRange"]; ok {
```
pratickchokhani, line 190: Add a validation where if both are present, then an error is returned.

profiles/target_profile.go
```diff
@@ -57,6 +58,13 @@ type TargetProfileConnection struct {
 type TargetProfile struct {
 	Ty   TargetProfileType
 	Conn TargetProfileConnection
+	DefaultIdentityOptions DefaultIdentityOptions
+}
+
+type DefaultIdentityOptions struct {
+	SkipRangeMin string
```
pratickchokhani, line 65: I presume int value is required here and we are also converting the string to int. Why are we storing String here?

profiles/target_profile.go
nmemond: Can you clarify what you mean by both? If it's skip range and start counter with, I don't think there's an issue with specifying both (and there may be cases where that's actually desired).

profiles/target_profile.go
nmemond: True, good point. I think I had made it a string because I had noticed other similar int-values were also strings. I'll look into making it an int, I don't anticipate any issues other than maybe the UI.

nmemond: One thing I did notice when starting on this, in my opinion the case where the skip range or start counter with values are not specified is a little less clear. If the string is empty, it's obvious that nothing has been specified, whereas if the int-value is 0, the reader has to know that both of these must be positive in order to know that 0 means nothing has been passed in.

I can still proceed with the change if you'd like.

nmemond: This does also complicate the UI, having to work with BigInts and add special handling for 0s. I'm not sure it's worth the effort, tbh.

profiles/target_profile.go
nmemond: Comment added, please review and let me know if any changes are needed!

PR #1255, opened 2025-12-29
Documentation update on RetryDLQ flow

aasthabharill:
https://b.corp.google.com/issues/460040370
documentation change in accordance with: https://github.com/GoogleCloudPlatform/DataflowTemplates/pull/3180

#### Steps: 
1. Review and resolve the issues causing the severe errors.
2. Create a temporary GCS bucket or folder to stage the severe error files (e.g., gs://<bucket-name>/dlq/severe-duplicate).
3. Set up a new Pub/Sub topic and subscription on the parent bucket (NOT on the severe-duplicate folder)
4. After confirming the Pub/Sub notification is active, move the files from the dlq/severe folder to the temporary folder (e.g., severe-duplicate).
6. Now, run the Dataflow template in retryDLQ mode, passing the new subscription ID in the dlqGcsPubSubSubscription parameter and ensure to pass gs://<bucket-name>/dlq/ as the deadLetterQueueDirectory, otherwise the errors will only be retried once. 

#### Background:
the code will create a severe bucket inside the dlq folder and reprocess the further errors by consuming through that folder instead of temporary, which is why its important to configure deadLetterQueueDirectory as the parent directory of severe-duplicate so the pubsub can later pick errors up from severe folder that will be created. If this is not correctly configured, the pubsub will pick up the errors at the beginning, do a one-pass and then stop as it would not be able to consume the new error files that the code created.

#### Reason for not using this approach
- Configuration is prone to errors which might be hard to debug for a customer
- Needs a completely new setup of pubsub and directory, whereas the one that is recommended can easily reuse resources across regular and retryDLQ runs

### DIFF
--- a/docs/troubleshoot/minimal.md
+++ b/docs/troubleshoot/minimal.md
@@ -129,6 +129,13 @@ These job parameters can be taken from the original job.
 This will reprocess the records marked as ‘severe' error records from the DLQ.  
 Before running the Dataflow job, check if the main Dataflow job has non-zero retryable error count. In case there are referential error records - check that the dependent table data is populated completely from the source database.
 
+The following parameters can be taken from the regular forward migration Dataflow job:
+region
+instanceId
+databaseId
+sessionFilePath
+deadLetterQueueDirectory
+
 Sample command to run the Dataflow job in retryDLQ mode is
 
 ```sh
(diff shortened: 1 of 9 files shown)


docs/troubleshoot/minimal.md
```diff
@@ -154,3 +154,32 @@ sessionFilePath
 deadLetterQueueDirectory
 dlqGcsPubSubSubscription
 ```
+
+
+#### Alternative: Retrying Severe Errors via the Regular Mode Pipeline
```
bharadwaj-aditya: It would make sense to keep this as a subheading for DLQ processing, please move it 1 level lower in the hierarchy.

PR #1256, opened 2025-12-30
Use Java Client Side Car for Cassandra Dual writes

VardhanThigle:
# Swap Go Spanner Cassandra Proxy with Java Client.
## Goal
Replace the `go-spanner-cassandra` proxy with the `java-spanner-cassandra` proxy in the `sources/cassandra` Docker image, while maintaining the existing dual-write proxy functionality (`zdm-proxy`). The Java Client being used will be maintained to accommodate newer options and features.

## Current Components of the Docker Image
- **Dockerfile**:
  - Base: `golang:1.24-bullseye`
  - Builds `zdm-proxy` (Go)
  - Builds `go-spanner-cassandra` (Go)
  - Final Image: `alpine:3.22` with binaries copied over.
- **Entrypoint**:
  - Runs `cassandra-spanner-proxy` with CLI flags derived from environment variables.
  - Runs `zdm-proxy` with a config file.
  - Monitors PIDs.

### 1. Dockerfile Restructuring (Multi-Stage Build)
We need a multi-stage build to handle both Go (for `zdm-proxy`) and Java (for `java-spanner-cassandra`) compilation.

**Stages:**
1.  **Go Builder**:
    -   Image: `golang:1.24-bullseye` (Keep existing).
    -   Action: Build `zdm-proxy` as before.
2.  **Java Builder**:
    -   **Base Image**: `maven:3.9-eclipse-temurin-17`.
    -   **Reasoning**: The `java-spanner-cassandra` client requires Java 8 or later. We will use **Java 17 LTS** as it is a modern, widely supported standard that offers better performance and container support than Java 8.
    -   Action: Clone `java-spanner-cassandra` and run `mvn clean install` to produce the shaded process/launcher jar.
3.  **Final Image**:
    -   Image: `alpine:3.22` (Keep existing).
    -   Action:
        -   Install `openjdk17-jre` (or headless version) + `bash` + `gettext` (for `envsubst`).
        -   Copy `zdm-proxy` binary from Go Builder.
        -   Copy `spanner-cassandra-launcher.jar` from Java Builder.
        -   Copy `entrypoint.sh`.
        -   Copy new `spanner.yaml` (template).

### 2. Configuration Strategy
We will map the existing environment variables to the new Java Proxy's configuration format using a template and `envsubst`. **This ensures that users do not need to change how they deploy the container.

#### Current Configuration Flow (Old)
1.  **User Input**: User passes environment variables to `docker run`: `SPANNER_PROJECT`, `SPANNER_INSTANCE`, `SPANNER_DATABASE`, `GRPC_CHANNELS` (optional).
2.  **Entrypoint Execution**: `entrypoint.sh` reads these variables directly.
3.  **Process Start**: `entrypoint.sh` constructs a command line string:
    ```bash
    /cassandra-spanner-proxy --db="projects/$SPANNER_PROJECT/instances/$SPANNER_INSTANCE/databases/$SPANNER_DATABASE" ...
    ```

#### Proposed Configuration Flow (New)
1.  **User Input**: User passes **identical** environment variables as before (`SPANNER_PROJECT`, etc.). No user-side change required.
2.  **Template Creation**: We bake a `spanner-cassandra-config.yaml` template into the image.
3.  **Entrypoint Execution**:
    -   `entrypoint.sh` uses `envsubst` to read the template and replace placeholders (e.g., `${SPANNER_PROJECT}`) with actual values from the current environment.
    -   References:
        -   The template will use standard env placeholders.
        -   `envsubst` allows dynamic generation of the complex YAML structure required by the Java proxy, which is more robust than passing many `-D` system properties for nested configs.
4.  **Process Start**:
    -   `entrypoint.sh` starts the Java process pointing to the *generated* config file.
    -   Command: `java -DconfigFilePath=/app/generated-config.yaml -jar ...`

#### Configuration Flexibility (FAQ)
**Q: Can a user have a config file in the code repo and only override a few parameters via env variables?**
**A: Yes.** The config file in the repo (`sources/cassandra/spanner-cassandra-config.yaml`) serves as the template.
-   **Static Values**: You can hardcode values in this file (e.g., `max_prepared_statements: 100`) if they should always be the same.
-   **Overridable Values**: For values you want to control via environment variables (like Project ID), you **must** use placeholders (e.g., `${SPANNER_PROJECT}`) in the file.
-   **Runtime Behavior**: When the container starts, `envsubst` will replace the placeholders with the actual environment variable values. Hardcoded values remain unchanged.
-   **Full Override**: If a user wants to supply a completely different config at runtime, they can mount a file to `/app/spanner-cassandra-config.yaml`, and the entrypoint will still process it with `envsubst` (allowing them to use their own placeholders if desired).

### 3. Entrypoint Script (`entrypoint.sh`)
Update `entrypoint.sh` to:
1.  Generate `generated-config.yaml` from `spanner-cassandra-config.yaml` using `envsubst`.
2.  Launch the Java proxy background process.
3.  Launch `zdm-proxy` (unchanged).
4.  Monitor processes (update PID tracking for Java process).

## Detailed Steps
1.  **Create `sources/cassandra/spanner-cassandra-config.yaml`**: Define the default template.
2.  **Update `sources/cassandra/Dockerfile`**:
    -   Add `maven:3.9-eclipse-temurin-17` build stage.
    -   Install `openjdk17-jre` and `gettext` in the final stage.
    -   Copy JARs and template.
3.  **Update `sources/cassandra/entrypoint.sh`**:
    -   Add `envsubst` step.
    -   Update process start command.

## Questions/Notes
-   **Java Version**: Proceeding with 17 (Client supports 8+).
-   **Config**: Proceeding with Template + Envsubst to maintain backward compatibility for user inputs while enabling static config in repo.

### Basic
1. Builds
2. Run built ZDM image
3. Insert 15 rows, modify 5 rows, delete 5 rows. Compare output with Spanner. 
Test passes. As of now the test is manual as Spanner Emulator does not support Cassandra Adapter.
### IOPS test with cassandra-stress
No regression in throughput with respect to main.
All latencies below are in milli seconds.
| Version of ZDM Image | Number of Threads | Number of GRPC Channels | IOPS        | Latency Mean | Latency median | Latency 95p | Latency 99p | Latency 99.9p | Latency Max |
|----------------------|-------------------|-------------------------|-------------|--------------|----------------|-------------|-------------|---------------|-------------|
| At PR                | 250               | 4                       | 1,789 op/s  | 137.5        | 134.5          | 157.3       | 198.8       | 364.6         | 534.5       |
| At PR                | 250               | 500                     | 1,790 op/s  | 139.3        | 138.7          | 151.8       | 159.6       | 248           | 351.5       |
| At Main              | 250               | 500                     | 1,744 op/s  | 142.4        | 141.7          | 156.2       | 164.6       | 178.4         | 345.8       |

- [X] Tests pass

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -2224,8 +2224,6 @@ google.golang.org/protobuf v1.28.1/go.mod h1:HV8QOd/L58Z+nl8r43ehVNZIU/HEI6OcFqw
 google.golang.org/protobuf v1.29.1/go.mod h1:HV8QOd/L58Z+nl8r43ehVNZIU/HEI6OcFqwMG9pJV4I=
 google.golang.org/protobuf v1.30.0/go.mod h1:HV8QOd/L58Z+nl8r43ehVNZIU/HEI6OcFqwMG9pJV4I=
 google.golang.org/protobuf v1.31.0/go.mod h1:HV8QOd/L58Z+nl8r43ehVNZIU/HEI6OcFqwMG9pJV4I=
-google.golang.org/protobuf v1.36.8 h1:xHScyCOEuuwZEc6UtSOvPbAT4zRh0xcNRYekJwfqyMc=
-google.golang.org/protobuf v1.36.8/go.mod h1:fuxRtAxBytpl4zzqUh6/eyUujkJdNiuEkXntxiD/uRU=
 google.golang.org/protobuf v1.36.9 h1:w2gp2mA27hUeUzj9Ex9FBjsBm40zfaDtEWow293U7Iw=
 google.golang.org/protobuf v1.36.9/go.mod h1:fuxRtAxBytpl4zzqUh6/eyUujkJdNiuEkXntxiD/uRU=
 gopkg.in/alecthomas/kingpin.v2 v2.2.6/go.mod h1:FMv+mEhP44yOT+4EoQTLFTRgOQ1FBLkstjWtayDeSgw=


sources/cassandra/entrypoint.sh
```diff
@@ -1,16 +1,29 @@
 #!/bin/bash
 
-# Start processes and store their PIDs
-/cassandra-spanner-proxy --db="projects/$SPANNER_PROJECT/instances/$SPANNER_INSTANCE/databases/$SPANNER_DATABASE" --grpc-channels=500 --tcp="0.0.0.0:9042" &
+# Generated configuration file from template using envsubst
+# envsubst does not support default values in the template (e.g. ${VAR:-default}).
+# We must set the defaults here in the shell before substitution.
+export GRPC_CHANNELS=${GRPC_CHANNELS:-4}
+export MAX_COMMIT_DELAY=${MAX_COMMIT_DELAY:-5}
```
rohitwali: isnt 0 the default?

sources/cassandra/entrypoint.sh
```diff
@@ -1,16 +1,29 @@
 #!/bin/bash
 
-# Start processes and store their PIDs
-/cassandra-spanner-proxy --db="projects/$SPANNER_PROJECT/instances/$SPANNER_INSTANCE/databases/$SPANNER_DATABASE" --grpc-channels=500 --tcp="0.0.0.0:9042" &
+# Generated configuration file from template using envsubst
+# envsubst does not support default values in the template (e.g. ${VAR:-default}).
+# We must set the defaults here in the shell before substitution.
+export GRPC_CHANNELS=${GRPC_CHANNELS:-4}
```
rohitwali, line 6: previous defaults were 500?

sources/cassandra/spanner-cassandra-config.yaml
```diff
@@ -0,0 +1,12 @@
+globalClientConfigs:
+  enableBuiltInMetrics: true
+  healthCheckEndpoint: "127.0.0.1:8080"
+
+listeners:
+  - name: "listener_1"
```
rohitwali, line 6: nit:spanner_cassandra_proxy_listener or something similar?

rohitwali (approved):
thanks for the changes. mostly LGTM, couple of clarifications

sources/cassandra/entrypoint.sh
VardhanThigle: We have taken this from the example config provided by the adapter client

https://github.com/googleapis/java-spanner-cassandra/blob/main/docs/config-options.md#config-options

sources/cassandra/entrypoint.sh
VardhanThigle: We have taken this from the example config provided by the adapter client

https://github.com/googleapis/java-spanner-cassandra/blob/main/docs/config-options.md#config-options

I am also checking with the adapter team.

sources/cassandra/spanner-cassandra-config.yaml
VardhanThigle: This is the config file format as needed by the adapter.
https://github.com/googleapis/java-spanner-cassandra/blob/main/docs/config-options.md#config-options

sources/cassandra/Dockerfile
```diff
@@ -15,30 +16,44 @@ WORKDIR /app/zdm-proxy/proxy
...
 
 # --- Final Image ---
 FROM alpine:3.22
 
-# Install bash, ca-certificates, openssl, and update packages
+# Install bash, ca-certificates, openssl, gettext (for envsubst), and Java 17 JRE
 RUN apk update && \
     apk upgrade && \
-    apk add --no-cache bash ca-certificates openssl
+    apk add --no-cache bash ca-certificates openssl gettext openjdk17-jre
```
bharadwaj-aditya, line 39: Just double checking. Is the cassandra spanner client tested on openjdk17 ?

sources/cassandra/entrypoint.sh
bharadwaj-aditya: My guess is we should be using multiplexed sessions here. In which case the number of connections comes down to 1.

VardhanThigle: That will be controlled by the java client - so currently checking with the client team on the recommendation here.

sources/cassandra/Dockerfile
VardhanThigle: It supports JDK > 8.
Our manual tests for this PR have been on 17 and no compilation or class loading issues faced there.

sources/cassandra/entrypoint.sh
mayurkale22: Default is 0 in the backend. I would suggest to remove setting this option

sources/cassandra/spanner-cassandra-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+    host: "127.0.0.1"
+    port: 9042
+    spanner:
+      databaseUri: "projects/${SPANNER_PROJECT}/instances/${SPANNER_INSTANCE}/databases/${SPANNER_DATABASE}"
+      numGrpcChannels: ${GRPC_CHANNELS}
+      maxCommitDelayMillis: ${MAX_COMMIT_DELAY}
```
mayurkale22: as mentioned in above comment, I would recommend to remove this option

sources/cassandra/entrypoint.sh
mayurkale22: default 4 is good enough for many cases, I would suggest not to override this option unless really required

sources/cassandra/Dockerfile
```diff
@@ -15,30 +16,44 @@ WORKDIR /app/zdm-proxy/proxy
...
 
-# --- Cassandra-to-Spanner CQL Proxy ---
+# --- Java Builder Stage ---
+FROM maven:3.9-eclipse-temurin-17 AS java-builder
+
 WORKDIR /app
 
-RUN git clone "https://github.com/googleapis/go-spanner-cassandra.git"
+# Clone the Java Spanner Cassandra repo
+# Using specific tag or main branch. For stability, usually better to pin, but instructions enable main.
```
mayurkale22, line 25: Using main branch, could be problematic for long run as the main branch is inherently unstable and subject to frequent, breaking changes. It would suggest to pin to latest released branch. Not a blocker for this PR.

sources/cassandra/README.md
```diff
@@ -12,10 +12,12 @@ This repository contains components for doing live migrations from Cassandra to
 - Handles dual writes to both Cassandra and Spanner
 - Ensures data consistency during migration
 
-### 2. Cassandra-Spanner Proxy
+### 2. [Java Spanner Cassandra Proxy](https://github.com/googleapis/java-spanner-cassandra)
 - Runs as a sidecar to ZDM proxy
 - Translates CQL (Cassandra Query Language) to Spanner API calls
 - Enables Cassandra-compatible applications to interact with Spanner
+- **Note:** This implementation replaces the deprecated Go-based proxy.
```
mayurkale22: Go-based proxy is really deprecated. I would just remove this note , if possible

sources/cassandra/README.md
```diff
@@ -26,7 +28,7 @@ The migration setup consists of:
 - **Origin**: Source Cassandra database
 - **Target**: Destination Cloud Spanner database
 - **ZDM Proxy**: Manages dual writes and read routing
-- **Cassandra-Spanner Proxy**: Translates CQL to Spanner API calls
+- **Java-Spanner Proxy**: Translates CQL to Spanner API calls
```
mayurkale22: Java-Spanner Proxy -> Spanner Cassandra Java Client?

sources/cassandra/entrypoint.sh
VardhanThigle: Thanks. Will keep it as is.

sources/cassandra/entrypoint.sh
VardhanThigle: Thanks. Will make 0 as default. I would like to expose the options supported by the java client to users while running the docker image, just in case they need it. So will retain the code that honors the env variable but make 0 as the default.

VardhanThigle: Defaulted to 0.

sources/cassandra/spanner-cassandra-config.yaml
VardhanThigle: I have defaulted to 0. Will sync more over chat.

sources/cassandra/Dockerfile
VardhanThigle: Makes sense. Pinned to `v1.1.0`.

sources/cassandra/README.md
VardhanThigle: Done. Thanks.

sources/cassandra/README.md
VardhanThigle: Done.

PR #1258, opened 2026-01-06
Fix Windows test failures: file handle leaks and OS-specific assertions

sm745052:
Fixes:
- File Handle Leaks in store_files.go
- Windows specific assertions in file_reader_test.go

### DIFF
--- a/file_reader/file_reader_test.go
+++ b/file_reader/file_reader_test.go
@@ -5,7 +5,6 @@ import (
 	"github.com/stretchr/testify/assert"
 	"os"
 	"runtime"
-	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


file_reader/file_reader_test.go
```diff
@@ -47,7 +49,11 @@ func TestFileReaderFile(t *testing.T) {
 			if tt.wantErr {
 				assert.Error(t, err)
 				if tt.expectedError != "" {
-					assert.Contains(t, err.Error(), tt.expectedError)
+					if runtime.GOOS == "windows" && tt.name == "Local file open error" {
```
pratickchokhani: What is the need for this?

pratickchokhani: While assigning to expected error, can we add a check for windows?

sm745052: done

PR #1322, opened 2026-04-28
updating go version  to fix CVE-2026-27144

VardhanThigle:

- [X] Tests pass
   - go mod tidy
   - go vet ./...
   - go test ./...

### DIFF
--- a/.github/workflows/integration-tests-against-emulator.yaml
+++ b/.github/workflows/integration-tests-against-emulator.yaml
@@ -173,7 +173,7 @@ jobs:
 
       - uses: actions/setup-go@v2
         with:
-          go-version: "1.25.8"
+          go-version: "1.25.9"
       - run: go version
       - run: go build
       - run: go test -v ./...
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           # Need to use a version less than 1.25 due to https://github.com/golang/go/issues/74462.
-          go-version: '>=1.23.0 <1.25.0'
+          go-version: '>=1.23.0 <1.25.9'
       - uses: actions/setup-node@v2
         with:
           node-version: 18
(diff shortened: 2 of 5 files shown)


.github/workflows/integration-tests-against-emulator.yaml
```diff
@@ -173,7 +173,7 @@ jobs:
 
       - uses: actions/setup-go@v2
         with:
```
shreyakhajanchi, line 175: we may need to change in lint.yaml and test-coverage.yaml as well

docs/install.md
```diff
@@ -91,7 +91,7 @@ Detailed instructions on how to install a new component in gCloud can be found [
 Building from source is only supported for MacOS and Linux based platforms.
 
 1. Install Go ([download](https://golang.org/doc/install)) on your development machine if it is not already installed, configure the [GOPATH](https://pkg.go.dev/cmd/go@master#hdr-GOPATH_environment_variable) environment variable if it is not already configured, and [test your installation](https://golang.org/doc/install#testing). <br/>
-    Required go version: 1.25.0+
+    Required go version: 1.29.0+
```
shreyakhajanchi: Isn't this incorrect?

shreyakhajanchi (commented):
Are there no changes in go.sum file?

VardhanThigle:
> Are there no changes in go.sum file?

Yes. That might be because the dependencies are already updated by dependbot.

docs/install.md
VardhanThigle: We are increasing to 29 as there's a vulnerability on 28. The previous update didn't change this to 28.

.github/workflows/integration-tests-against-emulator.yaml
VardhanThigle: Thanks. Added to both the files.

.github/workflows/lint.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           # Need to use a version less than 1.25 due to https://github.com/golang/go/issues/74462.
-          go-version: '>=1.23.0 <1.25.0'
+          go-version: '>=1.23.0 <1.25.9'
```
shreyakhajanchi, line 29: nit: the issue that the comment above mentions is in `1.25.0`, so if we are not facing the issue with the latest version , we should remove the comment

shreyakhajanchi (approved):

.github/workflows/lint.yaml
VardhanThigle: I think the comment is justifying the >=1.23.0 part for the linter. https://github.com/golang/go/issues/74462 is still open.


I will experiment removing the check on 1.23 and see if that works, but will do that as a followup as it's not blocking the release.

PR #1324, opened 2026-04-29
Adding 5K tables integration test for schema migration

sm745052:
Adds a 5,000-table integration test for MySQL direct connect and introduces a `BatchedInfoSchema` interface to allow bulk metadata fetching, improving performance for large schemas.

## Changes
- **`BatchedInfoSchema`**: New interface in `infoschema.go` for bulk fetching of columns, constraints, FKs, and indexes.
- **Parallel Batching**: Updates `GenerateSrcSchema` to process tables in parallel batches.
- **5k Test**: Adds `TestIntegration_MYSQL_DirectConnect_5000Tables` and 5,000 table SQL dump.
-   **Corrected `TestGetConstraints_CheckConstraintsTableAbsent`**: Updated the test to expect success when the `CHECK_CONSTRAINTS` table is missing. Now validates the fallback logic that uses a simpler query to fetch `TABLE_CONSTRAINTS` when the check constraints table is unavailable. Previously, the test was incorrectly asserting failure due to an incomplete mock setup for the fallback query.
-   **Ignored `ExpressionId` in Source Schema Assertions**: Modified the test helper `assertSrcColDef` in `internal/test_util.go` to ignore `ExpressionId` for both expected and actual columns. This avoids test brittleness caused by non-deterministic ID generation order during parallel batch processing.

## Batch Size Experiments (5k Tables)

| Batch Size | Result | Time |
| :--- | :--- | :--- |
| 100 | PASS | ~57s |
| 50 | PASS | ~94s |
| 10 | PASS | ~378s |
| 1 | TIMEOUT | >30m |

*Conclusion: Larger batch sizes significantly reduce schema processing time. The PR currently defaults to 50.*


sources/common/infoschema.go
```diff
@@ -36,17 +36,31 @@ type InfoSchema interface {
...
-	GetIndexes(conv *internal.Conv, table SchemaAndName, colNameIdMp map[string]string) ([]schema.Index, error)
 	ProcessData(conv *internal.Conv, tableId string, srcSchema schema.Table, spCols []string, spSchema ddl.CreateTable, additionalAttributes internal.AdditionalDataAttributes) error
 	StartChangeDataCapture(ctx context.Context, conv *internal.Conv) (map[string]interface{}, error)
 	StartStreamingMigration(ctx context.Context, migrationProjectId string, client *sp.Client, conv *internal.Conv, streamInfo map[string]interface{}) (internal.DataflowOutput, error)
 }
 
+// StandardInfoSchema supports per-table fetching of metadata.
+type StandardInfoSchema interface {
+	InfoSchema
+	GetColumns(conv *internal.Conv, table SchemaAndName, constraints map[string][]string, primaryKeys []string) (map[string]schema.Column, []string, error)
```
pratickchokhani, line 48: Add comments on to explain the input and output structures for all the functions below.

sources/common/infoschema.go
```diff
@@ -113,20 +127,82 @@ func (is *InfoSchemaImpl) GenerateSrcSchema(conv *internal.Conv, infoSchema Info
...
-		return res
-	}
 
-	r := task.RunParallelTasksImpl[SchemaAndName, SchemaAndName]{}
-	res, e := r.RunParallelTasks(tables, numWorkers, asyncProcessTable, false)
-	if e != nil {
-		logger.Log.Info(fmt.Sprintf("exiting due to error: %s , while processing schema for table %s\n", e, res))
-		return 0, e
+
+	if bis, ok := infoSchema.(BatchedInfoSchema); ok {
```
pratickchokhani, line 160: It seems that the following blocks are handling multiple logics. 
Consider breaking this or adding modulization.

sources/mysql/infoschema.go
```diff
@@ -174,36 +174,62 @@ func (isi InfoSchemaImpl) GetTables() ([]common.SchemaAndName, error) {
 	return tables, nil
 }
 
-// GetColumns returns a list of Column objects and names// ProcessColumns
-func (isi InfoSchemaImpl) GetColumns(conv *internal.Conv, table common.SchemaAndName, constraints map[string][]string, primaryKeys []string) (map[string]schema.Column, []string, error) {
```
pratickchokhani, line 175: Are we not retaining the implementation of StandardInfoSchema?

sources/mysql/infoschema.go
```diff
@@ -174,36 +174,62 @@ func (isi InfoSchemaImpl) GetTables() ([]common.SchemaAndName, error) {
...
-	q := `SELECT c.column_name, c.data_type, c.column_type, c.is_nullable, c.column_default, c.character_maximum_length, c.numeric_precision, c.numeric_scale, c.generation_expression, c.extra
+
+
+
+
```
pratickchokhani, line 177: All the implementations seems quite long. Can we add more modularisation to them?

sources/mysql/infoschema.go
sm745052: not for MySQL since that path will never be used.

sources/common/infoschema.go
sm745052: done

sources/common/infoschema.go
sm745052: although i made them modular, the functions are not reused . modularization here is just for modularity.

sources/mysql/infoschema.go
sm745052: same as above

pratickchokhani (approved):

PR #1344, opened 2026-05-27
Add support for TLS/mTLS connections in MySQL, PostgreSQL and Cassandra

nanditbansal:
**Summary**
This PR adds secure connection options (TLS and mTLS) for the three target source database systems (MySQL, PostgreSQL, and Cassandra) across both the command-line interface (CLI) and the web-based user interface (UI).

These changes ensure that users can securely connect to their source databases during schema migration by supplying CA certificates, client certificates, and client private keys. SQL Server and Oracle connection logic remain completely unchanged.

**Key Changes**

1. Connection Structs & Command Parsing (profiles)
Extended SourceProfileConnectionMySQL, SourceProfileConnectionPostgreSQL, and SourceProfileConnectionCassandra to support Sslmode, Sslrootcert, Sslcert, and Sslkey fields.
Populated these fields during profile parsing from both direct flags/parameters and standard environment variables (like PGSSLMODE, MYSQL_SSL_MODE, etc.).

2. Secure DSN Construction (profiles)
MySQL: Added registerMySQLTLSConfig helper to dynamically load and register custom tls.Config in memory via the standard go-sql-driver/mysql registry, mapping correct certificate checks for modes like verify-ca and skip-verify.
PostgreSQL: Mapped TLS parameters directly to DSN key-value pairs.

3. Cassandra Client Setup (accessors)
Extended GetOrCreateCassandraClusterClient and mapped the TLS/SSL parameters to standard gocql.SslOptions (mapping CA path, cert path, private key path, and host validation toggles).

4. Web Backend & UI Integration (webv2, ui)
Extended JSON models and structs in the web server backend to parse and store secure parameters from HTTP payloads.
Updated frontend typescript interfaces and Angular HTML forms to conditionally display secure connection inputs (SSL/TLS Settings) only when the user selects MySQL, PostgreSQL, or Cassandra as the database engine.

**Testing**
Unit Tests:
Added TestNewSourceProfileConnectionTLS to verify TLS properties parsing from parameters and env vars.
Updated TestGetSQLConnectionStr to dynamically generate and register a valid self-signed CA certificate in memory to test MySQL TLS/mTLS DSN builders without file dependencies.
Added new Cassandra TLS cases under TestGetOrCreateCassandraClusterClient.
Validation:
Verified all tests in the repository pass with go test ./....

### DIFF
--- a/accessors/cassandra/cassandra_accessor.go
+++ b/accessors/cassandra/cassandra_accessor.go
@@ -41,6 +41,10 @@ func NewCassandraAccessor(sourceProfile profiles.SourceProfile) (*CassandraAcces
 		cfg.DataCenter,
 		cfg.User,
 		cfg.Pwd,
+		cfg.Sslmode,
+		cfg.Sslrootcert,
+		cfg.Sslcert,
+		cfg.Sslkey,
 	)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to create Cassandra client: %w", err)
--- a/accessors/clients/cassandra/cassandra_client.go
+++ b/accessors/clients/cassandra/cassandra_client.go
@@ -36,7 +36,7 @@ var createSessionFromCluster = func(c *gocql.ClusterConfig) (GocqlSessionInterfa
 	return NewGocqlSessionImpl(session), nil
 }
 
-func GetOrCreateCassandraClusterClient(contactPoints []string, port int, keyspace, datacenter, user, password string) (CassandraClusterInterface, error) {
+func GetOrCreateCassandraClusterClient(contactPoints []string, port int, keyspace, datacenter, user, password string, sslmode, sslrootcert, sslcert, sslkey string) (CassandraClusterInterface, error) {
 	clusterConfigMux.Lock()
 	defer clusterConfigMux.Unlock()
 	
@@ -53,6 +53,14 @@ func GetOrCreateCassandraClusterClient(contactPoints []string, port int, keyspac
 	}
 	if datacenter != "" {
 		clusterCfg.PoolConfig.HostSelectionPolicy = gocql.DCAwareRoundRobinPolicy(datacenter)
+	}
+	if sslmode != "" && sslmode != "disable" {
+		clusterCfg.SslOpts = &gocql.SslOptions{
+			CaPath:                 sslrootcert,
+			CertPath:               sslcert,
+			KeyPath:                sslkey,
+			EnableHostVerification: sslmode == "verify-full",
+		}
 	}
 	globalClusterConfig = clusterCfg
 
--- a/profiles/common.go
+++ b/profiles/common.go
@@ -15,13 +15,17 @@
 package profiles
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/csv"
 	"fmt"
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/GoogleCloudPlatform/spanner-migration-tool/common/utils"
+	mysqldriver "github.com/go-sql-driver/mysql"
 	go_ora "github.com/sijms/go-ora/v2"
 	"github.com/GoogleCloudPlatform/spanner-migration-tool/logger"
 )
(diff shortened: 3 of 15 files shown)


accessors/clients/cassandra/cassandra_client.go
```diff
@@ -54,6 +54,14 @@ func GetOrCreateCassandraClusterClient(contactPoints []string, port int, keyspac
 	if datacenter != "" {
 		clusterCfg.PoolConfig.HostSelectionPolicy = gocql.DCAwareRoundRobinPolicy(datacenter)
 	}
+	if sslmode != "" && sslmode != "disable" {
+		clusterCfg.SslOpts = &gocql.SslOptions{
+			CaPath:                 sslrootcert,
+			CertPath:               sslcert,
+			KeyPath:                sslkey,
+			EnableHostVerification: sslmode == "verify-full",
```
VardhanThigle, line 62: There could be cases where only Server Authentication might be required. Infact  based on the customers seen, these are slightly more common than mtls

accessors/cassandra/cassandra_accessor.go
```diff
@@ -41,6 +41,10 @@ func NewCassandraAccessor(sourceProfile profiles.SourceProfile) (*CassandraAcces
 		cfg.DataCenter,
 		cfg.User,
 		cfg.Pwd,
+		cfg.Sslmode,
+		cfg.Sslrootcert,
+		cfg.Sslcert,
```
VardhanThigle, line 46: if this is the client side cert, could we please name it accordingly (likewise with the key)?

profiles/common.go
```diff
@@ -87,16 +91,65 @@ func ParseList(s string) ([]string, error) {
...
+			return "", fmt.Errorf("failed to read root CA cert: %w", err)
+		}
+		if ok := rootCertPool.AppendCertsFromPEM(pem); !ok {
+			return "", fmt.Errorf("failed to append root CA cert")
+		}
+		tlsConfig.RootCAs = rootCertPool
+	}
+
+	if sslcert != "" && sslkey != "" {
+		certs, err := tls.LoadX509KeyPair(sslcert, sslkey)
```
VardhanThigle, line 110: It might be a desired use case to provide keys via  secret manager.

For sensitive customers, we might also want to allow providing certificates via Certificate manager.

webv2/web.go
```diff
@@ -67,24 +67,55 @@ import (
...
+			Conn: profiles.SourceProfileConnection{
+				Ty: profiles.SourceProfileConnectionTypePostgreSQL,
+				Pg: profiles.SourceProfileConnectionPostgreSQL{
+					Host:        config.Host,
+					Port:        config.Port,
+					User:        config.User,
+					Pwd:         config.Password,
+					Db:          config.Database,
+					Sslmode:     config.Sslmode,
+					Sslrootcert: config.Sslrootcert,
```
VardhanThigle, line 104: Are we taking a file path here?

If yes that's not super compatible with the UI.

VardhanThigle (commented):
Have we tested these changes against real Cassandra outside of unit tetes, if yes, please add / ping the logs.

We need to support supplying keys via Secret manager as well ( potentially certificates via Certificate manager too for sensitive customers or for those who enforce automatic rotations - but certificates don't carry any secrets)

It's not a clear to me how is the UI ingesting the key material, is it taking it in clear text or is it just taking a path to the file? The former might be less ideal and we might need to handle it carefully, the later might not be ideal.

For UI it might make sense to allow keys via secret manager and certs via certificate manager. As an additional option to the user, we can implement an API workflow where cert and key files are uploaded to the Go server and then the migration uses a reference to the uploaded material.

accessors/cassandra/cassandra_accessor.go
nanditbansal: Sure, will make the change. Thanks

accessors/clients/cassandra/cassandra_client.go
nanditbansal: I'm not sure what you mean. Do we want to provide an additional functionality to customers to choose between TLS and mTLS?

webv2/web.go
nanditbansal: What would you recommend here?

VardhanThigle (commented):
I had a deep thought about this yesterday and given the over arching task, I think we have a scope to chalk out a small design on how we would handle certs and keys in SMT globally before making Cassandra specific changes.

manitgupta:
> I had a deep thought about this yesterday and given the over arching task, I think we have a scope to chalk out a small design on how we would handle certs and keys in SMT globally before making Cassandra specific changes.

Let us discuss this once before we spend more time on design on SMT specific work items. I am hoping we are prioritising changes to the DataflowTemplates first.

CC: @bharadwaj-aditya

PR #1346, opened 2026-05-27
Upgraded tidb parser version for vulnerability fixes.

pratickchokhani:


VardhanThigle (approved):

go.mod
```diff
@@ -23,8 +23,8 @@ require (
 	github.com/gorilla/mux v1.8.1
 	github.com/lib/pq v1.9.0
 	github.com/pganalyze/pg_query_go/v6 v6.1.0
-	github.com/pingcap/tidb v1.1.0-beta.0.20251126154744-e4e814fdc0af
-	github.com/pingcap/tidb/pkg/parser v0.0.0-20251126154744-e4e814fdc0af
```
manitgupta, line 27: We moved from a newer version to an older version? `202511` -> `202407`? How does that fix things?

Can you please update the description with details on what we did and how we have tested that it works?

pratickchokhani: The release version with TIDB is non-linear. Earlier I choose a latest release as the CVE note shows that the fix was added in May. 

With this, I am choosing the specific release mentioned in the CVE note.

go.mod
```diff
@@ -52,12 +52,17 @@ require (
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/internal/resourcemapping v0.53.0 // indirect
 	github.com/VividCortex/ewma v1.2.0 // indirect
 	github.com/cloudfoundry/gosigar v1.3.6 // indirect
+	github.com/cockroachdb/errors v1.11.1 // indirect
```
bharadwaj-aditya, line 55: why is there a dependency on cockroach ?

pratickchokhani: This is coming form TIDB

PR #1392, opened 2026-08-26
Use Spanner table count for `num_tables` metric

aasthabharill:
IMO to get an accurate representation of the number of tables migrated, we should use Spanner table count instead of source - as dropping tables from schema migration is allowed in SMT UI


darshan-sj (approved):
LGTM

PR #1395, opened 2026-09-02
Add support for UUID, money and other datatypes in postgres

SaurabhSingh2115:
Fixes #545473957

Add explicit type bindings within the conversion engine for all newly supported native types (e.g., UUID, MONEY, OID, BIT) to ensure the system leverages native storage structures.

Test Findings:

1. Bug -  serial2, serial4 and serial8 were becoming INT64 but silently losing their auto-increment. This PR added them to the type mapping in toddl.go but not to getAutoGenFromTypeName, so they never got an IDENTITY. Postgres treats them as exact aliases of smallserial/serial/bigserial, so this is just a miss.

2. Three e2e columns were asserting mappings that can't happen. uuid_to_bytes expected BYTES, but uuid only ever maps to STRING or UUID. money_to_int64 expected INT64, but money now maps to NUMERIC. And large_decimal_to_numeric was DECIMAL(40,10), which doesn't fit Spanner NUMERIC at all — so the old expectation was quietly asserting precision loss. Renamed the first two, and shrank the ranges on the large ones to (38,9), the biggest value that genuinely fits. Overflow behaviour is still covered in the golden test. Also dropped interval_to_int64 (interval only maps to STRING) and added oid and macaddr8


sources/postgres/toddl.go
```diff
@@ -180,6 +200,87 @@ func toSpannerTypeInternal(srcType schema.Type, spType string) (ddl.Type, []inte
...
+			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.Widened}
+		default:
+			return ddl.Type{Name: ddl.Bytes, Len: ddl.MaxLength}, nil
+		}
+	case "time", "time without time zone":
+		switch spType {
+		case ddl.String:
+			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.Widened}
+		default:
+			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, nil
```
sm745052: in default we do not raise the widened warning ?

sources/postgres/toddl.go
```diff
@@ -180,6 +200,87 @@ func toSpannerTypeInternal(srcType schema.Type, spType string) (ddl.Type, []inte
...
+			return ddl.Type{Name: ddl.Float64}, []internal.SchemaIssue{internal.Widened}
+		default:
+			return ddl.Type{Name: ddl.Int64}, nil
+		}
+	case "bit", "varbit", "bit varying":
+		switch spType {
+		case ddl.String:
+			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.Widened}
+		default:
+			return ddl.Type{Name: ddl.Bytes, Len: ddl.MaxLength}, nil
```
sm745052: same

sources/postgres/toddl.go
SaurabhSingh2115: fixed added warning for those which don't have native spanner typing

sources/postgres/toddl.go
```diff
@@ -128,34 +128,54 @@ func toSpannerTypeInternal(srcType schema.Type, spType string) (ddl.Type, []inte
...
 		}
-	case "numeric":
+	case "numeric", "decimal":
 		switch spType {
 		case ddl.String:
 			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.Widened}
 		default:
-			// TODO: check mod[0] and mod[1] and generate a warning
-			// if this numeric won't fit in Spanner's NUMERIC.
+			if len(srcType.Mods) > 0 {
```
sm745052, line 169: what if there is no Mod ? its a bare numeric

webv2/api/schema.go
```diff
@@ -1455,11 +1455,11 @@ func initializeTypeMap() {
 	}
 	// Initialize postgresTypeMap.
 	toddl = postgres.InfoSchemaImpl{}.GetToDdl()
-	for _, srcTypeName := range []string{"bool", "boolean", "bigserial", "bpchar", "character", "bytea", "date", "float8", "double precision", "float4", "real", "int8", "bigint", "int4", "integer", "int2", "smallint", "numeric", "serial", "smallserial", "text", "timestamptz", "timestamp with time zone", "timestamp", "timestamp without time zone", "varchar", "character varying", "path"} {
+	for _, srcTypeName := range []string{"bool", "boolean", "bigserial", "bpchar", "character", "bytea", "date", "float8", "double precision", "float4", "real", "int8", "bigint", "int4", "integer", "int2", "smallint", "numeric", "decimal", "serial", "serial4", "smallserial", "serial2", "serial8", "text", "timestamptz", "timestamp with time zone", "timestamp", "timestamp without time zone", "varchar", "character varying", "path", "uuid", "money", "oid", "bit", "varbit", "bit varying", "time", "time without time zone", "timetz", "time with time zone", "interval", "citext", "inet", "cidr", "macaddr", "macaddr8", "json", "jsonb"} {
```
sm745052: what does this do ? isnt UUID required here ?

sources/postgres/toddl.go
```diff
@@ -180,6 +200,87 @@ func toSpannerTypeInternal(srcType schema.Type, spType string) (ddl.Type, []inte
...
+		default:
+			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.Widened}
+		}
+	case "citext":
+		switch spType {
+		case ddl.String:
+			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.Widened}
+		default:
+			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.Widened}
+		}
```
sm745052: citext comes as `user defined` in information_schema table. can you please make a table and test these datatypes once in UI ?

sources/common/utils.go
```diff
@@ -197,6 +197,7 @@ var DATATYPE_TO_STORAGE_SIZE = map[string]int{
 	ddl.JSON:      ddl.StringMaxLength,
 	ddl.Numeric:   22,
 	ddl.Timestamp: 12,
+	ddl.UUID:      16,
```
sm745052, line 200: add mapping here as well
ui/src/app/app.constants.ts
export const DataTypes = {
  GoogleStandardSQL : ['BOOL','BYTES','DATE','FLOAT64','INT64','STRING', 'TIMESTAMP', 'NUMERIC', 'JSON'],
  PostgreSQL : ['BOOL','BYTEA','DATE','FLOAT8','INT8','VARCHAR', 'TIMESTAMPTZ', 'NUMERIC', 'JSONB']
}

sources/postgres/toddl.go
```diff
@@ -180,6 +200,87 @@ func toSpannerTypeInternal(srcType schema.Type, spType string) (ddl.Type, []inte
...
+		default:
+			return ddl.Type{Name: ddl.Int64}, nil
+		}
+	case "bit", "varbit", "bit varying":
+		switch spType {
+		case ddl.String:
+			return ddl.Type{Name: ddl.String, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.Widened}
+		default:
+			return ddl.Type{Name: ddl.Bytes, Len: ddl.MaxLength}, []internal.SchemaIssue{internal.Widened}
+		}
```
sm745052, line 234: why ddl.MaxLength ? is there a reasoning behind that ? if so please add in comments

sm745052 (commented):
added some comments .. can you please add e2e, integration tests ? also manually verify in UI (since we do not have UI test coverage) if the datatypes added are WAI. Add the test finding in PR description

webv2/api/schema.go
SaurabhSingh2115: source's data type name, there is uuid there

sources/postgres/toddl.go
SaurabhSingh2115: removed citext as discussed

sources/common/utils.go
SaurabhSingh2115: added

SaurabhSingh2115:
> added some comments .. can you please add e2e, integration tests ? also manually verify in UI (since we do not have UI test coverage) if the datatypes added are WAI. Add the test finding in PR description


Have added

sources/postgres/toddl.go
SaurabhSingh2115: fixed

sources/postgres/toddl.go
SaurabhSingh2115: I have attached warning to it: "Spanner does not support arbitrary precision numeric types. This type
mapping could lose precision and is not recommended for production use",

sources/postgres/toddl.go
```diff
@@ -48,9 +48,24 @@ func (tdi ToDdlImpl) ToSpannerType(conv *internal.Conv, spType string, srcType s
...
+// precision or scale, so its value may not fit the Spanner NUMERIC it maps to.
+// Only GoogleSQL is at risk: its NUMERIC is capped at (38,9), while PostgreSQL
+// dialect NUMERIC has the same range as PostgreSQL itself. money also maps to
+// NUMERIC without mods, but is bounded and reports the issue from its own case.
+func isUnconstrainedNumeric(srcType schema.Type, ty ddl.Type, spDialect string) bool {
+	if ty.Name != ddl.Numeric || spDialect == constants.DIALECT_POSTGRESQL {
+		return false
+	}
+	return len(srcType.Mods) == 0 && (srcType.Name == "numeric" || srcType.Name == "decimal")
+}
```
sm745052: I think we should send conv to `toSpannerTypeInternal`


adding issues here seems like an anti pattern

SaurabhSingh2115: Ig, but we were already adding issues for Arrays etc here

SaurabhSingh2115: also it does say 

// ToSpannerType maps a scalar source schema type (defined by id and
// mods) into a Spanner type. This is the core source-to-Spanner type
// mapping.  **toSpannerType returns the Spanner type and a list of type
// conversion issues encountered**

sm745052: issues added at `toSpannerType` are for general things like arrays not supported etc.

the case we are adding is very specific to one datatype and one dialect.

SaurabhSingh2115: fixed

sm745052 (approved):
thanks for the changes . LGTM

PR #1396, opened 2026-09-03
Map Postgres IDENTITY columns safely to Spanner

SaurabhSingh2115:
Fixes #545474127

Currently, SMT successfully detects legacy PostgreSQL sequences (e.g., SERIAL types) everywhere. However, the dump file parser entirely misses modern SQL-standard IDENTITY configurations (introduced in PostgreSQL 10+). If a user uploads a dump file from a modern database, their auto-increment columns are silently lost.

#### pg_dump
- Generated a real schema dump using PostgreSQL 16 pg_dump with ALWAYS and BY DEFAULT identity columns across public and custom schemas.
- Verified conversion outputs GENERATED BY DEFAULT AS IDENTITY (BIT_REVERSED_POSITIVE) in Spanner DDL instead of degrading to plain INT64.
- Verified legacy SERIAL columns continue to convert identically without regression.
- Live Database Ingestion (PG 16):

#### PostgreSQL < 10 Backward Compatibility (PG 9.6):
Tested against PostgreSQL 9.6 (where attidentity does not exist).
- Confirmed query error is caught cleanly without failing conversion or producing unexpected warnings, maintaining full support for legacy SERIAL sequences.

#### UI / Web Interface:

- Verified backend API endpoints (/connect, /convert/infoschema, /autoGenMap) return AutoGen=Identity for identity columns.
- Verified the UI properly displays the IdentitySkipRange note badge, enables auto-gen configuration controls (Skip Range / Start Counter With), and does not flag false DefaultValue issues for identity columns.

### DIFF
--- a/sources/postgres/infoschema.go
+++ b/sources/postgres/infoschema.go
@@ -244,6 +244,7 @@ func (isi InfoSchemaImpl) GetColumns(conv *internal.Conv, table common.SchemaAnd
                      = (e.object_catalog, e.object_schema, e.object_name, e.object_type, e.collection_type_identifier))
               where table_schema = $1 and table_name = $2 ORDER BY c.ordinal_position;`
 	serialCols := isi.getSerialColumns(conv, table)
+	identityCols := isi.getIdentityColumns(conv, table)
 	cols, err := isi.Db.Query(q, table.Schema, table.Name)
 	if err != nil {
 		return nil, nil, fmt.Errorf("couldn't get schema for table %s.%s: %s", table.Schema, table.Name, err)
@@ -273,7 +274,8 @@ func (isi InfoSchemaImpl) GetColumns(conv *internal.Conv, table common.SchemaAnd
 			}
 		}
 		isSerialColumn := slices.Contains(serialCols, colName)
-		ignored.Default = colDefault.Valid && !isSerialColumn
+		isIdentityColumn := slices.Contains(identityCols, colName)
+		ignored.Default = colDefault.Valid && !isSerialColumn && !isIdentityColumn
 		colId := internal.GenerateColumnId()
 		c := schema.Column{
 			Id:      colId,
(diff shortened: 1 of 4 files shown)


sources/postgres/infoschema.go
```diff
@@ -273,22 +274,49 @@ func (isi InfoSchemaImpl) GetColumns(conv *internal.Conv, table common.SchemaAnd
 			}
 		}
 		isSerialColumn := slices.Contains(serialCols, colName)
-		ignored.Default = colDefault.Valid && !isSerialColumn
+		isIdentityColumn := slices.Contains(identityCols, colName)
+		ignored.Default = colDefault.Valid && !isSerialColumn && !isIdentityColumn
```
sm745052, line 278: for an identity column, colDefault.Valid will be false

sources/postgres/infoschema.go
```diff
@@ -273,22 +274,49 @@ func (isi InfoSchemaImpl) GetColumns(conv *internal.Conv, table common.SchemaAnd
...
+        WHERE attrelid = $1::regclass AND attnum > 0 AND a.attidentity IN ('a', 'd');`
+	identityColsResult, err := isi.Db.Query(identityColsQuery, table.Schema+"."+table.Name)
+	if err != nil {
+		// Ignore error for backward compatibility: 'attidentity' was introduced in PostgreSQL 10
+		// to support SQL-standard identity columns (GENERATED ALWAYS / BY DEFAULT AS IDENTITY).
+		// In PostgreSQL 9.6 and older, this query fails because the 'attidentity' column does not exist.
+		// Since older versions do not have identity columns (they use SERIAL/sequences instead), we ignore it.
+		if strings.Contains(err.Error(), "attidentity") {
+			return []string{}
+		}
```
sm745052, line 305: strings.Contains(err.Error(), "attidentity") does work on 9.6 (I checked), but SQLSTATE 42703 is the stable contract

sources/postgres/infoschema.go
```diff
@@ -273,22 +274,49 @@ func (isi InfoSchemaImpl) GetColumns(conv *internal.Conv, table common.SchemaAnd
...
 
+func (isi InfoSchemaImpl) getIdentityColumns(conv *internal.Conv, table common.SchemaAndName) []string {
+	identityColsQuery := `SELECT a.attname FROM pg_attribute a
+        WHERE attrelid = $1::regclass AND attnum > 0 AND a.attidentity IN ('a', 'd');`
+	identityColsResult, err := isi.Db.Query(identityColsQuery, table.Schema+"."+table.Name)
+	if err != nil {
+		// Ignore error for backward compatibility: 'attidentity' was introduced in PostgreSQL 10
+		// to support SQL-standard identity columns (GENERATED ALWAYS / BY DEFAULT AS IDENTITY).
+		// In PostgreSQL 9.6 and older, this query fails because the 'attidentity' column does not exist.
+		// Since older versions do not have identity columns (they use SERIAL/sequences instead), we ignore it.
```
sm745052, line 302: do we know what is the minimum version of postgres supported ?

sources/postgres/infoschema.go
SaurabhSingh2115: fixed

sources/postgres/infoschema.go
SaurabhSingh2115: fixed

sources/postgres/infoschema.go
SaurabhSingh2115: it is present in the Postgres 10 docs but not in 9.6

https://www.postgresql.org/docs/10/catalog-pg-attribute.html#:~:text=attidentity,generated%20by%20default.

https://www.postgresql.org/docs/9.6/catalog-pg-attribute.html

sm745052 (approved):
LGTM .. please update the description with testing details (like tested in UI, etc)

PR #1401, opened 2026-09-16
Partitioned Table UI safeguards

SaurabhSingh2115:
Fixes #555025616

Testing details:

New UI behaviour 

1. By default the partitioned tables would be skipped and a warning will be displayed on the issues and suggestions board. 
image

2. Client can manually select them and click on the restore button to get them back to migration scope.
image
image

3. The specified warning will disappear if the client has restored all the partitioned tables back to the migration scope.
image


sm745052 (commented):
some testing details and UI screenshots on PR desc would be helpful.

schema/schema.go
```diff
@@ -45,6 +45,7 @@ type Table struct {
 	CheckConstraints []CheckConstraint
 	Indexes          []Index
 	Id               string
+	PartitionParent  string // Parent table name if this is a child partition, empty otherwise.
```
sm745052, line 48: // SetRowStats populates conv with the number of rows in each table.
func (is *InfoSchemaImpl) SetRowStats(conv *internal.Conv, infoSchema InfoSchema) {
	tables, err := infoSchema.GetTables()
	if err != nil {
		conv.Unexpected(fmt.Sprintf("Couldn't get list of table: %s", err))
		return
	}
	for _, t := range tables {
		tableName := infoSchema.GetTableName(t.Schema, t.Name)
		count, err := infoSchema.GetRowCount(t)
		if err != nil {
			conv.Unexpected(fmt.Sprintf("Couldn't get number of rows for table %s", tableName))
			continue
		}
		conv.Stats.Rows[tableName] += count
	}
}



will it double count ? if so can you correct it and add an unit test ?

SaurabhSingh2115: fixed

sm745052 (approved):
LGTM ...nit: change the warning text from
```
Table 'a': Partitioned tables are ignored
```
to
```
Table 'a': Its partitioned tables are ignored
```

